PR #36, opened 2017-11-27
Replace some instances of echo by printf

Firobe:
On some systems (including mine), the default behaviour of echo is to display '\n' verbatim, thus corrupting the file ("\nlet () = ...") and failing the build. One solution could be to add the -e option to echo but using printf is a more standard solution.


OCamlPro-Iguernlala:
Thanks for reporting and fixing the issue. Can you do the merge commit into `next` instead of `master` ?

Firobe:
It should be the case now.

OCamlPro-Iguernlala:
Thanks

PR #54, opened 2018-01-12
Fix generated cmdline parsing file

pinicarus:


pinicarus:
The file lib/util/cmdline_parser.ml generated from the file Makefile.users produces escaped new-lines that fail to compile

OCamlPro-Iguernlala:
Thanks for reporting the issue.
It's already reported in https://github.com/OCamlPro/alt-ergo/pull/36 and fixed by https://github.com/OCamlPro/alt-ergo/commit/f996b5cd446be80b9e9952d49863e5553d003e31

pinicarus:
Could you consider backporting your fix on 2.0.0 and tag a 2.0.1 version?
This would help distribution package maintainers to use a tagged version instead of maintaining separate patches.

Thanks.

PR #55, opened 2018-01-13
Use dynamic temporary files to run non-regression tests

pinicarus:
Building in some sandboxed environments prevents from using absolute paths


OCamlPro-Iguernlala:
Thanks for the improvement. Can you do the merge commit into `next` instead of `master` ?

OCamlPro-Iguernlala:
Thanks

PR #65, opened 2018-02-02
 add the ability to use the given timelimit per goal. This is important when we have multiple goals per file

OCamlPro-Iguernlala:
@mebsout What do you think about this ? I'm not sure for the GUI


OCamlPro-Iguernlala:
The change in the format of the result returned by the solver will/would affect other tools (eg Why3) that recognize Alt-Ergo's answer with regular expressions.
Should we propose a patch for all released Why3 versions and for the dev branch ?

OCamlPro-Iguernlala:
@bobot Yes, I changed the output format in order to keep compatibility with Why3.

For Frama-C and WP plugin, I don't know how Alt-Ergo' answer is parsed. Because now, we  print "( xxx secs)" for resolution time instead of just (xxx) and "(goal g)" at the end of the line to indicate the name of the goal.

PR #73, opened 2018-02-06
ignore multi-triggers with a size greater than a given constant (default const = 2)

OCamlPro-Iguernlala:
@bobot @correnson

I noticed a very bad user-given multi-trigger in axioms `separated_1` of file `Memory.why`. This makes Alt-Ergo stuck in the instantiation module, and I decided to ignore "big" multi-triggers.

Maybe you should rewrite the axiom, as follows:

```
axiom separated_1 : forall p:addr, q:addr. forall a:int, b:int
  [separated p a q b].
  separated p a q b ->
  forall i:int, j:int  [Mk_addr (base p) i, Mk_addr (base q) j].
 (Int.(<=) (offset p) (i)) /\ (Int.(<) (i) ((Int.(+) (offset
  p) (a)))) -> (Int.(<=) (offset q) (j)) /\ (Int.(<) (j) ((Int.(+) (offset
  q) (b)))) -> not ((Mk_addr (base p) i) = (Mk_addr (base q) j))

```

### DIFF
--- a/sources/lib/reasoners/matching.ml
+++ b/sources/lib/reasoners/matching.ml
@@ -424,12 +424,16 @@ module Make (X : Arg) : S with type theory = X.t = struct
     Debug.match_pats_modulo pat lsubsts;
     List.fold_left (match_one_pat env tbox pat) [] lsubsts
 
+  let trig_weight s t =
+    match T.view s, T.view t with
+    | {T.f=Symbols.Name _}, {T.f=Symbols.Op _}   -> -1
+    | {T.f=Symbols.Op _}  , {T.f=Symbols.Name _} -> 1
+    | _ -> (T.view t).T.depth - (T.view s).T.depth
+
+
   let matching env tbox pat_info =
     let pats = pat_info.trigger in
-    let pats_list =
-      List.stable_sort
-        (fun s t -> (T.view t).T.depth - (T.view s).T.depth) pats.F.content
-    in
+    let pats_list = List.stable_sort trig_weight pats.F.content in
     Debug.matching pats;
     let egs =
       { sbs = SubstT.empty;


OCamlPro-Iguernlala:
Actually, my heuristic is probably too restrictive. I loose a lot of proofs on our internal benchmarks. I'll look for another workaround.

correnson:
Thanks for the hint. The original trigger was (in Why-3 syntax):

```

  lemma separated_1 : forall p q : addr. forall a b i j : int
    [ separated p a q b , { base = p.base ; offset = i } ,  { base = q.base ; offset = j } ].
    separated p a q b ->
    p.offset <= i < p.offset + a ->
    q.offset <= j < q.offset + b ->
    { base = p.base ; offset = i } <>
    { base = q.base ; offset = j }
```

Just for information, why was it so bad ?

OCamlPro-Iguernlala:
I meant: if you have a multi-trigger `[t1, t2, t3]`, and if there are (for instance), 200 terms that match the first term of the trigger, 200 terms that match the second one, and 3 terms that match the last one, Alt-Ergo will try to generate the 200 x 200 x 3 instances. But in general, only few substitutions are generated at the end.

The workaround I merged sorts the terms of multi-triggers differently to try "to guide" E-matching.

correnson:
Ok! the original intent was to trigger an instantiation whenever two pointers are compared when there is a separation clause in the context, but we can not use `(<>)` in triggers, right ?

Alternatively, we can try to trigger something based on array access, instead. Can we use arrays in triggers?

OCamlPro-Iguernlala:
It was so bad because Alt-Ergo sorted the terms of the multi-triggers as follows (depending on depth of terms):

 `{ base = p.base ; offset = i } ;  { base = q.base ; offset = j },  separated p a q b` 

and if you have a lot of terms that match `{ base = p.base ; offset = i }` and  `{ base = q.base ; offset = j }`, we have the explosion in explained above.

Forcing Alt-Ergo to consider `separated p a q b` first (by rewriting the axiom or with my workaround) solves the problem.

And more generally, head symbols of  `{ base = p.base ; offset = i }` and  `{ base = q.base ; offset = j }` belong to an interpreted theory (records), and  this may be bad

OCamlPro-Iguernlala:
Yes, we can use arrays in triggers. Actually, the main issue here is that, depending on the context, instantiation of multi-triggers may make the E-matching module stuck ...

OCamlPro-Iguernlala:
We should probably allow some "semantic triggers in the generic instantiation engine, such as `<>`,  `<=` ...

PR #106, opened 2018-04-17
Handle ite in Alt-Ergo's backend

OCamlPro-Iguernlala:


OCamlPro-Iguernlala:
@mebsout @conchon naive reasoner in the backend for `If (Bool / Prop) then Term else Term`

PR #136, opened 2018-10-29
Separate Typing from Trigger inference

Gbury:
This PR tries and factorize trigger inference so that it can more easily be moved outside the typechecker. The trigger inference is still done in the typechecker, but only at one place, when typechecking a whole input file.


Gbury:
Some diffs may be large because I re-indented the concerned files with ocp-indent. Therefore it might be better to look at each commit separately.

Gbury:
PR rebased following the ocp-indent merge.

PR #143, opened 2018-11-06
Add documentation to the Ty module

Gbury:


sources/lib/structures/ty.mli
```diff
@@ -26,77 +26,204 @@
...
+  (** Bitvectors of a given length *)
   | Text of t list * Hstring.t
+  (** Abstract types applied to arguments. [Text (args, s)] is
+      the application of the abstract type constructor [s] to
+      arguments [args]. *)
   | Tfarray of t * t
+  (** Functional arrays. [TFarray (src,dst)] maps values of type [src]
+      to values of type [dst]. *)
   | Tnext of t
+  (** TODO: ? *)
```
OCamlPro-Iguernlala: experiments with a reachability theory. Can probably be removed (or ignored)

sources/lib/structures/ty.mli
```diff
@@ -26,77 +26,204 @@
...
+    The [value] field is mutated during unification,
+    hence distinct types should have disjoints sets of
+    type variables (see function {fresh}). *)
 
-and tvar = { v : int ; mutable value : t option }
 and trecord = {
   mutable args : t list;
+  (** parameters of the record
+      TODO: or are they the arguments applied
+            to the record constructor type ? *)
```
OCamlPro-Iguernlala: they are type variables, like in `Text` case

sources/lib/structures/ty.mli
```diff
@@ -26,77 +26,204 @@
...
-  mutable lbs :  (Hstring.t * t) list
+  (** Name of the record type *)
+  mutable lbs :  (Hstring.t * t) list;
+  (** List of fields of the record. Each field has a name,
+      and an associated type. *)
 }
+(** Record types. *)
 
+module Svty : Set.S with type elt = int
+(** Sets of type variables *)
```
OCamlPro-Iguernlala: indexed with the "int" ID

sources/lib/structures/ty.mli
```diff
@@ -26,77 +26,204 @@
...
+val print : Format.formatter -> t -> unit
+(** Printing function for types (does not print
+    the type of each fields for records). *)
+
+val print_list : Format.formatter -> t list -> unit
+(** Print function for lists of types (does not print
+    the type of each fields for records). *)
+
+val print_full : Format.formatter -> t -> unit
+(** Print function including the record fields. *)
```
OCamlPro-Iguernlala: and the constructors of enumerations I think

sources/lib/structures/ty.mli
Gbury: Aren't they rather the concrete types that replace the type variables ? (just as in the `Text` case)

sources/lib/structures/ty.mli
Gbury: After looking at the code, the contructors of an enumerations are never printed currently.

Some justification for that would be that, these constructors are always the same, contrary to record fields, whose actual type may change due to type variable substitution.

sources/lib/structures/ty.mli
OCamlPro-Iguernlala: Yes

PR #157, opened 2018-11-22
[WIP] Change build system

Gbury:
Switch to manual makefiles to ocp-build, and use ocp-autoconf. At the same time, perform a big cleanup of Makefiles (effectively, all the makefiles of `next` have been removed, and a new makefile written).

Remaining TODOS:
- [x] build plugins
- [x] install plugins
- [x] add tests/non-regression rules to makefiles
- [ ] add archi Makefile target
- [x] add public-release Makefile target

Still WIP, not ready to be merged.


Gbury:
This should be reviewable. The build system now uses `dune`.
The `archi` target has been removed because it is surprisingly difficult to adapt to a build system.

OCamlPro-Iguernlala:
It seems that the branch still has conflicts with next

Gbury:
Note that originally, I tried ocp-build, but later switched to dune. This was for the following reason:
- ocp-build only supports packs, and not prefixed libraries
- in case of packs, ocp-build is not currently able to build documentation.

About the prefix/libdir/sharedir: it is sort of necessary (at least part of it) because these values are used in `lib/util/config.ml` and are used for alt-ergo to print these directories when invoked with the correct command line option (the `-where` option). While this option could be removed, alt-ergo still search plugins in the `pluginsdir` directory, so one cannot get rid of this easily.

OCamlPro-Iguernlala:
I'm not expert of dune and/or  findlib. But if using  findlib.dynload would be cleaner, why not !

Gbury:
Rebased and ready to be reviewed.
I'll try and add the archi target soon.

Gbury:
All comments fixed, ^^

sources/alt-ergo.opam
```diff
@@ -0,0 +1,28 @@
...
+name: "alt-ergo"
+version: "2.2.0+"
+authors: "Alt-Ergo developers"
+maintainer: "OCamlPro <alt-ergo@ocamlpro.com>"
+license: "OCamlPro Non-Commercial Purpose License, version 1"
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+]
+depends: [
```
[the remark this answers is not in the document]
Gbury: done.

opam
```diff
@@ -10,19 +10,15 @@ bug-reports: "https://github.com/OCamlPro/alt-ergo/issues"
...
 build: [
-  ["autoconf"]
-  ["./configure" "-prefix" "%{prefix}%"]
-  [make]
+  ["ocp-build" "build"]
 ]
 
 install: [
-  [make "install" "MANDIR=%{man}%"]
+  ["ocp-build" "install"]
```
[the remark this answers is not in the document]
Gbury: I removed the root opam file, which isn't really useful.

sources/Makefile.config
```diff
@@ -0,0 +1,4 @@
+# Generated by configure
```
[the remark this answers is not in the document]
Gbury: done

Gbury:
Rebased (correctly this time).

non-regression/non-regression.sh
```diff
@@ -18,9 +18,10 @@ echo "[ with Tableaux-CDCL ]"
...
 
-echo "===============================================================================================\c";
+echo "===============================================================================================";
 echo "===============================================================================================";
 echo "[ WITH FM-SIMPLEX ]"
-echo "===============================================================================================\c";
 echo "===============================================================================================";
-./main_script.sh "-inequalities-plugin `pwd`/../sources/fm-simplex-plugin.cmxs"
+echo "===============================================================================================";
+./main_script.sh "-inequalities-plugin `pwd`/../sources/_build/install/default/share/plugins/fm-simplex-plugin.cmxs"
```
Coquera: Can we have/need a link for plugins in sources/ ?

sources/Makefile
```diff
@@ -1,3 +1,211 @@
...
+# ===========
+# Build rules
+# ===========
+
+# Build the alt-ergo lib
+lib: gen
+	$(DUNE) build $(DUNE_FLAGS) lib/AltErgoLib.{cma,cmxa,cmxs}
+
+# Build the cli/text alt-ergo bin
+bin: gen
```
Coquera: Can we have/need a link for binaries in sources/ ?

Gbury: @OCamlPro-Iguernlala suggested having symbolic links in the sources directory that would point to the relevant locations

OCamlPro-Iguernlala:
`make AB-Why3` doesn't work

OCamlPro-Iguernlala:
`make lib`, `make gen` not working

`make clean` does not remove generated file config.ml

Gbury:
Typo fixed.
Currently `make distclean` remove the generated files. However, note that because `Makefile` includes `Makefile.config` which is generated together with `config.ml` by the configure script, any invocation to `make` (including `make clean`) will re-generate `config.ml` if needed, so I'm not sure what good it is to remove it with clean.

Coquera:
> `make AB-Why3` doesn't work

Same for make fm-simplex

OCamlPro-Iguernlala:
- add old compilations flags if they are not set by dune:

```
LIGHT_OFLAGS = -annot -absname -bin-annot -short-paths -strict-sequence -g $(WARNINGS) -inline 100 $(INCLUDES)
```
- enable some optimizations:
```
OFLAGS = $(LIGHT_BFLAGS) -O3 -unbox-closures -for-pack AltErgoLib
```
- enable warnings of Makefile.users, and add others as TODOs (in a comment):
```
WARNINGS= -w +A-45-27-4-9-44-32-48-41-6-22

# xxxx: Warning 22: user warnings
# x6 : Warning 60: unused module MyNums. ==> pb with recursive modules

# x623: Warning 27: unused variable binders
# x619: Warning 4 : this pattern-matching is fragile.
# x291: Warning 9 : the following labels are not bound in this record pattern:
# x161: Warning 44: this open statement shadows the value identifier = (which is later used)
# x134: Warning 32: unused value take_max.
# x37 : Warning 48: implicit elimination of optional argument ?kind
# x35 : Warning 41: Other belongs to several types: abstract Sig.lit_origin
# x26 : Warning 6 : label is_le was omitted in the application of this function.
```

sources/dune
```diff
@@ -0,0 +1,12 @@
+(env
+ (dev
+  (flags (:standard
+           -annot -absname -bin-annot
+           -w +A-45-27-4-9-44-32-48-41-6-22
+           )))
+ (release
+  (flags (:standard
+           -annot -absname -bin-annot))
+  (ocamlopt_flags -O3 -unbox-closures))
```
Coquera: Why does the flags "-O3" and "-unbox-closures" are only enable in "release" and not both in "dev" and "release" ?

Gbury: Speed up build time ?

Coquera: Yeah sure. What happen if i want to test the performance of a new optimization in AE on my benchs, do i have to build a release to have the most of the performance offer by my compiler ? or do i use a switch without Flambda enable (since "-O3" has no effect if your switch does not have Flambda.) until i want to bench ?

OCamlPro-Iguernlala: I think we can put `-O3 -unbox-closures` in "dev" as well. Compilation of Alt-Ergo is quite fast, and even faster with dune. As @Coquera saind, these options have probably no effect in Flambda is disabled.

Gbury: done.

sources/Makefile
Gbury: I create the symbolics links (though they may point to non-existant files if the bins have not been compiled yet).

non-regression/non-regression.sh
Gbury: Are plugins sufficiently frequently used to need symlinks ?

OCamlPro-Iguernlala:
Hop ! big change !

PR #189, opened 2019-01-15
ADTs in the backend

OCamlPro-Iguernlala:


sources/lib/structures/ty.mli
```diff
@@ -54,6 +54,9 @@ type t =
       to values of type [dst]. *)
   | Tsum of Hstring.t * Hstring.t list
   (** Enumeration, with its name, and the list of its constructors. *)
+
+  | Tadt of Hstring.t * t list
```
Gbury: Could you add a little doc comment describing the semantic and/or use case of this new constructor ?

sources/lib/structures/ty.mli
```diff
@@ -76,15 +79,25 @@ and trecord = {
   mutable lbs :  (Hstring.t * t) list;
   (** List of fields of the record. Each field has a name,
       and an associated type. *)
+  constr : Hstring.t; (* for ADTs that become records. default is "{" *)
```
Gbury: Could the default be the empty string ?

sources/lib/structures/ty.mli
```diff
@@ -76,15 +79,25 @@ and trecord = {
...
 }
 (** Record types. *)
 
+type adt_constr =
+  (*  { constr : Hstring.t ; destrs : (Hstring.t * t) list }*)
+  Hstring.t * (Hstring.t * t) list
```
Gbury: The record option would be better than a tuple I think.
Also, maybe add some doc to explain what each field of the tuple stores ?

sources/lib/structures/ty.mli
```diff
@@ -76,15 +79,25 @@ and trecord = {
...
+
+
+type type_body =
```
Gbury: What is the use case envisionned for this type ? A quick search seems to say that this type is not used anywhere except the `type_body` function, particularly, it is not used outside of the `Ty` module.

sources/lib/structures/ty.mli
```diff
@@ -135,7 +148,12 @@ val tsum : string -> string list -> t
 (** Create an enumeration type. [tsum name enums] creates an enumeration
     named [name], with constructors [enums]. *)
 
-val trecord : t list -> string -> (string * t) list -> t
+val t_adt :
+  ?body: ((string * (string * t) list) list) option ->
+  string -> t list ->
+  t
```
Gbury: Could you document (even rather quickly) the expected meaning of the heavily nested list type ?

sources/lib/structures/typed.ml
```diff
@@ -52,6 +52,10 @@ type oplogic =
     OPand | OPor | OPxor | OPimp | OPnot | OPiff
   | OPif
 
+type pattern =
+  | Constr of { name : Hstring.t ; args : (Var.t * Hstring.t * Ty.t) list}
```
Gbury: Again, some documentation ?
Particularly, what is the semantic of the `Hstring.t` in each triple in the list of arguments ?

sources/lib/structures/typed.ml
```diff
@@ -79,6 +83,9 @@ and 'a tt_desc =
   | TTnamed of Hstring.t * ('a tterm, 'a) annoted
   | TTite of ('a tform, 'a) annoted *
              ('a tterm, 'a) annoted * ('a tterm, 'a) annoted
+  | TTsharp of bool * ('a tterm, 'a) annoted  * Hstring.t
```
Gbury: Could this be named something like projection ? The `sharp` reference is pretty specific to smt-lib and doesn't really help understand what this construction does.

sources/lib/structures/ty.mli
OCamlPro-Iguernlala: ok

PR #191, opened 2019-01-16
Abstract over input method

Gbury:
This PR creates the notion of input method.

An input method is a black box that outputs a list of typed statements, from a given filename. Currently, the only input method defined is the legacy one, using the AltErgoParsers lib and the builtin typechecker (as well as the psmt2 to native translation). Future input methods should include a direct psmt2 input method (to avoid typechecking twice), and a dolmen input method.

The choice of input method is currently done using a command line argument named "frontend".

Additionnally, this PR moves the generation of new ids for typed formulas from the typechecker to the `typed` module, as future input methods should not depend on the builtin typechecker, but only on the `typed` module.

### DIFF
--- a/sources/examples/lib_usage.ml
+++ b/sources/examples/lib_usage.ml
@@ -52,7 +52,7 @@ let goal_3 = PA.mk_goal Loc.dummy "toy_3" (PA.mk_not Loc.dummy eq1)
 
 let parsed = [goal_1; goal_2; goal_3]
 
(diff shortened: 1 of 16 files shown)


Gbury:
The `expr` part of the input method could be removed actually. I included it in order for the GUI to work initially, but since the type of parsed expression is not fixed, it was too much trouble to actually use it in the GUI.
So wait for me to remove it before merging.

PR #193, opened 2019-01-16
Better CI

Gbury:
This PR cleans up the big ugly script that was used in the travis CI.

Main changes:
- the travis.sh script is now split in multiple scripts added in the `extra` directory
- use the avsm ppa to install opam2 (instead of fetching the binary with wget)
- use travis builtin system to install external packages
- many different compiler versions are tested by separate travis jobs (in parralel) using the matrix feature of travis

Additionally, most shell scripts in the `extra` directory now use git to determine the root of the git repo and thus can now be called from anywhere (and not exclusively from the `extra` directory).


Gbury:
All tests now pass, this is ready for review, ^^

Gbury:
Ready for review, ^^

Issue #204, opened 2019-01-20
Add opam update & opam upgrade in travis ?


OCamlPro-Iguernlala:
I got an issue because the (newly released) version 0.2 of psmt2-frontend was not installed. I realized that this was due to caching .opam dir. Commenting the cache and commiting solved the problem. But a better solution would probably be to run opam update and opam upgrade to have latest versions of packages before compiling ?

PR #205, opened 2019-01-21
Run opam update if needed in the travis script

Gbury:
Should solve #204


Gbury:
The bash conditional works as expected, this is good to merge.

PR #214, opened 2019-01-23
Fix potential typo in arrays_rel.ml

### LINKED ISSUE #209: Suspicious unused variable in arrays_rel.ml
In file `arrays_rel.ml` around line 261 (cf https://github.com/OCamlPro/alt-ergo/blob/next/sources/lib/reasoners/arrays_rel.ml#L261 ) a fold is done on a set, but the folded function actually ignores the element of the set given to it, and thus repeats the same thing for each element of the set...

Gbury:
Should close #209 

cc @mebsout  when he returns.


Gbury:
Rebased on next. The non-regression tests should now pass.

Maybe this would deserve some benchmarking before merging ?

OCamlPro-Iguernlala:
@Gbury I just started benchmarking

PR #215, opened 2019-01-25
Refactor main loop

Gbury:
The main loop of `main_text` is changed toward something more incremental wrt the input statements.
Parsing is now supposed to return a lazy sequence of statements, which are each treated individually in the main loop.

Some caveats with the current implementation:
- The code for parsing hasn't been modified much so as it stands, parsing input files still builds the list of parsed statements, which is then transformed into a sequence (that should be fixable in later PRs)
- There are still quite a few lists used (particularly in the main loop state), which are not used very efficiently (notably some potentially very long lists are concatenated and reversed). Further work should probably change these to vectors for better performance of appending and iterating.
- Some logic about splitting typed statements into cnf is duplicated between `main_text` (with the new main loop), and `main_gui` (which still uses the function in the typechecker).


Gbury:
Ready for review @OCamlPro-Iguernlala

.travis.yml
```diff
@@ -15,7 +15,7 @@ addons:
     sources:
       - avsm
     packages:
-      - opam>=2
+      - opam>=2.0.1
```
Coquera: Why do you need this version ?

Gbury: It's not that specific version, but some builds failed because of the `--disable-sandox` option, and using this lower bound seemed to fix the problem. But we could probably drop this modification and look for a cleaner fix if the problem re-appears.

PR #216, opened 2019-01-29
More documentation

Gbury:
This PR first adds some documentation to some module interfaces.

More importantly, it sets up automatic documentation build and checking in the travis CI, as well as automatic uploading of the generated html documentation on the gh-pages branch.


Gbury:
Good to review and merge.
Note that there will likely be some fixes to do on next once this is merged as travis automatically disables gh-pages upload on pull request (thus the auto-upload of documentation in gh-pages cannot be tested in a PR).

PR #220, opened 2019-02-08
Add targets to build release version of single targets

### LINKED ISSUE #219: add targets make install-lib install-gui ...
We should be able to do
```
$ ./configure <target>
$ make <target>
$ make install-<target>
```
Without relying on OPAM.

Gbury:
Should close #219

### DIFF
--- a/sources/INSTALL.md
+++ b/sources/INSTALL.md
@@ -5,14 +5,19 @@
   LablGtk2 and the widget GSourceView2 to compile the GUI. You may
   need superuser permissions to perform the installation.
 
-#### Common Steps
+#### Configuration
 
-  1. Configure with "./configure" to generate Makefile.config
+  1. Configure with "./configure" to generate Makefile.config,
+  in order to build everything (lib, parsers, binaries, and GUI).
 
   2. Alternatively, you can configure with "./configure -prefix
   some-absolute-path-prefix" to add a prefix for installation
   directories. You may also want to use "make gen && cat lib/util/config.ml"
   to see directories where things will be installed.
+
+  3. You can use "./configure <package>" to select which packeg you
+  can to build. "<package>" may be one of: alt-ergo-lib, alt-ergo-parsers,
+  alt-ergo, altgr-ergo.
 
 The steps below will build and install native or bytecode binaries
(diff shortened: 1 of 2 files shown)


sources/INSTALL.md
```diff
@@ -5,15 +5,20 @@
...
-  1. Configure with "./configure" to generate Makefile.config
+  1. Configure with "./configure" to generate Makefile.config,
+  in order to build everything (lib, parsers, binaries, and GUI).
 
   2. Alternatively, you can configure with "./configure -prefix
   some-absolute-path-prefix" to add a prefix for installation
   directories. You may also want to use "make gen && cat lib/util/config.ml"
   to see directories where things will be installed.
 
+  3. You can use "./configure <package>" to select which packeg you
```
OCamlPro-Iguernlala: Typo

sources/INSTALL.md
```diff
@@ -5,15 +5,20 @@
...
+  can to build. "<package>" may be one of: alt-ergo-lib, alt-ergo-parsers,
```
OCamlPro-Iguernlala: can -> want ?

sources/INSTALL.md
```diff
@@ -25,17 +30,29 @@ depending on whether ocamlopt is installed or only ocamlc is detected.
 
   3. Uninstall with "make uninstall"
```
OCamlPro-Iguernlala: I don't remember if make uninstall works if we only install some targets ? To be tested

sources/INSTALL.md
```diff
@@ -49,12 +66,14 @@ The steps below will build and install additional plugins (extension
...
 
-  2. Install with "make install"
+  2. The Fm-Simplex plugin is currently built and installed
+  at the same time as the alt-ergo binary.
 
 #### The profiler plugin
 
 This plugin has been "inlined" in Alt-Ergo sources.
 
+
```
OCamlPro-Iguernlala: add some text for the Why3 parser plugin (which should not be put in alt-ergo-parsers package due to licensing considerations)

sources/INSTALL.md
OCamlPro-Iguernlala: Ah ! I see below that you also added targets to remove specific packages

PR #239, opened 2019-04-02
WIP : Cube test first version

Stevendeo:
Cubetest for Alt-Ergo experiments

### DIFF
--- a/sources/lib/util/options.ml
+++ b/sources/lib/util/options.ml
@@ -134,6 +134,9 @@ module M = struct
   let unsat_mode = ref false
   let inline_lets = ref false
 
+  let cubefast = ref ""
+
+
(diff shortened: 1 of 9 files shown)


sources/lib/util/options.ml
```diff
@@ -250,6 +253,28 @@ module M = struct
...
+      cubefast := cf_max
+    | "both" | "b"  ->
+      cubefast := cf_both
+    | _ -> (
+        raise
+          (Arg.Bad
+             ("Bad value '" ^
+              str ^
+              "' for option -cubefast"))
+      )
```
Gbury: Could the `cubefast` option actually be a sum type rather than encode its options in a string ?

PR #240, opened 2019-05-10
Preludes and plugins relative to the executable file

Gbury:
This PR does a few things things:
- first, a bugfix: the preludes (b-set-theory and fpa-theory), weren't installed, which the second commit fixes
- second, the paths for the directories where alt-ergo looks for the plugins and preludes, is now relative to the executable file, in order for alt-ergo to be relocatable. This is done in the first commit, which also fixes these paths (which previously referred to the "lib" directory instead of the "share" directory).


Gbury:
Good to go once CI is green.

PR #241, opened 2019-05-14
Solving issue #200

Stevendeo:


Gbury:
Iirc, the idea was to change the definition of types in module `Ty`, so that we have two distinct notion: type definitions (e.g. defintion of a record type, or an enumeration), and types, whereas currently the two are more or less mixed together, which is annoying.

OCamlPro-Iguernlala:
Are you sure that the two functions are exactly the same ? Because the function you deleted is supposed to print types in a format that can be parsed back by Alt-Ergo

Stevendeo:
I don't think they are exactly the same, can the used one be modified or is its output also used somewhere else ?

Stevendeo:
Update : replaced Ty.print occurences by Ty.print_full from original function

PR #250, opened 2019-09-16
Correct steps bound issue

OCamlPro-Coquera:
Correct steps bound issue where when the option is set the number of total steps differs from normal behavior. Use int instead of Int64 for steps_bounds

### DIFF
--- a/sources/tools/gui/main_gui.ml
+++ b/sources/tools/gui/main_gui.ml
@@ -497,7 +497,7 @@ let update_status image label buttonclean env s steps =
     let time = Options.Time.value () in
     if not satmode then Loc.report std_formatter d.st_loc;
     if satmode then printf "@{<C.F_Red>unsat@}@."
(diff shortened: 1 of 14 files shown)


sources/tools/gui/main_gui.ml
```diff
@@ -497,7 +497,7 @@ let update_status image label buttonclean env s steps =
     let time = Options.Time.value () in
     if not satmode then Loc.report std_formatter d.st_loc;
     if satmode then printf "@{<C.F_Red>unsat@}@."
-    else printf "@{<C.F_Green>Valid@} (%2.4f) (%Ld)@." time steps;
+    else printf "@{<C.F_Green>Valid@} (%2.4f) (%L)@." time steps;
```
Gbury: This should be `%d` and not `%L` (`%L` is specific to scanf).

Gbury:
While we're at it, maybe we could add a check during argument parsing to ensure the step bounds is positive ?

PR #251, opened 2019-09-23
WIP Remove constraints over ocp-indent.1.6.1 and indent ocaml.warnings pr…

OCamlPro-Coquera:
…operly

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ before_install:
   # Init opam, and the default switch with the right ocaml version
   - opam init --disable-sandboxing --compiler=${OCAML_VERSION}
   - eval `opam config env`
-  # Install odoc & ocp-indent
+  # Install odoc & ocp-indent>1.6.1
   - opam install odoc ocp-indent
   # Set the verbose option *after* compiling the compiler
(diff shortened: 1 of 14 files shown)


.travis.yml
```diff
@@ -39,7 +39,7 @@ before_install:
   - opam init --disable-sandboxing --compiler=${OCAML_VERSION}
   - eval `opam config env`
   # Install odoc & ocp-indent
-  - opam install odoc ocp-indent.1.6.1
+  - opam install odoc ocp-indent
```
Gbury: There is a slight problem with that change because it doesn't constrain at all the version of `ocp-indent`, however `ocp-indent` gives different results when indenting `[@ocaml.ppwarning]` annotation, and thus depending on the whims of the solver when installing `ocp-indent`, we could have either `1.6.1` or another version, which would result in CI failing sometimes just because the wrong version of `ocp-indent` vas installed (if I understand correctly).
It would probably be better to try and install a version of `ocp-indent > 1.6.1`, but I'm not sure if it's doable with the command line...

OCamlPro-Coquera:
Should be done with #264

PR #252, opened 2019-10-07
Add compatibility package `stdlib-shims` and replace `Pervasives.*` b…

OCamlPro-Coquera:
…y `Stdlib.*`

### DIFF
--- a/sources/lib/frontend/typechecker.ml
+++ b/sources/lib/frontend/typechecker.ml
@@ -37,7 +37,7 @@ module S = Set.Make(String)
 module HSS = Hstring.Set
 
 module MString =
-  Map.Make(struct type t = string let compare = Pervasives.compare end)
+  Map.Make(struct type t = string let compare = String.compare end)
 
 module Types = struct
 
@@ -86,7 +86,7 @@ module Types = struct
       List.for_all2
         (fun pp x ->
            match pp with
-           | PPTvarid (y, _) -> Pervasives.(=) x y
+           | PPTvarid (y, _) -> Stdlib.(=) x y
            | _ -> false
         ) lpp lvars
     with Invalid_argument _ -> false
@@ -105,7 +105,7 @@ module Types = struct
           to_tyvars := MString.add s nty !to_tyvars;
           nty
       end
-    | PPTexternal (l, s, loc) when Pervasives.(=) s "farray" ->
+    | PPTexternal (l, s, loc) when String.equal s "farray" ->
       let t1,t2 = match l with
         | [t2] -> PPTint,t2
         | [t1;t2] -> t1,t2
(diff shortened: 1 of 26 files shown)


sources/lib/frontend/typechecker.ml
```diff
@@ -37,7 +37,7 @@ module S = Set.Make(String)
 module HSS = Hstring.Set
 
 module MString =
-  Map.Make(struct type t = string let compare = Pervasives.compare end)
+  Map.Make(struct type t = string let compare = Stdlib.compare end)
```
Gbury: Here, using `String.compare` (or at least a string-specialised version of comparison) would probably be better I think.

sources/lib/frontend/typechecker.ml
```diff
@@ -105,7 +105,7 @@ module Types = struct
           to_tyvars := MString.add s nty !to_tyvars;
           nty
       end
-    | PPTexternal (l, s, loc) when Pervasives.(=) s "farray" ->
+    | PPTexternal (l, s, loc) when Stdlib.(=) s "farray" ->
```
Gbury: Same here concerning specialisation ?

sources/lib/reasoners/fun_sat.ml
```diff
@@ -75,7 +75,7 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
         SE.fold
           (fun f mp ->
              let w = var_inc +. try ME.find f mp with Not_found -> 0. in
-             stable := !stable && Pervasives.(<=) w 1e100;
+             stable := !stable && Stdlib.(<=) w 1e100;
```
Gbury: Float comparison ? (By the way, do we care about the behaviour of this comparison on nan ? because iirc there are some tricky differences between compare and infix comparisons on floats)

OCamlPro-Coquera: Stdlib.(infix op) by Stdlib.compare

sources/lib/structures/ty.ml
```diff
@@ -602,12 +602,12 @@ let instantiate lvar lty ty =
...
 
-let compare_subst = M.compare Pervasives.compare
+let compare_subst = M.compare Stdlib.compare
 
-let equal_subst = M.equal Pervasives.(=)
+let equal_subst = M.equal Stdlib.(=)
 
 module Svty =
-  Set.Make(struct type t = int let compare = Pervasives.compare end)
+  Set.Make(struct type t = int let compare = Stdlib.compare end)
```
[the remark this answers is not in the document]
OCamlPro-Coquera: I specializied the Map.make(<int>) inside util.ml and add Set.make(<int>) and replace every occurrences of Set(Map).Make(<int>) with Util.SI (MI)

PR #256, opened 2019-10-09
Add option "-gc-policy" to set the allocation policy of the garbage c…

OCamlPro-Coquera:
…ollector

### DIFF
--- a/sources/lib/util/options.ml
+++ b/sources/lib/util/options.ml
@@ -231,6 +231,9 @@ module M = struct
     else
       raise (Arg.Bad ("-steps-bound argument should be positive"))
 
+  let set_gc_policy n =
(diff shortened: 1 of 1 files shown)


OriginLabs-Iguernlala:
I'd say yes: 0 = next fit, 1 = first fit and 2 = best fit, or something like that

OCamlPro-Coquera:
> I don't think this is a good idea because if someday another allocation policy is implemented, it will not be usable because of this code.

If someday another policy is implemented we will need to update this code. This will forced us to test and benchs the new policy and maybe set it by default instead of just allowing it. This code warn the user that only 3 policy are available because we only test this 3. I don't want to allow things that we have not test.

PR #260, opened 2019-11-28
Patches for steps behavior

kanigsson:
These are our patches for improving alt-ergo steps behavior in some cases.

Note that this patch reintroduces the bug that we reported on issue #250 (same reproducer as on that issue, communicated privately), so likely some fix is needed.

### DIFF
--- a/sources/lib/reasoners/intervalCalculus.ml
+++ b/sources/lib/reasoners/intervalCalculus.ml
@@ -1494,10 +1494,10 @@ let assume ~query env uf la =
       (fun ((env, eqs, new_ineqs, rm) as acc) (a, root, expl, orig) ->
          let a = normal_form a in
          Debug.assume a expl;
+         Steps.incr (Interval_Calculus);
          try
            match a with
            | L.Builtin(_, ((L.LE | L.LT) as n), [r1;r2]) ->
-             Steps.incr (Builtin);
              incr nb_num;
              let p1 = poly_of r1 in
              let p2 = poly_of r2 in
--- a/sources/lib/reasoners/intervalCalculus.ml
+++ b/sources/lib/reasoners/intervalCalculus.ml
@@ -1497,6 +1497,7 @@ let assume ~query env uf la =
          try
            match a with
            | L.Builtin(_, ((L.LE | L.LT) as n), [r1;r2]) ->
+             Steps.incr (Builtin);
              incr nb_num;
              let p1 = poly_of r1 in
              let p2 = poly_of r2 in
--- a/sources/lib/util/steps.ml
+++ b/sources/lib/util/steps.ml
@@ -48,8 +48,8 @@ let incr k =
       naive_steps := !naive_steps + n;
   end;
   if steps_bound () <> -1
-  && ((Pervasives.compare !steps ((steps_bound ())) > 0)
-      || (Pervasives.compare !naive_steps ((steps_bound ())) > 0)) then
+  && ((Stdlib.compare !steps ((steps_bound ())) > 0)
+      || (Stdlib.compare !naive_steps ((steps_bound ())) > 0)) then
     begin
       printf "Steps limit reached: %d@."
         (if !naive_steps > 0 then !naive_steps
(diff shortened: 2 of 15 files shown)


sources/lib/reasoners/intervalCalculus.ml
```diff
@@ -1497,6 +1497,7 @@ let assume ~query env uf la =
          try
            match a with
            | L.Builtin(_, ((L.LE | L.LT) as n), [r1;r2]) ->
+             Steps.incr (Builtin);
```
Gbury: Is the `Builtin` step counter defined now ? i.e. do we have a specification as to what kinds of steps it counts ? because here it appears to deal with arithmetic ?

sources/lib/util/steps.ml
```diff
@@ -0,0 +1,73 @@
+open Options
+open Format
+
+type incr_kind = Matching | Omega | Fourier | Uf | Builtin | Ac | Naive of int
```
Gbury: Some internal explanations of what each step kind refers to would be useful, for instance what does `Naive 3` means ?

sources/lib/util/steps.ml
```diff
@@ -0,0 +1,73 @@
...
+let naive_steps = ref 0
+let steps = ref 0
+let mult_f = ref 0
+let mult_m = ref 0
+let mult_s = ref 0
+let mult_uf = ref 0
+let mult_b = ref 0
+let mult_a = ref 0
+
+let incr k =
```
Gbury: Maybe add a comment stating the the "mutlipliers" are here to homogeneize the global step counter by (strangely enough) dividing each step kind by its multiplier.

sources/lib/util/steps.ml
```diff
@@ -0,0 +1,73 @@
...
+  steps := 0;
+  mult_f := 0;
+  mult_m := 0;
+  mult_s := 0;
+  mult_uf := 0;
+  mult_b := 0;
+  mult_a := 0
+
+let get_steps () =
+  max !naive_steps !steps
```
Gbury: Why is there two global counters: the `steps` one and the `naive_steps` one ? A comment explaining the desing and goal/usefulness would be useful.

sources/lib/util/steps.mli
```diff
@@ -0,0 +1,12 @@
+type incr_kind = Matching | Omega | Fourier | Uf | Builtin | Ac | Naive of int
```
Gbury: These steps kind **must** be documented, ^^
(as a rule of thumb, any change/additions in interfaces should ensure that the modified parts have proper documentation)

sources/lib/reasoners/intervalCalculus.ml
OCamlPro-Coquera: This step counter is named `Builtin` because of the previous line (1499), it indeed deal with arithmetic. It can be rename to `Arith` if you prefere ?

Gbury: Well, first yes, a name more related to the fact that it deals with arithmetic would be better, but basically, it bugs me that we would have 3 different counters for arithmetic: Omega, Fourier, and builtin. Do we want/need to distinguish the three ? If so for what reasons (which could be spelled out in the doc of the steps module).

sources/lib/util/steps.ml
```diff
@@ -0,0 +1,89 @@
+open Options
```
Gbury: Missing header

sources/lib/util/steps.mli
```diff
@@ -0,0 +1,21 @@
+type incr_kind =
```
Gbury: There is a missing header for this file.
Also, all interface files should start with doc comments of the form:
```ocaml
(* ... header ..... *)

(** Module_Name

    Long description of the rationale and intent for the module, generic explanations about the way it behaves, about
    some of the design choices about it, etc...
*)
```

Also, some ocamldoc titles would be useful, e.g.:
```ocaml
(** {1 Steps counters} *)
```

sources/lib/reasoners/intervalCalculus.ml
OCamlPro-Coquera: We want to distinguish them because they have different multipliers.

PR #261, opened 2020-01-10
Fix issue with dynlink of parsers

OCamlPro-Coquera:


sources/Makefile
```diff
@@ -76,12 +76,16 @@ fm-simplex:
 	$(DUNE) build $(DUNE_FLAGS) \
 		_build/install/default/share/alt-ergo/plugins/fm-simplex-plugin.cma \
 		_build/install/default/share/alt-ergo/plugins/fm-simplex-plugin.cmxs
+		ln -sf _build/install/default/share/alt-ergo/plugins/fm-simplex-plugin.cma fm-simplex-plugin.cma
+		ln -sf _build/install/default/share/alt-ergo/plugins/fm-simplex-plugin.cmxs fm-simplex-plugin.cmxs
```
Gbury: Wrong indentation, it should start at the same column as the `$(DUNE)` a few lines above

sources/Makefile
```diff
@@ -76,12 +76,16 @@ fm-simplex:
...
 
 # Ab-Why3 plugin
 AB-Why3:
 	$(DUNE) build $(DUNE_FLAGS) \
 		_build/install/default/share/alt-ergo/plugins/AB-Why3-plugin.cma \
 		_build/install/default/share/alt-ergo/plugins/AB-Why3-plugin.cmxs
+		ln -sf _build/install/default/share/alt-ergo/plugins/AB-Why3-plugin.cma AB-Why3-plugin.cma
+		ln -sf _build/install/default/share/alt-ergo/plugins/AB-Why3-plugin.cmxs AB-Why3-plugin.cmxs
```
Gbury: Wrong indentation here also

PR #262, opened 2020-01-12
Add --static option to configure for statically linked compilation

OCamlPro-Coquera:


sources/configure.ml
```diff
@@ -116,6 +118,7 @@ let () =
   let () = Format.fprintf fmt "# Generated by configure@." in
   let () = Format.fprintf fmt "prefix=%s@." !prefix in
   let () = Format.fprintf fmt "libdir=%s@." !libdir in
+  let () = Format.fprintf fmt "profile=%s@." (if !static then "static-release" else "release") in
```
Gbury: With these modifications, the only two profile used will be `release`and `static-release`, ignoring the `dev` profile entirely, which is not optimal I think.

sources/configure.ml
```diff
@@ -15,6 +15,7 @@
 let prefix = ref ""
 let libdir = ref ""
 let mandir = ref ""
+let static = ref false
```
Gbury: I think it'd be better to have a variant `type profile = Dev | Release | Static` instead of a simple boolean (among other things to correctly allow to build in dev mode, see the other comment).

PR #268, opened 2020-01-21
Allowing Alt-Ergo to be compiled from 4.04.0 to 4.09.0

OCamlPro-mattiasdrp:


Gbury:
While you're at it, could you add Ci for versions of ocaml up to 4.09 ? (just look at the `.travis.yml` file at the root of the repo and copy-paste the few lines required)

PR #270, opened 2020-02-07
WIP: Options completely translated to cmdliner

OCamlPro-mattiasdrp:
This breaks the options syntax since Unix standard doesn't allow long names with one hyphen (-). All options are now preceded by two hyphens (--)


Gbury:
I didn't have ttime to review the whole patch yet, but with a quick look, there is a problem: the options are parsed and initialized in the initialization code of the `Options` module, which is wrong for a few reasons:
- it is a very very bad practice to perform important semantic actions in the initialization code of a module, and alt-ergo regularly suffers because of modules that does this (e.g. we saw a parser that parses an input file as part of its initialization code, and well it breaks as soon as we want to changes a few things).
- due to the current dependencies order, this means that some functor instantiations which use first class modules will be done before parsing the options, which is problematic.

OCamlPro-mattiasdrp:
I actually don't understand how these functors are allowed to be instantiated before the options are parsed. If they depend on certain options they should be instantiated after the module has been initialised. I changed it for now but it goes against cmdliner philosophy for me and it goes against my understanding of a command parsing module.

OCamlPro-mattiasdrp:
Question for the main users of alt-ergo: which options should be available with one letter only?

OCamlPro-Coquera:
> Question for the main users of alt-ergo: which options should be available with one letter only?

```
-h      --help
-v      --verbose
-d      --debug

-t n    --timelimit n

-g      --greedy

-s      --save-used-context
-r      --replay-used-context

-u      --unsat-core
-m      --all-models

-i      --default-lang
-o      --output-format (option to add)
```

OCamlPro-Coquera:
Add to Exit status :
 ```
 142 on timeout errors
 1 on errors
```

File: sources/lib/util/options.ml [hunk 1]
```diff
@@ -26,926 +26,1609 @@
...
+    profiling : bool;
+    profiling_period : float;
+    profiling_plugin : string;
+    verbose : bool;
+  }
+
+type quantifiers_opt =
+  {
+    arith_matching : bool;
+    disable_flat_formulas_simplification : bool;
```
OCamlPro-Coquera: Should be in sat_opt

sources/lib/util/options.ml
```diff
@@ -26,926 +26,1609 @@
...
+    timelimit : float;
+    timelimit_per_goal : bool;
+  }
+
+type output_opt =
+  {
+    all_models : bool;
+    complete_model : bool;
+    interpretation : int;
+    model : bool;
```
OCamlPro-Coquera: Could we rename models options into :
``` --model=default|all|complete ```

File: sources/lib/util/options.ml [hunk 3]
```diff
@@ -26,926 +26,1609 @@
...
+    unsat_core : bool;
+    unsat_mode : bool;
```
OCamlPro-Coquera: Replace this option with `-o -output-format=VAL` where VAL can be
- native (default)
- smtlib 
- and SZS for futur support of CASC/tptp

sources/lib/util/options.ml
```diff
@@ -26,926 +26,1609 @@
...
+
+  let docs = s_execution in
+
+  let default_input_lang =
+    let doc =
+      "Set the default input language to 'lang'. Useful when the extension \
+       does not allow to automatically select a parser (eg. JS mode, GUI \
+       mode, ...)" in
+    Arg.(value & opt string !vdefault_input_lang &
+         info ["default-lang"] ~docs ~doc) in
```
OCamlPro-Coquera: Replace with `-i --input-lang`

sources/lib/util/options.ml
```diff
@@ -26,926 +26,1609 @@
...
+         However, unit eliminations are always done" in
+    let docv = "VAL" in
+    Arg.(value & opt string dv & info ["fm-cross-limit"] ~docv ~docs ~doc) in
+
+  let interpretation_timelimit =
+    let doc = "Set the time limit to $(docv) seconds for model generation\
+               (not supported on Windows)." in
+    let docv = "SEC" in
+    Arg.(value & opt (some float) None &
+         info ["interpretation-timelimit"] ~docv ~docs ~doc) in
```
OCamlPro-Coquera: Maybe rename this options into `--timelimit-interpretation`

sources/lib/util/options.ml
OCamlPro-mattiasdrp: I don't see what you mean here ;-)

sources/parsers/psmt2_to_alt_ergo.ml
```diff
@@ -356,10 +356,10 @@ module Translate = struct
     | Cmd_Assert(assert_term) ->
       (translate_assert (pos command) assert_term) :: acc
     | Cmd_CheckEntailment(assert_term) ->
-      Options.set_unsat_mode false;
+      Options.set_output_mode Native;
```
OCamlPro-Coquera: I've made an error here, this should not set the output to Native. Could you change the code so that the Output mode should be done when we analyse the extension of the input file.
By default :
- .smt2 | .psmt2 -> Smtlib
- .why | .mlw | .ae -> Native

sources/lib/util/options.ml
```diff
@@ -26,926 +26,1613 @@
...
+    timelimit : float;
+    timelimit_per_goal : bool;
+  }
+
+type output_opt =
+  {
+    interpretation : int;
+    model : model;
+    unsat_core : bool;
+    unsat_mode : output;
```
OCamlPro-Coquera: Could you rename this with output_format ?

sources/lib/util/options.ml
```diff
@@ -26,926 +26,1613 @@
...
+    disable_weaks : bool;
+    enable_assertions : bool;
+    gc_policy : int;
+  }
+
+type limit_opt =
+  {
+    age_bound : int;
+    fm_cross_limit : Numbers.Q.t;
+    interpretation_timelimit : float;
```
OCamlPro-Coquera: Could you rename this with timelimit_interpretation

sources/lib/util/options.ml
```diff
@@ -26,926 +26,1613 @@
...
+    (* TODO : Link this to Alt-Ergo numbers *)
+    let dv = Numbers.Q.to_string !vfm_cross_limit in
+    let doc = Format.sprintf
+        "Skip Fourier-Motzkin variables elimination steps that may produce \
+         a number of inequalities that is greater than the given limit. \
+         However, unit eliminations are always done" in
+    let docv = "VAL" in
+    Arg.(value & opt string dv & info ["fm-cross-limit"] ~docv ~docs ~doc) in
+
+  let interpretation_timelimit =
```
OCamlPro-Coquera: use timilimit_interpretation instead of interpretation_timelimit

sources/lib/util/options.ml
```diff
@@ -26,926 +26,1613 @@
...
+       Unknown, before instantiation, or before every decision or \
+       instantiation. A negative value (-1, -2, or -3) will disable \
+       interpretation display. Note that $(b, --max-split) limitation will \
+       be ignored in model generation phase" in
+    let docv = "VAL" in
+    Arg.(value & opt int !vinterpretation &
+         info ["interpretation"] ~docv ~docs ~doc) in
+
+  let model =
+    let doc = "Experimental support for models on labeled terms" in
```
OCamlPro-Coquera: -Complete for complete model
-All for all model

sources/lib/util/options.ml
```diff
@@ -26,926 +26,1613 @@
...
+  Term.(ret (const mk_theory_opt $
+             disable_adts $ inequalities_plugin $ no_ac $ no_contracongru $
+             no_fm $ no_nla $ no_tcp $ no_theory $ restricted $
+             tighten_vars $ use_fpa
+            )
+       )
+let main =
+
+  let file =
+    let doc = "Source file. Must be suffixed by .mlw or .why" in
```
OCamlPro-Coquera: "... .why, .smt2 or psmt2"

sources/lib/util/options.mli
```diff
@@ -26,205 +26,729 @@
...
+val set_debug_use : bool -> unit
+
+(** Set {!val:profiling} *)
 val set_profiling : float -> bool -> unit
 
-(** additional setters *)
+(** Set {!val:timers} *)
+val set_timers : bool -> unit
+
+(** Additional setters *)
```
OCamlPro-Coquera: Sould be 
(** {3 Additional setters} *)

sources/lib/util/options.mli
```diff
@@ -26,205 +26,729 @@
...
+(** Default to [false] *)
+
+(** {3 Particular getters : functions that are immediately executed} *)
 
-(** particular getters : functions that are immediately executed **************)
 val exec_thread_yield : unit -> unit
 val exec_timeout : unit -> unit
 
+(** Displays the used rules *)
 val tool_req : int -> string -> unit
```
OCamlPro-Coquera: This is not a getter, and more of a usage function

PR #280, opened 2020-03-09
fix cmdliner dependency

OCamlPro-Coquera:


Gbury:
The fix is good, but it's somewhat annoying to have the lib depend on cmdliner...
@OCamlPro-mattiasdrp  how hard/annoying would it be to split the cmdliner term for options away from `options.ml` and into something like `tools/text/main_options.ml` ? I guess this would reduce safety becasue it would force all options to be settable, :/

OCamlPro-mattiasdrp:
Is it just for a dependency issue or for another reason? I actually don't know why the command line parsing was implemented in the library since a lib should be usable directly in your code so even in the previous version only the setters could be used.

PR #285, opened 2020-03-20
Improve how input file format is handle

OCamlPro-Coquera:


sources/lib/util/options.ml
```diff
@@ -132,12 +132,15 @@ let vno_backward = ref false
 let vno_sat_learning = ref false
 let vinstantiate_after_backjump = ref false
 let vdisable_weaks = ref false
-let vinput_format = ref "why"
 let vanswers_with_loc = ref true
 
-type output = ONative | OSmtlib | OSZS
+type input = INative | ISmtlib | IWhy3 (* | ISZS *)
```
Gbury: Since disambiguation of adt with OCaml, prefixing constructors name is not neded nor really useful, and I find it particularly annoying, so I'd much rpefer removing the first `I` (and simialry for the `O` in output formats).

sources/lib/util/options.ml
```diff
@@ -132,12 +132,15 @@ let vno_backward = ref false
...
+let vinfer_input_format = ref true
```
Gbury: I think it would be nice if we started documenting not only the interface of modules (for which I prasie this PR which adds the necessary doc comments, :) ), but also the inside of modules, for instance, by adding a comment stating what those reference are supposed to hold (semantics, goals, etc...). This would really immensely help future code modifications (by instance to spot bugs if the use of the ref doe snot correspond to its semantics/intended goal, etc..).

sources/lib/util/options.ml
Gbury: Additionally, it might be useful to have all type definitions at the beginning of a module rather than spread all around a file ?

Gbury: Also, I'd rather name this `input_format` to distinguish it from a representation of inputs file/pb (which might be : `type input = File of string | Stdin` for instance).

sources/lib/util/options.ml
```diff
@@ -381,7 +391,34 @@ let mk_execution_opt frontend input_format parse_only parsers
...
+      if not found then
+        let m = "Why3 format is only available when AB-Why3 parser is loaded"
+        in
+        raise (Error (false, m));
+      else
+        IWhy3
+    (*     | Some fmt when match_format fmt = "szs" -> ISZS *)
+    | Some fmt ->
+      let m = "Option --input does not accept the argument \"" ^ fmt in
+      raise (Error (false, m))
```
Gbury: Please, please use cmdliner's already existing mechanism for parsing arguments.
Basically, this would mean splitting this function in two: one for parsing strings into the `input` type, which ideally would be used only to create a cmdliner's argument converter (see https://erratique.ch/software/cmdliner/doc/Cmdliner.Arg.html#VALconv ), and another function, the setter for the argument which could check the availability of parsers.
Furthermore, use of the `Str` module is at best not recommanded or even deprecated, particularly in this case, where additionally trying to find a parser with "why3" in its name is a particularly atrocious and a brittle hack.
Overall, this brings about the following point:
- either we use a regular ADT to represent input formats, in which case it pretty much precludes the possibility of registering custom parsers for languages not in the definition of the ADT. In this case, why3 would be pretty much the only reasonable use of a custom parser (and we could maybe load it automatically ?)
- or, we use an extensible variant type for input_formats, but this brings about quite a lot of pain to the mechanism which would parse the input format argument string into this extensible variant
- or, if we really must allow and support custom parsers for custom formats (which I somehow doubt given the complexity of producing a parsed AST that would then be correctly typed), well... we can discuss other options.

sources/lib/util/options.ml
```diff
@@ -923,13 +966,13 @@ let parse_execution_opt =
...
-       mode, ...)." in
+    let doc = Format.sprintf
+        "Set the default input format to $(docv) and must be %s. \
+         Useful when the extension does not allow to automatically select \
+         a parser (eg. JS mode, GUI mode, ...)."
+        (Arg.doc_alts ["native"; "smtlib"; "why3"]) in
     let docv = "FMT" in
-    Arg.(value & opt string !vinput_format &
-         info ["i"; "input"] ~docv ~doc) in
+    Arg.(value & opt (some string) None & info ["i"; "input"] ~docv ~doc) in
```
Gbury: It would be really really really really ..... really better to use cmdliner's converter here rather than string.

sources/parsers/parsers.ml
```diff
@@ -132,10 +155,24 @@ let parse_input_file file =
...
+       (* | "szs" -> Options.set_output_format OSZS *)
+       | _ ->
+         if Options.infer_input_format () then
+           Format.eprintf "Warning: This extension is not supported@."
+         else if Options.input_native () then
+           Options.set_output_format ONative
+         else if Options.input_smtlib () then
+           Options.set_output_format OSmtlib
+         else if Options.input_why3 () then
+           Options.set_output_format OWhy3
```
Gbury: This nesting of if/then/elses seems really strange, why not pattern_match on the input_format since it exists somewhere ?

sources/lib/util/options.ml
```diff
@@ -34,6 +34,7 @@ exception Error of bool * string
 (* Declaration of all the options as refs with default values *)
 
 type model = MNone | MDefault | MAll | MComplete
+type in_out_format = Native | Smtlib | Why3 (* | SZS *)
```
Gbury: We should keep two different types: `input_format` and `output_format`.

sources/lib/util/options.ml
```diff
@@ -132,12 +133,18 @@ let vno_backward = ref false
 let vno_sat_learning = ref false
 let vinstantiate_after_backjump = ref false
 let vdisable_weaks = ref false
-let vinput_format = ref "why"
 let vanswers_with_loc = ref true
 
-type output = ONative | OSmtlib | OSZS
+(** Alt-Ergo infers automatically the input format according to the
+    file extension and set the corresponding parser. *)
+let vinfer_input_format = ref true
```
Gbury: Two things here:
- No need for doc comments in .ml files (i.e. use single `*` after the opening parenthesis).
- the comment as it is is not helpful from an implementation perspective because it doesn't answer the following question: what does `vinfer_input_format` does ? rather than a generic comment like that, something better would be:
`vinfer_input_format controls whether input format should be inferred by alt-ergo. If true, the vinput_format is basically ignored until format inferrence is done at which point it is set <I suppose ?>; if false, what happens ?(fill in what should actually happen / the semantic)`. Additionally, if possible some reference about where it is intended to be set and used would be interesting but not obligatory. And the same should be done for `vinput_format`.

sources/lib/util/options.ml
```diff
@@ -132,12 +133,18 @@ let vno_backward = ref false
...
+let vinput_format = ref Native
+
+(** Alt-Ergo infers automatically the output format according to the
+    file extension and set the corresponding output format. *)
```
Gbury: Same as above.

sources/lib/util/options.ml
Gbury: Basically, references and generally things which are done via side effects and whose changes should be synchronised should have documentation about what each state means, particularly corner cases where for instance, for some reasons (such as if specified by the user options on the command line), inference is activated **and** a specific input format is set: will it be ignored, will it supersede inference, etc...

sources/lib/util/options.ml
```diff
@@ -148,6 +155,13 @@ let vsave_used_context = ref false
 let vprofiling_period = ref 0.
 let vprofiling = ref false
 
+let match_format f =
+  match f with
+  | "native" | "ae" | "why" | "mlw" -> "native"
+  | "smtlib" | "smt" | "smt2" | "psmt" | "psmt2" -> "smtlib"
+  | "why3" | "Why3" -> "why3"
+  (* | "szs" | "SZS" -> "szs" *)
+  | s -> s
```
Gbury: With the split of type between input format and output format, this function will also need to be split, and in that case please differentiate the following things:
- the format itself
- the unique and canonic name of the a given input or output format (i.e. `smtlib2` (the `2` here is important because version 3 is coming with a change in semantics as I understand it), `tptp`, `alt-ergo`, etc..)
- the extension(s) supported by a given format.
- optionally (though I'd prefer if there was none), the aliases recognised to be the same as a format's canonic name.

sources/lib/util/options.ml
MonsieurPi: There's a button just above the comment box that allows you to make a suggestion that can immediately be committed (available with ctrl+g). Try to use it, it makes the review easier and faster to read ;-)

OCamlPro-mattiasdrp: There's a button just above the comment box available with ctrl+g as well which allows you to suggest a commit immediately. It makes the job easier for the person requesting your review and it's easier to read for other reviewers. Try to use it when you have a direct change in mind like this one where you provided what you think the comment should be.

sources/lib/util/options.ml
OCamlPro-Coquera: Since you don't want disambiguation such as `INative`, etc, and you style want two types you have this kind of errors :
`This expression has type output_format but an expression was expected of type input_format`

sources/lib/util/options.ml
Gbury: I'll keep that in mind for when it is applicable, but the problem is I don't know precisely what the message should be ! And that's exactly the problem: I don't know the semantics of these two references and how they are intended to interact with each other, and the current message does not answer these questions.

sources/lib/util/options.ml
Gbury: When you have these errors, you can add a few type hints here and there (usually, specifying the type of a function argument or of a match construct is more than enough).

sources/lib/util/options.ml
```diff
@@ -381,7 +420,24 @@ let mk_execution_opt frontend input_format parse_only parsers
...
+  let input_format : input_format = match input_format with
+    | None -> Native
+    | Some fmt ->
+      match match_input_format fmt with
+      | Native -> Native
+      | Smtlib2 -> Smtlib2
+      | Why3 ->
+        Format.eprintf "[warning] Why3 format is only available when \
+                        AB-Why3 parser is loaded@.";
+        Why3
```
Gbury: If I understand correctly, this warning will *always* appear even when the why3 parser has indeed been loaded, right ?
I think this is wrong and that a better behaviour would be the following: when the why3 parser is loaded no warning should appear, and when it isn't loaded, this should be an error.

sources/lib/util/options.ml
Gbury: Again, you should use cmdliner's convention to convert the argument into an `input_format`here instead of keeping a string and raison errors when you set the refs.

sources/lib/util/options.mli
```diff
@@ -507,8 +527,12 @@ val output_native : unit -> bool
 val output_smtlib : unit -> bool
 (** Default to [false] *)
 
+(** [true] if the output format is set to [Why3] *)
+val output_why3 : unit -> bool
+(** Default to [false] *)
```
Gbury: These functions (all of the `output_lang`) should disappear, and instead the call-sites should be replaced with a match (or an equality test) on the result of  `input_format ()`.

sources/parsers/parsers.ml
```diff
@@ -47,37 +47,63 @@ let register_parser ~lang new_parser =
...
+      with Not_found ->
+        debug_no_parser_for_input_format "smtlib";
+        exit 1
+    end
+  | Options.Why3 -> begin
+      try List.assoc ".why3" !parsers
+      with Not_found ->
+        debug_no_parser_for_input_format "why3";
+        exit 1
+    end
```
Gbury: There is basically the same code repeated thrice here, you should probably abstract that into a function of its own and call it instead of having copies of the same code, ^^

sources/parsers/parsers.ml
```diff
@@ -47,37 +47,63 @@ let register_parser ~lang new_parser =
...
-  let i = Options.input_format () in
   match lang_opt with
-  | Some lang ->
-    begin
-      try List.assoc lang !parsers
-      with Not_found ->
-        debug_no_parser_for_extension lang;
-        try List.assoc i !parsers
+  | Some lang -> begin
+      if Options.infer_input_format () then
```
Gbury: You should add a warning here I think, or at least warn when a user has specified both to infer the file format and a specific file format.

sources/lib/util/options.ml
```diff
@@ -148,6 +164,29 @@ let vsave_used_context = ref false
...
+  (* | "szs" | "SZS" -> "szs" *)
+  | s -> Unknown(s)
+
+let match_extension e =
+  match e with
+  | "ae" | "why" | "mlw" -> "native"
+  | "smt2" | "psmt2" -> "smtlib"
+  | "why3" -> "why3"
+  (* | "szs" | "SZS" -> "szs" *)
+  | s -> s
```
Gbury: This should return an `input_format` and not a string. Basically, if you later want to match on a value returned by a function (which is the case with this function), strings are a bad idea...

sources/parsers/parsers.ml
```diff
@@ -132,10 +158,25 @@ let parse_input_file file =
   in
   try
     (if Options.infer_output_format () then
-       match ext with
-       | ".smt2" | ".psmt2" -> Options.set_output_format OSmtlib
-       | ".why" | ".mlw" | ".ae" -> Options.set_output_format ONative
-       | _ -> Format.eprintf "Warning: This extension is not supported@."
+       let ext =
+         try Str.string_after ext 1
```
Gbury: The `Str` module is almost deprecated(slow, outdated, barely maintained, etc...) and absolutely not warranted in this case. Here, you can use `Filename.extension` (since ocaml >= 4.04), or simply use `String.rindex_from` followed by a `String.sub` to have greater compatibility across ocaml versions.

sources/lib/util/options.ml
OCamlPro-Coquera: That was in the first version, but since you didn't want me to use Str to search in the dynamically linked parser name "why3" i removed it.

sources/lib/util/options.mli
OCamlPro-Coquera: `output_format ()`*

sources/parsers/parsers.ml
OCamlPro-Coquera: I you had read carefully this PR you should know that input format inference is only done when no input format is set.

Gbury: The problem overall, is that there are way too much places when the inference flag is used to change behaviour, optimally there should only be one place where it is used.
Here, there seems to be a language provided so why is there a need to have different behaviour based on whether inference is activated ?

sources/parsers/parsers.ml
```diff
@@ -47,37 +47,63 @@ let register_parser ~lang new_parser =
     end;
   parsers := (lang, new_parser) :: !parsers
 
+let debug_no_parser_for_no_extension () =
+  if verbose() then
+    eprintf
+      "error: no extension found, can't infer input format@."
```
Gbury: Errors, and even more importantly, fatal errors (after which alt-ergo exits) shouldn't print only in verbose mode I think.

sources/lib/util/options.mli
OCamlPro-mattiasdrp: Well, it won't actually work if you want to do it outside of options since (=) is set to be of type int -> int -> int at the end of options.ml. Any equality test outside of options not involving integers won't work hence the need to define these flag getters here

sources/parsers/parsers.ml
```diff
@@ -47,37 +47,63 @@ let register_parser ~lang new_parser =
...
-      try List.assoc i !parsers
-      with Not_found ->
-        debug_no_parser_for_input_format i;
-        exit 1
-    end
+    if Options.infer_input_format () then begin
+      debug_no_parser_for_no_extension ();
+      exit 1 end
+    else
+      get_input_parser ()
```
Gbury: This whole block seems very strange..; the way I read it:

if no language is provided (because we matched `None`), then, if inference is activated, print an error (if verbose, which is a problem, the error should always be printed since we exit right after..., additionally, using exit directly is **BAD**, this should be an exception catched at top-level), else, if no inference is actiavted (and we still don't have a language), try and get the input parser ?

 Either the code is wrong, or there should be a lot more documentation around here to explain what is going on and what this should be doing.

sources/lib/util/options.ml
Gbury: Well, altogether this strongly suggests that here (during parsing of options), is a really bad place to emit a warning like this...
This warning should rather be emitted in the parsers file, where we can know whether we have a specific parser for why3 (coming from the AB-why3 plugin or another provided by the user), or if we don't have one, and we instead try and fallback on the native format parser, while emitting a warning that it might be an error (if the provided file is in why3 and not alt-ergo native format), and that anyway the extension has been deprecated for alt-ergo native files.

sources/lib/util/options.mli
Gbury: Well, in that case there are two good solutions (compared to the bad one show here):
- instead of using esuality, always use a match on the `output_format` which actually is reasonable and can help greatly future-proofing the code, using the exhaustive match check of the compiler.
- or, the option module can export a comparison and/or equality function on types where it might be useful

OCamlPro-mattiasdrp: I'm actually curious to know why this one is "bad" since we're just hiding behind functions the kind of outputs that are available and the way they're implemented. Your first solution would actually make the implementation visible everywhere and any change of it would mean changing it in every file while here it means only changing 3 functions. Your second option has the same problem. I kind of like the actual solution (which was implemented a long time ago)

Gbury: Well because basically, everywhere you use one of these, you should actually do something different, but still something for each language: basically, the output format (resp. the input format), should **only** be used to print (resp. parse/typecheck) differently based on the format choosen. Other considerations should have a dedicated logic (which can use the formats to decide some things), which has a name, documentation, and specified semantics, otherwise it's an unbeliveable mess to try and understand let alone modify code. This is the case currently, but it would be good if we could fix that when we touch the relevant code.

OCamlPro-mattiasdrp: Ah well, this has then nothing to do with the current PR and it would add noise for nothing then.

Gbury: Well, the current PR is a good time to fix that part and we should absolutely do it now, else we won't do it. Basically every time we touch a piece of code, if the modification is not time sensitive (such as fixing a soundness bug), we should clean up the related part of the code, else we'll be stuck with atrocious, unbearable code on which every modification is a nightmare because there is no doc or clear semantics.

For instance, imagine when we will add smtlib v3 which should release soon if I recall correctly. I expect the semantics to really not differ much between smtlib2 and smtlib3, at least on the fragment which is handled by alt-ergo. So basically, with things the way they are, everywhere there is a `output_smtlib2 ()`, one should add a `|| output_smtlib3 ()`, which is not only tiresome but also error prone, sinc one could very easily miss one such occurrence, and review on the github API will easily miss such missing occurrences. With pattern matches, it becomes a lot easier to spot everywhere this need to be updated.

sources/parsers/parsers.ml
OCamlPro-Coquera: Too much places, how's that ?

Gbury: There are at least three invocations of `Options.infer_input_format` in this file (two in this function, and one later): basically, I'm saying this is two more than should be. We need one such call at the place where we decide which parser should be used, but I don't see why we would need more than that one and only call. And neither the code nor the (lack of) documentation around explains these extra calls.

sources/parsers/parsers.ml
```diff
@@ -47,38 +49,36 @@ let register_parser ~lang new_parser =
...
-    eprintf
-      "error: no parser registered for the provided input format %S ?@." s
+let find_parser ext format =
+  try List.assoc ext !parsers
+  with Not_found ->
+    raise (ParserError ("Error: no parser registered for the provided \
+                         input format %S ?@."^format))
+
+let set_output_format f =
+  if Options.infer_output_format () then
```
Gbury: Actually one question: how is the output inference supposed to work ?

Gbury: And why is it relevant in the parsers ?

sources/parsers/parsers.ml
```diff
@@ -47,38 +49,36 @@ let register_parser ~lang new_parser =
...
+
+let get_input_parser f =
+  set_output_format f;
+  match f with
+  | Options.Native -> find_parser ".ae" "native"
+  | Options.Smtlib2 -> find_parser ".smt2" "smtlib2"
+  | Options.Why3 -> find_parser ".why3" "why3"
+  | Options.Unknown s -> find_parser s s
 
 let get_parser lang_opt =
```
Gbury: could you rename this argument `ext_opt` or something similar, this name is actually very confusing, and all the more with the current PR which introduces a real notion of language (aka format).

sources/parsers/parsers.ml
OCamlPro-Coquera: If output format is not set with the `-o` option then we try to infer the output format accordingly with the input format. If `-o` is not set the flag infer_output_format is true. Then we match the format f, if it's a known format, we set the output format to this format f. If not, we print a message for the user to warn him about the fact that the output format is not known. Since this function is called in get_input_parser, the output inference match the input. If no input is set with `i-` option we try to find a format from file extension, this format will be used in both input and output format. Same if `-i` is set. So by default if you set an input format we will print result according this input format.

sources/parsers/parsers.ml
OCamlPro-Coquera: I think Github UI does not point to the right place ? I rename `f` in `fmt` since this is a format and `ext` in `ext_opt` in `find_parser`

Gbury: I meant the argument `lang_opt` of `get_parser`, ^^

sources/parsers/parsers.ml
Gbury: Well, that might be an interesting thing to document somewhere, :)

sources/parsers/parsers.ml
OCamlPro-Coquera: AH ok ^^, done

sources/parsers/parsers.ml
OCamlPro-Coquera: I've documented it a bit more in options. but since it's done in parser i will add some few line in parsers too

sources/lib/util/options.ml
OCamlPro-Coquera: Since .why and .mlw are set to depreciated, i've remove them from known extension of Alt-Ergo native parser. If this extension are found the match_extension will return the Why3 format. At parsing, if no parser exist for why3 we will print a warning telling user that they should use the AB-Why3 plugin and this extension are depreciated, and then continue with native parser. Does this sound good for you ?

Gbury: Sounds good indeed.

sources/lib/util/options.mli
OCamlPro-Coquera: Done

sources/lib/util/options.ml
```diff
@@ -1080,24 +1151,26 @@ let parse_output_opt =
...
-    let doc = Format.sprintf
+    let doc =
+      Format.sprintf
         "Answer unsat/sat/unknown instead of Valid/Invalid/I don't know. \
-         $(docv) must be %s. It must be noticed that not specifying an \
-         output format will let Alt-Ergo set it according to the file's \
-         extension."
-        (Arg.doc_alts ["native"; "smtlib"]) in
+         $(docv) must be %s. It must be noticed that not specifying an output \
+         format will let Alt-Ergo set it according to the file's extension."
```
Gbury:
```suggestion
         format will let Alt-Ergo set it according to the input file's extension."
```

sources/lib/frontend/frontend.ml
```diff
@@ -330,9 +330,10 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
         report_loc Loc.dummy time steps
 
   let print_status status steps =
-    if Options.output_smtlib () then print_status_output_smtlib status steps
-    else if Options.output_native () then print_status_valid_mode status steps
-    else assert false           (* will be useful for szs *)
+    let open Options in
```
Gbury: you don't need this module opening I think, the type inference is smart enough to know the return type of `Options.output_format ()` if I recall correctly.

sources/lib/structures/errors.ml
```diff
@@ -185,7 +185,7 @@ let report fmt = function
 
 let error e l = raise (Error(e,l))
 let warning e l =
-  if Options.output_smtlib () then Format.fprintf err_formatter "; ";
+  if Options.output_format () = Options.Smtlib2 then Format.fprintf err_formatter "; ";
```
Gbury: As mentionned by @OCamlPro-mattiasdrp this shouldn't be allowed, better to match on the input format, and possibly add a comment specifying why we need to add this in the case of smtlib2 I guess it's for the location to be in a commented line ?

OCamlPro-Coquera: Done

sources/lib/frontend/frontend.ml
OCamlPro-Coquera: Done

PR #286, opened 2020-03-23
2.3.2 minor

OCamlPro-Coquera:
This minor release will resolve this following issues :

  -  Fix preludes installation
  -  Fix altgr-ergo building rule
  -  Fix issue with dynlink of parsers


Gbury:
Minor note: you could add a `dev` mention on the change line for the build rule of altgr-ergo (it only affected dev who use the makefile, since the installation process does not use the makefile).

PR #287, opened 2020-03-23
New build system with proper dune-project and tree structure

OCamlPro-mattiasdrp:
I'd like you to try the new "make fmt" which allows to format the code with ocamlformat. It's the biggest change in this PR.


configure.ml
```diff
@@ -55,70 +58,70 @@ let opam_var v =
   let _ = Unix.close_process_in ch in
   s
 
+let concat = Filename.concat
```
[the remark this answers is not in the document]
OCamlPro-mattiasdrp: > Having generated files checked into the repository is really a sure way to have them out-of-sync at one time, but theses files have to be part of the repo in order to pin the repo. At the very least, a CI test checking that they are in sync is needed.

I don't understand this part.

> while I'm at it, for any PR, please separate commit with semantics changes from commits which reformats code, especially on large scale, because else the review becomes almost impossible

That's not the case in this PR, no?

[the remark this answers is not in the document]
OCamlPro-mattiasdrp: Forgot about the alias, I used it to test something and forgot to revert it.

dune-project
```diff
@@ -0,0 +1,113 @@
+(lang dune 2.4)
+
+; Since we want to generate opam files we need to provide informations ;
```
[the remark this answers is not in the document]
OCamlPro-mattiasdrp: Yes we do, I think, I personally prefer when all the modifications are possible in one file for all the other files.

As for ocamlformat, it's just available doing "make fmt", it's not mandatory and it's just here as a feature I think we should test for a while and see what's possible with it and what's not.

configure.ml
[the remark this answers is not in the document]
Gbury: > > Having generated files checked into the repository is really a sure way to have them out-of-sync at one time, but theses files have to be part of the repo in order to pin the repo. At the very least, a CI test checking that they are in sync is needed.
> 
> I don't understand this part.
> 

When files that are generated by some command, are also committed into the repo, you can be absolutely 100% certain that at some point someone will modify the `dune-project`, commit that change, forget to re-generate the `.opam` files, and push the change to the `dune-project` file without the associated changes that should also occur in the `opam` files, at which point the files are out-of-sync on the repo, and it will introduce strnage changes in unrelated commits... Basically commiting generated file is **BAD** and dangerous in general, and we definitely need to have the .opam files committed currently.

> > while I'm at it, for any PR, please separate commit with semantics changes from commits which reformats code, especially on large scale, because else the review becomes almost impossible
> 
> That's not the case in this PR, no?

Well, in this PR you have at least three changes:
- some moving around of files, and some necessary changes associated to these moves (and these moves **only**)
- some changes to files in order to change some things (these are not strictly related to the move of files)
- some files have been modified because of the ocamlformat reformatting

What I'm saying in that one should make sure in PRs such as this that these three kind of changes in are separate commits in order to make the PR reviewable. Because currently there are hundreds of files modified, most of which should only contain formatting changes (i.e. no semantic changes) which needn't be reviewed with much attention if we trust ocamlformat, but there are also other changes (for instance the changes to the .opam file build intructions) that have to be reviewed carefully. The problem is that in its current state, this is very complicated to do because all changes are thrown in together and it's not easy to spot which change is which....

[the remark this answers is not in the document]
Gbury: Okay so it seems there are no ocamlformat re-formatting, i just thought so because of all the unnecessary changes to the configure script.
It might be useful to have more accurate and complete description of the changes either in a PR description (or detailed in commit messages), ^^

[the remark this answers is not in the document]
OCamlPro-mattiasdrp: > When files that are generated by some command, are also committed into the repo, you can be absolutely 100% certain that at some point someone will modify the dune-project, commit that change, forget to re-generate the .opam files, and push the change to the dune-project file without the associated changes that should also occur in the opam files, at which point the files are out-of-sync on the repo, and it will introduce strnage changes in unrelated commits... Basically commiting generated file is BAD and dangerous in general, and we definitely need to have the .opam files committed currently.

Ah, yes, I agree. 

No files have been changed by ocamlformat so I disagree with the rest of your message.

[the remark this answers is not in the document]
OCamlPro-mattiasdrp: Yep, I should put a better commit message, as of now it's just a work in progress on which I didn't ask for a review, I just wanted you to notice I started working on it, don't worry, when I'll need you it'll be well documented ;-)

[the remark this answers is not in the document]
Gbury: Well, currently, the diff of the PR makes it quite hard to see that you actually broke the opam file (the missing configure step), because with the move and the generation of files combined, you don't see, nor usually review the diff of the `.opam` files because it doesn't appear as a diff (you see one file removed and one file added, but not them side by side), and it is easy to miss them when reviewing. Whereas, if the move of files we in a commit of itw own, and the addition of dune-generated files was in another commit of its own, it would be much much much easier to review.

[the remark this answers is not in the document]
OCamlPro-mattiasdrp: I'm pretty sure one much is enough. ;-)
And as I already said, once everything is done I will commit the changes group by group, don't start to review them now, it's just a loss of time ;-)

dune-project
```diff
@@ -0,0 +1,113 @@
...
+(package
+ (name alt-ergo-parsers)
+ (synopsis "The Alt-Ergo SMT prover parser library")
+ (description "This is the parser library used in the Alt-Ergo SMT solver.
+
+Alt-Ergo is an automatic theorem prover of mathematical formulas. It was developed at LRI, and is now maintained at OCamlPro.
+
+See more details on http://alt-ergo.ocamlpro.com/"
+ )
+ (license "OCamlPro Non-Commercial Purpose License, version 1, LGPL-2.1-only")
```
OCamlPro-Coquera: Only the Package Alt-Ergo have both OCP and LGPL licences due to the AB-Why3 plugin.

OCamlPro-Coquera:
```suggestion
 (license "OCamlPro Non-Commercial Purpose License")
```

dune-project
```diff
@@ -0,0 +1,113 @@
...
+(package
+ (name alt-ergo-lib)
+ (synopsis "The Alt-Ergo SMT prover library")
+ (description "This is the core library used in the Alt-Ergo SMT solver.
+
+Alt-Ergo is an automatic theorem prover of mathematical formulas. It was developed at LRI, and is now maintained at OCamlPro.
+
+See more details on http://alt-ergo.ocamlpro.com/"
+ )
+ (license "OCamlPro Non-Commercial Purpose License, version 1, LGPL-2.1-only")
```
OCamlPro-Coquera:
```suggestion
 (license "OCamlPro Non-Commercial Purpose License, version 1")
```

dune-project
```diff
@@ -0,0 +1,113 @@
...
+(package
+ (name altgr-ergo)
+ (synopsis "The GUI for the Alt-Ergo SMT prover")
+ (description "Altgr-Ergo is the graphical interface for the Alt-Ergo SMT prover.
+
+Alt-Ergo is an automatic theorem prover of mathematical formulas. It was developed at LRI, and is now maintained at OCamlPro.
+
+See more details on http://alt-ergo.ocamlpro.com/"
+ )
+ (license "OCamlPro Non-Commercial Purpose License, version 1, LGPL-2.1-only")
```
OCamlPro-Coquera:
```suggestion
 (license "OCamlPro Non-Commercial Purpose License, version 1")
```

src/lib/structures/expr.ml
```diff
@@ -2037,6 +2037,7 @@ module Triggers = struct
     | _ -> false
 
   let term_definition s e =
+    (* eprintf "String : %s, Expr : %a@." s print e; *)
```
OCamlPro-Coquera: This is not part of this PR ?

OCamlPro-Coquera:
> I'd like you to try the new "make fmt" which allows to format the code with ocamlformat. It's the biggest change in this PR.

I like the use of ocamlformat. It will help us to have a standardised code like we tried to have previously with ocp-indent and the style checker. Could you add a check for ocamlformat in the CI ?

sources/dune
```diff
@@ -1,9 +0,0 @@
-(env
- (dev
-  (flags (:standard -bin-annot -w -22))
-  (ocamlopt_flags -g -O3 -unbox-closures))
- (release
-  (flags (:standard -bin-annot -w -22))
-  (ocamlopt_flags -O3 -unbox-closures))
-)
-
```
Gbury: This file was removed, but I don't see its contents added anywhere...

src/lib/structures/expr.ml
OCamlPro-Coquera:
```suggestion
```

src/lib/structures/expr.ml
```diff
@@ -2049,6 +2050,7 @@ module Triggers = struct
       if head_is_name s e then e
       else
         let s = TSet.filter (head_is_name s) (max_pure_subterms e) in
+        eprintf "List : %a@." print_list (TSet.elements s);
```
OCamlPro-Coquera:
```suggestion
```

src/lib/structures/expr.ml
```diff
@@ -2098,13 +2100,19 @@ module Triggers = struct
...
+            let () = eprintf "e1 : %a@.e2 : %a@." print e1 print e2 in
             if equal e e1 then e2 else if equal e e2 then e1 else f
           | _ -> f
         in
+        let () = eprintf "defn : %a@." print defn in
         let tt = max_terms defn e in
+        List.iter (eprintf "tt : %a@." print) tt;
+
         let tt = List.fast_sort (fun a b -> depth b - depth a) tt in
         filter_good_triggers (vterm, vtype) @@ triggers_of_list [[e]; tt]
```
OCamlPro-Coquera:
```suggestion
      | (Dpredicate s | Dfunction s), _ ->
        let e = term_definition s f in
        let defn = match f with
          | { f = (Sy.Form Sy.F_Iff | Sy.Lit Sy.L_eq) ; xs = [e1; e2]; _ } ->
            if equal e e1 then e2 else if equal e e2 then e1 else f
          | _ -> f
        in
        let tt = max_terms defn e in
        let tt = List.fast_sort (fun a b -> depth b - depth a) tt in
        filter_good_triggers (vterm, vtype) @@ triggers_of_list [[e]; tt]
```

.travis.yml
```diff
@@ -46,7 +46,7 @@ before_install:
 
 # Install dependencies
 install:
-  - opam pin add --no-action ./sources
+  - opam pin add --no-action ./src
```
Gbury: Isn't the goal of this PR that the root of the repo can be pinned instead of the source directory ?

OCamlPro-Coquera:
```suggestion
  - opam pin add --no-action .
```

alt-ergo-lib.opam.template
```diff
@@ -0,0 +1,8 @@
+# This part comes from the template. Please edit alt-ergo.opam.template
+# and not alt-ergo-lib.opam which is generated by dune
+
+build: [
+  ["./configure" name]
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
```
Gbury, line 7: You should keep the in this template the part the dune auto-generated in the commits before, that dealt with the `{with-doc}` part at least I think (the test part is much less useful because we do not run the tests through dune and we'd need to discuss whether the non-regressions tests should be run by opam).

configure.ml
Gbury, line 61: Could you remove this unneeded alias ?

configure.ml
```diff
@@ -55,70 +58,70 @@ let opam_var v =
...
-      if prefix_set
-      then Filename.concat !prefix "man"
-      else opam_var "man");
+      if prefix_set then concat !prefix "man" else opam_var "man");
   ()
 
 (* Output config options into lib/util/config.ml *)
 let () =
-  let f = Filename.concat (Filename.concat "lib" "util") "config.ml" in
+  let f = concat "src" @@ concat "lib" @@ concat "util" "config.ml" in
```
Gbury, line 75: This change is a bit gratuitous I think, :/

configure.ml
```diff
@@ -55,70 +58,70 @@ let opam_var v =
...
+       abs_exe_path@,\
+       |> Filename.dirname@,\
+       |> follow Filename.parent_dir_name@,\
+       |> follow \"share\"@,\
+       |> follow \"alt-ergo\"@]@,\
+       @,\
+       let pluginsdir = datadir |> follow \"plugins\"@,\
+       @,\
+       let preludesdir = datadir |> follow \"preludes\"@]@."
+  in
```
Gbury, line 110: Please revert his change, it makes the printed code much much less readable for no real improvement.

dune-project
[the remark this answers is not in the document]
Gbury: Well, now that we have both the dun-project file and the template files, we technically have one more file than before but anyway, :p

pre-commit
```diff
@@ -0,0 +1,14 @@
...
+CHANGED=$?
+
+if [ $CHANGED ]; then
+    echo
+    echo -e "\e[41m!!! dune-project has changed !!!\033[0m"
+    echo
+    echo "You should apply"
+    echo "make"
+    echo "to ensure all the .opam files are changed accordingly"
+fi
```
Gbury, line 14: As it is this check is not enough and particularly annoying. It will often print false alert/warnings, and never really prevent bad commits (sinc it never exits with exit code different from 0).
Waht it should rather do is if the dune-project file has changed, then check whether freshly generated `.opam` files have a diff with the current one, and exit if the diff exists.

rsc/extra/check_makefile.sh
```diff
@@ -10,7 +10,7 @@ git_repo=`git rev-parse --show-toplevel`
 local_install_dir=$1
 
 # Cd into the source directory
```
OCamlPro-Coquera, line 12: Already in project root directory, this command is maybe not needed anymore. 
We can let it here, just to be sure and change comment
```suggestion
# Cd into the root project directory
```

PR #297, opened 2020-04-08
Add contributions guidelines section

OCamlPro-Coquera:


CONTRIBUTING.md
```diff
@@ -0,0 +1,23 @@
...
+The following terms apply to all such contributions:
+
+* We require all pull requests to pass :
+
+	* Our non regression suite, you can run `make non-regression` to check if your changes affect the performance and answers of Alt-Ergo.
+
+	* Our style checking, every lines should length 80 columns maximum and should be indent with ocp-indent.
+	  We strongly recommand to use the following command line so the indent and style checking will be done automatically when commiting.
+
+	```cp rsc/extra/pre-commit-git-hook .git/hooks/pre-commit```
```
Gbury, line 15: Note, this script bugs when using git worktrees, maybe we should fix that...

CONTRIBUTING.md
```diff
@@ -0,0 +1,23 @@
...
+
+* Pull request should be rebased and up to date from the `base` branch to be merged.
```
Gbury, line 17: This is quite ambiguous, people could try and look for a branch named `base` to make PRs against. One way to avoid the ambiguity is to introduce the concept of the `base`branch in the first paragraph when you specify that contributions are welcome as PRs, agains some base branch (which would typically be `next`, some particular dev branch, or in rare cases `master`).

CONTRIBUTING.md
```diff
@@ -0,0 +1,23 @@
...
+
+* All commits will be squashed into one when merged into the `base` branch. A PR with only one commit is appreciated since it facilitates the review.
```
Gbury, line 19: I don't really agree with this comment. Some changes are way way way better reviewed if commits are clean and each do a specified task. So while specifying the commits will be squashed when the PR is merged is fine, I wouldn't recommand PR to come as single commits. Rather Prs should come with something like a clean commit history, with each commit properly named and annotated with a message.

CONTRIBUTING.md
OCamlPro-Coquera: OK, i will wait for your suggestion and commit it then, thanks.

CONTRIBUTING.md
OCamlPro-Coquera: Same as above, thanks

PR #299, opened 2020-04-14
Build js

ACoquereau:
You can test the different bulding rules :
`make js-node` to build a node js compatible version and then run `node alt-ergo.js <options> <input_file>`
`make js-example` to build the simple js example using the web worker. In www you will find a minimal webpage. You may need to run `python3 -m http.server 8080` in www for symlink and then open localhost:8080 in your browser.

This PR merge some files in the common internal library. The alt-ergo-js opam package make 3 different binaries :
1) alt-ergo-js-node compatible with node js
2) alt-ergo-js-worker build a web worker running Alt-Ergo
3) alt-ergo-js-example build a small Js example using the Alt-Ergo webworker


src/bin/text/dune
```diff
@@ -10,6 +10,7 @@
   (libraries    alt-ergo-lib alt-ergo-parsers stdlib-shims)
   (link_flags   (:include flags.dune))
   (modules      Main_input Main_text)
+  (modes byte native)
```
Gbury, line 13: I think we'd prefer `(modes best)` here

src/bin/js/missing_primitives.js
```diff
@@ -0,0 +1,31 @@
...
+//Provides: unix_times
+//Requires: unix_gettimeofday
+function unix_times () {
+	var utime = unix_gettimeofday ();
+	return BLOCK(0, utime, utime, utime, utime)
+}
+//Provides: unix_setitimer
+function unix_setitimer () {
+  return BLOCK(0, 0, 0, 0)
+}
```
Gbury, line 13: I think we'd like to have a real `setitimer` implementation to be able to have timeouts in the js version, if it's reasonably possible ?

OCamlPro-Coquera:
> I think the best solution would be the first (i.e. `make js`) one, but using symlinks for the copied files, that way there's no worry about them going out-of-sync.
> 
> Also, we'd need to test the js version in the CI pipeline (both that it builds and that it works (e.g. by running the non-regression tests)).

Symlinks is a good idea, i was unable to build with symlinks during my tests (failed for unknown reason) i will look into it.

I agree with the test of the js version in the CI. Testing build will be easy, but to test that it works we will need something like node-js installed in the CI to run the js. Do you know how to add this kind of tool ?

src/bin/js/missing_primitives.js
OCamlPro-Coquera: I agree, i don't know if it's possible but i will look into it

src/bin/text/dune
OCamlPro-Coquera: Done

src/bin/gui/dune
```diff
@@ -12,6 +12,7 @@
   (modules      Gui_session Gui_config Annoted_ast Connected_ast
                 Gui_replay Main_gui)
   (link_flags   (-linkall))
+  (modes (best exe))
```
zapashcanon, line 15: IMHO this would be better to juste remove this.

src/bin/text/dune
```diff
@@ -10,6 +10,7 @@
   (libraries    alt-ergo-lib alt-ergo-parsers stdlib-shims)
   (link_flags   (:include flags.dune))
   (modules      Main_input Main_text)
+  (modes (best exe))
```
zapashcanon, line 13: Same here, it would be better to just let dune handle this.

rsc/extra/check_js_test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -e
```
zapashcanon, line 1: I have two little suggestions. :D

First is to use:

`#!/bin/sh` instead of `#!/bin/bash`

The second is to replace:

`#!/bin/bash -e` by

```sh
#!/bin/bash

set -eu
```

So that, one can launch the script with e.g. `sh script.sh` and still get the `-e` behavior. Also with adding `-u`, the script will fail on uninitialized variables.

rsc/extra/check_js_test.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Check that the makefile targets work
+echo "=+= [check_makefile.sh] building ... =+="
+opam install zarith_stubs_js
+make js
+
+# Run the tests
+cd $git_repo/non-regression
+sh ./run_valid.sh "nodejs ../alt-ergo-js" "2"
+sh ./run_invalid.sh "nodejs ../alt-ergo-js" "2"
```
zapashcanon, line 26: The `sh` is not needed in these two lines.

Makefile
```diff
@@ -273,6 +273,21 @@ html: doc
...
+# ========================
+# Javascript building rule
+# ========================
+
+# Build alt-ergo-js (dev)
+# zarith_stubs_js package is needed for this rule
+# note that --timeout option is unavailable due to the lack of js primitives
+# use of input zip file is also unavailable
+js: gen
+	$(DUNE) build $(DUNE_FLAGS) _build/default/src/bin/text/Main_js.bc.js
```
zapashcanon, line 285: This can just be `$(DUNE) build $(DUNE_FLAGS) src/bin/text/Main_js.bc.js`

Makefile
```diff
@@ -273,6 +273,21 @@ html: doc
...
+	ln -sf _build/default/src/bin/text/Main_js.bc.js alt-ergo-js
+
+alt_js: gen
+	$(DUNE) build $(DUNE_FLAGS) _build/default/src/bin/js/Main_text.bc.js
```
zapashcanon, line 289: This can just be `$(DUNE) build $(DUNE_FLAGS) src/bin/js/Main_text.bc.js`

.travis.yml
```diff
@@ -20,6 +20,7 @@ addons:
       - zlib1g-dev
       - libgtk2.0-dev
       - libexpat1-dev
+      - nodejs
```
Gbury, line 23: It actually might be better to change the `language` at the beginning of the travis file, rather than install node for each run: indeed, currently there is no builtin support for ocaml as a language in travis, so the language used is C because it's the most lightweight, but now that we test js, we might as well use its native support.

.travis.yml
```diff
@@ -80,6 +81,11 @@ jobs:
       env: OPAMBUILDTEST=false OCAML_VERSION=4.04.2
 
     # === Build Tests ===
+    # Javascript (no tests)
+    - stage: "test"
+      name: "js (no tests)"
+      script: rsc/extra/check_js_test.sh
+      env: OPAMBUILDTEST=false OCAML_VERSION=4.09.0
```
Gbury, line 88: It would be better to put that test at the end, rather than at the start of the `test`stage (jobs are executed in the order specified iirc, and we'll likely want to test the regular builds first).

src/bin/text/main_js.ml
```diff
@@ -0,0 +1,14 @@
...
+(*     Alt-Ergo: The SMT Solver For Software Verification                     *)
+(*     Copyright (C) 2018-2018 --- OCamlPro SAS                               *)
+(*                                                                            *)
+(*     This file is distributed under the terms of the license indicated      *)
+(*     in the file 'License.OCamlPro'. If 'License.OCamlPro' is not           *)
+(*     present, please contact us to clarify licensing.                       *)
+(*                                                                            *)
+(******************************************************************************)
+
+include Common
```
Gbury, line 12: Why the include ? For a one-line/one-function-call file, it's better to simply use the fully qualified name.

src/lib/util/options.ml
```diff
@@ -545,7 +545,7 @@ let mk_sat_opt bottom_classes disable_flat_formulas_simplification
          match s with
          | "" -> set
          | s -> Util.SS.add s set
-      ) Util.SS.empty (Str.split (Str.regexp ",") ndo)
+      ) Util.SS.empty (Re.Str.split (Re.Str.regexp ",") ndo)
```
Gbury, line 548: Rather than use Re, which is a bit much for simply splitting a string on characters, why not use `String.split_on_char` (or re-implement it, given that it's pretty simple).

.travis.yml
OCamlPro-Coquera: Ok, i hope this solution does not add too much weight in the CI

.travis.yml
OCamlPro-Coquera: I've put it at the beginning since it's only in dev/test phase, i'll of course put it at the end when i'ill have a working script.

src/bin/text/dune
```diff
@@ -28,6 +28,29 @@
   )
 )
 
+(executable
+  (name         Worker)
```
zapashcanon, line 32: Do you think you could just use lowercase everywhere ? Otherwise [terrible things](https://github.com/BinaryAnalysisPlatform/core-lwt/pull/5) are going to happen... :p (My link is juste about using `_` or `-` but same goes for lower/uppercase.

src/bin/text/dune
```diff
@@ -28,6 +28,29 @@
...
+  (libraries    common zarith_stubs_js js_of_ocaml js_of_ocaml-compiler lwt js_of_ocaml-lwt
+)
+  (modules      Worker)
```
zapashcanon, line 35: here too

src/bin/text/dune
```diff
@@ -28,6 +28,29 @@
...
+  (modes        js)
+  (preprocess (pps js_of_ocaml-ppx lwt_ppx))
+  (js_of_ocaml
+    (flags --no-source-map --linkall --pretty --disable genprim)
+  )
+)
+
+(executable
+  (name         Main)
```
zapashcanon, line 44: here too

src/bin/text/dune
OCamlPro-Coquera: Yep it's still wip, i'll clean this at the end

zapashcanon: OK. :)

src/bin/js/missing_primitives.js
OCamlPro-Coquera: Setting a timeout will be possible at top level see the `main_js.ml` file.

src/bin/js/index_js.mld
```diff
@@ -0,0 +1,6 @@
+{1 Alt-ergo-js}
+
+A Javascript version of the alt-ergo package is available with the command [make js]. The command [make js] will create the javascript file [alt-ergo-js].
+To run this command you will need this aditionnal dependencies [js_of_ocaml-compiler] and [zarith_stubs_js].
+
+The [--timeout] option is unavailable along with the use of input zip file.
```
Gbury, line 6: Are the options really unavailable, or are they simply ignored ?

OCamlPro-Coquera: Timeout is ignored

src/bin/js/dune
```diff
@@ -0,0 +1,43 @@
...
+)
+
+(library
+  (name options_rec)
+  (modules      options_rec)
+)
+
+(executable
+  (name         worker_js)
+  (libraries    options_rec alt-ergo-lib alt-ergo-parsers common zarith_stubs_js js_of_ocaml js_of_ocaml-compiler lwt js_of_ocaml-lwt
```
hhugo, line 23: Why do you depend on `js_of_ocaml-compiler` ? it doesn't seem to be used anywhere

src/lib/util/options.ml
```diff
@@ -545,7 +545,7 @@ let mk_sat_opt bottom_classes disable_flat_formulas_simplification
          match s with
          | "" -> set
          | s -> Util.SS.add s set
-      ) Util.SS.empty (Str.split (Str.regexp ",") ndo)
```
hhugo, line 548: Note that `Str` is now supported by js_of_ocaml.3.6

src/bin/js/dune
```diff
@@ -0,0 +1,43 @@
...
+)
+  (modules      worker_js)
+  (modes        js)
+  (preprocess (pps js_of_ocaml-ppx lwt_ppx))
+  (js_of_ocaml
+    (flags --no-source-map --linkall --disable genprim)
```
hhugo, line 29: What are the reason for passing the two flags `--linkall` and `--disable genprim` ?

src/bin/js/index_js.mld
```diff
@@ -0,0 +1,7 @@
+{1 Alt-ergo-js}
+
+A Javascript version of the alt-ergo package is available with the command [make js]. The command [make js] will create the javascript file [alt-ergo-js].
+To run this command you will need this aditionnal dependencies [js_of_ocaml-compiler] and [zarith_stubs_js].
```
hhugo, line 4: your js_of_ocaml deps seems to be `js_of_ocaml`, `js_of_ocaml-compiler` and `js_of_ocaml-lwt`

rsc/extra/check_js_test.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# Cd into the root project directory
+cd $git_repo
+
+# Configure using the local installation directory
+./configure --prefix=$local_install_dir
+
+# Check that the makefile targets work
+echo "=+= [check_makefile.sh] building ... =+="
+opam install zarith_stubs_js js_of_ocaml-compiler
```
hhugo, line 21: should you install `js_of_ocaml` and `js_of_ocaml-lwt` as well ?

rsc/extra/travis_js_test.sh
```diff
@@ -0,0 +1,38 @@
...
+local_install_dir=$1
+
+# Cd into the root project directory
+cd $git_repo
+
+# Configure using the local installation directory
+./configure --prefix=$local_install_dir
+
+# Install requiered package for js compilation
+opam install zarith_stubs_js js_of_ocaml-compiler js_of_ocaml lwt js_of_ocaml-lwt js_of_ocaml-ppx lwt_ppx
```
Gbury, line 20: It would be better to pin (with no action) the opam package (which probably needs to be created) for the js version of alt-ergo, and then install its dependencies, that way the script also checks that the specified dependencies are enough to compile it.

src/bin/js/dune
OCamlPro-Coquera: That was old dependencies that i forgot to clean, that should be good know

src/lib/util/options.ml
OCamlPro-Coquera: Good, this could be usefull

Gbury:
Small nitpick, but it would probably be better to have one `alt-ergo-js` opam package which would install the various sub libs, etc... that is needed instead of multiplying the packages ?

src/bin/js/dune
OCamlPro-Coquera: None, it's old code from tests

rsc/extra/check_js_test.sh
OCamlPro-Coquera: Yes some dependencies were missing. This is now manage in the travis script using the dependencies of the js packages

rsc/extra/travis_js_test.sh
OCamlPro-Coquera: I've add 3 packages for js, when the tests will pass i will merge in one as you suggested

src/bin/js/index.html
```diff
@@ -0,0 +1,18 @@
...
+
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Alt-Ergo JS</title>
+    <meta name="description" content="Alt-Ergo Js">
+    <meta name="author" content="Alt-Ergo dev team">
+    <link rel="stylesheet" href="style.css">
+    <!-- ENTRY POINT -->
+    <script type="text/javascript" src="alt-ergo-worker.js"></script>
```
hhugo, line 11: do you really need to include the worker here ?

src/bin/js/main_js.ml
```diff
@@ -0,0 +1,130 @@
...
+module Html = Dom_html
+
+let document = Html.window##.document
+
+let file = ref "goal g : true"
+let extension = ref ".ae"
+let timeout = ref 100.
+
+let concat l =
+  List.fold_left (fun acc s -> Format.sprintf "%s\n %s@." acc s) "" l
```
hhugo, line 14: `String.concat` should do the work.

src/bin/js/index.html
OCamlPro-Coquera: No, it was debug info to ensure the worker was accessible. It is not needed anymore, thanks to point it out

.travis.yml
OCamlPro-Coquera: JS testing phase is now done at the end. Since this tests take a lot of time ~20 minutes, I'm only testing the js-node on the 4.09.0 version of OCaml. Do you think we should test more versions .

src/bin/js/index_js.mld
OCamlPro-Coquera: I've updating this documentation and divided it for my 3 rules and corresponding deps

rsc/extra/travis_js_test.sh
OCamlPro-Coquera: Done

.travis.yml
Gbury: Testing only on a single ocaml version should be fine.

rsc/extra/ocpChecker/dune-project
```diff
@@ -0,0 +1 @@
+(lang dune 1.5)
```
Gbury, line 1: We don't need this file, as this should be part of the alt-ergo project.

OCamlPro-Coquera: Fixed

OCamlPro-Coquera:
This PR is no longer WIP. It is open to comments, discussions and improvements :)

You can test the different bulding rules :
- `make js-node` to build a node js compatible version and then run `node alt-ergo.js <options> <input_file>`
- `make js-example` to build the simple js example using the web worker. In `www` you will find a minimal webpage. You may need to run `python3 -m http.server 8080` in `www` for symlink and then open localhost:8080 in your browser.

src/bin/js/index.mld
```diff
@@ -0,0 +1,31 @@
...
+Zip files are not recognised file extension.
+
+{2 The Alt-ergo web worker}
+
+A web worker of the alt-ergo solver is available with the command [make js-worker].
+This web worker takes a string corresponding to the input file to solve, and a set of options to set. It uses Js_of_ocaml Worker's and Lwt. At the end of solving it returns a couple of string lists.
+The first one corresponds to the status of the different goals to prove.
+The second one corresponds to errors that have occured during solving.
+
+To run this command you will need the followind aditionnal dependencies: [js_of_ocaml] [js_of_ocaml-lwt] and [zarith_stubs_js].
```
OCamlPro-Coquera, line 21:
```suggestion
To run this command you will need the following aditionnal dependencies: [js_of_ocaml] [js_of_ocaml-lwt] and [zarith_stubs_js].
```

OCamlPro-Coquera:
The worker will need the work done in #305 to retrieve and return specific information such as models are unsat-cores using the different formatters declared in the Options module

ACoquereau:
This PR has been divided in #453, #454 and #456

PR #317, opened 2020-07-10
Fix dune warning flags

OCamlPro-Coquera:
Disable warning as error and optimisation for the dune dev profile. Add new profile to enable all warnings. Add some known warning flag as error in release profile since this warnings should be resolve for release

### DIFF
--- a/src/dune
+++ b/src/dune
@@ -1,16 +1,13 @@
 (env
- (dev
-  ;; This profil is used (default) to dev without optimisation and blocking warnings.
+ (dev ;; This profil is used (default) to dev without optimisation and blocking warnings.
   (flags
    (:standard -bin-annot -w -22 -warn-error -A))
   (ocamlopt_flags -g))
- (warnings
-  ;; This profile can be used to simply enable all warnings.
+ (warnings ;; This profile can be used to simply enable all warnings.
   (flags
    (:standard -bin-annot -w +A -warn-error -A))
   (ocamlopt_flags -g))
- (release
-  ;; The release profile has optimisation enable.
+ (release ;; The release profile has optimisation enable.
   (flags
    (:standard -bin-annot))
   (ocamlopt_flags -O3 -unbox-closures))
--- a/src/dune
+++ b/src/dune
@@ -12,6 +12,6 @@
  (release
   ;; The release profile has optimisation enable.
   (flags
-   (:standard -bin-annot -w -A+K+X+Y+Z-22 -warn-error +X+K+Y+Z))
+   (:standard -bin-annot))
   (ocamlopt_flags -O3 -unbox-closures))
 )
--- a/src/dune
+++ b/src/dune
@@ -1,9 +1,17 @@
 (env
  (dev
+  ;; This profil is used (default) to dev without optimisation and blocking warnings.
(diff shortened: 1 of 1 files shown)


src/dune
```diff
@@ -1,9 +1,17 @@
...
+  ;; This profile can be used to simply enable all warnings.
+  (flags
+   (:standard -bin-annot -w +A -warn-error -A))
+  (ocamlopt_flags -g))
  (release
+  ;; The release profile has optimisation enable.
   (flags
-   (:standard -bin-annot -w -22))
-  (ocamlopt_flags -O3 -unbox-closures)))
+   (:standard -bin-annot -w -A+K+X+Y+Z-22 -warn-error +X+K+Y+Z))
```
Gbury, line 15: This is wrong, as it will very likely break the build on future releases of the compiler, especially considering the face that alt-ergo is released as sources rather than pre-compiled binaries. I'd very much prefer that for the release profile, you keep the dune defaults, which are good on this case.

OCamlPro-Coquera: Fixed

OCamlPro-Coquera:
@Gbury  The release profile has been reverted. For the comment, i prefer this way than above. I can agree to put them next to the profile name, as you prefer.

src/dune
```diff
@@ -1,9 +1,17 @@
 (env
  (dev
+  ;; This profil is used (default) to dev without optimisation and blocking warnings.
```
OCamlPro-Coquera, line 3:
```suggestion
 (dev ;; This profil is used (default) to dev without optimisation and blocking warnings.
```

src/dune
```diff
@@ -1,9 +1,17 @@
...
   (flags
-   (:standard -bin-annot -w -22))
-  (ocamlopt_flags -g -O3 -unbox-closures))
+   (:standard -bin-annot -w -22 -warn-error -A))
+  (ocamlopt_flags -g))
+ (warnings
+  ;; This profile can be used to simply enable all warnings.
```
OCamlPro-Coquera, line 8:
```suggestion
 (warnings ;; This profile can be used to simply enable all warnings.
```

src/dune
```diff
@@ -1,9 +1,17 @@
...
+  (flags
+   (:standard -bin-annot -w +A -warn-error -A))
+  (ocamlopt_flags -g))
  (release
+  ;; The release profile has optimisation enable.
```
OCamlPro-Coquera, line 13:
```suggestion
 (release ;; The release profile has optimisation enable.
```

src/dune
```diff
@@ -1,9 +1,14 @@
...
- (release
+   (:standard -bin-annot -w -22 -warn-error -A))
+  (ocamlopt_flags -g))
+ (warnings ;; This profile can be used to simply enable all warnings.
   (flags
-   (:standard -bin-annot -w -22))
-  (ocamlopt_flags -O3 -unbox-closures)))
+   (:standard -bin-annot -w +A -warn-error -A))
+  (ocamlopt_flags -g))
+ (release ;; The release profile has optimisation enable.
```
Gbury, line 10:
```suggestion
 (release ;; The release profile has optimizations enabled.
```

Issue #315, opened 2020-07-09
stack overflow


JuliaLawall:
The file https://pages.lip6.fr/Julia.Lawall/ab.why (generated by why3) produces a stackoverflow in with alt-ergo 2.3.0.  The following backtrace is reported:

Fatal error: exception Stack overflow
Raised by primitive operation at file "list.ml", line 88, characters 32-39
Called from file "list.ml", line 88, characters 32-39
Called from file "list.ml", line 88, characters 32-39
Called from file "list.ml", line 88, characters 32-39
Called from file "list.ml", line 88, characters 32-39
Called from file "list.ml", line 88, characters 32-39
Called from file "list.ml", line 88, characters 32-39
Called from file "list.ml", line 88, characters 32-39
Called from file "lib/structures/expr.ml", line 1901, characters 13-54
Called from file "lib/structures/expr.ml", line 1944, characters 11-59
Called from file "lib/structures/expr.ml", line 2151, characters 10-65
Called from file "lib/reasoners/matching.ml", line 625, characters 20-71
Called from file "lib/reasoners/matching.ml", line 654, characters 32-51
Called from file "map.ml", line 319, characters 19-42
Called from file "map.ml", line 319, characters 26-41
Called from file "map.ml", line 319, characters 26-41
Called from file "map.ml", line 319, characters 26-41
Called from file "lib/reasoners/instances.ml", line 288, characters 27-65
Called from file "lib/reasoners/satml_frontend.ml", line 370, characters 6-68
Called from file "lib/reasoners/satml_frontend.ml", line 868, characters 12-40
Called from file "list.ml", line 117, characters 24-34
Called from file "lib/reasoners/satml_frontend.ml", line 926, characters 27-58

OCamlPro-Coquera:
Thanks for your report.
The bug seems to happen at trigger computation, we will look into it.

OriginLabs-Iguernlala:
Yes, I confirm. This happens when trying to infer triggers for axiom `conclusion_trans8`. I don't know if this (really big) axiom, attached below, is written by hand or auto-generated. It would help if the user could provide the triggers (in particular if she/he has an idea on how the axiom should be instantiated/used during proof).

[conclusion_trans8.ae.zip](https://github.com/OCamlPro/alt-ergo/files/4907443/conclusion_trans8.ae.zip)

(For AE developers):

[...]

PR #324, opened 2020-07-11
Improve multi triggers inference

OriginLabs-Iguernlala:
This PR fixes the `Stack Overflow` in https://github.com/OCamlPro/alt-ergo/issues/315

As said in that issue, the `Stack Overflow` is raised when trying to infer multi-triggers for a big axiom, with the following characteristics:
```
4 bound vars:
  'p1~3818'
  'p2~3819'
  'p3~3820'
  'c~3821'
number of (non-var) terms containing these vars 31
size of powerset from these terms 1584291
size of smalltest multi-trigger = 3
size of biggest multi-trigger = 20
```
Since we have 4 bound vars (and no type variables), we expect the biggest multi-trigger that covers all these variables to have at most 4 elements. But here, a debug trace showed that `size of biggest multi-trigger = 20`.

This situation is due to the use of functions `Set.equal` (in `  if SSet.equal bv1 bv2 && Svty.equal vty1 vty2 then`) to test if a term `t` covers new variables (or type variables) of a multi-trigger that is being constructed (ie. partial). Actually, the test should be:
`  if SSet.subset bv1 bv2 && Svty.subset vty1 vty2 then`

```
     | (t, bv1, vty1)::l ->
        let llt, llt_ok =
          SLLT.fold
            (fun (l, bv2, vty2) (llt, llt_ok) ->
               if SSet.equal bv1 bv2 && Svty.equal vty1 vty2 then
                 (* t doesn't bring new vars *)
                 llt, llt_ok
               else
```

With this fix, we see in the logs that things are better: 

```
bound vars:
  'p1~3818'
  'p2~3819'
  'p3~3820'
  'c~3821'
number of terms containing these vars 35
size of powerset from these terms 3627
size of smalltest multi-trigger = 3
size of biggest multi-trigger = 4
```

**TODO*:* Test this PR on our internal testsuite before/after merging it. It may improve performances, but it may also cause regressions!

### DIFF
--- a/src/lib/structures/expr.ml
+++ b/src/lib/structures/expr.ml
@@ -1855,7 +1855,7 @@ module Triggers = struct
         let llt, llt_ok =
           SLLT.fold
             (fun (l, bv2, vty2) (llt, llt_ok) ->
-               if SSet.equal bv1 bv2 && Svty.equal vty1 vty2 then
+               if SSet.subset bv1 bv2 && Svty.subset vty1 vty2 then
                  (* t doesn't bring new vars *)
                  llt, llt_ok
                else
--- a/src/lib/structures/expr.ml
+++ b/src/lib/structures/expr.ml
@@ -1855,17 +1855,18 @@ module Triggers = struct
         let llt, llt_ok =
           SLLT.fold
             (fun (l, bv2, vty2) (llt, llt_ok) ->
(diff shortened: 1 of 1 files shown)


OCamlPro-Coquera:
Testsuite ok.

PR #328, opened 2020-07-12
fix print status for timeout

OCamlPro-Coquera:


OriginLabs-Iguernlala:
So, it was really a simple refactoring/mechanic copy/paste bug ?

OCamlPro-Coquera:
Yep copy paste error, my bad :)

Issue #245, opened 2019-06-19
Assert failure when parsing a file in 2.3.0


vprevosto:
The attached file, generated by Frama-C/WP, is accepted by Alt-Ergo 2.2.0, but leads to an assertion failure in 2.3.0:
```
Fatal error: exception File "lib/structures/expr.ml", line 2056, characters 15-21: Assertion failed
```

A quick search seems to indicate that the issue lies in the recursive definition of function `L_count` (lines 612-621). Recursive functions are a nasty thing, and I wouldn't mind if Alt-Ergo would refuse this definition, but I'd expect a better error message.

[mjrty_no_majority_ensures_result_no_cand_Alt-Ergo.txt](https://github.com/OCamlPro/alt-ergo/files/3306892/mjrty_no_majority_ensures_result_no_cand_Alt-Ergo.txt)

OCamlPro-mattiasdrp:
Sorry for the delay, 

I'm working on your issue. I was able to reduce it to a minimal problem : 

```
logic match_bool : bool, 'a, 'a -> 'a

function L_count (x : int) : int =
    let x = (x - 1) : int in
    let x_1 = L_count(x) in
    match_bool(true, x_1, x_1)

goal mjrty_no_majority_ensures_result_no_cand:
  forall i1, i2 : int.
  (0 < i1) -> (0 < i2)
```
This will result in the assertion failure. It allowed me to realise that this assertion failure doesn't appear in Alt-Ergo 2.2.0 so I'll try to fix it as fast as possible.

OCamlPro-mattiasdrp:
Working on fixing the problem still. Found out this behaviour:

With my minimal example, the term is created like this:
```
(let 'x_1~3' = L_count(('x~2' - 1)) in
(L_count('x~2') = match_bool(true,'x_1~3','x_1~3')))
```
And it raises the assertion failure because the term contains `L_count(x)` and `L_count(x-1)`.
Changing `let x = L_count(x-1)` for `let x = L_count(x)` will make the program run smoothy.

If I write the same kind of function but with two different parameters:

```
function L_count (x : int) : int =
    let x_1 = L_count(x - 1) in
    let x_2 = L_count(x + 1) in
    match_bool(true, x_1, x_2)
```
It is created as the following term:
```
(L_count('x~2') = match_bool(true,L_count(('x~2' - 1)),L_count(('x~2' + 1))))
```

And this doesn't create any bug.

So I tried inlining my minimal example:
```
[...]

PR #336, opened 2020-07-16
defining terms of recursive functions

OriginLabs-Iguernlala:
fixes https://github.com/OCamlPro/alt-ergo/issues/245

But the example provided in that issue triggers another bug, which should be fixed by https://github.com/OCamlPro/alt-ergo/pull/337


OriginLabs-Iguernlala:
testsuite ok

Issue #335, opened 2020-07-16
Unexpected free variables in SAT solver


OriginLabs-Iguernlala:
In the given example, Alt-Ergo crashes with

```
Fatal error: exception File "src/lib/reasoners/satml.ml", line 811, characters 8-14: Assertion failed
```
The logs show something like:
```
...
Entry of pre_assume: Given (lemma: L_count forall 'Mint_0~39':( <record>addr,int) farray, 'a~40': <record>addr, 'length_0~41':int, 'cand_0~42':int[].
 (let [sko = !?___let4('x~43','cand_0~42','a~40','Mint_0~39')] 'x_1~44' =
L_count2('Mint_0~39','a~40','x~43','cand_0~42') in
(L_count('Mint_0~39','a~40','length_0~41','cand_0~42') = match_bool(eqb('length_0~41',0),0,match_bool(eqb('Mint_0~39'[shift_uint32('a~40',('length_0~41' - 1))],'cand_0~42'),(1 + 'x_1~44'),'x_1~44')))))
...
```

For the function:
```
logic L_count2 :
    (addr,int) farray,
    addr,
    int,
    int -> int

function L_count
    (Mint_0:(addr,int) farray,
    a:addr,
    length_0:int,
    cand_0:int) :
    int =
    let x = (length_0 - 1) : int in
    let x_1 = L_count2(Mint_0, a, x, cand_0) : int in
    match_bool((eqb(length_0, 0)), 0,
      (match_bool((eqb(Mint_0[shift_uint32(a, x)], cand_0)), (1 + x_1), x_1)))
```
The variable `x~43` above is not quantified (free), and probably corresponds to the internal name of the variable `x` introduced with `let x = (length_0 - 1) : int in`

The variable `x` (`x~43`⁾ has likely been inlined, but not everywhere. This bug is probably due to the fact that the function that returns the `free_vars` of an expression doesn't consider those appearing in skolem terms.

[prev.ae.zip](https://github.com/OCamlPro/alt-ergo/files/4932644/prev.ae.zip)

PR #337, opened 2020-07-17
Fix lifting/purification of expressions with let-in

OriginLabs-Iguernlala:
Fixes https://github.com/OCamlPro/alt-ergo/issues/335

let-in were not always re-constructed in the correct order.

For instance:

let x = e1 in
let y = h(x, e2) in
f(x,y) < g(y,x)

may become

let y = h(x, e2) in
let x = e1 in
f(x,y) < g(y,x)

which is incorrect, because 'x' escapes its scope.

### DIFF
--- a/src/lib/structures/expr.ml
+++ b/src/lib/structures/expr.ml
@@ -2298,6 +2298,15 @@ let is_pure e = e.pure
 
 module Purification = struct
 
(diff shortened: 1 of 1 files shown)


OriginLabs-Iguernlala:
testsuite ok

Issue #333, opened 2020-07-14
Do not use colors in solver's final output


OriginLabs-Iguernlala:
It would probably be better to output the solver's answer (sat / unsat / unknwon / valid, ...) without ANSI colors.

For me, it should be as simple as doing `echo $output | grep ^unsat$` to test if the solver answered unsat. But, this doesn't work if the output is colored.

OCamlPro-Coquera:
The options `no-colors-in-output` is here to disabling colors in outputs. You should be able to perform your commands with this option, no ?

OCamlPro-Coquera:
We can also add an option `--clean-ouputs`, and `-c`, that disable colors, headers (for debug) and formatting so you will have exacly what you had previously and we will have something more user friendly without this option ?

OriginLabs-Iguernlala:
Yes, I can. But you cannot/shouldn't ask people who already integrated Alt-Ergo in their tools to activated an option to fix such issues imo.

OriginLabs-Iguernlala:
Also, I noticed that when Alt-Ergo answers eg `sat` or `unsat` when fed with an SMT file, the comment:
```
; File .... : Valid (xxx seconds) (y steps)
```
disapeared. So now, one should perfix commands with `time` to get resolution time :-/

OCamlPro-Coquera:
Since the next version of Alt-Ergo will break retro compatibility because of new options format, from `-<option>` to `--<option>` and `.why` files are marked as obselete (use `.ae` instead) users that integrated Alt-Ergo will be forced to check and change their options. I think a simple option like `-c` to have clean output can be possible no ?

PR #339, opened 2020-07-17
Fix solver output

OriginLabs-Iguernlala:
fixes https://github.com/OCamlPro/alt-ergo/issues/333

4 commits, 4 (small) changes in printer.ml

- no ANSI color in solver's final answer
- refactoring: provide the fmt for print_status_loc and print_status_value
- refactor print_status to show stats even in unsat mode (commented and in std_err)
    eg.
```   
    ; File "foo.smt2", line 1, characters 1-17: Inconsistent assumption (0.0040) (0 steps)
    ; File "foo.smt2", line 2, characters 1-12: Valid (0.0041) (0 steps) (goal g)
    unsat
```    
   where the two first lines are commented and printed to std debug/err.

- [printer] log print_status_{inconsistent, preprocess} in fmt_dbg instead of fmt_std
    For instance, if we have something like:
 ```
    File "bar.ae", line 2, characters 1-14: Inconsistent assumption (0.0010) (0 steps)
    File "bar.ae", line 2, characters 1-14: Valid (0.0010) (0 steps) (goal g)
```
  The first line is written to fmt_dbg (= std_err in default settings)

### DIFF
--- a/src/lib/util/printer.ml
+++ b/src/lib/util/printer.ml
@@ -264,20 +264,15 @@ let print_status_loc loc =
         "%a@,"
         Loc.report loc
 
-let print_status_value v color =
-  if Options.get_output_with_colors () then
-    fprintf (Options.get_fmt_std ())
-      "@{<%s>@{<bold>%s@}@}"
-      color v
-  else fprintf (Options.get_fmt_std ())
-      "%s" v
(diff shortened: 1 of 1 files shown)


OriginLabs-Iguernlala:
testsuite ok

Issue #340, opened 2020-07-19
soundness bug in LIA


OriginLabs-Iguernlala:
This example has status sat, but Alt-Ergo says it's unsat.

[convert-jpg2gif-query-1197.smt2.zip](https://github.com/OCamlPro/alt-ergo/files/4943372/convert-jpg2gif-query-1197.smt2.zip)

OriginLabs-Iguernlala:
minimal example that reproduces the bug:
```
(declare-fun a () Int)
(declare-fun b () Int)
(declare-fun c () Int)
(declare-fun d () Int)

(assert (= (+ a (* b (- 524288)) (* c (- 4294967296)) (- d)) (- 1024)))
(assert (<= a 0))
(assert (>= c 0))
(assert (>= b 0))
(assert (>= d 0))
(check-sat)
```

OriginLabs-Iguernlala:
Even simpler:
```
(declare-fun a () Int)
(declare-fun b () Int)
(declare-fun c () Int)
(declare-fun d () Int)
(declare-fun p (Int Int Int Int) Bool)

(assert (p a b c d)) ; just here for ordering of variables
(assert (<= a 0))
(assert (>= c 0))
(assert (>= b 0))
(assert (>= (+ (* (- 16384) c) (* (- 2) b) a 1) 0))
(check-sat)
```

OriginLabs-Iguernlala:
The original bug should be fixed by PR #342 

But there is still another bug (with the original example) if option `--max-split=-1` is activated

PR #342, opened 2020-07-19
Bug fix in intervals deduction from FM results

OriginLabs-Iguernlala:
In some situations, the constant c in an inequatily on integers:
p + c <= 0, is not an integer. In this case, it should be replaced with:
p + floor(c) <= 0

fixes https://github.com/OCamlPro/alt-ergo/issues/340


OriginLabs-Iguernlala:
testsuite ok

PR #343, opened 2020-07-22
fix simplification of distincts

OriginLabs-Iguernlala:
bad simplification in eg `distinct (a, a, b)` should return `false` (`true` for `n_ary_eq`) instead of `distinct(a,b)`


Gbury:
Shouldn't it return `false` instead in your example ? The PR code seems correct, but I think you more likely wanted to talk about the n-ary equality (which also affect n-ary pairwise disequality such as `distinct`).

OriginLabs-Iguernlala:
Yes :-D

PR #358, opened 2020-08-19
Doc-Alt-Ergo

LouisRustenholz:
Adding a documentation for Alt-Ergo, built with Sphinx.

Alt-Ergo's native language documentation is included.


.travis.yml
```diff
@@ -40,10 +42,15 @@ before_install:
   - eval `opam config env`
   # Install odoc & ocp-indent
   - opam install odoc ocp-indent.1.8.1
+  # Install sphinx libs
+  - pip install --user sphinx
+  - pip install --user recommonmark
+  - pip install --user sphinx_rtd_theme
+  - pip install --user sphinx-markdown-tables
```
Gbury, line 49: Is there any way to add these as package deps in the `addons` section (which is cleaner) ? I.e. do these packages exist as ubuntu-packages or alternatively, does some additional ppa allows to use them, in the same vein as the `avsm` ppa adds versioned opam packages.

.travis.yml
```diff
@@ -40,10 +42,15 @@ before_install:
...
   # Set the verbose option *after* compiling the compiler
   # and installing odoc, else the log oveflows
   - export OPAMVERBOSE=1
-
+
```
Gbury, line 53: Spurious end-of-line spaces added

docs/sphinx_docs/API/index.md
```diff
@@ -0,0 +1,3 @@
+
+API documentation
+=================
```
Gbury, line 3: Is there any way to add a link to the odoc-generated API doc here ?

rsc/extra/deploy_doc.sh
```diff
@@ -8,7 +8,9 @@
 # files. Travis will automatically commit the changes
 # and push it to the repo
 
+# Legacy ODOC directory hidden in a ODOC_DIR folder
 VERSION=$1
+OLD_ODOC_DIR := "odoc"
```
Gbury, line 13: As far as I can tell from the PR, the sphinx doc is complementary to the odoc documentation: the sphinx doc is more of a cli-user facing doc, whereas the API doc is a dev and lib-user facing doc. In particular, I didn't see any thing that would make sphinx generate the API doc (but maybe I'm wrong).
Assuming I'm right, I find the use of `OLD` and `Legacy` wording particularly misleading considering and actually wrong.

Gbury (changes requested):
Looks good (modulo some small comments).
Some things are missing I think:
- update INSTALL.md to add the doc-building deps related to sphinx (and specify for other dependencies whether they are for build or other steps, and distinguish ocaml dependencies from external deps such as python)
- Add a section about documentation building to `INSTALL.md` actually, ^^

Additionally, is there an up-to-date preview of what the rendered doc looks like (including the index page copied by the deploy_doc script) ?

PR #361, opened 2020-09-11
Next+smt2 status

OCamlPro-Coquera:
This pr aim to add a check for the status of the smt2 files. A warning / error is raised if the file is known to be sat (resp unsat) and Alt-ergo return unsat (resp sat).

### DIFF
--- a/src/bin/text/main_text.ml
+++ b/src/bin/text/main_text.ml
@@ -163,8 +163,9 @@ let () =
         let l, env = I.type_parsed state.env p in
         List.fold_left (typed_loop all_used_context) { state with env; } l
       with
-        Errors.Error e ->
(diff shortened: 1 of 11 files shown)


src/bin/text/main_text.ml
```diff
@@ -163,8 +163,9 @@ let () =
         let l, env = I.type_parsed state.env p in
         List.fold_left (typed_loop all_used_context) { state with env; } l
       with
-        Errors.Error e ->
-        Printer.print_err "%a" Errors.report e;
+      | Errors.Error e ->
+        if e != Warning_as_error then
+          Printer.print_err "%a" Errors.report e;
```
OCamlPro-Coquera, line 168:
```suggestion
```

PR #305, opened 2020-05-12
Refactoring formatter and printer

OCamlPro-Coquera:
Add options `--err-formatter` and `--std-formatter` to set the output formatter used to print results and other informations.

Additionaly, in the library, we can now specify a different formatter for debug, warning, model, unsat-core, usefull to ouput only model to a buffer for example.

Every print in Alt-Ergo is run through the new printer module. Alt-Ergo now print informations to the corresponding formatter such as debug informations to the debug formatter set in the Options module. Even this was not the goal of this PR, some print have been update and reformat. Some colors have also been added in the printing.

PR #362, opened 2020-09-14
Add some laziness in print_dbg

### LINKED ISSUE #321: Put some laziness in Printer.print_dbg ?
I see a possible regression in performances due to introduction of module `Printer`. I'm thinking of this kind of usage of `Printer.print_dbg`:

```
  print_dbg ~debug "%d" (List.length l);
```

With the old (basic) mecanism, this code would be:
```
if debug then
  Format.fprintf fmt "%d" (List.length l)
``` 

But, with the new mechanism, `List.length l` is executed even if the debug flag is false.

OCamlPro-Coquera:
Add guard (ite) before print with compute like `List.length`
should resolve #321


OCamlPro-Coquera:
revert changes in #305 regarding the guard of the debug printing

PR #364, opened 2020-09-23
Fix opam call in configure.ml

AltGr:
it's bad practice to call `opam` without `--readonly` from within an opam package, and could cause failures or deadlocks.


OCamlPro-Coquera:
Thanks @AltGr

PR #366, opened 2020-10-06
Add greedier instantiation round in both sat solver

OCamlPro-Coquera:
The goal of this PR is to delay the triggers-vars heuristic when Alt-Ergo perform a greedy instantiation round. This heuristic can be costly (time) but in certain cases allow a faster resolution.

This changes has been tested for both SAT. The global performance (time) has been improved


src/lib/reasoners/satml_frontend.ml
```diff
@@ -890,6 +899,7 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
      greedy = true;
     }
 
+
```
OCamlPro-Coquera, line 902:
```suggestion
```

PR #371, opened 2020-10-29
Incremental smt2

ACoquereau:


src/lib/structures/commands.ml
```diff
@@ -58,6 +60,8 @@ let print_aux fmt = function
       name print_goal_sort sort Expr.print e
   | ThAssume t ->
     Format.fprintf fmt "th assume %a" Expr.print_th_elt t
+  | Push -> Format.fprintf fmt "Push"
+  | Pop ->  Format.fprintf fmt "Pop"
 
 let print fmt decl = print_aux fmt decl.st_decl
```
OriginLabs-Iguernlala, line 66: I probably never noticed that a function `print` exists in commands.ml. Has it been added recently ?

src/bin/text/main_text.ml
```diff
@@ -157,10 +157,11 @@ let () =
   let all_used_context = FE.init_all_used_context () in
   if Options.get_timelimit_per_goal() then
     FE.print_status FE.Preprocess 0;
+  let assertion_stack = ref (Stack.create ()) in
```
OriginLabs-Iguernlala, line 160: ref is probably not needed

src/lib/frontend/parsed_interface.ml
```diff
@@ -91,6 +91,13 @@ let mk_theory_axiom loc name expr =
 let mk_theory_case_split loc name expr =
   Axiom (loc, name, Util.Default, expr)
 
+(** Declaration of stack assertions commands *)
+
+let mk_push loc =
+  Push loc
+
+let mk_pop loc =
+  Pop loc
```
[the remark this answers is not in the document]
OriginLabs-Iguernlala: I don't know if it's better to be generic in Alt-Ergo and support `push n` and `pop n` or not. But, in any case, we can already continue reviewing and merging this version.

src/parsers/psmt2_to_alt_ergo.ml
```diff
@@ -401,8 +423,8 @@ module Translate = struct
     | Cmd_SetLogic _ -> not_supported "set-logic"; acc
     | Cmd_SetOption _ -> not_supported "set-option"; acc
     | Cmd_SetInfo _ -> not_supported "set-info"; acc
-    | Cmd_Push _ -> not_supported "push"; assert false
-    | Cmd_Pop _ -> not_supported "pop"; assert false
+    | Cmd_Push n -> (translate_push n (pos command)) @ acc
+    | Cmd_Pop n -> (translate_pop n (pos command)) @ acc
```
OriginLabs-Iguernlala, line 427: - you can have a single function `translate_push_pop` that takes a function as a parameter (`mk_push` or `mk_pop`)
- `acc` could be given as an argument to `translate_xxx` to avoid lists concatenation (`@`)

src/parsers/psmt2_to_alt_ergo.ml
```diff
@@ -349,6 +354,23 @@ module Translate = struct
       List.map2 translate_datatype_decl sort_dec datatype_dec
     with Invalid_argument _ -> assert false
 
+  let rec list_init acc n f =
+    if n > 0 then
+      list_init (f :: acc) (n - 1) f
+    else acc
+
+  let translate_push n pos =
+    try let n = int_of_string n in
```
OriginLabs-Iguernlala, line 363: provide directly `n` of type int to `translate_xxx` to be able to make this function recursive and inline `list_init`.

src/parsers/psmt2_to_alt_ergo.ml
```diff
@@ -401,8 +423,8 @@ module Translate = struct
     | Cmd_SetLogic _ -> not_supported "set-logic"; acc
     | Cmd_SetOption _ -> not_supported "set-option"; acc
     | Cmd_SetInfo _ -> not_supported "set-info"; acc
-    | Cmd_Push _ -> not_supported "push"; assert false
-    | Cmd_Pop _ -> not_supported "pop"; assert false
+    | Cmd_Push n -> (translate_push n (pos command)) @ acc
```
OriginLabs-Iguernlala, line 426: let n : int = `nb_contexts n` to do:
- `int_of_string n`, or
- fail with the message in `must_not_happen`

Then, `translate_xxx` could be called with `n : int`

OriginLabs-Iguernlala (changes requested):

src/lib/frontend/parsed_interface.ml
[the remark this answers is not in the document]
ACoquereau: I have made some change to push `push n` and `pop n` through the lib until the SATs. I think is better to have a generic support for `push` and `pop` command as you suggest

src/lib/structures/commands.ml
ACoquereau: You probably never noticed this function because it's in a comment :) maybe never used.

src/parsers/psmt2_to_alt_ergo.ml
ACoquereau: Since i don't lift `push` and `pop` commands anymore i directly use `mk_push pos n`.

src/parsers/psmt2_to_alt_ergo.ml
ACoquereau: translate_push_pop adds only command now, we avoid lists concatenation

src/parsers/psmt2_to_alt_ergo.ml
ACoquereau: Should be ok with previous comments and changes

src/lib/frontend/frontend.ml
```diff
@@ -143,10 +143,10 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
   let process_decl print_status used_context ((env, consistent, dep) as acc) d =
     try
       match d.st_decl with
-      | Push -> SAT.push env, consistent, dep
-      | Pop ->
+      | Push n -> SAT.push env n, consistent, dep
+      | Pop n ->
         (* Reset the constitency for each pop command *)
-        SAT.pop env, true, dep
+        SAT.pop env n, true, dep
```
OriginLabs-Iguernlala, line 149: I haven't noticed it during the first review: is it correct to always return `dep` ?

OriginLabs-Iguernlala (commented):
- I don't know if I wrote it somewhere: Can `--timeout-per-goal` option be used in incremental mode ?

- Another question related to save/replay used-context: an extra work is probably needed. This is probably already done for native-lang's multi-goals feature, and we could adapt it for push/pop

src/lib/structures/typed.ml
```diff
@@ -358,6 +360,10 @@ let rec print_tdecl fmt = function
          ~pp:(print_rwt print_term)) l
   | TGoal (_, sort, name, f) ->
     Format.fprintf fmt "goal %s: @[<hov>%a@]" name print_formula f
+  | TPush ->
+    Format.fprintf fmt "push"
+  | TPop ->
+    Format.fprintf fmt "pop"
```
OriginLabs-Iguernlala, line 366: maybe fix these lines, even if it's dead code ?

src/lib/structures/typed.mli
```diff
@@ -297,6 +297,11 @@ and 'a tdecl =
       declares a type [t], with parameters [vars], and with
       contents [body]. This new type may either be abstract,
       a record type, or an enumeration. *)
+  | TPush of Loc.t * int
+  (** Push new assertions level onto the assertion  stack *)
+  | TPop of Loc.t * int
+  (** Pop an assertions level from the assertion stack *)
+
```
OriginLabs-Iguernlala, line 304: comments not clear: push/pop `n` assertion levels

src/parsers/psmt2_to_alt_ergo.ml
```diff
@@ -31,6 +31,11 @@ module Translate = struct
     | None -> Loc.dummy
     | Some p -> p
 
+  let must_not_happen loc s =
+    let s = Format.sprintf
+        "psmt2-frontend typing should unsure that case can't happen : %s" s in
```
OriginLabs-Iguernlala, line 36: unsure ... -> ensure that this case

src/lib/frontend/typechecker.ml
```diff
@@ -2143,9 +2144,30 @@ let type_user_defined_type_body ~is_recursive env acc (loc, ls, s, body) =
...
+    f s
+  else
+    let _ = f s in
+    push_pop_env f (n-1) s
+
+let rec type_decl (acc, env) d assertion_stack =
   Types.to_tyvars := MString.empty;
   match d with
+  | Push (loc,n) ->
+    push_pop_env (Stack.push env) n assertion_stack;
```
OriginLabs-Iguernlala, line 2158: Should we check that `n >= 0` or ` n >= 1` ?

src/lib/frontend/typechecker.ml
```diff
@@ -2143,9 +2144,30 @@ let type_user_defined_type_body ~is_recursive env acc (loc, ls, s, body) =
     assert (not is_recursive); (* Abstract types are not recursive *)
     acc, env
 
-let rec type_decl (acc, env) d =
+let rec push_pop_env f n s =
+  if n <= 1 then
+    f s
+  else
+    let _ = f s in
```
OriginLabs-Iguernlala, line 2151: `let () = f s in`

ACoquereau:
>     * I don't know if I wrote it somewhere: Can `--timeout-per-goal` option be used in incremental mode ?
Yes, it works pretty well
 
>     * Another question related to save/replay used-context: an extra work is probably needed. This is probably already done for native-lang's multi-goals feature, and we could adapt it for push/pop

I did not look into this issue but you're right we need to fix this

src/lib/frontend/typechecker.ml
ACoquereau: I have added a check for `n >= 0` in the typechecker and in psmt2_to_alt_ergo.ml since the smtlib standard only spécify that n should be `>= 0`

src/lib/frontend/typechecker.ml
ACoquereau: Can't use `()` because `f s` return an env when `f` is `Stack.pop`

src/lib/frontend/typechecker.ml
```diff
@@ -2148,17 +2148,21 @@ let rec push_pop_env f n s =
   if n <= 1 then
     f s
   else
-    let _ = f s in
+    let _env = f s in
```
OriginLabs-Iguernlala, line 2151: It's not env in case `f` is `Stack.push` but of type `unit`.

ACoquereau: That's why `let _ = ` was a good solution for me :)

src/lib/reasoners/satml.ml
```diff
@@ -1754,4 +1782,27 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
...
+    g.is_guard <- false;
+    g.neg.is_guard <- false;
+    assert (not g.var.na.is_true); (* atom not false *)
+    if g.var.pa.is_true then (* if already decided  *)
+      begin
+        cancel_until env (g.var.level - 1); (* undo its decision *)
+        (* all previous guards are decided *)
+        env.next_dec_guard <- Vec.size env.increm_guards
+      end;
+    enqueue env g.neg env.next_dec_guard None
```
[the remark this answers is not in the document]
ACoquereau: Right, it should be `enqueue env g.neg 0 None`, done

src/lib/frontend/typechecker.ml
```diff
@@ -2143,9 +2144,20 @@ let type_user_defined_type_body ~is_recursive env acc (loc, ls, s, body) =
...
   match d with
+  | Push loc ->
+    Stack.push env !assertion_stack;
+    let td = {c = TPush(loc); annot = new_id () } in
+    (td,env) :: acc ,env
+  | Pop loc -> begin
+      try
+        let env = Stack.pop !assertion_stack in
+        let td = {c = TPop(loc); annot = new_id () } in
+        (td,env) :: acc, env
```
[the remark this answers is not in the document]
ACoquereau: It seems that env in `(td,env)` is not used in `type_parsed` (used in main_text.ml) but used in type_file (used in the gui). This need some tests

src/lib/frontend/frontend.ml
ACoquereau: No i don't think it's correct indeed. Since we force the reset of constintency for `pop` command, should we retrieve the `dep` that was computed before the n assertion levels we pushed ?

src/lib/reasoners/satml.ml
```diff
@@ -460,7 +460,15 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
     a.var.vpremise <- []
 
   let enqueue_assigned env a =
-    if a.neg.is_guard then raise (Unsat None);
+    if a.neg.is_guard then begin
+      (* if the negation of a is (still) a guard, it should be forced to true
+         during the first decisions.
+         If the SAT tries to deduce that a.neg is true (ie. a is false),
+         then we have detected an inconsistency. *)
+      assert (a.var.level <= env.next_dec_guard);
```
OriginLabs-Iguernlala, line 468: Is it `env.next_dec_guard` or just `Vec.size env.increm_guards` ?

src/lib/reasoners/satml.ml
```diff
@@ -570,7 +577,15 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
   let enqueue env a lvl reason =
     assert (not a.is_true && not a.neg.is_true &&
             a.var.level < 0 && a.var.reason == None && lvl >= 0);
-    if a.neg.is_guard then raise (Unsat None);
+    if a.neg.is_guard then begin
+      (* if the negation of a is (still) a guard, it should be forced to true
+         during the first decisions.
+         If the SAT tries to deduce that a.neg is true (ie. a is false),
+         then we have detected an inconsistency. *)
+      assert (a.var.level <= env.next_dec_guard);
```
OriginLabs-Iguernlala, line 585: Is it `env.next_dec_guard` or just `Vec.size env.increm_guards` ?

src/lib/frontend/frontend.ml
OriginLabs-Iguernlala: Yes, I think we should do this

src/lib/frontend/typechecker.ml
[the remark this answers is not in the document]
OriginLabs-Iguernlala: Yes, only used in the GUI

src/lib/frontend/typechecker.ml
OriginLabs-Iguernlala: Ok, let's keep `let _ = ...`

src/lib/frontend/typechecker.ml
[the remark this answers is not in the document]
ACoquereau: The gui raise :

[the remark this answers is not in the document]
ACoquereau: Type_parsed is used with split_goals (typechecker.ml) in the main_gui  but this function was not yet updated with the addition of incremental commands. If i read correctly the env associated with the td should be the env at td assertion level. So i will make some change to save old env as you suggested if you agree ?

src/lib/reasoners/satml.ml
ACoquereau: I think it's better to use `env.next_dec_guard`. This value is incremented when we decide (`pick_branch_lit`) a guard. We want to ensure that if we enqueue `a` and `a.neg` is a guard, this guard `a.neg` was decided before the current next_guard_lvl.

src/lib/frontend/frontend.ml
ACoquereau: Done. I also fixed the consistency issue retrieving the consistency of the previous assertion level

src/lib/reasoners/satml_frontend.ml
```diff
@@ -839,23 +852,24 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
         (*SAT.update_lazy_cnf activate ~dec_lvl;*)
         SAT.assume env.satml unit nunit f ~cnumber:0 activate ~dec_lvl;
       with
-      | Satml.Unsat (lc)  -> raise (IUnsat (env, make_explanation lc))
+      | Satml.Unsat (lc) -> raise (IUnsat (env, make_explanation lc))
       | Satml.Sat -> assert false
 
-
-  let assume_aux_bis ~dec_lvl env l =
+  let pre_assume_aux env l =
```
OriginLabs-Iguernlala, line 858: introduction of function `pre_assume_aux` is probably not needed if we don't use it (anymore) in `push`

src/lib/reasoners/fun_sat.ml
```diff
@@ -1694,11 +1706,76 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+        );
+    let rec push_aux acc n =
+      if n <= 0 then acc
+      else
+        let b = E.fresh_name Ty.Tbool in
+        Stack.push b acc.incremental.stack_guard;
+
+        Stack.push (unit_facts acc) acc.unit_facts_cache;
+
+        let incremental =
```
OriginLabs-Iguernlala, line 1725: - Can probably be simplified only `incremental` field is updated and returned`
- We can probably use `Util.loop` for the recursion

src/lib/reasoners/fun_sat.ml
```diff
@@ -1694,11 +1706,76 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+        );
+    let rec pop_aux acc n =
+      if n <= 0 then acc
+      else
+        let neg_b = Stack.pop acc.incremental.stack_guard in
+        let b =
+          if Stack.is_empty acc.incremental.stack_guard then
+            None
+          else Some (Stack.top acc.incremental.stack_guard)
+        in
```
OriginLabs-Iguernlala, line 1752: Don't understand why we have `let neg_b = ...` above.

Because, if `b` is of the form `Some c`, `c` is equal to `neg_b` (in both cases, we have `Stack.pop acc.incremental.stack_guard`)

src/lib/reasoners/fun_sat.ml
```diff
@@ -137,6 +137,11 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
 
   end
 
+  type incremental = {
+    current_guard: E.t option;
+    stack_guard: E.t Stack.t;
+    guards: SE.t;
```
OriginLabs-Iguernlala, line 143: From the usage I've seen, `guards` in actually `neg_old_gards`

src/lib/reasoners/fun_sat.ml
```diff
@@ -1694,11 +1706,76 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+    pop_aux env to_pop
+
   let unsat env gf =
     Debug.is_it_unsat gf;
     try
-      if Options.get_tableaux_cdcl () then
+      let pushed_assertions =
+        SE.fold (fun e acc ->
+            (mk_gf e "" true true, Ex.empty) :: acc
+          ) env.incremental.guards [] in
```
OriginLabs-Iguernlala, line 1770: `env.guards` can probably directly be a list of `(Expr.gformula * Ex.t)`

src/lib/reasoners/fun_sat.ml
```diff
@@ -1794,7 +1878,8 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
        inst = Inst.add_predicate env.inst gf dep }
     else
       begin
-        assert (not (ME.mem a_t env.ground_preds));
+        if Stack.is_empty env.incremental.stack_guard then
```
OriginLabs-Iguernlala, line 1881: pred_def (of ground and non-ground predicates), should take guards into account (see detailed commented in satml_frontend.ml)

OriginLabs-Iguernlala (changes requested):
There are probably other refs inside or outside fun_sat.ml and satml_frontend.ml that need to be saved (on push) and restored (on pop). Like for instance,
- steps counter
- fied `model_gen_mode` in fun_sat.ml
- other local refs related to model generation in fun_sat.ml
```
  let all_models_sat_env = ref None
  let latest_saved_env = ref None
  let terminated_normally = ref false
```
In fun_sat for instances, we should probably have a function similar to `reset_refs ()` to save/restore refs. The data can probably be directly stored in `incremental.stack_guard`. A similar solution should also work for satml_frontend.

src/lib/reasoners/fun_sat.ml
ACoquereau: This was needed in an older version and it's not needed anymore

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1023,9 +1037,63 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
     if get_model () then fails "model"
 
+  let push env to_push =
+    let rec push_aux acc n =
+      if n <= 0 then acc
+      else
+        let b = E.fresh_name Ty.Tbool in
+        let mf_b = mk_gf b in
+        let acc, pending = pre_assume_aux acc [mf_b] in
+
```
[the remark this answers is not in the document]
ACoquereau: The `SAT.push` function need an `Atom` (the guard) I create this guard by using the `pre_assume_aux` function  with `new_vars`. It should be better to create this atom in the SAT.push function, and lighter in the satml_frontend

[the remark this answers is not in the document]
OriginLabs-Iguernlala: You need to 
- use `Satml_frontend.simplify` to translate the fresh guard of type `Expr.t` to `Satml_types.t`
- either use `Satml_types.cnf_abstr` like in `Satml_frontend.pre_assume` (that translates the atom to a CNF), or write a simpler function that wraps `Satml_types.atom_of_lit`.

In this case, `Satml_types.atom_of_lit` should be exported with a better interface first (currently `new_vars` is a ref modified locally, it'd be better to take it as argument without a ref and return it in the result of `Satml_types.atom_of_lit`

Then `Satml_types.atom_of_lit`, will contruct the `Satml_type.atom` corresponding to the fresh atom, and return the set of new vars.  In satml_frontend, the field `new_vars` should be updated accordingly (see `pre_assume`).

OriginLabs-Iguernlala (approved):

OriginLabs-Iguernlala (changes requested):

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1044,9 +1035,67 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+        SAT.push acc.satml atom_guard;
+        Stack.push expr_guard acc.guards.stack_guard;
+        Steps.push_steps ();
+        {acc with guards =
+                    {acc.guards with
+                     current_guard = expr_guard;} }
+      )
+      ~max:to_push
+      ~elt:()
+      ~init:env
```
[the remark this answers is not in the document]
ACoquereau: We need the satml in `SAT.push  acc.satml atom_guard`.

PR #401, opened 2021-01-07
WIP: Should be benched: More deductions on arrays

iguerNL:


OriginLabs-Iguernlala:
CI fails because the patch need improvements to avoid infinite loops (deduction of the same fact in a loop)

PR #417, opened 2021-01-18
Fix printing of status when in interpretation mode and add clean fmt …

ACoquereau:
…in the print_fmt function


OriginLabs-Iguernlala:
The commit is pushed directly to optim-alt branch

PR #435, opened 2021-02-02
Improve purification process in expressions

iguerNL:
Mainly for "Let x = e1 in e2" in cases where e1 is a cascade of Ites


iguerNL:
This fixes the "bad" behavior on optimization file generated from opam (and preprocessed with ctrl-ergo), in which alt-ergo tries to purify a formula (that is already pure) and spends a lot of time (~ 70 seconds) on it.

The evaluation of the fix on our internal test-suite is in progress...

ACoquereau (approved):
Ok, if the results of the benchs are also ok

Issue #370, opened 2020-10-08
[build] add CI check for opam files


ACoquereau:
Check that opam files are up to date with dune-project informations

PR #436, opened 2021-02-04
Replace Travis CI with GH-actions

ACoquereau:
This work aims to replace the old travis CI.

Travis workflow was divided in 3 stages: `pre-check`, `test` and `documentation`.

- The `pre-check` stage is now done in the `format` workflow that include a `check-indentation` and a `check-style` jobs.
This light workflow is done for every `push`. Note that it's not blocking as `pre-check` was. The goal is to be able to test basic building and testing regardless of the code formatting. 

- The `documentation` stage is now done in the `documentation` workflow.
 For every `push` we try to build de ocaml documentation. Same as the `format` workflow, it is not blocking on error for testing.
 But it is blocking for the second job, the sphinx documentation building. This job only runs when a PR is open or on the `next` and `main` branchs.
 When the code is pushed on `next` a `deploy` job is triggered after the 2 previous one are ok. This job deploy the ocaml and sphinx documentation on the `gh-pages` branch. 
 The ocaml and sphinx building produce artifacts that allow developers to check if the produced documentation is ok.

- The `test` stage is done in the `build ubuntu` workflow.
For every `push` the code is build and installed with `opam` on an ubuntu system. If no error occurs, our non-regression testsuite is launched. 
More Checks are done when a PR is opened. We build and test more versions of the ocaml compiler.

Additional tests are now available:
- `build make` workflow test makefile rules. Basic `make` rule is test for every `push`. `Install/Uninstall` and packages rules are tested when a PR is opened.

- `build windows` and `build macos` workflows are only triggered when the code is pushed on either the `main` or `next` branch. As the `build ubuntu` workflow, we try to build with opam and test our solver.

- `build linux` only runs on `main` and `next` branch. This simple workflow aim to build on a minimalist opam docker container.


ACoquereau:
Note that `opam pin` and `opam depext` steps will not be mandatory when the workflows will use action with opam 2.1

ACoquereau:
This PR should also include check that resolve #370

ACoquereau:
> This PR should also include check that resolve #370
Done with  76012f0538716c6e6ea30b9171a6ba7fcef7759d

ACoquereau:
Workflows that runs only on the `next` branch can be preview in https://github.com/ACoquereau/alt-ergo/actions where a similar PR has been merged into next

.github/workflows/build_make.yml
```diff
@@ -0,0 +1,210 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Retrieve opam cache
+        uses: actions/cache@v2
+        id: cache-opam
+        with:
+          path: ~/.opam
+          key: v1-${{ runner.os }}-alt-ergo-${{ matrix.cache-name }}-${{ env.OCAML_DEFAULT_VERSION }}-${{ hashFiles('*.opam') }}
```
Gbury, line 33: is it really useful to have the matrix.cache-name part in a variable rather than manually inlined here ? I'd think the only interesting use of a variable is if we one day introduce more different builds in the matrix which need different caches, or do you have another use in mind ?

.github/workflows/build_make.yml
```diff
@@ -0,0 +1,210 @@
...
+        run: opam depext -y alt-ergo-lib alt-ergo-parsers alt-ergo altgr-ergo
+        # if: steps.cache-opam.outputs.cache-hit != 'true'
+
+      - name: Install deps
+        run: opam install -y ./*.opam --deps-only --with-test
+        if: steps.cache-opam.outputs.cache-hit != 'true'
+
+      - name: Fixup
+        run: opam upgrade --fixup
+        if: steps.cache-opam.outputs.cache-hit == 'true'
```
Gbury, line 58: Same as above, add more comments to most steps to better explain what they do

.github/workflows/build_make.yml
```diff
@@ -0,0 +1,210 @@
...
+        run: opam exec -- make
+
+      - name: Check opam files
+        # Check that opam files are unchanged
+        run: |
+          if git diff --exit-code *.opam ; then
+          echo "opam files ok" ;
+          else
+          exit 1 ;
+          fi
```
Gbury, line 71: That's a nice addition to the CI. Did you have time to look at github bots that could suggest the diff as changes to be included in the PR ?

.github/workflows/build_make.yml
```diff
@@ -0,0 +1,210 @@
...
+      - name: Build
+        run: |
+          opam depext -y alt-ergo-lib alt-ergo-parsers alt-ergo altgr-ergo
+          opam install -y ./*.opam --deps-only --with-test
+          opam exec -- make
+
+      - name: Install/Uninstall Alt-Ergo-lib
+        run: |
+          opam exec -- make install-lib
+          opam exec -- make uninstall-lib
```
Gbury, line 190: If you really want to test the installation/uninstallation process, then you should:
- after installing, use the installed version (there was an old script doing that iirc) and/or check the existence of files at the installation destination
- after uninstalling, ensure that no leftover files are left

This also applied to the the parser lib, the binary (which is easier to test), and the gui (more tricky to test)

.github/workflows/build_ubuntu.yml
```diff
@@ -0,0 +1,240 @@
+name: Build Ubuntu
```
Gbury, line 1: This is a bit confusing compared to the `build_linux` workflow, some comments (and renaming of workflows/jobs/others), would be helpful in understanding the different things tested. Is this test specific to ubuntu as in "check that alt-ergo can build using the versions of ocaml and deps that are available on ubuntu (as opposed to generic linux)" ?

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,114 @@
+name: Documentation
+
+# Trigger the documentation build for any push, pull request and any branch only deploy on next
```
Gbury, line 3: Trigger the documentation -> what part ? the build ? the upload ? a bit more specificity would help (I'd guess it's check that the documentation builds in this case).

.github/workflows/format.yml
```diff
@@ -0,0 +1,77 @@
+name: Format
```
Gbury, line 1: Typically such tests are more often called as lint/linter checks; it's be a better name considering that Format is a stdlib module and so for an OCaml project, it's a bit ambiguous to see that as a standalone name.

Gbury (changes requested):
Generally good but:
- the files lack a lot of documentation as comments about what is done and why
- there are a lof of tests and files compared to most other CI setups I've seen, and it's easy to get lost, so there should be a markdown file somewhere that explains and details the various tests that are done by the CI (and when these are done), and mentions some of the tecnical details (for instance list all the cache keys and explain which ones are used/created/updated by which jobs).
- what is the `unchanged file with check annotation` part of the PR (basically sphinx warnings about files unchanged by this PR) ? Is it a new thing activated by this PR ? Github marks that as beta, so it's possible it's juste github trying to add things on its own, but it's be nice to be sure.

.github/workflows/build_make.yml
ACoquereau: I've tried to test and use diff-action (https://github.com/marketplace/actions/get-diff-action) on my personnal fork. I was able to easily get the diff for `.opam` files. But the diff was between the `base branch` and the `target branch`. I will try to change the `from and to` https://github.com/technote-space/get-diff-action#from-to

ACoquereau:
>     * what is the `unchanged file with check annotation` part of the PR (basically sphinx warnings about files unchanged by this PR) ? Is it a new thing activated by this PR ? Github marks that as beta, so it's possible it's juste github trying to add things on its own, but it's be nice to be sure.

This part of the PR is automatically generated with the sphinx build action.
The sphinx documentation was added with this warning, I'll try to clean them.

.github/workflows/documentation.yml
ACoquereau: Is the new comments ok to you ?

.github/workflows/documentation.yml
```diff
@@ -1,110 +1,166 @@
...
-
-      - run: opam pin add . -y --no-action
-      - run: opam depext -y alt-ergo-lib alt-ergo-parsers alt-ergo altgr-ergo
-        # if: steps.cache-opam.outputs.cache-hit != 'true'
-      - run: opam install -y ./*.opam --deps-only --with-test
-        # if: steps.cache-opam.outputs.cache-hit != 'true'
+      # Add OPAMYES=true to the environment, this is usefill to replace `-y` option
+      # in any opam call
+      - name: Setup opam variables
+        run: export OPAMYES=true
```
Gbury, line 52: It's probably better to set the env as part of the env for the job (as in https://github.com/Gbury/dolmen/blob/master/.github/workflows/build.yml#L53 for instance), so that would mean putting the following right after the `runs-on` setting:
```
env:
  OPAMYES: 'true'
```

.github/workflows/documentation.yml
```diff
@@ -1,110 +1,166 @@
...
+
+      # Install external dependencies if the cache is not retrieved
+      # remove this step when opam 2.1 will be used
+      - name: opam install depext
+        run: opam depext alt-ergo-lib alt-ergo-parsers alt-ergo altgr-ergo
+        if: steps.cache-opam.outputs.cache-hit != 'true'
+
+      # Install dependencies if the cache is not retrieved
+      - name: opam install deps
+        run: opam install ./*.opam --deps-only --with-test
```
Gbury, line 67: Not sure the `--with-test` is useful here, however a `'--with-doc` would be useful here, and for the depext (or as an env var in the same way as for `OPAMYES`)

.github/workflows/documentation.yml
```diff
@@ -1,110 +1,166 @@
...
+        if: steps.cache-opam.outputs.cache-hit != 'true'
+
+      # Install odoc if the cache is not retrieved
+      - name: opam install odoc
+        run: opam install odoc
+        if: steps.cache-opam.outputs.cache-hit != 'true'
```
Gbury, line 73: Odoc should become a `{with-doc}` dependency so that this step is no longer needed

.github/workflows/documentation.yml
```diff
@@ -1,110 +1,166 @@
...
+
+  # For every push on next, retrieve ocaml and sphinx documentation
+  # and publish them on gh-pages branch
   deploy_docs:
-    # Deploy the documentation only on next
     name: Deploy documentation
+
     if: github.ref == 'refs/heads/next'
 
     needs: sphinx_docs
```
Gbury, line 140: Even if this works because of transitive dependencies, if possible I think it'd be better to explicitly state that this job needs the `build_docs` job (since one step explicitly downloads an artifact from that job).

Gbury (changes requested):

ACoquereau:
> > ```
> > * what is the `unchanged file with check annotation` part of the PR (basically sphinx warnings about files unchanged by this PR) ? Is it a new thing activated by this PR ? Github marks that as beta, so it's possible it's juste github trying to add things on its own, but it's be nice to be sure.
> > ```
> 
> This part of the PR is automatically generated with the sphinx build action.
> The sphinx documentation was added with this warning, I'll try to clean them.

Clean in #440

.github/workflows/build_make.yml
ACoquereau: I've used this for previous tests. This is not usefull anymore.

.github/workflows/build_make.yml
ACoquereau: I've added a script that test if the bins (bin and gui install rules) and the libs (lib and parsers install rules) are accessible either with ocamlfind for a lib or by running `<bin> --version`.
This command must succed after the `install` step and must fails when we run the script after the `uninstall` step.

.github/workflows/build_make.yml
ACoquereau: > I've tried to test and use diff-action (https://github.com/marketplace/actions/get-diff-action) on my personnal fork. I was able to easily get the diff for `.opam` files. But the diff was between the `base branch` and the `target branch`. I will try to change the `from and to` https://github.com/technote-space/get-diff-action#from-to

This solution did not worked.
I've look at https://github.com/marketplace/actions/action-git-diff-suggestions as suggested. The results are way better.
Note that this action only blocking if a PR is open. I suppose this  is still acceptable ?

.github/workflows/build_macos.yml
```diff
@@ -0,0 +1,122 @@
...
+          path: ~/.opam
+          key: v1-${{ runner.os }}-alt-ergo-${{ env.OCAML_DEFAULT_VERSION }}-${{ hashFiles('*.opam') }}
+
+      - name: Use OCaml ${{ env.OCAML_DEFAULT_VERSION }}
+        uses: avsm/setup-ocaml@v1
+        with:
+          ocaml-version: ${{ env.OCAML_DEFAULT_VERSION }}
+
+      - name: Run non regression tests
+        run: opam exec -- rsc/extra/non_regression.sh
```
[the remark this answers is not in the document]
ACoquereau: It only re-uses the previous opam cache which already contains the alt-ergo binary. The checkout and ocaml-setup are needed to retrieve the non-regression script, files, and opam (used to run the non-regression script).

At this time no deps or depexts are needed to test this project. I agree with you on the fact that we need to convert this tests to use `dune`. An issue is open (#441) and should be fixed in the futur. Since then, i've added some steps to install specific deps and depexts.

.github/workflows/build_ubuntu.yml
ACoquereau: `build_linux` is now `build_docker` it should be clearer

ACoquereau:
@Gbury, in 5f57c4e3db4fa31350477c4df0dda80f05a2cb6c, I've merged the `tests` and `upload` jobs with the `install` ones. It make all workflow run faster and it's way simpler to read and maintain.

PR #442, opened 2021-02-13
[WIP][Don't merge] re-implement all-models feature differently

iguerNL:
This PR partially rely on a modification in psmt2-frontend.

- two new options (files option.ml*, parse_command.ml and models.ml):
  + all-models: to enable generation of all (propositional) models (+ concrete theories interpretation if --interpretation is set)
  + show-propositional-model: to enable display of propositional when --interpretation or --all-models are enabled
 the first option is not used in models.ml

- cleaning in theory.ml now that we're able to filter the boolean model from the SAT (in case of check-all-sat SMT2 command),

- add a function get_propositional_model in SAT solvers (used when all-models in set but not --interpretation) (sat_solver_sig.ml*, fun_sat.ml, satml_frontend.ml)

- other modified files, except frontend.ml: modifications to support the new SMT2 command `(check-all-sat <list of boolean variables>)` added to psmt2-frontend. This command is not part of the standard, but is implemented in MathSAT5. Useful to generate all models by filtering by the set of given bool variables.

- file frontend.ml contains the main modification to implement the feature.
+ the function `process_decl` is now recursive. It doesn't directly return on SAT/Unknown, but calls an auxiliary function `process_unknown` to analyze the situation:

1. check if we are in `all-sat` mode (in which case, return `Some {set of filter}`) or not (in which case, return `None`),
2. get the computed model if any (a Model.t computed with --interpretation, or a partial Model.t containing only the propositional part if in --all-sat mode).
3. the model is filtered, if in `all-sat mode` and some filter is given.
4. the filtered model is printed
5. if in `all-sat mode`, function `process_decl` is called recursively on the negation of the filtered propositional model to force the SAT to explore more branchs.

done: Would be nice to go beyond timeouts in interpretations generation (can be caused by loops due to non-linear arithmetic), and continue exploring other branches,

TODO:
- Not working with Fun_sat
- Some extensions needed in the GUI for modifications made in the frontend

### DIFF
--- a/src/lib/frontend/frontend.ml
+++ b/src/lib/frontend/frontend.ml
@@ -149,10 +149,11 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
     | Some SAT.ProofSearch -> "ProofSearch"
     | Some SAT.ModelGen -> "ModelGen"
 
-  let print_model env timeout =
-    let get_m = Options.get_interpretation () in
+  let print_model ?(all_sat=false) model_opt timeout =
+    let pp_prop_model = all_sat || Options.get_show_prop_model () in
+    let get_m = Options.get_interpretation () || pp_prop_model in
     let s = timeout_reason_to_string timeout in
-    match SAT.get_model env with
+    match model_opt with
     | None ->
       if get_m then
         Printer.print_fmt (Options.get_fmt_err ())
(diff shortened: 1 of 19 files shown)


src/lib/frontend/frontend.ml
```diff
@@ -250,29 +258,105 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
...
+        (* 1. case where all-bool-models wrt. given filter is
+           requested *)
+        let m = Lazy.force m in
+        let propositional = filter_by_all_sat m.propositional filter in
+        let m = { m with propositional } in
+        print_model (Some (lazy m)) timeout_kind;
+        let f =
+          Expr.Set.fold
+            (fun e acc -> E.mk_and e acc false 0) propositional E.vrai
+        in
```
ACoquereau, line 349: Add some explanation on what is done here. `not f` (propositional model) is added as a query to force resolution `-` the previous sat model

ACoquereau (approved):

ACoquereau (approved):

PR #445, opened 2021-02-17
Caching the result of X.make

Stevendeo:


src/lib/reasoners/uf.ml
```diff
@@ -553,7 +553,7 @@ module Env = struct
       ) env (X.leaves mkr)
 
   let init_term env t =
-    let mkr, ctx = X.make t in
+    let mkr, ctx = X.make_semantic t in
     let rp, ex = normal_form env mkr in
     let env =
       {env with
```
iguerNL, line 559: There are other few (at least one) usage of X.make in Uf.ml that could be replaced with X.make_semantic

iguerNL: The same remark applies for usage of X.make in intervalCalculus.ml and ccx.ml

iguerNL (changes requested):
I think adding a cache is good, but we can do things slightly better.
- do not a function `make_semantic` anymore, but keep using `X.make`
- In shostak.ml, instead of the current modification, we'll rather have at the end of the file:
```
module HE = Hashtbl.Make(Expr)
module Combine = struct
  include CX
  let make =
    let cache = HE.create 1024 in
    fun t ->
      ....
``` 

This way, we enforce the use of the caching version of X.make outside of Shostak theories combination.
end
```

iguerNL:
Replaced with https://github.com/OCamlPro/alt-ergo/pull/446

PR #453, opened 2021-03-10
Add --numbers-lib option at configure. This options allow user to cho…

ACoquereau:
…ose between Zarith (default) and Nums library to handle numbers in Alt-Ergo

### DIFF
--- a/configure.ml
+++ b/configure.ml
@@ -22,6 +22,25 @@ let static = ref false
 
 let pkg = ref ""
 
+(* this option add the possibility to choose the library that handle numbers in Alt-Ergo between zarith and nums *)
+type numbers_lib =
+  | Nums
+  | Zarith
+
+let print_numbers_lib = function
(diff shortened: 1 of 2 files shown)


configure.ml
```diff
@@ -181,7 +206,10 @@ let () =
 
 (* run dune to check that dependencies are installed *)
 let () =
-  let p_opt = match !pkg with "" -> "" | s -> Format.asprintf "-p %s" s in
+  let p_opt =
+    match !pkg with
+    | "" -> "--only-packages=alt-ergo-lib,alt-ergo-parsers,alt-ergo,altgr-ergo"
+    | s -> Format.asprintf "-p %s" s in
```
ACoquereau, line 212:
```suggestion
  let p_opt = match !pkg with "" -> "" | s -> Format.asprintf "-p %s" s in
```

PR #302, opened 2020-04-21
Options reorganisation

OCamlPro-mattiasdrp:
This PR aims to separate the options provided by the lib from the parsing of the command line.

All options now have a proper getter/setter and can be accessed with `get_<name_option>` and set with `set_<name_option>`

A new internal lib has been created, common, that contains the common functions for the binaries (here, only the parse_command functions which use cmdliner)

This will allow anyone to use the library and to be able to set any option they want (useful for try alt-ergo, for example).

PR #456, opened 2021-03-17
Js worker

ACoquereau:
This PR is the second part of the split of #299 (part 1 : #454 ).

This PR adds some functions in parsing and for context retrieving (functions that were already used in previous version of AE and in the previous version of Try-Alt-Ergo).

This PR adds a web-worker that runs Alt-Ergo with the make rule `make js-worker`. This webworker uses json file as input and output. As the nodejs compatible version of AE, this make rule needs Js_of_ocaml. It also uses the data_encoding library to encode and decode json file. 
Instead of #299 the webworker (and js-node) will not be package with opam. This will simplify every other build.


src/lib/frontend/input.mli
```diff
@@ -34,6 +34,9 @@ module type S = sig
   type parsed
   (** The type of a parsed statement. *)
 
+  val parse_file : file:string -> format:string option -> parsed Seq.t
+  (** Parse a file as a string with the given format or the input_format set *)
+
```
iguerNL, line 39: Maybe rename the `file` label to `content` to be more explicit that it's the content of the file ?

src/lib/frontend/frontend.ml
```diff
@@ -276,8 +276,8 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
       Printer.print_status_unsat ~validity_mode
         (Some loc) (Some time) (Some steps) (get_goal_name d);
       if get_unsat_core() &&
-         not (get_debug_unsat_core()) &&
-         not (get_save_used_context())
+         not (get_debug_unsat_core())
+         (* && not (get_save_used_context()) *)
```
iguerNL, line 280: Why removing this test ? it's here to not print the UNSAT core if the user asked to save the used context (name of axioms) in a `.used` file.

src/bin/js/worker_interface.ml
```diff
@@ -0,0 +1,995 @@
...
+(*     present please contact us to clarify licensing.                        *)
+(*                                                                            *)
+(******************************************************************************)
+open Js_of_ocaml
+open Data_encoding
+
+(** Types extract from AltErgoLib Utils.util and Utils.options *)
+
+type model = MNone | MDefault | MAll | MComplete
+
```
iguerNL, line 17: Not sure we should redefine all the types in options.ml here. We can directly use those defined in Options. This will (hopefully) allow to modify this file when types in options.ml evolve (thanks to the compiler)

iguerNL (commented):
I had a first look at the modifications. We should maybe discuss some of them:

It's maybe a lot of modifications, but couldn't we put type `options` in options.ml ? it may have pros and cons. We can discuss it.

Using ez_jslib for the Js_of_ocaml part would allow to "hide" the object-oriented style for future maintainers.

I think `retrieve_used_context` is an old code that have been simplified. Now, a `RootDep` (Explanation.elt) is added in Frontend if unsat-core generation is enabled.

```
  let mk_root_dep f_name =
    if Options.get_unsat_core () then Ex.singleton (Ex.RootDep f_name)
    else Ex.empty
```

So At the end, you have just to inspect the explanation of unsatisfiability to extract used context (but this only works for Tableaux SAT solvers I guess, as we didn't implement it in CDCL yet).

```
(* In Explanation.ml *)
let print_unsat_core ?(tab=false) fmt dep =
  iter_atoms
    (function
      | RootDep s ->
        if tab then Format.fprintf fmt "  %s@." s (* tab is too big *)
        else Format.fprintf fmt "%s@." s
      | Dep _ -> ()
      | Bj _ | Fresh _ | Literal _ -> assert false
    ) dep
```

src/lib/frontend/frontend.ml
ACoquereau: This has been revert since we do not use save_used_context anymore

src/bin/js/index.mld
```diff
@@ -5,8 +5,55 @@
...
+
+To run [make js-worker] you will need the following additionnal dependencies: [js_of_ocaml] [js_of_ocaml-lwt] [data-encoding] and [zarith_stubs_js].
+
+The [--timeout] option is ignored due to a lack of js primitives
+Zip files are not recognised file extension.
+
+{2 The Alt-ergo web worker small example}
+
+A small example of how to use the Alt-Ergo web worker can be build with the command [make js-example]. This command also makes the web worker if it has not already been built. It produces a [www] directory with an [index.html] file.
+
```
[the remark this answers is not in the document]
ACoquereau: The worker, worker_interface and example are documented in the sphinx doc.

ACoquereau:
> I think `retrieve_used_context` is an old code that have been simplified. Now, a `RootDep` (Explanation.elt) is added in Frontend if unsat-core generation is enabled.
> 
c7f44e8 uses RootDep instead of retrieve_used_context. This new version have simplified this PR with the removal of retrieve_used_context part

src/bin/js/worker_interface.ml
ACoquereau: I choose to create worker_interface as independent as possible from the alt-ergo-lib options file. OCaml users of the webworker can easily get this helper file and use the functions to convert options to json file or recover results from a json file sent by the worker.

Put the type `options` in options.ml was an idea raised at the time of #302, to improve the usability of the Alt-Ergo-lib. This solution should need some discussion for a future improvement.

Makefile
```diff
@@ -316,7 +316,24 @@ js-node: gen
...
+# zarith_stubs_js, js_of_ocaml and js_of_ocaml-lwt packages are needed for this rule
+js-worker: gen
+	$(DUNE) build $(DUNE_FLAGS) --profile=release $(BJS_DIR)/worker_js.bc.js
+	ln -sf $(DEFAULT_DIR)/$(BJS_DIR)/worker_js.bc.js alt-ergo-worker.js \
+
+# Build a small web example using the alt-ergo web worker
+# This example is available in the www/ directory
+# zarith_stubs_js, js_of_ocaml and js_of_ocaml-lwt js_of_ocaml-ppx lwt_ppx packages are needed for this rule
+js-example: gen js-worker
+	$(DUNE) build $(DUNE_FLAGS) --profile=release $(BJS_DIR)/worker_example.bc.js
```
iguerNL, line 329: `gen is already a dependency of js-worker. Can be removed here

PR #462, opened 2021-05-05
Purification bugfix

Stevendeo:
In few cases, skolemization outputs an non-purified expression, i.e. an expression with formulas inside terms that are not defined as let-bindings. This causes Alt-Ergo to fail on few examples.
This PR adds a purification step while building quantified formulas, which fixes the issue.

* [x] Fix
* [x] Bench


Gbury (approved):
The fix seems safe as far as semantics goes, and if it fixes the bug, it's good.
However, I'm a bit concerned about the potential exponential complexity that can occur with this patch (it requires a lot of nested quantification, so it should presumably not happen very often, but we should try and benchmark this change).

Stevendeo:
I have succesfully benched it on large formulas and I left a 'todo' note in the comments about it.

Issue #460, opened 2021-04-16
Unsound regression in 2.4.0


kanigsson:
[repr.ae.txt](https://github.com/OCamlPro/alt-ergo/files/6322106/repr.ae.txt)

Hello,

it seems that the unsoundness in #248 has been reintroduced in 2.4.0. The reproducer filed there creates an internal assertion failure, but the attached similar reproducer proves:
```
alt-ergo repr.ae
File "repr.ae", line 25, characters 3-57: Valid (0.0056) (27 steps) (goal defqtvc)
```
Older versions of alt-ergo don't prove the VC (commit e573c3dc929370ed403b40fc1db79985c63be140).

The two axioms are pretty clearly not contradictory (one can see that alt-ergo doesn't derive a contradiction just from the axioms by changing the `2 <= y` in the goal to `3 <= y`). The goal states that `y` is greater or equal to 2 even though we know close to nothing about y, except that is in in a range that starts at 1 and that it is used to define z.

kanigsson:
Hello, I just noticed that the same unsoundness is also present in the `next` branch.

Gbury:
I just submitted a patch to fix this issue. This should be merged very soon and be part of the next release, ^^

Gbury:
@kanigsson now that the fix has been merged in the `next` branch, can you try to see if that solved your issue ?

Gbury:
@kanigsson ping ? (without any feedback, i'll close this issue soon since PR #465 fixes the problem on the example provided).

kanigsson:
I tested only the branch `fix_multiple_rounding`, and it seems to fix this bug here and the one I reported privately. Thanks a lot. I let you close the ticket if appropriate.

Gbury:
Thanks, closing this issue since #465 and #487 have been merged

PR #465, opened 2021-06-30
Fix issue #460

Gbury:
This PR should fix issue #460. It reworks the fix that was done in PR#342, which was intended to fix issue #340, but also introduced the bug found by #460.

The first commit adds regression tests for issues #340 and #460, and the second commit fixes the bug (see the message in the comments for more information).


Gbury:
So, there were a few problems with the non-regressions scripts:
- first, the scripts did not run alt-ergo on `.smt2` files, so I've added that to the scripts
- as a consequence, one test in `valid/quantifiers` was not tested and was actually failing, I moved this test to the challenge section
- I reworked the main script to have a more compact output and more explicitly show when a test fail

At one point, we should ensure that the same script runs in the CI and when using `make`, but that can be done later.

Anyway, now the tests should pass now, ^^

Gbury:
So, travis CI has changed quite a bit, and the current setup stopped working when travis-ci.org stopped, i.e. June 15th, so now that we have the github actions, maybe we could remove the  `.travis.yml` and disable travis CI (or we can enable travsi CI on travis-ci.com iiuc) ? cc @Stevendeo

PR #472, opened 2021-08-06
Fix issue 350

### LINKED ISSUE #350: [Bug] Assertion failure in lib/reasoners/shostak.ml when reasoning on bitv

Stevendeo:
This is a test for fixing issue #350.

The fix consists to produce in `Bitv` fresh variables from `Expr` instead of internal variables of `Bitv`. In such way, Shostak knows that the variable is uninterpreted for `Bitv` and we respect the invariant enforced by the `assert false`.


src/lib/reasoners/shostak.ml
```diff
@@ -500,9 +500,11 @@ struct
       List.filter (fun (p,_) ->
           match p.v with
           | Ac _ -> true | Term _ -> SX.mem p original
-          | _ ->
-            Printer.print_err "%a" CX.print p;
-            assert false
```
Stevendeo, line 505: @iguerNL 
I did not find what enforces this invariant ; do you remember what should have forced it to be either Ac or a Term ?

Halbaroth (approved):
Fine!

Gbury:
@Stevendeo do you have any insight/explanation as to what's happening, and why the changes fixes the issue ?

Halbaroth:
It is broken with Dolmen frontend... I am fixing it.

Stevendeo:
> @Stevendeo do you have any insight/explanation as to what's happening, and why the changes fixes the issue ?

@Gbury I do not remember, but I'm fairly certain the fix does not solve anything ; that's why I left it a draft.

Stevendeo:
@Gbury I do not remember, I actually forgot I did not put it as WIP. If it works, I'm guessing it is a miracle.

Halbaroth:
You did. I thought you knew what you did ;) I will investigate so :D

Halbaroth:
After a long dive in the codebase yesterday, I can confirm that the Steven's fix is not correct. 

The issue comes from the way Alt-Ergo manages `uninterpreted semantic values` in the Shostak module. By `uninterpreted semantic value`, I mean a semantic value whose the top symbol is uninterpreted. For instance:
- `x`
- `f(34)` 
- `f(g(5)+5)`

But `x+y` doesn't fall into this category since its top symbol is interpreted in the `Arith` theory (even if, it is not interpreted in the usual sense since `x` and `y` are variables).

Now, you have two important functions defined in the Shostak module: `solve` and `leaves`. The `leaves` function returns the direct uninterpreted subvalues of a semantic value. For instance:
- `leaves(x) = []`
- `leaves(f(g(5)+5)) = [g(5)]`

The function `solve` tries to solve a system of equations between semantic values and in case of success, it returns a sequence of substitutions. More precisely, these substitutions replace some leaves in the members of the equalities by semantic values. So the substitutions are of the form : 
``` r1 --> R1, r2 --> R2, ...```
where `r1, r2, ...` are *uninterpreted* semantic value.

Unfortunately, the Bitvector theory produces wrong substitutions. The module produces substitutions whose the left member is interpreted. By miracle, these terms are not necessary to solve the equations, so if you break the invariant above, AE successes.

Actually, it makes sense that Bitv produces such substitutions. The module has it own variable type. For the global theory, a variable produces in Bitv looks like an interpreted semantic value of Bitv and the Shostak module has no way to know that it´s actually a variable since the type of semantic value of Bitv is abstract.

I had notice this strange design before. I am thinking now how to change it properly.

Halbaroth:
I am testing the patch on Marvin.

Halbaroth:
No regression found :)

bclement-ocp:
I don't think the fix is correct:

 - First `Var` is used for quantified variables, not skolem variables, so it should be `Name` that is used, not `Var`
 - Second, the size [here](https://github.com/OCamlPro/alt-ergo/pull/472/files#diff-f5c5d1437bc32718cda79735910071ec38afb11cd0701bae3498f276861d7a36R591-R593) is incorrect and unsound
 - Third, we return fresh variables as leaves, but we don't take that into consideration when applying substitutions; I am not sure why this actually works without changing `is_mine`.

In any case, even with the fix for the size mentioned above and the use of `Name` instead of `Var` this causes the unsoundness in #664.

bclement-ocp:
OK, I figured it out – as I suspected, the unsoundness in #664 comes from this point:

> Third, we return fresh variables as leaves, but we don't take that into consideration when applying substitutions; I am not sure why this actually works without changing is_mine.

However I was slightly wrong — `is_mine` doesn't need to be changed because `subst_rec` is changed, but `leaves` is not changed — so we lie to the Shostak module and hide the fact that we depend on the variables, and they never get substituted.

Closing this, I will open a new PR.

PR #478, opened 2021-08-17
Solving issue 475

### LINKED ISSUE #475: Assertion failure in "src/lib/reasoners/satml_frontend.ml", line 1019, when using the CDCL solver
[c_1130_l_bis.ae.txt](https://github.com/OCamlPro/alt-ergo/files/6982313/c_1130_l_bis.ae.txt)

Running the command:
```
./alt-ergo --sat-solver CDCL c_1130_l_bis.ae
```
Causes the following error:
```
Fatal error: exception File "src/lib/reasoners/satml_frontend.ml", line 1019, characters 13-19: Assertion failed
```

And when the other solvers are used (by replacing CDCL with Tableaux or Tableaux-CDCL or CDCL-Tableaux) then Alt-Ergo responds with "I don't know".

Some additional interesting points:
  - if we remove the axiom ```ax_1```, AE says: "I Don't Know".
  - if we replace the nested pattern matching between the lines 20 and 23 with ```buqv0``` or ```v2_1_2``` or ```v2_1_1``` or ```false```, AE says: "I Don't Know", if we replace it with ```true``` AE says: "Valid".
  - if we replace the line 26 with ```true``` or ```false```, AE says: "I Don't Know".

Stevendeo:


Halbaroth:
I forgot Steven suggested this patch years ago. After exploring the issue #475 by myself, I arrived to the same conclusion. 
It seems we had an exception handler during the instantiation pass of SatML frontend in order to catch the exception `Unsat` there. Indeed, some theories of AE are not complete, which means their reasoners cannot detect all the contradictions in the current boolean model of the SAT solver (modulo these theories) during the theory constraint propagation. In this case, we could detect a contradiction during the instantiation of some axioms.

It seems we forgot to reraise also the exception `I_dont_know` in this handler. This one is raised if the instantiation engine doesn't make any progress. The `updated` variable above shows when the environment does not change after some instantiation passes.
https://github.com/OCamlPro/alt-ergo/blob/4c581c56bf8987a1f7a3f607a7d5d78a117171f6/src/lib/reasoners/satml_frontend.ml#L995-L999

I guess the bug is really rare because the current code of `unsat_rec` does not catch the exception `I_dont_know` during its first iteration.

src/lib/structures/errors.ml
```diff
@@ -95,7 +95,11 @@ exception Error of error
 
 let error e = raise (Error e)
 
-let typing_error e loc =
+let typing_error ?ae_loc e loc =
+  let () =
+    match ae_loc with
+    | None -> ()
+    | Some ae_loc -> Format.printf "Error at %s@." ae_loc in
```
bclement-ocp: This has nothing to do here

src/lib/structures/errors.mli
```diff
@@ -109,7 +109,7 @@ exception Error of error
 val error : error -> 'a
 
 (** Raise the input {!typing_error} as {!Typing_error} *)
-val typing_error : typing_error -> Loc.t -> 'a
+val typing_error : ?ae_loc:string -> typing_error -> Loc.t -> 'a
```
bclement-ocp: Ditto.

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1015,9 +1015,9 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
         unsat_rec env ~first_call:false
       with
       | Satml.Unsat lc -> raise (IUnsat (env, make_explanation lc))
+      | I_dont_know _ as e -> raise e
```
```suggestion
      | I_dont_know _ as e ->
          let bt = Printexc.get_raw_backtrace () in
          Printexc.raise_with_backtrace e bt
```

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1015,9 +1015,9 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
       | _ -> assert false
```
bclement-ocp: Let's leave a comment here explaining why we don't expect any exception here.

Alternatively, is there a reason to enforce that the code within never raises an exception? Catch-all `try ... with` are usually not a good idea.

src/lib/structures/errors.ml
Halbaroth: I agree but it was like this in this PR and I won't create a separated PR for such a tiny and inoffensive modification.

src/lib/reasoners/satml_frontend.ml
Halbaroth: Pierre told me it was the right way to reraise an exception in OCaml in order to preserve the backtrace as much as possible. Sould we do this modification everywhere in the codebase?

bclement-ocp: Ah, right, actuall we don't do anything with `e` so that should be fine, my bad.

src/lib/structures/errors.ml
bclement-ocp: I am not asking to make a separate PR for this change, I am asking that we remove it. This is clearly a temporary print that was added for debugging and shouldn't make it into the actual code, especially since it is not used at all.

Halbaroth: Ok :)

src/lib/reasoners/satml_frontend.ml
Halbaroth: In my opinion, catching all the exceptions is ok if you raise an uncatchable exception in the handler as `assert false`. We can try without catching all the exceptions but we need to test it :)

bclement-ocp: Testing is a bit pointless because if testing *with* the `assert false` does not fail, it means no exception is raised, and so we will not see the difference, no?

Having the catch-all `assert false` here means that if the user hits `Ctrl-C` (`Sys.Break`), or if the timeout triggers within this specific piece of code (`Util.Timeout` -- it is inside the solver so I believe the timeout is active while we execute this code), we will die with an assertion failure instead, which is unexpected and prevents normal handling.

By the way, there are no uncatchable exceptions:
```ocaml
# try assert false with Assert_failure _ -> () ;;
- : unit = ()
```

Halbaroth: I know there is no uncatchable exceptions but catching an assert false is a big mistake :) 

You are right. It should actually fail :)

Halbaroth:
It is ready :)

bclement-ocp:
There is still an `assert false` in there, my understanding of the discussion is that we agreed we should remove it (e.g. in case a timeout occurs)?

Halbaroth:
Oups I pushed it on my repository. I forgot it was a Steven's PR :)

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1015,8 +1015,7 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
         unsat_rec env ~first_call:false
       with
       | Satml.Unsat lc -> raise (IUnsat (env, make_explanation lc))
-      | _ -> assert false
-
+      | I_dont_know _ as e -> raise e
```
bclement-ocp: If we remove the `assert false` branch we no longer need this, it will be the default behavior

Halbaroth:
It is ready ;)

bclement-ocp:
There are still no changes to this PR, and the `I_dont_know` case is still there.

Halbaroth:
There is changes https://github.com/OCamlPro/alt-ergo/pull/478/commits/d770377ea449f705f7a437ff1f2fb772e8ebe9c0 but sure we can remove the I_dont_know case!

bclement-ocp:
Weird, GitHub did not show me any new commits since my last comment… But it also shows that commit before my comment, which it was definitely not yesterday since it exactly addresses the comment.

Anyways, that looks good now :tada:

bclement-ocp (approved):

PR #484, opened 2021-09-09
fix [partial-match] warning in psmt2_to_alt_ergo

hra687261:


Gbury:
the last commit doesn't belong to this PR, and should probably be submitted as a separate PR

Issue #248, opened 2019-09-09
unsoundness on non-linear arithmethics


kanigsson:
Hello Alt-ergo team,

We believe we have found an unsoundness on this small file:
```
logic div : int, int -> int

logic bit_index : int

predicate in_range(x: int) = (((- 2147483648) <= x) and (x <= 2147483647))

predicate in_range2(x: int) = ((1 <= x) and (x <= 17179869183))

axiom cdiv_cases :
 (forall n:int. forall d:int [div(n, d)].
 (((0 <= n) -> ((0 < d) -> (div(n, d) = (n / d))))
 and (((n < 0) -> ((0 < d) -> (div(n, d) = (-((-n) / d))))))))

axiom Div_1 : (forall x:int. (div(x, 1) = x))

goal VC_1 : (in_range2 (bit_index) ->
 (in_range ((div((bit_index - 1), 8) + 1))))

(*goal VC_2 : (not
 (in_range ((div((17179869183 - 1), 8) + 1))))*)
```

Our comments:

- `VC_1` proves with option `-max-split 5`, without it alt-ergo suffers from an internal crash
- when we comment `VC_1` and uncomment `VC_2`, it proves too, which is bad since `VC_2` is `(not VC_1)` when `bit_index` is `17179869183`
- `cdiv_cases` cannot be unsound since it gives a value to an otherwise undefined symbol
- `Div_1` can be proved from `cdiv_cases` if we comment `VC1` and `VC2` and change the axiom to  be a goal, so it cannot introduce an unsoundness either.
- We have a slightly modified version of alt-ergo, can you tell us if you can reproduce the unsoundness on the regular 2.3 version?

Gbury:
We can reproduce using the `next` branch, so it is likely the bug is also there in `2.3`.
The internal crash has been fixed in fd7c757c917a56b2449d7badc5f8f4bf719cb109
We are still looking at how and why alt-ergo manages to prove VC1 when it is incorrect (doing it by hand, we have that : `div(17179869183 - 1, 8) + 1 = 2147483648`, thus VC1 should not be provable).

kanigsson:
Were you able to find anything? This is a quite urgent issue for us, as it directly impacts the soundness of SPARK.

OriginLabs-Iguernlala:
@Gbury and @OCamlPro-Coquera managed to locate the origin of the issue. We'll see what is the best fix

kanigsson:
Thank you for the update.

[...]

PR #487, opened 2021-10-08
Avoid multiple rounding of intervals for polynomials

Gbury:
Another instance of multiple rounding causing unsoundness. In essence the fix is very similar to what was done in PR #249 to fix Issue #248 .

### DIFF
--- a/src/lib/reasoners/intervalCalculus.ml
+++ b/src/lib/reasoners/intervalCalculus.ml
@@ -448,7 +448,7 @@ let generic_find ?(only_inf=false) xp env =
        interval, the function will raise, in order to let its caller know the
        interval is not consistent/reachable/satisfiable.
        This interacts badly with callers who only care about the lower bound,
-       as these callers do not catch the inconsisten exception, and this can
+       as these callers do not catch the inconsistent exception, and this can
        lead to the solver returning Unsat/Valid even when it is incorrect to
        do so. (see issues #460 #340, and PRs #342 and #465). *)
     let ip =
@@ -494,11 +494,18 @@ let generic_add x j use is_mon env =
 module Debug = struct
   open Printer
 
-  let assume a expl =
+  let add p =
+    if get_debug_fm () then
+      print_dbg
+        ~module_name:"IntervalCalculus" ~function_name:"add"
+        "@[<v 2>New polynome added: %a]" P.print p
+
+  let assume ~query a expl =
     if get_debug_fm () then
       print_dbg
         ~module_name:"IntervalCalculus" ~function_name:"assume"
-        "@[<v 2>We assume: %a@,explanations: %a@]"
+        "@[<v 2>%s We assume: %a@,explanations: %a@]"
+        (if query then "[query]" else "")
         LR.print (LR.make a)
         Explanation.print expl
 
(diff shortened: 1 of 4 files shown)


src/lib/reasoners/intervalCalculus.ml
```diff
@@ -675,17 +682,21 @@ let mult_bornes_vars vars env ty =
...
-
+  (* in the case of integer polynomes, ensure that we only round once
+     at the end, and not at every add/scale operation on intervals. *)
+  let rational_interval =
+    List.fold_left
+      (fun i (a, x) ->
+         let i_x, _  =
+           try MX.n_find x env.monomes
+           with Not_found ->
+             if monomes_inited then assert false;
```
Stevendeo: Style : could be simplified by `assert (not monomes_inited)`

src/lib/reasoners/intervals.ml
```diff
@@ -73,7 +73,7 @@ module Debug = struct
       print_interval fmt e;
       List.iter (fprintf fmt " U %a" print_interval) l
 
-  let print fmt { ints; expl = e; _ } =
+  let print fmt { ints; expl = e; is_int = _; } =
```
Stevendeo: Style : is this change necessary?

src/lib/reasoners/intervals.ml
```diff
@@ -492,29 +492,34 @@ let scale_interval_neg n (b1, b2) =
 
 let affine_scale ~const ~coef uints =
   Options.tool_req 4 "TR-Arith-Axiomes scale";
-  if Q.equal coef Q.one then
-    { uints with ints = List.map (translate const) uints.ints; }
```
Stevendeo, line 496: Why did you remove the case where the coef is one?

src/lib/reasoners/polynome.ml
```diff
@@ -168,7 +168,7 @@ module Make (X : S) = struct
       else begin
         M.iter
           (fun t n -> fprintf fmt "%s*%a " (Q.to_string n) X.print t) p.m;
-        fprintf fmt "%s" (Q.to_string p.c);
+        fprintf fmt "+ %s" (Q.to_string p.c);
```
Stevendeo: Style : if `p.c < 0`, you may want not to print `+`

src/lib/reasoners/intervals.ml
Gbury: Because it was actually incorrect: if the constant for the translation is rational, then integer intervals need to be rounded even when the coefficient for the scaling is `1`.

Additionally, even if the constant for the translation is an integer, it is useful to use this function as a "round this interval" function, which is precisely what the new `coerce` function does (it performs an `affine_scale` of `coef=1` and `const=0`).

src/lib/reasoners/polynome.ml
Gbury: fixed

src/lib/reasoners/intervalCalculus.ml
Gbury: done

src/lib/reasoners/intervals.ml
Gbury: not necessary, I reverted the change

Gbury:
I pushed a commit to address the review.

Stevendeo:
> I pushed a commit to address the review.

Thanks! 
Now the CI complains because of the indentation but after this it is fine for me

Issue #482, opened 2021-08-23
Assertion failure in "src/lib/structures/expr.ml", line 2863


hra687261:
[co_2112_l.ae.txt](https://github.com/OCamlPro/alt-ergo/files/7032462/co_2112_l.ae.txt)

Running Alt-Ergo on the attached file causes an assertion failure in the formula purification function in ```expr.ml```.

if we replace the pattern matching between the lines 15 and 18 with ```ff(eqadt_adt_10)```, then the exception is not raised.

hra687261:
Fixed by #489

hra687261:
Fixed by #489

PR #489, opened 2021-10-22
Fix issue 482

hra687261:
Purification of functional array expressions, fixes issue #482.


src/lib/structures/expr.ml
```diff
@@ -2433,6 +2433,30 @@ module Purification = struct
           | _, (B_lemma _ | B_skolem _ | B_none | B_let _) -> assert false
         end
 
+      (* When e is an access to a functional array
+         in which the stored values are booleans *)
+      | Sy.Op Get ->
+        begin match e.xs with
+          | [x; y] ->
```
Gbury: Please give more informative names to these variables. For instance, to better distinguish which is the expression for the array, and which is the expression for the looked-up index.

src/lib/structures/expr.ml
```diff
@@ -2433,6 +2433,30 @@ module Purification = struct
...
+            let x' =
+              if is_pure x then x
+              else
+                let x', lets = purify_term x SMap.empty in
+                mk_lifted x' lets
```
Gbury: wouldn't it be better to lift the required let-bindings at the top ? i.e. something like: `let x', x'_lets = .... in let y', y'_lets = ... in mk_lifted ...` ?

Gbury (changes requested):
I tihk the fix is good, but I left a few comments. Also, while we're at it, could you change the `assert false` that are present in the purification functions into e.g. `failwith "unexpected expression in purification function"` so that the error message is less cryptic when such bugs occur again ?

src/lib/structures/expr.ml
hra687261: Ok, I replaced ```x``` with ```fa``` for functional array, and ```y``` with ```i``` for index.

src/lib/structures/expr.ml
hra687261: Yes, you're right. It's done.

Gbury (approved):

PR #490, opened 2021-11-03
Added contex reinitialisation functions

hra687261:
The purpose of this PR is to add the necessary functions to allow resolution context re-initialization in Alt-Ergo.
Since Alt-Ergo internally relies on side effects through references and hash-tables, A user that uses it as a library might want to reset the resolution context to its initial state, aka the state in which the modules are after their initialization.
The current version re-initializes some of the refs with hard-coded values which is not ideal but it works.

### DIFF
--- a/src/bin/gui/main_gui.ml
+++ b/src/bin/gui/main_gui.ml
@@ -231,8 +231,10 @@ let pop_model sat_env () =
       ~hpolicy:`AUTOMATIC
       ~packing:pop_w#vbox#add () in
   let buf1 = GSourceView3.source_buffer () in
-  let tv1 = GSourceView3.source_view ~source_buffer:buf1 ~packing:(sw1#add)
-      ~wrap_mode:`CHAR () in
+  let tv1 =
+    GSourceView3.source_view
+      ~source_buffer:buf1 ~packing:(sw1#add) ~wrap_mode:`CHAR ()
+  in
   let _ = tv1#misc#modify_font font in
   let _ = tv1#set_editable false in
   let model_text = asprintf "%a@." (SAT.print_model ~header:false) sat_env in
@@ -1165,21 +1167,17 @@ let start_gui all_used_context =
              st_ctx annoted_ast dep actions resulting_ids in
          connect env;
 
-         let remove_ctx_button =
-           GButton.toggle_tool_button ~label:" Remove context"
-             ~stock:`CUT ~packing:toolbar#insert ()
-         in
+         let remove_ctx_button = GButton.toggle_tool_button
+             ~label:" Remove context" ~stock:`CUT ~packing:toolbar#insert () in
          ignore (
-           remove_ctx_button#connect#clicked ~callback:(remove_context env)
+           remove_ctx_button#connect#clicked ~callback:(remove_context env);
          );
-
-         let run_button = GButton.tool_button ~label:" Run Alt-Ergo"
-             ~stock:`EXECUTE () in
+         let run_button =
+           GButton.tool_button ~label:" Run Alt-Ergo" ~stock:`EXECUTE () in
          toolbar#insert run_button;
 
          let stop_button =
-           GButton.tool_button ~label:" Abort" ~stock:`STOP ()
-         in
+           GButton.tool_button ~label:" Abort" ~stock:`STOP () in
          stop_button#misc#hide ();
 
          (* TODO: Use toolbar#insert instead of insert_space *)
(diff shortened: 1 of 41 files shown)


Stevendeo (approved):
Seems good to me

hra687261:
@Stevendeo I'll mark this as a draft for now since I wanted to do the context reinitialisation more dynamically and avoid using literals for the integer ids.

src/bin/gui/main_gui.ml
```diff
@@ -1195,33 +1193,27 @@ let start_gui all_used_context =
...
+             ~packing:(toolbox#pack ~fill:true) ()
          in
          toolsearch#set_icon_size `DIALOG;
 
          let tool_item = GButton.tool_item ~packing:toolsearch#insert () in
          let search_box =
-           GPack.hbox ~spacing:5 ~border_width:5 ~packing:tool_item#add ()
-         in
+           GPack.hbox ~spacing:5 ~border_width:5 ~packing:tool_item#add () in
          ignore(GMisc.image ~icon_size:`LARGE_TOOLBAR
```
Stevendeo: I disagree with this indentation. the `let` and the `in` should have the same indentation, so that the block of code can be seen directly.

src/lib/reasoners/fun_sat.ml
```diff
@@ -1955,4 +1967,32 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+    Satml_types.Flat_Formula.reinit_cpt ();
+    Ty.reinit_decls ();
+    IntervalCalculus.reinit_cache ();
+    Inst.reinit_em_cache ();
+    Expr.reinit_cache ();
+    Hstring.reinit_cache ();
+    Shostak.Combine.reinit_cache ();
+    Uf.reinit_cache ()
+
+  let _ =
```
Stevendeo: Why not a `let ()`?

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1159,6 +1159,29 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+    Satml_types.Flat_Formula.reinit_cpt ();
+    Ty.reinit_decls ();
+    IntervalCalculus.reinit_cache ();
+    Inst.reinit_em_cache ();
+    Expr.reinit_cache ();
+    Hstring.reinit_cache ();
+    Shostak.Combine.reinit_cache ();
+    Uf.reinit_cache ()
+
+  let _ =
```
Stevendeo: Same as before; `let () =`

src/lib/reasoners/sat_solver_sig.mli
```diff
@@ -70,6 +70,11 @@ module type S = sig
   val print_model : header:bool -> Format.formatter -> t -> unit
 
   val reset_refs : unit -> unit
+
+  (** [reinit_ctx ()] reinitializes the solving context.
+      The order in which the calls are done is important. *)
```
Stevendeo: Can you document (in the code) why the order is important?

Gbury:
This looks ok once @Stevendeo 's comments are addressed.

src/bin/gui/main_gui.ml
hra687261: I'm not a fan of it either. I only indented that way because the rest of the file is indented like that. I'll revert the changes, but fixing the indentation and style in other places in the files should be done in a separate PR so I won't do it here.

src/lib/reasoners/sat_solver_sig.mli
hra687261: Honestly I am not sure if it is still important. It was important in the first version of context reinitialization when I reanitialized by recreating the top-level values that are created at the start of AE, but now that the context is saved and the caches are just recovered from the backed up values, the order should not be important anymore. I will check to be sure and update the comment.
Either way Ideally users should only reinitialize the context through the exposed functions in the sat solvers and not use the internal functions of the modules unless they know what they are doing.

src/lib/reasoners/fun_sat.ml
hra687261: No reason, I changed it.

src/lib/reasoners/satml_frontend.ml
hra687261: Done.

src/lib/reasoners/sat_solver_sig.mli
hra687261: As I thought, the order is no longer important. I updated the comment.

hra687261:
It's done.

PR #491, opened 2021-11-25
Added Dolmen as a frontend 

hra687261:
The purpose of this PR is to make it possible to use Dolmen as a front-end to Alt-Ergo and eventually make it the default front-end instead of the current one. 

To use Dolmen as a frontend: 
```./alt-ergo --frontend dolmen```

### CODE REVIEW

File: src/lib/frontend/d_cnf.ml [hunk 1]
```diff
@@ -0,0 +1,1143 @@
...
+  struct
+    type t = int
+    let compare = Int.compare
+  end
+  )
+
+(** Helper function: returns the basename of a dolmen path, since in AE
+    the problems are contained in one-file (for now at least), the path is
+    irrelevant and only the basename matters *)
+let get_basename DStd.Path.(Local { name; } | Absolute { name; _ }) =
```
Gbury: you should at the very least, assert that the absolute path is empty, or else that will cause problems later

hra687261: Done.

hra687261:
This PR will have to wait for a new release of dolmen before being reviewed/merged as it uses its latest version

Halbaroth (approved):

PR #495, opened 2021-12-23
Enable type inference for Menhir.

fpottier:
This is required for compatibility with Menhir 20211215.


Gbury:
Thanks a lot ! I'll take a closer look and merge as soon as I get back from my holidays, ^^

Gbury (approved):

Gbury:
Thanks for the fix, :)

PR #506, opened 2022-06-09
Updating min version of psmt2-frontend

Stevendeo:


bclement-ocp:
This change is already present on `next`, closing

PR #510, opened 2022-07-04
Added an `answer` statement to AE's native lang

hra687261:
This is an attempt to add an `answer` statement which would state the expected result of trying to prove the goal that comes after it. 
This is particularly useful for non-regression tests and bench-marking.


Gbury:
I don't think this is a good idea for a few reasons:
1) it adds new keywords to the lexer, which means that any (syntactically correct) problem that currently uses these words as regular variable names ill stop being syntactically correct, which is not good.
2) what's the use of having that **in alt-ergo** ? it can be useful for tests/benchmark software to check the results of provers, but I don't see the use in provers themselves
3) it's a kind of metadata about the problem, and not part of the input problem
4) most of the time it's not present in input problems, because the information is not known (e.g. program verification), so really it's only ever really useful in the context of benchmarks, where we have problems with known expected result.

The solution usually taken for this kind of things is to put that information in comments, or using already existing syntax construction for metadata, e.g. for smtlib:
```smtlib
(set-info :status unsat)
```
or for tptp:
```tptp
% Status   : Theorem
```

We can explore having a similar kind of standard comment syntax for alt-ergo, but I don't see the point of having a dedicated syntax for it (especially considering how syntax changes can break existing problems), or why (or how) alt-ergo should parse and use such information.

hra687261:
Ok, it's true that it doesn't need to be added to AE's syntax since AE doesn't need nor use that information.
But it would be great to have a standard comment syntax to express it.
For now I guess I'll use comments with smt-lib2 syntax.

PR #514, opened 2022-07-07
Fix odoc cyclic dependency

Stevendeo:


Gbury:
I don't think that's the right fix: it will force users to install odoc when installing alt-ergo. To fix the cyclic dependency in the CI, it's more likely that we need to fix the CI workflow commands.

Stevendeo:
Indeed, changing the ocaml version was sufficient to solve the issue. I'm not sure how to save the cache so for now, `opam install --deps-only` is called everytime, maybe once it is merged it will be updated?

Gbury (approved):

PR #518, opened 2022-07-11
Fix CI

Stevendeo:


.github/workflows/build_docker.yml
```diff
@@ -26,6 +30,9 @@ jobs:
       # to `ocaml` with the container parameters
       - run: sudo chmod a+wx .
 
+      - name: Reclaim cache directory
+        run: sudo chown -R 1001:1001 .
```
Gbury: what kind of problem does that fix ?

Gbury (commented):
what's the exact motivation for this PR ?

Stevendeo:
The CI is currently broken, this PR aims at fixing the different CI issues.

.github/workflows/build_docker.yml
Stevendeo: The docker image build fails ; right now the error is `Error: EACCES: permission denied, mkdir '/__w/_temp/9c8f979e-f394-49cf-ae44-be4cb680bb8a'` so I guessed there was some kind of ownership issue. Right now, I'm playing with CI configuration to start the Docker check on a specific set of PRs (the PR of branches named `fix-ci`).

hra687261:
This PR needs to be merged otherwise all new PRs and PRs that have recent commits will keep having failing CIs

Gbury (approved):

Gbury:
Well, looks like this actually broke CI once merged cc @Stevendeo , :p

hra687261:
It actually failed in a **new** way. The error is not the same as the one it fixes.
I created a PR #525 which should fix the new CI error

Issue #509, opened 2022-06-23
Build fails on many architectures


SnarkBoojum:
From the [Debian build logs](https://buildd.debian.org/status/package.php?p=alt-ergo), there are failures on many architectures with the same error:
```
File "src/plugins/AB-Why3/dune", line 24, characters 5-22:
24 |     (ABWhy3Plugin.cmxs as plugins/AB-Why3-plugin.cmxs)
          ^^^^^^^^^^^^^^^^^
Error: No rule found for src/plugins/AB-Why3/ABWhy3Plugin.cmxs
File "src/plugins/fm-simplex/dune", line 17, characters 5-25:
17 |     (FmSimplexPlugin.cmxs as plugins/fm-simplex-plugin.cmxs)
          ^^^^^^^^^^^^^^^^^^^^
Error: No rule found for src/plugins/fm-simplex/FmSimplexPlugin.cmxs
```
and I have to admit I don't know why some architectures know how to build that file and not others... perhaps a case of buggy opt/native support?

SnarkBoojum:
I can confirm my hunch: the renaming lines for .cmxs files in the two `src/plugins/*/dune` files are the problem, as the files don't exist on non-native architectures!

Stevendeo:
PR #524 fixes it

PR #524, opened 2022-09-15
Fix #504 and #509

### LINKED ISSUE #504: Plugins don't belong to /usr/share
While working on the Debian alt-ergo package, the tools complained:
```
E: alt-ergo: arch-dependent-file-in-usr-share usr/share/alt-ergo/plugins/AB-Why3-plugin.cmxs
E: alt-ergo: arch-dependent-file-in-usr-share usr/share/alt-ergo/plugins/fm-simplex-plugin.cmxs
```
and indeed, if those files are arch-dependent, they should go somewhere else ; perhaps in /usr/lib/alt-ergo ?

hra687261:
Changing the installation section for plugins and preludes from `share` to `lib` (to fix issues #504 and #509).
If some users use the plugins and look for them in `/usr/share` they won't find them, so if this PR is merged it needs to be clearly stated in the next release that the installation section for plugins and preludes was changed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,26 +103,26 @@ gui: gen
 # fm-simplex plugin
 fm-simplex:
 	$(DUNE) build $(DUNE_FLAGS) \
-		$(INSTALL_DIR)/default/share/alt-ergo/plugins/fm-simplex-plugin.cma \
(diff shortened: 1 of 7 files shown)


Stevendeo (approved):

Stevendeo:
Seems go to me!

PR #526, opened 2022-09-21
New test scripts

Halbaroth:
This PR consists in replacing the old bash scripts used for running the non-regression tests by a mighty Ocaml script.

I moved most of the tests from `/non-regression` directory to `/tests`. ~~I didn't remove completely the former directory because 
I don't know if we should keep some of its tests. For instance, I guess keeping the tests about parsing will not make sense as soon as we have replace the former parser by the one provided by Dolmen.~~

The new workflow for running non-regression tests is as follow:
1. (optional) add extra tests in the directories `tests/unsat` `tests/sat` and `tests/unknown` according to the expected result.
2. Run `make gentest` to regenerate the dune files.
3.  Run `make runtest` to perform tests.

I added CI support. Run `make runtest-ci` to perform tests for CI.

Any suggestion is very welcome :D


hra687261:
Suggestion (Inspired by what is done in https://git.frama-c.com/pub/colibrics):
As it was done in the older version, tests can be kept in folders which separate them by their expected result (`../theory/{sat|unsat|unknown}/*.{ae|smt2}`).
That way, you won't need expect files (which all contain the same stuff anyway), you just need to generate rules in those folders that will tell Alt-Ergo that it should respond with sat, unsat or unkown. It can be done through a command line option that tells AE to expect some result, and if it got a different result, it can print an error message. (You can probably also print the command that led to the error using `Sys.argv`).
You can have an executable that will generate a dune file in each folder, the dune file will contain the rules of the tests for all the files in the folder (with the right expect, determined by the folder's name). So adding a new test will simply consist of adding a its file to the right folder.

Halbaroth:
I modified the script to use the directory name as an indicator of the expected result of the test. I didn't add any extra option in AE since I can use the stanza `system` of Dune as follow (for a test expecting unsat):
```
(rule
 (alias runtest)
 (deps (:input NAME_OF_TEST))
 (package alt-ergo-lib)
 (action
  (chdir %{workspace_root}
   (ignore-stdout
    (ignore-stderr
     (with-accepted-exit-codes 0
      (system "alt-ergo --output=smtlib2 --timelimit=1 %{input} | grep -e '^unsat'")))))))
```

hra687261:
LGTM, although I'm not sure about the deletion of the `challenges`, `incorrect` and `gui` tests.
`incorrect` and `gui` don't seem very useful (at least not as non-regression tests, if someone works on the GUI they'll need to add more tests, separately from these ones), but I think `challenges` can be kept (as a goal to reach).

Halbaroth:
Most of incorrect tests are about the old parser, so it is pointless to keep them since we don't plan to modify the parser soon and the parser will be replaced completely by Dolmen. I agree to put back the challenge tests but I guess we don't know the expected results of these tests. Otherwise, they had been put in the valid subdirectory of the old non-regression directory?

hra687261:
Since they are in the `valid` folder it's probably safe to assume that they are valid.
Looking at a few of them:
- https://github.com/OCamlPro/alt-ergo/blob/4c02677244ba610466d5068011ea4dd69fc91577/non-regression/challenges/valid/challenge-case-split-non-pred-booleans__KO.ae#L1-L8
- https://github.com/OCamlPro/alt-ergo/blob/4c02677244ba610466d5068011ea4dd69fc91577/non-regression/challenges/valid/testfile-arith_mult_interval005__KO.ae#L1-L3
- https://github.com/OCamlPro/alt-ergo/blob/4c02677244ba610466d5068011ea4dd69fc91577/non-regression/challenges/valid/challenge_with_rational_division__KO.ae#L1-L6

They are obviously Valid.

This one ~also seems valid~ probably isn't:
-  https://github.com/OCamlPro/alt-ergo/blob/4c02677244ba610466d5068011ea4dd69fc91577/non-regression/challenges/valid/is_it_valid.ae#L1-L9

Because apparently if `e = A` then `e#bb` is unspecified (I didn't find much documentation on it).
For example:
```
type t = A | B of { i : int }

goal g1:
  let n = B(0) in
  n#i = 0

goal g2:
  let n = A in
  n#i = 0

goal g3:
  let n = A in
  n#i = 0 or n#i <> 0
```  
Results in:
```
File "tmp.ae", line 15, characters 3-29: Valid (0.0006) (1 steps) (goal g1)
File "tmp.ae", line 19, characters 3-26: I don't know (0.0011) (2 steps) (goal g2)
File "tmp.ae", line 23, characters 3-40: Valid (0.0017) (0 steps) (goal g3)
```
Therefore, in the test `(e#bb ? X or e#bb ? Y)` is always true but `e ? B` isn't always true. So the goal is not valid.
It would be interesting to see if AE manages to generate a counter example on it (@Stevendeo).

Halbaroth:
In my opinion, the non-regression tests shouldn't fail in the normal situation. I agree to bring back challenges but I do think we should create a separate alias for them. For instance, we could run non-regression tests as follow:
```make non-regression```
and challenge tests as follow:
```make challenge```

For the challenge tests, we could display a message if a test succeed. If so, we move the test in the non-regression directory.

hra687261:
Yes it can be done that way. Or the ones that don't pass can be put in a folder named `unknown` and I am not sure but I think that if the response changes (from unknown to unsat or sat) then benchpress will list the test as `improved`.

Halbaroth:
If some of them pass, we can put them in the non-regression folder. I don't use benchpress in the above script but we can add them on Marvin for instance.

Halbaroth:
> Because apparently if `e = A` then `e#bb` is unspecified (I didn't find much documentation on it). For example:
> 
> ```
> type t = A | B of { i : int }
> 
> goal g1:
>   let n = B(0) in
>   n#i = 0
> 
> goal g2:
>   let n = A in
>   n#i = 0
> 
> goal g3:
>   let n = A in
>   n#i = 0 or n#i <> 0
> ```
> 
> Results in:
> 
> ```
> File "tmp.ae", line 15, characters 3-29: Valid (0.0006) (1 steps) (goal g1)
> File "tmp.ae", line 19, characters 3-26: I don't know (0.0011) (2 steps) (goal g2)
> File "tmp.ae", line 23, characters 3-40: Valid (0.0017) (0 steps) (goal g3)
> ```

It does not make sense that ```n#i``` is a valid term. I hope such a weird syntax is forbidden in SMT language. In my opinion, the proper answer should be a type error because the type of `n` is not a record but a variant type. I will write the challenge in SMT.

Halbaroth:
If we put the challenge tests on Marvin, I think that we can remove them of the tests folder. I plan to add categories in Benchtop, so it will be clear that a new challenge has been solved in Benchtop in the category `challenge`. My point is these challenges have little chance to be solved by a minor change in AE. So we can perform them only before merging a new PR.

hra687261:
Yes, I agree, the ones that pass (I think two of them do) can be put in the non-regression tests as valid, the others should be moved to marvin. 
Regarding `n#i` I agree that it shouldn't be allowed and should result in a type error, but I think it was intentionally designed that way in Alt-Ergo's native language, I don't know the reasoning behind it though.

Halbaroth:
Guillaume explained to me the meaning of the projector `n#i`. It's the dirty way to implement the ADT without pattern matching. The same trick is used in SMT lib. Since the semantic of `n#i` is not defined when `n` is not the constructor `B`, AE could produce a model as you noticed.

Gbury:
If you want to understand this a bit more, you can grep for "selector" in the [smtlib language reference pdf](https://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2021-05-12.pdf)

Halbaroth:
I put the successful challenges back. The failed ones will be tested on Marvin.

Halbaroth:
Is it ok for you?

hra687261 (approved):

hra687261:
Yes, sorry, all good by me.

PR #530, opened 2022-10-14
Models, rebased

Stevendeo:


alt-ergo-lib.opam
```diff
@@ -1,6 +1,6 @@
 # This file is generated by dune, edit dune-project instead
 opam-version: "2.0"
-version: "dev"
+version: "models"
```
Gbury: that should not be changed by this PR: the version should stay dev as this is going to be merged into `next` which is the dev version

src/bin/common/parse_command.ml
```diff
@@ -894,26 +933,27 @@ let parse_output_opt =
...
   let interpretation =
-    let doc =
-      "Experimental support for counter-example generation. Possible \
-       values are 1, 2, or 3 to compute an interpretation before returning \
-       Unknown, before instantiation (1), or before every decision (2) or \
-       instantiation (3). A negative value (-1, -2, or -3) will disable \
-       interpretation display. Note that $(b, --max-split) limitation will \
-       be ignored in model generation phase." in
+    let doc = Format.sprintf
+        "Experimental support for counter-example generation. \
```
Gbury: The doc should not say `Experimental`: maybe something along the lines of `best effort` would better signify that it should be correct, but there is no strong guarantee on the generated model.

src/bin/common/parse_command.ml
```diff
@@ -936,8 +976,32 @@ let parse_output_opt =
...
+  in
+
+  let mdls =
+    let doc =
+      "Simply activates the models in alt-ergo. This is achieved by setting \
+       some parameters by default: interpretation = every; instanciation \
+       heuristic = normal; \
+       sat-solver = tableaux"
+    in
+    Arg.(value & flag & info ~doc ~docs ["models"])
```
Gbury: I think the default should be to provide one model, and use the `before_end` setting. Outputting more than one model by default will confuse users. And so, also remove the `s` to the option name.

src/bin/common/parse_command.ml
```diff
@@ -1238,8 +1304,16 @@ let parse_fmt_opt =
     Arg.(value & opt formatter_conv Stderr & info ["err-formatter"] ~docs ~doc)
   in
 
+  let mdl_formatter =
+    let doc = Format.sprintf
+        "Set the model formatter used by default to output model and
+         interpretation. Possible values are %s."
+        (Arg.doc_alts ["stdout"; "stderr"; "<filename>"]) in
+    Arg.(value & opt formatter_conv Stdout & info ["mdl-formatter"] ~docs ~doc)
```
Gbury: The name `formatter` is meaningless for users of alt-ergo (it's only pertinent for those that program in ocaml). Something like `model-output` would be clearer (both for the doc string and the option name).

src/lib/frontend/models.mli
```diff
@@ -0,0 +1,22 @@
...
+(** {1 Models module} *)
+
+(** Print the given counterexample on the given formatter with the
+    corresponding format setted with Options.get_output_format *)
+val output_concrete_model :
+  Format.formatter ->
+  Expr.Set.t ->
+  Profile.V.t Profile.P.t ->
+  Profile.V.t Profile.P.t ->
+  Profile.V.t Profile.P.t ->
```
Gbury: these three arguments should be named, and some doc added to state what they are, otherwise, the function is simply impossible to understand.

src/lib/reasoners/fun_sat.ml
```diff
@@ -1313,7 +1259,7 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
            | E.Let _ | E.Iff _ | E.Xor _ ->
              Printer.print_err
                "Currently, arbitrary formulas in Hyps
-                are not Th-reduced";
+are not Th-reduced";
```
Gbury: bad indentation ?

src/lib/structures/expr.ml
```diff
@@ -427,36 +427,65 @@ let rec print_silent fmt t =
     begin
       match lit, xs with
       | Sy.L_eq, a::l ->
-        fprintf fmt "(%a%a)"
-          print a (fun fmt -> List.iter (fprintf fmt " = %a" print)) l
+        if get_output_smtlib () then
```
Gbury: Rather than matching for `get_output_smtlib ()` in each case, it'd be nicer to have two functions, say `print_silent_smtlib` and `print_silent_ae`, and then have
```
let print_silent ... = if get_output_smtlib () then print_silent_smtlib ... else print_silent_ae ...
```

src/lib/structures/expr.mli
```diff
@@ -168,14 +168,14 @@ val is_fresh : t -> bool
 val is_fresh_skolem : t -> bool
 val is_int : t -> bool
 val is_real : t -> bool
-val type_info : t -> Ty.t
-
+val [@inline always] type_info : t -> Ty.t
```
Gbury: these annotations are useless in `.mli` files afaik.

src/lib/structures/profile.mli
```diff
@@ -0,0 +1,28 @@
+(******************************************************************************)
+(*                                                                            *)
+(*     Alt-Ergo: The SMT Solver For Software Verification                     *)
+(*     Copyright (C) 2020-2020 --- OCamlPro SAS                               *)
+(*                                                                            *)
+(*     This file is distributed under the terms of the license indicated      *)
+(*     in the file 'License.OCamlPro'. If 'License.OCamlPro' is not           *)
+(*     present, please contact us to clarify licensing.                       *)
+(*                                                                            *)
+(******************************************************************************)
```
Gbury: Lack of doc: what is this module ? what use does it have ?

src/parsers/native_parser.mly
```diff
@@ -44,7 +44,7 @@
 %token AND LEFTARROW RIGHTARROW AC AT AXIOM CASESPLIT REWRITING
 %token BAR HAT
 %token BOOL COLON COMMA PV DISTINCT DOT SHARP ELSE OF EOF EQUAL
-%token EXISTS FALSE VOID FORALL FUNC GE GOAL GT CHECK CUT
+%token EXISTS FALSE VOID FORALL FUNC GE CHECK_VALID CHECK_SAT GT CHECK CUT
```
Gbury, line 47: Iirc, we have some kind of doc for the native ae language somewhere (in a sphinx doc I think ?), and so, this change to the language should also change what little doc we have for it, ^^

Gbury (changes requested):
I haven't reviewed the model producing mechanism in depth, but that should be okay for a first PR, we can (and should) add tests later to verify that the model output is not too broken on at least simple cases.

src/lib/structures/expr.mli
Stevendeo: Does that mean if the mli does specify this inlining but the ml does not, the function will not be inlined?

Gbury: If I'm correct, in an `.mli` file, the attribute is simply ignored (i.e. I don't think the inlining code ever looks at attributes on the type declaration, or even has access to the type declaration).

docs/sphinx_docs/Usage/index.md
```diff
@@ -26,6 +26,22 @@ Alt-Ergo supports file extensions:
...
+There is two ways to activate model generation:
+- `with the --interpretation=VALUE`, where VALUE can be equal to:
+  * "none", and alt-ergo will not generate models (by default);
+  * "first", and alt-ergo will output the first model it finds;
+  * "every", alt alt-ergo will compute a model before each decision
+  * "last", and alt-ergo will output the last model it computes before returning 'unknown'.
+  Note that this mode only works with the option `--sat-solver tableaux`.
+
+- with the `--model option`, setting the interpretation option to 'last', and the sat-solver
+to 'tableaux.
```
Gbury: maybe put this option first, as this is the recommended way to generate models (basically the `--interpretation` option is for advanced users/devs)

src/bin/common/parse_command.ml
```diff
@@ -1238,8 +1303,16 @@ let parse_fmt_opt =
     Arg.(value & opt formatter_conv Stderr & info ["err-formatter"] ~docs ~doc)
   in
 
+  let model_output =
+    let doc = Format.sprintf
+        "Set the model formatter used by default to output model and
```
Gbury: A better doc string would be something like: "Set the output used for the model generation. Possible values are ..."

src/lib/frontend/typechecker.ml
```diff
@@ -1958,9 +1958,9 @@ let type_goal acc env_g loc sort n goal =
 
 
 let rec type_and_intro_goal acc env sort n f =
-  let b = (* smtfile() || smt2file() || satmode()*) false in
+  (* let b = (\* smtfile() || smt2file() || satmode()*\) false in *)
```
Gbury: maybe remove that line ?

src/lib/reasoners/shostak.ml
```diff
@@ -583,6 +587,25 @@ struct
...
+        else
+          begin
+            let eq = List.filter (fun (_t, r) -> is_bool_const r) eq in
+            match eq with
+            | (e,_r)::_ -> Some (e, false) (* false <-> not a case-split *)
+            | [] ->
+              let dist = List.filter (fun r -> is_bool_const r) distincts in
+              match dist with
+              | {v = Term e; _}::_ ->
+                Some (Expr.neg e, true) (* safety: consider it as case-splut *)
```
Gbury:
```suggestion
                Some (Expr.neg e, true) (* safety: consider it as case-split *)
```

src/lib/structures/expr.ml
```diff
@@ -377,175 +377,380 @@ let print_binders =
...
+ *   Format.pp_print_list ~pp_sep:(fun fmt _ -> Format.fprintf fmt sep) pp fmt
+ *
+ * let print_list pp fmt = print_list_sep "," pp fmt *)
+
+module SmtPrinter = struct
+
+  let rec print_formula fmt form xs bind =
+    match form, xs, bind with
+    | Sy.F_Unit _, [f1; f2], _ ->
+      fprintf fmt "@[(%a /\\@ %a)@]" print_silent f1 print_silent f2
```
Gbury: these cases do not output correct smtlib2, either they should be fixed (e.g. conjunction is `(and foo bar)` not `(foo /\ bar)`), or at the very least, these cases should raise en exception.

src/lib/structures/expr.mli
Gbury: You should remove the attributes from the mli files: they are literally ignored/unused by the compiler, and they are surprising/quite distracting when reading an `.mli` file.

src/lib/structures/profile.mli
Gbury: The doc helps, but why is this module called `Profile` ? Naively, the name would make me think of progiling information, which this is definitely not.

src/lib/util/version.ml
```diff
@@ -29,7 +29,7 @@
 (* WARNING: a "cut" is performed on the following file in the Makefile.
    DO NOT CHANGE its format *)
 
-let _version="dev"
+let _version="models"
```
Gbury: revert to `dev` (and make note that this string must be changed when preparing a new release).

Gbury (changes requested):

src/bin/common/parse_command.ml
```diff
@@ -936,8 +975,32 @@ let parse_output_opt =
     Arg.(value & opt (some format_conv) None & info ["o"; "output"] ~docv ~doc)
   in
 
+  let model_type =
+    let doc =
+      Format.sprintf
+        "Control the output model type of the solver, $(docv) must be %s."
+        (Arg.doc_alts [ "value"; "constraint" ])
```
Halbaroth: The documentation is wrong, It is `constraints` and not `constraint`. Perhaps adding more information about these options? I didn't find the meaning of `constraints` in the doc directory also. Actually I would like to minimize some values in a generated models but I didn't find how to do that.

src/lib/structures/expr.mli
Stevendeo: I disagree with you here; I find useful to see directly when a function will be inlined without having to see the implementation.

Gbury: In that case, you can always mention it in the doc comments.
To expand a bit, here are some reasons why I think it's bad to have such attributes in .mli files:
- it is currently silently ignored by the compiler, but there are upcoming patches that would emit warnings on ignored attributes, warnings that dune would likely treat as errors during dev.
- I'm not sure why you  would want to know whether a function will be inlined. The only situation where it is really important/critical is when the function occurs in a hot path, and when the function body is really small (with a size/runtime comparable to that of a function call). In such cases, if you're interested in performances, you already need to know what the implementation of the function does (or at the very least, its complexity), so you need knowledge of the function's body, which defeats the purpose of having the annotation in the mli file.
- inline annotations/attributes are only read/acted upon by flambda, so most of the time (as long as your switch did not explicitly enable flambda), all such annotations are ignored, so you cannot and should not rely on functions being inlined, and thus having that info in the mli file seems... ineffective.
- very small functions such as these ones, will already be unconditionally inlined, even with closure; and with flambda, the decision of whether to inline is made at the call-sites rather than at the definition site, so again, annotating the interface doesn't make much sense.

Overall, I'd say that the goal of being a bit more aware of performance (which I assume is why you find it useful to know whether a function will be inlined), would be more served by providing an estimation of the complexity (in terms of run time/memory) in the doc comment of the function, and rely on the fact that very simple/trivial functions will 1) either be naturally inlined by the compiler without any annotation (because the ocaml compiler is not that bad), or 2) will have its definition/implementation annotated to guide the flambda inliner, if it is pertinent (which depends on the implementation, not the interface).

Stevendeo: * If in the future dune will emit warnings on ignored attributes, then it is a good idea to write them in the `.mli` as a safeguard when it comes to update the code don't you think?
* It is just a question of documentation and clarification. The annotation was here in the first place actually, and the more information I can get in the .mli the less I am likely to make wrong assumptions in general.
* I could as well remove all `.mli`s before compiling and I would still be able to compile the project the exact same way.
* I did not know that ^^

I will save us some efforts and remove this annotation anyway.

Gbury: Thanks (also note that, since the compiler ignores the attribute, nothing prevents you from adding that annotation to a function that will not in fact, be inlined).

src/bin/common/parse_command.ml
Stevendeo: Constraints & optimizing models are not the same. The `constraints` mode outputs the set of expressions alt-ergo managed to prove (cf Fun_sat.t, field `gamma`, extracted from `extract_prop_model`).
This is not the optimizer.

Gbury (approved):

PR #535, opened 2022-11-23
Fix ci

Halbaroth:
This PR fixes the CI again.
It was a mess... 

- I rolled back some modifications I did in the PR #526. Using `grep` to retrieve the expected answer was not a good idea and created more trouble. So I use again the promote mechanism of Dune. 
- I put `(dune (3.0 >=))` has a requirement on all packages. Maybe it is not fine for all users. 
- I created a new package `alt-ergo-js` with all the dependencies. Now, `alt-ergo` requires that the version of `js_of_ocaml` satisfies `<= 4.0.0`. It seems the primitive `unix_gettimeoftheday` is not available in the last release `4.1.0`. I have to check and I might submit a bug issue on the js_of_ocaml repository. This solution is temporary.
- Léo told me the Why3 team uses AE compiled with `js_of_ocaml`. The above condition could break something?
- I seriously plan to rewrite CI and the Makefile completely in a new PR this week. 
- After updating `ocplib-simplex`, the windows CI was broken. It seems the repository https://github.com/fdopen/opam-repository-mingw is discontinued but still in use in the github action https://github.com/ocaml/setup-ocaml. The issue has been pointed out in the ticket https://github.com/fdopen/opam-repository-mingw/issues/108. The solution that I've found consist in using both the usual opam repository and the mingw variant to retrieve the ocaml compiler. More precisely, I put in the `build_windows.yml` file:
```yaml
      - name: Use OCaml ${{ env.OCAML_DEFAULT_VERSION }}
        uses: ocaml/setup-ocaml@v2
        with:
          ocaml-compiler: ${{ env.OCAML_DEFAULT_VERSION }}
          opam-repositories: |
            windows: https://github.com/fdopen/opam-repository-mingw.git
            default: https://github.com/ocaml/opam-repository.git
```


tools/dune
```diff
@@ -1,4 +1,7 @@
 (executable
-  (name gentest)
-  (modules gentest)
-  (libraries unix))
+ (name gentest)
+ (modules gentest)
+ (libraries
+  unix
+ )
+)
```
Stevendeo, line 5: Ahem :)

alt-ergo-js.opam
```diff
@@ -0,0 +1,41 @@
...
+
+See more details on https://alt-ergo.ocamlpro.com/"""
+maintainer: ["Alt-Ergo developers"]
+authors: ["Alt-Ergo developers"]
+homepage: "https://alt-ergo.ocamlpro.com/"
+doc: "https://ocamlpro.github.io/alt-ergo"
+bug-reports: "https://github.com/OCamlPro/alt-ergo/issues"
+depends: [
+  "ocaml" {>= "4.05.0"}
+  "dune" {>= "3.0" & >= "3.0"}
```
Stevendeo: Something in the `dune` files may lead to repeating the version twice.

Stevendeo: This happens in the other opam files

tools/dune
Stevendeo: Actually, the best version would be :
``` 
 (libraries unix)
)
```

Stevendeo (approved):
Good enough ;)

PR #540, opened 2022-12-21
Fix infinite loop in Expr.SMTPrinter when the debug option is set

hra687261:


src/lib/structures/expr.ml
```diff
@@ -493,7 +493,7 @@ module SmtPrinter = struct
         Format.fprintf fmt "| %a@,"  (Util.print_list ~sep:"," ~pp:print) l;
       ) trs
 
-  and print_verbose fmt t = print fmt t
```
Halbaroth, line 496: This code is quite strange. Are `print_verbose` and `print` equal now?

hra687261: The difference between `print_verbose` and `print_simple` used to be that `print_verbose` printed the type as well as the expression while `print_simple` only printed the expression, but since the printing of the types was removed not long ago, making `print_verbose` call `print` (which is just a dispatcher) leads to an infinite loop when the `verbose` flag is set.
So yeah, I don't know if we want to add back the printing of types, or just have one printing function, or print other information, but for now `print_verbose` and `print_simple` do the same thing. It's not great to have useless code but it's better than an infinite loop.

Halbaroth (approved):

PR #542, opened 2022-12-23
Fix js of ocaml

Halbaroth:
This PR resolves the compilation error about the missing primitive `unix_gettimeoftheday`. Since the release 4.1.0 of `js_of_ocaml`, the function has been renamed as `caml_unix_gettimeoftheday`. See https://github.com/ocsigen/js_of_ocaml/issues/1370.

- I also fixed the command `make js`.
- I set the option `--no-source-map` for the release profile only. Indeed, according to the `js_of_ocaml` documentation, producing a mapping between the OCaml source and the generated Javascript gives a bigger and slower binary. Yet, the option could be useful for debugging purposes.
- I tried to remove the below warning message throwing by `make`:
```
Warning: '--source-map' is enabled but the bytecode program was compiled with no debugging information.
Warning: Consider passing '-g' option to ocamlc.
```
This error is produced during the compilation with `js_of_ocaml` of the `Zarith` library. The warning oddly disappears if we turn off the separation compilation in `js_of_ocaml` but we got a bunch of new warning messages about missing primitives:
```
There are some missing primitives     
Dummy implementations (raising 'Failure' exception) will be used if they are not available at runtime.
You can prevent the generation of dummy implementations with the commandline option '--disable genprim'
Missing primitives provided by +dynlink.js:
  caml_dynlink_add_primitive
  caml_dynlink_get_current_libs
  caml_dynlink_lookup_symbol
  caml_dynlink_open_lib
Missing primitives provided by +toplevel.js:
  caml_get_section_table
  caml_realloc_global
  caml_reify_bytecode
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ parsers: gen
 	$(DUNE) build $(DUNE_FLAGS) @$(PARSERS_DIR)/all
 
 js: gen
-	$(DUNE) build $(DUNE_FLAGS) -p alt-ergo-js
+	$(DUNE) build $(DUNE_FLAGS) @$(BJS_DIR)/all
 
 fm-simplex: gen
 	$(DUNE) build $(DUNE_FLAGS) @$(PLUGINS_DIR)/fm-simplex/all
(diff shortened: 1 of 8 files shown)


Gbury (approved):
Looks good

PR #546, opened 2023-01-24
Fix regressions for the 2.4.3

Halbaroth:


AllanBlanchard:
FYI, I have a switch that follows this branch, so if you want information about regressions/improvements for future commits on this branch I can test them against the Frama-C WP tutorial and extract alt-ergo files if some examples seem to be interesting for you. If you want to test it yourself, I can also provide a branch of the tutorial for this.

Also I have worked on the Sum example of the issue, CVC4 1.8 proves it (before I had 1.7).  If it can help, to finish the failing proof with Alt-Ergo, I needed to add this information:

```c
/*@
  requires i < len ;
  requires array[i] < INT_MAX ;
  requires \valid(array + (0 .. len-1));
  assigns array[i];
  ensures sum(array, 0, len) == sum{Pre}(array, 0, len)+1;
*/
void inc_cell(int* array, size_t len, size_t i){
  //@ assert sum(array, 0, len) == sum(array, 0, i)+sum(array, i, len) ;
  //@ assert sum(array, i, len) == sum(array, i, i+1)+sum(array, i+1, len) ;
  array[i]++ ;
  //@ assert sum(array, 0, len) == sum(array, 0, i)+sum(array, i, len) ; // < ADDED
  //@ assert sum(array, i, len) == sum(array, i, i+1)+sum(array, i+1, len) ; // < ADDED

  //@ assert unchanged{Pre, Here}(array, 0, i);
  //@ assert unchanged{Pre, Here}(array, i+1, len);
}
```

Leading to the following AE goal: [sum-T-wp_goal.txt](https://github.com/OCamlPro/alt-ergo/files/10713829/sum-T-wp_goal.txt).

AllanBlanchard:
Also maybe a changed file has not been committed, I had to change two calls to `Util.string_of_th_ext ext` (the symbol does not exist).

AllanBlanchard:
If I find regressions related to this branch do you prefer that I list them here or in a related issue?

Halbaroth:
> If I find regressions related to this branch do you prefer that I list them here or in a related issue?

Thanks for your help and sorry for the late reply. I prefer to discuss the regression on your issue #505 ;)

Halbaroth:
After trying a lot to fix all the regressions, it seems the problem is harder than I thought. My patch introduces a lot of new regressions. I will write a new issue to explain in details what's happened and what we can do to solve it but all the solutions are time-consuming. This PR blocks everything. Thus, I decide to revert the flatten optimization introducing by Mohamed. This should fix your regressions @AllanBlanchard

Issue #548, opened 2023-02-08
Wrong type-checking with mutually recursive definitions


claudemarche:
With the attached file [MutualDef.zip](https://github.com/OCamlPro/alt-ergo/files/10683938/MutualDef.zip), Alt-Ergo version 2.4.2 type-checker wrongly complains with 
```
[Error]File "/home/cmarche/MutualDef.ae", line 10, characters 33-34:
Typing Error: this expression should have type prop
```
Similar definitions without mutual recursion are accepted normally.

hra687261:
Apparently it is not treated as a mutually recursive function definition but as a normal definition, so instead of:
```
f(x) = f_body
and g(y) = g_body
```
it is seen as:
```
f(x) = (f_body and g(y) = g_body)
```
with `g` and `y` as uninterpreted symbols. Which is why `f_body` is expected to be of type prop.
Which is annoying, because by looking at the parser their doesn't seem to be support for parsing mutually recursive functions. Although it shouldn't be too hard to add.

claudemarche:
Indeed, I didn't realized it was a parsing problem at first!

Since Alt-Ergo supports mutually recursive ATDs, it is expected that it supports also mutually recursive definitions.

A possible solution to avoid the parsing ambiguity is to use another keyword than `and`, say `with` like in Coq or WhyML. Or do not use `and` for logical connective, but `/\`.

hra687261:
Fixed by #549. 
So to add support for mutually-recursive definitions without adding a new keyword and without introducing ambiguity, it was decided that the syntax would look like this from now on:
```
function f (...): ... = ...
and function g(...): ... = ...
and predicate p(...) = ...
```

Halbaroth:
Fixed by #550 for future Dolmen front-end.

PR #549, opened 2023-02-08
Added support for mutually recursive function and predicate definitions

hra687261:
A proposition to fix the issue #548
I chose to add the `with` keyword as suggested by @claudemarche.
I don't see why a predicate can't be mutually recursive with a function and vice-versa, therefore I added one constructor for both.
Of course for now Dolmen doesn't support this, therefore the CI won't pass


Halbaroth:
Great! It's quite strange that the feature has not been added before since everything works well with such a modification. 
Besides, some people use custom parsers. Modifying the legacy parser API could break temporary their stuff.  It would be better to only add this feature in Dolmen and tell Claude to use the Dolmen front-end now.

Gbury:
one alternative syntax option would be to require something like `function f(x) : ret = ... and function g(y) : ... = ...`, i.e. the addition of the `function` keyword at each function definition should be enough to make the parsing non ambiguous, and not break anything that is currently accepted.

hra687261:
How does this addition break anything? What am I missing? The `with` keyword was just added and didn't exist before and mutually recursive functions where simply not supported. This addition doesn't change anything to the definition of non-mutually-recursive functions and predicates.

Although in a purely subjective manner and from an aesthetic point of view I find the `with` keyword more "elegant" than `and function ...`/`and predicate ...`, I don't really have a strong opinion on the subject and don't mind changing it. 

The only ambiguity I see is that there is no distinction between predicates and functions that return bools, but semantically that doesn't change anything internally AFAIK.
https://github.com/OCamlPro/alt-ergo/blob/5f63a49c793e777f11d12228ed2aa3195d5fbbb5/src/lib/frontend/typechecker.ml#L2228-L2242

Gbury:
> How does this addition break anything? What am I missing? The `with` keyword was just added and didn't exist before and mutually recursive functions where simply not supported. This addition doesn't change anything to the definition of non-mutually-recursive functions and predicates.

Adding a keyword is breaking in the sense that before, `with` was a valid name for a variable and/or a function, but once it is a keyword, it may not be used as a variable name anymore (changing the set of keywords of a language is almost always breaking).

hra687261:
True, I'll change it.

Halbaroth:
I prefer the Guillaume's syntax but I haven't a strong opinion too. In my opinion, users expect some changes when we will deprecate the legacy parser but we should'nt break the legacy parser before this change ;)

Halbaroth:
If your PR is ready for reviewing, let me know and I'll review it right now ;)

src/lib/frontend/typechecker.ml
```diff
@@ -2154,6 +2156,51 @@ let type_user_defined_type_body ~is_recursive env acc (loc, ls, s, body) =
     assert (not is_recursive); (* Abstract types are not recursive *)
     acc, env
 
+let type_fun_aux_1 env loc n ?ret_ty l  =
```
Gbury: is it possible to find a better name for this function (and for `type_fun_aux_2`) ? something that would better describe what the function is doing...
Something along the lines of `declare` for `aux_1` and define for `aux_2` if I understand the code correctly.

src/parsers/native_parser.mly
```diff
@@ -229,11 +238,21 @@ algebraic_args:
...
-and_recursive_opt:
+and_recursive_ty_opt:
   | { [] }
   | AND ty_vars = type_vars ty = ident EQUAL enum = list1_constructors_sep_bar
-others = and_recursive_opt
-    { (($startpos, $endpos), ty_vars, ty, enum) :: others}
+      others = and_recursive_ty_opt
+      { (($startpos, $endpos), ty_vars, ty, enum) :: others}
+
+with_recursive_def_opt:
```
Gbury:
```suggestion
and_recursive_def_opt:
```

hra687261:
@Halbaroth it is!

src/lib/frontend/typechecker.ml
hra687261: Done.

src/parsers/native_parser.mly
hra687261: Done.

Halbaroth (approved):

PR #554, opened 2023-03-22
[WIP] attempt to add support for more FPA and BV primitives

hra687261:
This is a quick attempt to add to support for more FPA and BV primitives.
The idea is to something similar to what is currently done in FPA arithmetic. So a command line option  `--use-bv` was added, similar to `--use-fpa`.
As you know, FPA support in Alt-Ergo is essentially axiomatized and requires adding a file of primitives (or just redeclaring the logic functions).
That can't be done with BV for now at least, because the type `bitv[n]` expects `n` to be a non-negative integer. I don't know if we want to change that and what it would entail.
For now, I tried to add `bit2nat` and `nat2bv`, I had to add the `bvget` symbol which is an access on a bitvector and returns an integer (1 or 0) .
I tried to do something similar to what is done in [records.ml](https://github.com/OCamlPro/alt-ergo/blob/963e40c76daa3e598296f75c7341c0ed4f301319/src/lib/reasoners/records.ml) ~~but the current version is not sound. It can be seen on the added tests.
The issues seems to be either related to values not being propagated correctly, or the CC not realizing some values are equal. Using the debug options `-dcc` and `-duf` shows this.~~
This is probably not the best way to support theory, but this can be a placeholder for now (if it ends up working).


Gbury:
As far as I know, one of the reasons why some of the float reasoning is hidden behind the `--use-fpa` flag is that the presence of the flag makes alt-ergo automatically include the theory file for fpa (see `src/preludes/fpa-theory-*.ae`). This file contains an axiomatisation of floating point arithmetic, and thus it adds a significant number of quantified axioms, which might (or might not) change the behaviour of alt-ergo even on problems not using floats.

On the other hand, as far as I can tell, the new `--use-bv` flag should not have that drawback. All it does is enable translation of some functions into their builtin symbols, and some more reasoning. Therefore, I'm not sure that it should be a flag, or at least it should probably be on by default.

hra687261:
> presence of the flag makes alt-ergo automatically include the theory file for fpa

Unfortunately no, you still have to manually add the prelude file, the `--use-fpa` option just tells AE to translate the FPA primitives to Alt-Ergo symbols interally. But it should be done automatically or at least it should have a default prelude file that is loaded whenever `--use-fpa` is set and which can be overwritten by `--prelude` option.

> On the other hand, as far as I can tell, the new --use-bv flag should not have that drawback. All it does is enable translation of some functions into their builtin symbols, and some more reasoning. Therefore, I'm not sure that it should be a flag, or at least it should probably be on by default.

True, I'll remove it before making the PR ready for review.

Another issue is that there doesn't seem to be a mechanism to simply add primitives to Alt-Ergo's legacy front-end, so the new BV primitives need to be declared as uninterpreted symbols (like in the examples [testfile-bitv0025.ae](https://github.com/hra687261/alt-ergo-fuzz/blob/83474cdb2061c91de6fe81acb4647c552e284dc0/tests/bitv/testfile-bitv0025.ae)) or they need to be added as keywords to the language that will be parsed and type-checked.

Edit: The unsoundness has been fixed. ~~On the other hand AE for now can't prove this: [testfile-bitv0026.ae](https://github.com/hra687261/alt-ergo-fuzz/blob/83474cdb2061c91de6fe81acb4647c552e284dc0/tests/bitv/testfile-bitv0026.ae) which is problematic, I'm thinking of ways to add that `a = b <=> a^{0} = b^{0} & ... & a^{n-1} = b^{n-1}` for `a` and `b` two bit-vectors of type `bitv[n]`.~~

hra687261:
These changes are included in #669 669

PR #558, opened 2023-04-02
Test the dolmen front-end only once

hra687261:
If the point is only to test that the dolmen front-end parses and types correctly, then the tests should be ran with it only once. Currently all the tests, with all the combinations of options, are run twice with the two front-ends.


hra687261:
Looks like a few tests (that don't even use the dolmen frontend) are timing out. That's not likely to be caused by this PR.

Halbaroth (approved):
LGTM

PR #559, opened 2023-04-03
Prevent tests from running during the build passes

### LINKED ISSUE #557: Compiling runs all the tests
Doing a `make` or `dune build` runs all the non-regression tests which takes a good amount of time and significantly slows down development. The tests shouldn't be ran if not explicitly asked to or the size of the test-suite needs to be significantly decreased.
On the other hand doing  a `dune exec -- alt-ergo ...` doesn't run the tests, which is good but still, someone should be able to build without running the binary.

Halbaroth:
After this PR, if you run `make bin` or `make lib`, no tests are run anymore. You have to type `make runtest`, respectively `make runtest-ci`, to run quick tests, respectively CI tests.

This PR should fix #557

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,7 +149,7 @@ gentest: $(wildcard tests/**/*)
 
 # Run non-regression tests.
 runtest: gentest bin
-	dune build @runtest
+	dune build @runtest-quick
 
 # Run non-regression tests for the CI.
 runtest-ci: gentest bin
--- a/Makefile
+++ b/Makefile
@@ -112,7 +112,8 @@ plugins: gen
 # on "lib", "bin" and "gui", since dune can
 # parralelize more
 all: gen
-	$(DUNE) build $(DUNE_FLAGS)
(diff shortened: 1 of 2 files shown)


Halbaroth:
Is it ok for you?

hra687261:
Not ideal, running `make gentest` followed by `make` runs all the tests while building the executable, which means that there's a good chance it will do that after every build (following an edit).
But there doesn't seem to be an alternative, I didn't do a lot of searching but I didn't find much about it.

Halbaroth:
Why do you run `make`? Most of the time, you don't need to build the gui for instance. It makes sense that `make` runs all the targets, including the tests.

Halbaroth:
I found another solution. Basically, I gave all the desired targets to dune directly. If you run `make` now, Dune will compile everything but doesn't run tests. Is it fine?

hra687261:
Great!

hra687261 (approved):

PR #560, opened 2023-04-03
Fix CI

hra687261:


Halbaroth:
I can reproduce the error using the same version of OCaml (4.10.0) and js_of_ocaml (5.1.1). The error disappears with js_of_ocaml 5.0.1. 

A short guideline for this kind of problem with js_of_ocaml:
1. Update your opam list: `opam update`
2. Create a local opam switch and install the same version of OCaml and js_of_ocaml
3. Try to reproduce the error with this switch
4. If it doesn't work... you can pray
5. If it works, downgrade js_of_ocaml until the error message disappears

hra687261:
That seems correct, thanks!
This will need a real fix at some point, but for now I think it's safe to set an upper bound for js_of_ocaml.
Also the version of ocaml that is used for the CI can/should probably be increased to something like 4.12.0.

Halbaroth (approved):

PR #561, opened 2023-04-03
Fix bitv extract and definition of functions

Halbaroth:
This PR fixes three things:
- The size of a bitvector after applying the extract operator was not correctly computed in the Dolmen frontend.
- The current implementation of AE refuses any term of the form `a = b` where `a` and `b` are two terms of type `Ty.Tbool`. Such equalities have to be transformed into equivalences `a <-> b`. It was not done for the definition of functions whose the return type is boolean. 
- The printer of legacy Typed AST didn't print correctly the concatenation of bitvectors.

The fact we didn't notice these regressions before means we have no proper tests for bitvectors!

### DIFF
--- a/src/lib/frontend/d_cnf.ml
+++ b/src/lib/frontend/d_cnf.ml
@@ -621,10 +621,11 @@ let rec mk_expr ?(loc = Loc.dummy) ?(name_base = "")
             in
             E.mk_term (Sy.Op Sy.Minus) [e1; aux_mk_expr x] ty
 
-          | B.Bitv_extract { i; j; n }, [x] ->
+          | B.Bitv_extract { i; j; _ }, [x] ->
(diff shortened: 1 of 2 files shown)


hra687261:
LGTM!

hra687261 (approved):

hra687261:
Looks like the reason why it was not noticed before is that in [bitv.ml](https://github.com/OCamlPro/alt-ergo/blob/963e40c76daa3e598296f75c7341c0ed4f301319/src/lib/reasoners/bitv.ml) the size of the resulting bit-vector is not taken from the type, but recalculated from the parameters:
https://github.com/OCamlPro/alt-ergo/blob/963e40c76daa3e598296f75c7341c0ed4f301319/src/lib/reasoners/bitv.ml#L228-L241

Issue #477, opened 2021-08-16
Unsoundness bug (cu_685)


hra687261:
[c_685_l.smt2.txt](https://github.com/OCamlPro/alt-ergo/files/6992923/c_685_l.smt2.txt)
[c_685_l.ae.txt](https://github.com/OCamlPro/alt-ergo/files/6992924/c_685_l.ae.txt)

The .ae and .smt2 files are supposed to be equivalent.

Alt-Ergo responds with "Valid" (on both the ".ae" and the ".smt2" files) when the Tableaux or the Tableaux-CDCL solver is used, it takes too long (more than 3 minutes) to answer when the CDCL-Tableaux solver is used, and answers with "I don't know" when the CDCL solver is used.

CVC5 answers with "sat" on the ".smt2" file, which is contradictory to Alt-Ergo's answer (With the Tableaux and the Tableaux-CDCL solvers).

Commentary:
- If we replace the line 13 in the ".smt2" file (and the line 12 in the ".ae" file) with ```iuqv0``` or a constant integer, then Alt-Ergo answers with "I don't know" which is the expected response.

PR #562, opened 2023-04-04
[arith] Add contextual explanation when tightening equalities

### LINKED ISSUE #479: Unsoundness bug (cu_735)
[c_735_l_bis.smt2.txt](https://github.com/OCamlPro/alt-ergo/files/6998557/c_735_l_bis.smt2.txt)
[c_735_l_bis.ae.txt](https://github.com/OCamlPro/alt-ergo/files/6998558/c_735_l_bis.ae.txt)

Responses:

|   &nbsp;  |   .smt2   |    .ae    |
|-----------|-----------|-----------|
|   AE(T)   |   Valid   |   Valid   |
|   AE(T-C) |   Valid   |   Valid   |
|   AE(C-T) |   Valid   |   Valid   |
|   AE(C)   |   Valid   |   Valid   |
|   CVC5    |    sat    |   &nbsp;  |

T: Tableaux
C: CDCL
AE: Alt-Ergo (used-solver)

Comments:
  - Replacing the line 13 in the ".smt2" file (line 11 in the ".ae" file) with ```1``` which is equal to ```ui_3``` divided by itself if ```ui_3``` is not zero, makes AE respond with "I don't know" with the four solvers. Removing any or both of the axioms (1st and 2nd assertions) does that too.

bclement-ocp:
When asserting an equality [r1 = r2] between two terms, we are tightening the intervals of both terms to the common intersection, and add the explanations of the equality to the new bounds (this is necessary to preserve soundness).

The intersection function tries to be smart, and immediately raises a "not consistent" exception if it finds that the intervals are disjoint, and it does so *before* we added the explanations to the bounds, which may cause us to incorrectly backjump decisions that are actually relevant 💣

Fixes #479

### DIFF
--- a/src/lib/reasoners/intervalCalculus.ml
+++ b/src/lib/reasoners/intervalCalculus.ml
@@ -1517,7 +1517,25 @@ let tighten_eq_bounds env r1 r2 p1 p2 origin_eq expl =
       Debug.tighten_interval_modulo_eq_begin p1 p2;
       let i1, us1, is_mon_1 = generic_find r1 env in
       let i2, us2, is_mon_2 = generic_find r2 env in
-      let j = I.add_explanation (I.intersect i1 i2) expl in
+      (* We are assuming [origin_eq] with explanation [expl], where [origin_eq]
+       * is an equality between terms [r1] and [r2] with respective intervals
+       * [i1] and [i2].
+       *
+       * The shared interval for both [r1] and [r2] after the assumption
+       * becomes the intersection of [i1] and [i2], and must account for the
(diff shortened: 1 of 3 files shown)


Halbaroth (approved):
Great! The benchmarks are running :)

Gbury (approved):

bclement-ocp:
I believe this may also fix #477.

At least I can reproduce #477 on next but not with this patch. I have not investigated further, but it seems coherent, since the cases for #479 and #477 are similar (comparison with an `ite`).

Halbaroth:
No regression found! Congratulation :D
I also thought your fix solved the issue #477

PR #569, opened 2023-04-12
Use dune-build-info for version number

bclement-ocp:
Currently making alt-ergo releases involves changing the version in version.ml and running a bunch of scripts.

This means that it is rarely done, and that executables built the "regular" way (i.e. not for a release) always report "dev" as their version, which is less than ideal for determining from what source a specific executable was built.

This patch uses the dune-build-info library to get the commit version that was used when building alt-ergo. Because the dune-build-info version is only available when the executable has been either promoted or installed, and we usually don't want to install dev binaries, the alt-ergo binary is now promoted into the source directory (in src/bin/text/Main_text.exe).  This binary has the current git hash embedded as version through dune-build-info.  To facilitate its use, the toplevel `alt-ergo` symlink comes back, and now points to `src/bin/text/Main_text.exe`.

In addition, the public-release Makefile target now replaces the tags %%VERSION_NUM%%, %%VCS_COMMIT_ID%% and %%BUILD_DATE%% (with the same behavior as `dune subst` or `dune-release`, except for the 3rd one that doesn't exist in these tools) with their appropriate values and no longer resort to brittle cuts and the like.  Although I don't know how important this is because `make public-release` is clearly broken right now (e.g. it tries to copy files from the directory licences instead of licenses).


bclement-ocp:
Note that the current version we get on next from dune-build-info is `2.1.0-684-g30e1367` which is wrong. But this is next, so that's fine (although it really should be 2.4.2-something, ideally).

For this to work for releases the only requirement is to use annotated tags to tag the releases as recommended by https://git-scm.com/docs/git-tag (which I am not sure we have been doing, but is just a matter of passing `-a` to `git tag`). I believe this is already what GitHub releases do, although I am not 100% sure.

Halbaroth (commented):
Fine :D

Halbaroth (approved):

PR #572, opened 2023-04-14
clean shostak

lefessan:
Names `X1`...`X8` should be replaced by the theories they correspond to, since there are pattern-matchings relying on their meaning.


Halbaroth:
Can you run the linter please? I will accept your first PR :)

Halbaroth (approved):

PR #579, opened 2023-04-17
Remove (version dev) field in dune-project

bclement-ocp:
The presence of an explicit "version" field in dune-project causes dune-build-info to use the value from the "version" field, which defeats the purpose of using it for dev builds.

We re-added the "version" field because `opam install . --deps-only` did not work, because opam was not able to install version "dev" of alt-ergo-lib and alt-ergo-parsers (which makes sense since it is defined in the same repo). We are now ignoring dependencies on these packages when running `make deps` and `make dev-switch` instead.


Halbaroth (approved):
LGTM

PR #580, opened 2023-04-17
Use a weak table for the Shostak.Combine cache

bclement-ocp:
Backports b2ca04f8be5c27d1f54781ba64ca07d22c8e0538 from 2.4.x into next


Halbaroth (approved):
Since the regression is due to the weak table, I agree to merge this PR and we will move the test on Marvin.

Halbaroth:
LGTM

PR #583, opened 2023-04-19
CI fix for dune-build-info

bclement-ocp:

PR #584, opened 2023-04-19
Merge builds for all OSes

bclement-ocp:
This simplifies the CI infrastructure by de-duplicating the different
build_XXX.yml into a single build.yml.

The resulting build.yml is simpler than the previous ones in several
ways:

 - Action ocaml/setup-ocaml already takes care of caching opam
   information, so we don't need to cache it a second time

 - The alt-ergo binary was only generated as an artifact on direct
   pushes on main/master but we now have a PR-based workflow, so this
   part of the workflow was not used. It can be reintroced later if needed.

 - We no longer build with the huge matrix of different OCaml releases
   because, again, they were not used --- they all built with the
   OCAML_DEFAULT_VERSION instead of using the actual version from the
   matrix. As a side note, dolmen is a dependency that requires ocaml >=
   4.08 so there is no chance we can build with 4.05 anymore.

   From an environmental point of view, it is also unnecessary to build
   everything with all these OCaml versions all the time. We can
   reintroduce a (working) set of OCaml versions if necessary (it will
   just require adding the appropriate `include` entries to the `matrix`)

Note that Windows builds are currently broken because dolmen 0.8.1 is
not available in the Windows opam:
  https://github.com/ocaml-opam/opam-repository-mingw/tree/sunset

NB: This PR currently includes #583


Halbaroth (approved):
LGTM

PR #591, opened 2023-04-20
Support preludes with the dolmen frontend

### LINKED ISSUE #585: Option `--prelude` does not work with `--frontend dolmen`
Consider:

```bash
$ cat real.ae
goal g : real_of_int(0) = 0.0
$ alt-ergo  --use-fpa --prelude fpa-theory-2019-10-08-19h00.ae real.ae
File "real.ae", line 1, characters 10-30: Valid (0.0061) (0 steps) (goal g)
$ alt-ergo --frontend dolmen --use-fpa --prelude fpa-theory-2019-10-08-19h00.ae real.ae
File "real.ae", line 1, character 9-20:
1 | goal g : real_of_int(0) = 0.0
             ^^^^^^^^^^^
Error Unbound identifier: `real_of_int`
```

hra687261:
Fixes #585
While adding the fix I noticed an error in Dolmen:
https://github.com/Gbury/dolmen/blob/a5632a000a8c4ed10c7653574fdf9d9af11a0b05/src/standard/expr.ml#L1848-L1850
It expects the in_interval trigger to only support integers, while in the fpa preludes files it also supports reals.
In the legacy frontend it's apparently polymorphic and just expects the bounds to be of the same type as the expression for which the check is done:
https://github.com/OCamlPro/alt-ergo/blob/959f56880f1dbb9426aed842c345d00c0a9493df/src/lib/frontend/typechecker.ml#L1252-L1258
I will first add the fix in Dolmen before this PR can be merged.

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -525,7 +525,13 @@ let main () =
       Options.with_timelimit_if (not (Options.get_timelimit_per_goal ()))
       @@ fun () ->
 
-      let st, g = Parser.parse_logic [] st (State.get State.logic_file st) in
+      let preludes =
+        List.map (fun f -> Dolmen_std.Statement.include_ f [])
+          (Options.get_preludes ())
+      in
+      let st, g =
+        Parser.parse_logic preludes st (State.get State.logic_file st)
+      in
       let all_used_context = FE.init_all_used_context () in
       let finally = finally ~handle_exn in
       let st =
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -526,7 +526,7 @@ let main () =
       @@ fun () ->
 
       let preludes =
-        List.map (fun f -> Dolmen_std.Statement.include_ f [])
+        List.map Dolmen_std.Statement.import
           (Options.get_preludes ())
       in
       let st, g =


bclement-ocp:
@hra687261 What is the status on this?

My understanding is that prelude support with the dolmen frontend works in this PR, but the current FPA prelude is not properly parsed by Dolmen, regardless of prelude support. If that is correct, we should merge this so that the `--prelude` option works (even if the FPA prelude will not until Gbury/dolmen#137 is merged).

src/bin/common/solving_loop.ml
```diff
@@ -525,7 +525,13 @@ let main () =
       Options.with_timelimit_if (not (Options.get_timelimit_per_goal ()))
       @@ fun () ->
 
-      let st, g = Parser.parse_logic [] st (State.get State.logic_file st) in
+      let preludes =
+        List.map (fun f -> Dolmen_std.Statement.include_ f [])
```
bclement-ocp: Nit: I think we can do

```suggestion
        List.map Dolmen_std.Statement.import
```

hra687261:
That's correct, although since the preludes contain some features that are not that frequently used elsewhere (triggers, semantic triggers and filters) I think it would be a good idea to bench AE using `--use-fpa --prelude ...` with both front-ends to make sure that the behavior doesn't change.

bclement-ocp (approved):
:tada:

bclement-ocp:
Ignore the test JS/Make test failures, see #631 

Yes we should bench / test with the `--use-fpa` and the preludes to ensure no regression between frontends. That part will have to wait for Gbury/dolmen#137 however! Opening an issue

PR #593, opened 2023-04-20
Removing alt-ergo-js.opam

Halbaroth:
It was a mistake to add this package. This PR removes it.
If you want to compile Alt-Ergo into js, you have to do:
```
make deps
make js-deps
make js-node
```
The target `js-deps` will install the required dependencies with proper versions in your current switch.
If the CI is happy, I will update the documentation.


bclement-ocp:
I don't really know how try-alt-ergo is made, but otherwise it looks OK to me.

bclement-ocp (approved):

PR #598, opened 2023-05-16
Fix memory leak

Halbaroth:
This PR fixes the memory leak in the module `Vec.ml`. The leak comes from the function: 
```ocaml 
let pop t = assert (t.sz >=1); t.sz <- t.sz - 1
```
The pop function does not replace the popped value by the dummy value.

Basically, I replace the old module `Vec.ml` by its clean version from mSAT repository. 
I also replaced as much as I can the for-loops in the codebase by iterators from the Vec module.
Yet, as I noticed months ago, I cannot abstract the vector type because of this code in Satml: 
```ocaml
  let rec dummy_var =
    { vid = -101;
      pa = dummy_atom;
      na = dummy_atom;
      level = -1;
      index = -1;
      reason = None;
      weight = -1.;
      sweight = 0;
      seen = false;
      vpremise = [] }
  and dummy_atom =
    { var = dummy_var;
      timp = 0;
      lit = dummy_lit;
      watched = {Vec.dummy=dummy_clause; data=[||]; sz=0};
      neg = dummy_atom;
      is_true = false;
      is_guard = false;
      aid = -102 }
  and dummy_clause =
    { name = "";
      atoms = {Vec.dummy=dummy_atom; data=[||]; sz=0};
      activity = -1.;
      removed = false;
      learnt = false;
      cpremise = [];
      form = vraie_form }
```

### DIFF
--- a/src/lib/util/vec.ml
+++ b/src/lib/util/vec.ml
@@ -9,144 +9,144 @@
 (*                                                                            *)
 (******************************************************************************)
 
-type 'a t = { mutable dummy: 'a; mutable data : 'a array; mutable sz : int }
+type 'a t = {
+  mutable data : 'a array;
+  mutable sz : int;
+  dummy: 'a;
+}
 
-let make capa d = {data = Array.make capa d; sz = 0; dummy = d}
+let make n ~dummy = {data = Array.make n dummy; sz = 0; dummy}
 
-let init capa f d =
-  {data = Array.init capa (fun i -> f i); sz = capa; dummy = d}
+let[@inline] create ~dummy = {data = [||]; sz = 0; dummy}
 
-let from_array data sz d = {data = data; sz = sz; dummy = d}
+let[@inline] clear vec = vec.sz <- 0
 
-let from_list l sz d =
-  let l = ref l in
-  let f_init _ = match !l with [] -> assert false | e::r -> l := r; e in
-  {data = Array.init sz f_init; sz = sz; dummy = d}
+let[@inline] shrink vec i =
+  assert (i >= 0);
+  assert (i <= vec.sz);
+  for j = vec.sz - i to vec.sz - 1 do
+    Array.unsafe_set vec.data j vec.dummy
+  done;
+  vec.sz <- vec.sz - i
 
-let clear s = s.sz <- 0
+let[@inline] pop vec =
+  assert (vec.sz > 0);
+  let x = Array.unsafe_get vec.data (vec.sz - 1) in
+  Array.unsafe_set vec.data (vec.sz - 1) vec.dummy;
+  vec.sz <- vec.sz - 1;
+  x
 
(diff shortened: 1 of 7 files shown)


src/lib/util/vec.ml
```diff
@@ -9,144 +9,144 @@
...
-
-#endif
-*)
+    let elt = Array.unsafe_get vec.data i in
+    if not (elt == vec.dummy) then
+      acc := f !acc elt
+  done;
+  !acc
+
+let to_list a = Array.to_seq a.data |> List.of_seq
```
bclement-ocp, line 139: I'd expect this to return a list of length [a.sz], not [Array.length a.data].

bclement-ocp (changes requested):
This does two (three?) things at once:

 - Change the implementation of [Vec] to use that from mSAT (or a derived thereof, see below)
 - Change all uses of direct [for] loop iterations to [Vec.iter]
 - Change [Vec.make] to take the [dummy] value as a keyword argument

I'd expect the 3rd one to be inconsequential, but the first and second point -- although they look mostly correct to me, except for `Vec.to_list` as noted -- can both be causes for subtle issues and it would be good to separate them.

It looks like this is already the case, but the history is messy -- it would be better to keep both commits separated in the history. Can you either make two separate PRs with these changes, or make sure that this PR has clean history with exactly two commits (one for each change) that we won't squash when merging?

By the way, it looks like the [to_list] definition is different from that in mSAT. Can you clarify the difference  between the new implementation of [Vec] and the one from mSAT?

Halbaroth:
Actually, we couldn't use directly the implementation of Vec in mSAT since we use a dummy field in our vector structure and mSAT does not. Thus it is complicated to separate the first and third task. The second task is entirely done in the commits [3c39562](https://github.com/OCamlPro/alt-ergo/pull/598/commits/3c39562db374c963ecf8a3737e76dab00bfd6859) and [72db3fd](https://github.com/OCamlPro/alt-ergo/pull/598/commits/72db3fdc9a83b47d922b4f2a921fd574abb5c49a).

src/lib/util/vec.ml
Halbaroth: You are right. I will fix it.

bclement-ocp:
I don't think "replace the old module `Vec.ml` by its clean version from mSAT repository" is an accurate description of the changes in that case. Especially since the goal of this PR is to prevent a memory leak that, after checking, is actually present in mSAT's `Vec.ml` as well.

This is just a third implementation of `Vec` at that point, with its own bugs (see `to_list`, and from a cursory look `exists` seems to be able to call the predicate on the `dummy` values as well), that requires an actual careful review to be merged, whereas the fix for the stated bug:
> The pop function does not replace the popped value by the dummy value.

should be a one-line change.

Halbaroth:
Ok

PR #515, opened 2022-07-07
Fixing issue 52

Stevendeo:
Handles #52

PR #599, opened 2023-05-16
Remove polymorphism compares

Halbaroth:
This PR removes all the polymorphic comparisons but on integers. 
- I added a prelude that is automatically open in every modules of `Alt_ergo_lib`.
- I also removed some polymorphic comparisons in Bitv forgotten by Steven in PR #515 .


src/lib/reasoners/bitv.ml
```diff
@@ -104,7 +117,7 @@ module Shostak(X : ALIEN) = struct
 
   let compare_simple_term = compare_alpha_term (fun st1 st2 ->
       match st1, st2 with
-      | Cte b,Cte b' -> compare b b'
+      | Cte true, Cte true | Cte false, Cte false -> 0
```
bclement-ocp: We should use:

```suggestion
      | Cte b, Cte b' -> Bool.compare b b'
```

src/lib/reasoners/bitv.ml
```diff
@@ -119,13 +132,15 @@ module Shostak(X : ALIEN) = struct
           if c2 <> 0 then c2 else compare_xterm t1 t2
     )
 
+  let compare_abstract = Lists.compare compare_simple_term
+
   let compare_solver_simple_term = compare_alpha_term (fun st1 st2 ->
       match st1, st2 with
-      | S_Cte b, S_Cte b' -> compare b b'
+      | S_Cte true, S_Cte true | S_Cte false, S_Cte false -> 0
```
bclement-ocp: See above, use `Bool.compare`

src/lib/reasoners/bitv.ml
```diff
@@ -233,11 +249,12 @@ module Shostak(X : ALIEN) = struct
 
     let string_to_bitv s =
       let tmp = ref[] in
-      String.iter(fun car -> tmp := (car<>'0',1)::(!tmp)) s;
+      String.iter(fun car -> tmp := (Char.compare car '0' <> 0, 1)::(!tmp)) s;
```
bclement-ocp: Use `Char.(car <> '0')` instead

src/lib/reasoners/bitv.ml
```diff
@@ -617,9 +644,9 @@ module Shostak(X : ALIEN) = struct
         |a::b::r ->
           begin
             match a.bv,b.bv with
-            |Cte bol,Cte bol' when bol = bol' ->
+            | Cte true, Cte true | Cte false, Cte false ->
```
bclement-ocp: Use `when Bool.(bol = bol')`

src/lib/util/lists.ml
```diff
@@ -62,3 +66,13 @@ let to_seq l =
   in
   aux l
 
+let rec compare cmp l1 l2 =
```
bclement-ocp, line 70: This is defined in OCaml's standard library as `List.compare`, but sadly only from `4.12`. Maybe leave a `TODO` to remove this and use the standard `List.compare` in the future?

bclement-ocp (changes requested):
LGTM with the noted changes — let's try to use existing monomorphic or parameterized equality/comparison functions when possible.

(as a side note: ideally we would use `ppx_deriving`... maybe we should start doing so for new types -- actually looks like this is planned in #538 )

src/lib/reasoners/bitv.ml
Halbaroth: Is it faster or that is a matter of taste? I wrote `Bool.compare` also but since Steven used the other style, I thought it could be faster.

src/lib/util/lists.ml
Halbaroth: I know but sure I will add a comment :)

src/lib/reasoners/bitv.ml
bclement-ocp: The performance difference should be negligible (`Bool.compare` is most probably inlined anyways), but the version with `Bool.compare` has less mental overhead: since it is a catch-all case, one doesn't have to keep track of the cases `true, false` and `false, true` while reading the rest of the code, and it is more obviously correct.

src/lib/reasoners/bitv.ml
Halbaroth: I got a warning: 
```
File "src/lib/reasoners/bitv.ml", line 647, characters 34-38:
647 |             | Cte b1, Cte b2 when Bool.(b1 = b2) ->
                                        ^^^^
Warning 33 [unused-open]: unused open Stdlib.Bool.
```

bclement-ocp: Sorry, should be `Bool.equal b1 b2`, I got confused

Halbaroth:
Done :)

bclement-ocp (approved):

Issue #590, opened 2023-04-20
Removing the support for the num library


Halbaroth:
The current version of AE allows using Zarith or Num for arbitrary-precision arithmetic. As @hra687261 told me, the support of Num is not discontinued as it is explained on their repository: 

> This is a legacy library. It used to be part of the core OCaml distribution (in otherlibs/num) but is now distributed separately. New applications that need arbitrary-precision arithmetic should use the Zarith library (https://github.com/ocaml/Zarith) instead of the Num library, and older applications that already use Num are encouraged to switch to Zarith. Zarith delivers much better performance than Num and has a nicer API.

I suggest to remove completely the support of Num for the next release.

PR #600, opened 2023-05-17
Remove nums

Halbaroth:
This PR removes the support for the deprecated Nums library as suggested in #590.


bclement-ocp:
The parser and ast are still using num in this PR I believe, is that intentional?

Halbaroth:
No, I didn't notice it... I modified the legacy parser to use only Zarith.

src/parsers/psmt2_to_alt_ergo.ml
```diff
@@ -115,8 +115,8 @@ module Translate = struct
     in
     aux sort.ty
 
-  let better_num_of_string s =
-    begin match String.split_on_char '.' s with
+  (* -- unused -- let better_num_of_string s =
```
bclement-ocp: Let's just remove this. It is in the git history if we really need it.

bclement-ocp (approved):
This looks good to me, please just remove the old code from the comments.

I believe `decimal_number` (and `hexa_number`) could be implemented with `Q.of_string` (`Q.of_string ("0x" ^ x)`, respectively), but the specialized implementation is probably a tiny bit faster. Not that it matters much…

src/parsers/psmt2_to_alt_ergo.ml
Halbaroth: Sure but we will remove the legacy parser so I did the minimal effort.

PR #602, opened 2023-05-23
Refactoring the debug commands

Halbaroth:
This PR refactors the way that AE parses debug commands. In the previous version, AE used a distinct flag for each debug option without using the enum convertor of the cmdliner library. For instance, this means we had to type something as follow:
``` 
alt-ergo input.ae --dsat --dcc --dcombine
```
After this PR, we can type both
```
alt-ergo input.ae --debug sat,cc,combine
``` 
and
```
alt-ergo input.ae -d sat -d cc -d combine
```
Instead of using a level of verbosity per debug flag, I create a new option `debug-verbosity`. So if you want more verbosity for the matching module for instance, type:
```
alt-ergo input.ae -d matching --debug-verbosity 2
```

### DIFF
--- a/src/bin/common/parse_command.ml
+++ b/src/bin/common/parse_command.ml
@@ -150,80 +150,157 @@ let formatter_printer fmt formatter =
 
 let formatter_conv = Arg.conv ~docv:"FMT" (formatter_parser, formatter_printer)
 
-type rule = RParsing | RTyping | RSat | RCC | RArith | RNone
+module Rule = struct
+  type t = RParsing | RTyping | RSat | RCC | RArith | RNone
 
-let value_of_rule = function
-  | RParsing -> 0
-  | RTyping -> 1
-  | RSat -> 2
-  | RCC -> 3
-  | RArith -> 4
-  | RNone -> -1
+  let all = [RParsing; RTyping; RSat; RCC; RArith; RNone]
 
-let rule_parser = function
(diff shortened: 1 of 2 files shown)


bclement-ocp:
That looks nice! I am not sure about the shared `debug_verbosity` setting as it prevents having different verbosities for different modules (rare case, but still can be useful).

Can we do somethink like the `verbose` flag which increase verbosity when passed (`-vv` is more verbose than `-v` etc.)? So that we could do `alt-ergo -d matching` to get verbosity level 1 and `alt-ergo -d matching -d matching` (or maybe even `alt-ergo -dd matching`?) to get level 2.

Halbaroth:
I agree adding a separate option is not that great. I thought counting the number of occurrence but it is quite ugly and inconvenient to have to repeat the name of the flag. I like your second proposition. You can increase the level of the verbosity with only one character for all the used flags and you keep the possibility to have a different level per flag.

Halbaroth:
Done :)

bclement-ocp:
Looks good! Does `cmdliner` parses `-ddd` properly here, or do we need `--ddd` with double dash?

bclement-ocp (approved):

Halbaroth:
Alas, it seems we need the double dash... I will check if there is an issue about this on the cmdliner repository.

PR #604, opened 2023-05-24
Actually use a weak cache for Shostak.Combine

bclement-ocp:
PR #580 was hilariously wrong, but nobody noticed. In that implementation, we only keep a hold on the key, but not the [weak_t] value --- so it is immediately dead and will be collected in the next GC cycle.

This PR use `Ephemeron.K1.Make` instead, which is the appropriate structure for weak hash tables.


Halbaroth (approved):
LGTM

PR #605, opened 2023-05-24
Better list comparisons

Halbaroth:
This PR removes the old comparison function `Util.cmp_lists`.  The new function `Lists.compare` is faster and similar to the function added in Ocaml 4.12.

This PR includes #599


bclement-ocp:
It's not clear that `Lists.compare` is faster actually, did you benchmark it? The way this is written seems to try to exploit exceptions to bypass nested recursion levels (there is no repeated checks for `c <> 0` in `compare_triggers` for instance.

bclement-ocp:
Note to self: holding on reviewing until #610 is done

Halbaroth:
We decide during the last dev meeting to limit the usage of ppx derivers as `ppxlib` isn't compatible with `Flambda 2`.

PR #606, opened 2023-05-25
Fix the leak memory in Vec

Halbaroth:
This PR fixes the memory leak in the module `Vec.ml`. The leak comes from the function: 
```ocaml 
let pop t = assert (t.sz >=1); t.sz <- t.sz - 1
```
The pop function does not replace the popped value by the dummy value.


bclement-ocp (approved):
LGTM.

PR #607, opened 2023-05-25
New vec module

Halbaroth:
This PR is proposition for a new vector module.

Basically, I replace the old module `Vec.ml` by its clean version from mSAT repository. 
I also replaced as much as I can the for-loops in the codebase by iterators from the Vec module.
Yet, as I noticed months ago, I cannot abstract the vector type because of this code in Satml: 
```ocaml
  let rec dummy_var =
    { vid = -101;
      pa = dummy_atom;
      na = dummy_atom;
      level = -1;
      index = -1;
      reason = None;
      weight = -1.;
      sweight = 0;
      seen = false;
      vpremise = [] }
  and dummy_atom =
    { var = dummy_var;
      timp = 0;
      lit = dummy_lit;
      watched = {Vec.dummy=dummy_clause; data=[||]; sz=0};
      neg = dummy_atom;
      is_true = false;
      is_guard = false;
      aid = -102 }
  and dummy_clause =
    { name = "";
      atoms = {Vec.dummy=dummy_atom; data=[||]; sz=0};
      activity = -1.;
      removed = false;
      learnt = false;
      cpremise = [];
      form = vraie_form }
```

### DIFF
--- a/src/lib/util/iheap.ml
+++ b/src/lib/util/iheap.ml
@@ -33,8 +33,9 @@ type t = {heap : int Vec.t; indices : int Vec.t }
 let dummy = -100
 
 let init sz =
-  { heap    =  Vec.init sz (fun i -> i) dummy;
-    indices =  Vec.init sz (fun i -> i) dummy}
+  let lst = List.init sz (fun i -> i) in
+  { heap    =  Vec.of_list lst ~dummy;
+    indices =  Vec.of_list lst ~dummy}
 
 let left i   = (i lsl 1) + 1 (* i*2 + 1 *)
 let right i  = (i + 1) lsl 1 (* (i+1)*2 *)
@@ -106,7 +107,7 @@ let filter s filt cmp =
     end
     else Vec.set s.indices (Vec.get s.heap i) (-1);
   done;
-  Vec.shrink s.heap (lim - !j) true;
+  Vec.shrink s.heap (lim - !j);
   for i = (lim / 2) - 1 downto 0 do
     percolate_down cmp s i
   done
@@ -146,6 +147,6 @@ let remove_min cmp ({heap=heap; indices=indices} as s) =
   Vec.set heap 0 (Vec.last heap); (*heap.last()*)
   Vec.set indices (Vec.get heap 0) 0;
   Vec.set indices x (-1);
-  Vec.pop s.heap;
+  ignore(Vec.pop s.heap);
   if Vec.size s.heap > 1 then percolate_down cmp s 0;
   x
--- a/src/lib/util/iheap.ml
+++ b/src/lib/util/iheap.ml
@@ -98,15 +98,15 @@ let increase cmp s n =
 
 let filter s filt cmp =
   let j = ref 0 in
+  Vec.iter (fun elt ->
+      if filt elt then begin
+        Vec.set s.heap !j elt;
+        Vec.set s.indices elt !j;
+        incr j
+      end
+      else Vec.set s.indices elt (-1)
+    ) s.heap;
   let lim = Vec.size s.heap in
-  for i = 0 to lim - 1 do
-    if filt (Vec.get s.heap i) then begin
-      Vec.set s.heap !j (Vec.get s.heap i);
-      Vec.set s.indices (Vec.get s.heap i) !j;
-      incr j;
-    end
-    else Vec.set s.indices (Vec.get s.heap i) (-1);
-  done;
   Vec.shrink s.heap (lim - !j);
   for i = (lim / 2) - 1 downto 0 do
     percolate_down cmp s i
--- a/src/lib/util/iheap.ml
+++ b/src/lib/util/iheap.ml
@@ -98,14 +98,15 @@ let increase cmp s n =
 
 let filter s filt cmp =
   let j = ref 0 in
-  Vec.iter (fun elt ->
-      if filt elt then begin
-        Vec.set s.heap !j elt;
-        Vec.set s.indices elt !j;
-        incr j
-      end
-      else Vec.set s.indices elt (-1)
-    ) s.heap;
+  for i = 0 to Vec.size s.heap - 1 do
+    let elt = Vec.get s.heap i in
+    if filt elt then begin
+      Vec.set s.heap !j elt;
+      Vec.set s.indices elt !j;
+      incr j
+    end
+    else Vec.set s.indices elt (-1)
+  done;
   let lim = Vec.size s.heap in
   Vec.shrink s.heap (lim - !j);
   for i = (lim / 2) - 1 downto 0 do
--- a/src/lib/util/vec.ml
+++ b/src/lib/util/vec.ml
@@ -34,11 +34,17 @@ type 'a t = {
   dummy: 'a;
 }
 
+let[@inline] size vec = vec.sz
+
 let make n ~dummy = {data = Array.make n dummy; sz = 0; dummy}
 
 let[@inline] create ~dummy = {data = [||]; sz = 0; dummy}
 
-let[@inline] clear vec = vec.sz <- 0
+let[@inline] clear vec =
+  for i = 0 to size vec - 1 do
+    Array.unsafe_set vec.data i vec.dummy
+  done;
+  vec.sz <- 0
 
 let[@inline] shrink vec i =
   assert (i >= 0);
@@ -58,8 +64,6 @@ let[@inline] pop vec =
 let[@inline] last vec =
   assert (vec.sz > 0);
   Array.unsafe_get vec.data (vec.sz - 1)
-
-let[@inline] size vec = vec.sz
 
 let[@inline] is_empty vec = vec.sz = 0
 
--- a/src/lib/util/vec.ml
+++ b/src/lib/util/vec.ml
@@ -139,7 +139,7 @@ let exists p vec =
   try
     for i = 0 to size vec - 1 do
       let elt = Array.unsafe_get vec.data i in
-      if p elt && not (elt == vec.dummy) then raise Terminate
+      if not (elt == vec.dummy) && p elt then raise Terminate
     done;
     false
   with Terminate -> true
@@ -155,8 +155,8 @@ let fold f acc vec =
   done;
   !acc
 
-let to_list a = Array.to_seq a.data |> List.of_seq
 let to_array a = Array.sub a.data 0 a.sz
+let to_list a = Array.to_seq (to_array a) |> List.of_seq
 
 let of_list l ~dummy : _ t =
   match l with
--- a/src/lib/util/vec.ml
+++ b/src/lib/util/vec.ml
@@ -163,6 +163,10 @@ let sort vec f : unit =
   Array.fast_sort f arr;
   vec.data <- arr
 
-let pp ?(sep=", ") pp fmt a =
-  let pp_sep fmt () = Format.fprintf fmt "%s@," sep in
-  Format.pp_print_list ~pp_sep pp fmt (to_list a)
+let internal_iter f vec =
+  for i = 0 to size vec - 1 do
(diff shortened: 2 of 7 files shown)


src/lib/util/iheap.ml
```diff
@@ -33,8 +33,9 @@ type t = {heap : int Vec.t; indices : int Vec.t }
 let dummy = -100
 
 let init sz =
-  { heap    =  Vec.init sz (fun i -> i) dummy;
-    indices =  Vec.init sz (fun i -> i) dummy}
+  let lst = List.init sz (fun i -> i) in
+  { heap    =  Vec.of_list lst ~dummy;
+    indices =  Vec.of_list lst ~dummy}
```
bclement-ocp, line 38: Why this change?

src/lib/util/iheap.ml
```diff
@@ -97,16 +98,16 @@ let increase cmp s n =
...
 let filter s filt cmp =
   let j = ref 0 in
+  Vec.iter (fun elt ->
+      if filt elt then begin
+        Vec.set s.heap !j elt;
+        Vec.set s.indices elt !j;
+        incr j
+      end
+      else Vec.set s.indices elt (-1)
+    ) s.heap;
```
bclement-ocp: It is often implicitly understood that combinators such as `Vec.iter` do not modify the underlying container, so please don't use `Vec.iter` when the iterated function modifies the container and use a `for` loop in that case.

src/lib/util/vec.ml
```diff
@@ -28,144 +28,144 @@
...
+  dummy: 'a;
+}
 
-let init capa f d =
-  {data = Array.init capa (fun i -> f i); sz = capa; dummy = d}
+let make n ~dummy = {data = Array.make n dummy; sz = 0; dummy}
+
+let[@inline] create ~dummy = {data = [||]; sz = 0; dummy}
+
+let[@inline] clear vec = vec.sz <- 0
```
bclement-ocp: Should this overwrite the values with the dummy?

src/lib/util/vec.ml
```diff
@@ -28,144 +28,144 @@
...
+
+let[@inline never] grow_to vec cap : unit =
+  assert (Array.length vec.data < Sys.max_array_length);
+  let cap =
+    min Sys.max_array_length (max 4 cap)
+  in
+  let arr' = Array.make cap vec.dummy in
+  Array.blit vec.data 0 arr' 0 (Array.length vec.data);
+  vec.data <- arr';
+  assert (Array.length vec.data > vec.sz);
```
bclement-ocp: Should probably `assert (cap > vec.sz) before blitting instead

src/lib/util/vec.ml
```diff
@@ -28,144 +28,144 @@
...
+  vec.sz <- vec.sz + 1
+
+let[@inline] get vec i =
+  assert (0 <= i && i < vec.sz);
+  let res = Array.unsafe_get vec.data i  in
+  if res == vec.dummy then raise Not_found;
+  res
+
+let[@inline] set vec i elt =
+  assert (not (elt == vec.dummy));
```
bclement-ocp, line 110: `invalid_arg "Vec.set"` would be better

src/lib/util/vec.ml
```diff
@@ -28,144 +28,144 @@
...
+  vec.data.(i) <- elt;
+  vec.sz <- max vec.sz (i+1)
+
+let[@inline] fast_remove vec i =
+  assert (i>= 0 && i < vec.sz);
+  Array.unsafe_set vec.data i @@ Array.unsafe_get vec.data (vec.sz - 1);
+  vec.sz <- vec.sz - 1
```
bclement-ocp, line 118: Leaks memory

src/lib/util/vec.ml
```diff
@@ -28,144 +28,144 @@
...
+  let i = ref 0 in
+  while !i < size vec do
+    if f (Array.unsafe_get vec.data !i) then incr i else fast_remove vec !i
+  done
 
-let from_array data sz d = {data = data; sz = sz; dummy = d}
+let sort vec f : unit =
+  let sub_arr =
+    if is_full vec then vec.data else Array.sub vec.data 0 vec.sz
+  in
```
bclement-ocp: This is `to_array` I believe?

Also I don't think we ever sort vectors anyways…

src/lib/util/vec.ml
```diff
@@ -28,144 +28,144 @@
...
-  if fill_with_dummy then
-    for i = t.sz - i to t.sz - 1 do
-      t.data.(i) <- t.dummy
+let exists p vec =
+  let exception Terminate in
+  try
+    for i = 0 to size vec - 1 do
+      let elt = Array.unsafe_get vec.data i in
+      if not (elt == vec.dummy) && p elt then raise Terminate
     done;
```
bclement-ocp:
```suggestion
iter (fun elt -> if p elt then raise Terminate) vec
```

src/lib/util/vec.ml
```diff
@@ -28,144 +28,144 @@
...
-    return j < ts.size();
-}
-
-#endif
-*)
+    let elt = Array.unsafe_get vec.data i in
+    if not (elt == vec.dummy) then
+      acc := f !acc elt
+  done;
+  !acc
```
bclement-ocp:
```suggestion
let acc = ref acc in iter (fun e -> acc := f !acc e) vec; !acc
```

src/lib/util/vec.ml
```diff
@@ -28,144 +28,144 @@
...
-let clear s = s.sz <- 0
+let[@inline] iter f = iteri (fun _ elt -> f elt)
 
-let shrink t i fill_with_dummy =
-  assert (i >= 0 && i<=t.sz);
-  if fill_with_dummy then
-    for i = t.sz - i to t.sz - 1 do
-      t.data.(i) <- t.dummy
+let exists p vec =
+  let exception Terminate in
```
bclement-ocp: Is there a reason to use a local exception here? If not it is slightly more efficient to declare the exception outside of `exists` (I know I am guilty of doing this as well, but `let exception` registers a new exception each time `exists` is called I believe)

src/plugins/fm-simplex/simplex.ml
```diff
@@ -372,29 +372,22 @@ module Simplex (C : Coef_Type) = struct
...
           subst_in_p (snd (Vec.get sbt j)) (Vec.get sbt i)
         done;
       done;
-      let l1 = ref [] in
-      let l2 = ref [] in
-      for i = 0 to Vec.size sbt - 1  do l1 := (Vec.get sbt i)  :: !l1 done;
-      for i = 0 to Vec.size zsbt - 1 do l2 := (Vec.get zsbt i) :: !l2 done;
-      !l2, !l1
-
+      Vec.to_list zsbt, Vec.to_list sbt
```
bclement-ocp: I am not sure it matters (it very well may), but I think this builds the lists in the reverse order compared to the previous implementation.

bclement-ocp (commented):
I am not convinced by the removal of the `size` argument when creating arrays as it causes inefficiencies by iterating over the list another time just to compute its length, but I think this is correct otherwise.

src/lib/util/vec.ml
Halbaroth: I will check if satml assumes the vector is really erased after this operation.

src/lib/util/vec.ml
Halbaroth: We can remove useless function, or at least comment them.

src/lib/util/vec.ml
Halbaroth: No there is no reason to do that.

src/lib/util/vec.ml
Halbaroth: Using `invalid_arg` would be better?

src/lib/util/vec.ml
Halbaroth: The reason why we use assertions instead of exceptions is we can disable them for performance purposes.

src/lib/util/vec.ml
bclement-ocp: This is an internal function only used by the `Vec` module, doesn't need an `invalid_arg`. It is just cleaner to never create a `vec` in inconsistent state (we could also do `assert (Array.length arr' > vec.sz)`; in any case we should do it before setting `vec.data`.

src/lib/util/vec.ml
bclement-ocp: I don't think we actually do, but that makes sense.

Halbaroth:
> I am not convinced by the removal of the `size` argument when creating arrays as it causes inefficiencies by iterating over the list another time just to compute its length, but I think this is correct otherwise.

I noticed that when I wrote the PR but in my opinion, it is better to have a simpler API and  the overhead is probably tiny.

src/plugins/fm-simplex/simplex.ml
Halbaroth: I suggest to not preserve the old order but add a comment there?

src/lib/util/iheap.ml
Halbaroth: It is because the API of `Vec` changes and there is no `init` function anymore. Probably the older code is a bit faster.

src/lib/util/vec.ml
Halbaroth: There is two options:
- erasing the array itself. This could increase the number of allocations and change the behavior of the GC and performances
- replace all the values in the current array by the dummy value.

I think the second option is less dangerous.

src/plugins/fm-simplex/simplex.ml
Halbaroth: I keep the new code. If we observe regressions, I will check this behavior change again.

Halbaroth:
After testing on our internal set of tests, it seems this PR doesn't introduce regressions. We got +18-9 without correction for the weak table but +2-0 after correcting it.

src/lib/util/vec.ml
```diff
@@ -28,147 +28,145 @@
...
-  pop t
-
-
-let sort t f =
-  let sub_arr = Array.sub t.data 0 t.sz in
-  Array.fast_sort f sub_arr;
-  t.data <- sub_arr
-
-let iter vec f =
+let internal_iter f vec =
```
bclement-ocp, line 173: Why not either use `iter` directly or use this implementation for `iter`?

src/plugins/fm-simplex/simplex.ml
bclement-ocp: As far as I know we do not use the FM-Simplex plugin in benches… we probably will not notice anything without using it ;)

Given that this is code that applies to substitutions and the code seems to be sensible to ordering (see use of `downto` in the for loop above), I'd prefer if we made sure to keep the order the same, especially since this is code that we do not use and don't know very well. If we don't, we really should test it to make sure this doesn't break it.

bclement-ocp (commented):
Looks like the new implementation is slightly faster than the old one then :)

Looks good to me, just a question regarding iter + the fm-simplex module.

src/lib/util/vec.ml
Halbaroth: The exposed iterator `iter` ignores the dummy value, which means we don't print the dummy values silently if we use it. As this printer is only for debugging purposes, I prefer printing them. An alternative implementation consists in printing the indices of the elements.

bclement-ocp: OK, makes sense — a bit uncomfortable with printing those but as we discussed offline we will try to remove the dummy values in the reachable part of the vector entirely.

Halbaroth:
I create a new function `to_rev_list` to preserve the appropriate order in the FM-Simplex plugin. :)

bclement-ocp (approved):
Sorry, thought I had approved — let's merge

PR #553, opened 2023-03-13
Rebase of Optim Alt-Ergo

Halbaroth:

PR #608, opened 2023-05-26
Set option support

Halbaroth:
This PR aims to make AE fit better the SMT-LIB standards by adding the support for `set-option`.

- [x] support `:regular-output-channel` and `:diagnostic-output-channel`
- [x] support `:produce-models`
- [x] partial support `:verbosity`
- [x] partial support `:reproducible-resource-limit`
- [x] support `:produce-unsat-cores`

I don't plan to support other standard options in this PR. Maybe we could add `print-success` later. 

I had to refactor the way AE manages its output channels in order to be able to change the channel during the execution of smt2 scripts.

### DIFF
--- a/src/bin/common/parse_command.ml
+++ b/src/bin/common/parse_command.ml
@@ -581,12 +581,9 @@ let mk_opts file () () debug_flags ddebug_flags dddebug_flags rule () halt_opt
     `Ok true
   end
 
-let mk_output_channel_opt std_output deprecated_std_output err_output
-    deprecated_err_output mdl_output =
+let mk_output_channel_opt std_output err_output mdl_output =
   Options.Output.(create_channel std_output |> set_std);
-  Options.Output.(create_channel deprecated_std_output |> set_std);
   Options.Output.(create_channel err_output |> set_err);
-  Options.Output.(create_channel deprecated_err_output |> set_err);
   Options.Output.(create_channel mdl_output |> set_mdl);
   `Ok()
 
@@ -1208,24 +1205,36 @@ let parse_theory_opt =
        )
 
 let parse_fmt_opt =
-
   let docs = s_fmt in
   let docv = "CHANNEL" in
 
-  let std_output, deprecated_std_output =
+  let merge_formatters default preferred deprecated =
+    match preferred, deprecated with
+    | Some fmt, _ -> fmt
+    | None, Some fmt -> fmt
+    | None, None -> default
+  in
+
+  let std_output =
     let doc =
       Format.sprintf
         "Set the standard output used by default to print the results,
-    models and unsat cores. Possible values are %s."
+          models and unsat cores. Possible values are %s."
         (Arg.doc_alts ["stdout"; "stderr"; "<filename>"])
     in
     let deprecated = "this option is depreciated. Please use --std-output." in
-    Arg.(value & opt string "stdout" & info ["std-output"] ~docv ~docs ~doc),
-    Arg.(value & opt string "stdout" & info ~deprecated ["std-formatter"]
-           ~docv ~docs ~doc)
+    let std_output =
+      Arg.(value & opt (some' string) None & info ["std-output"] ~docs
+             ~doc ~docv)
+    in
+    let std_formatter =
+      Arg.(value & opt (some' string) None  & info ["std-formatter"]
+             ~deprecated ~docs ~docv)
+    in
+    Term.(const (merge_formatters "stdout") $ std_output $ std_formatter)
   in
 
-  let err_output, deprecate_err_output =
+  let err_output =
     let doc =
       Format.sprintf
         "Set the error output used by default to print error, debug and
@@ -1233,9 +1242,15 @@ let parse_fmt_opt =
         (Arg.doc_alts ["stdout"; "stderr"; "<filename>"])
     in
     let deprecated = "this option is depreciated. Please use --err-output." in
-    Arg.(value & opt string "stderr" & info ["err-output"] ~docv ~docs ~doc),
-    Arg.(value & opt string "stderr" & info ~deprecated ["err-formatter"]
-           ~docv ~docs ~doc)
+    let err_output =
+      Arg.(value & opt (some' string) None & info ["err-output"] ~docs
+             ~doc ~docv)
+    in
+    let err_formatter =
+      Arg.(value & opt (some' string) None & info ["err-formatter"]
+             ~deprecated ~docs ~docv)
+    in
+    Term.(const (merge_formatters "stderr") $ err_output $ err_formatter)
   in
 
   let model_output =
@@ -1247,8 +1262,8 @@ let parse_fmt_opt =
     Arg.(value & opt string "stdout" & info ["model-output"] ~docv ~docs ~doc)
   in
 
-  Term.(ret (const mk_output_channel_opt $ std_output $ deprecated_std_output
-             $ err_output $ deprecate_err_output $ model_output))
+  Term.(ret (const mk_output_channel_opt $ std_output $ err_output
+             $ model_output))
 
 let main =
 
--- a/src/bin/common/parse_command.ml
+++ b/src/bin/common/parse_command.ml
@@ -581,9 +581,12 @@ let mk_opts file () () debug_flags ddebug_flags dddebug_flags rule () halt_opt
     `Ok true
   end
 
-let mk_output_channel_opt std_output err_output mdl_output =
+let mk_output_channel_opt std_output deprecated_std_output err_output
+    deprecated_err_output mdl_output =
   Options.Output.(of_filename std_output |> set_std);
+  Options.Output.(of_filename deprecated_std_output |> set_std);
   Options.Output.(of_filename err_output |> set_err);
+  Options.Output.(of_filename deprecated_err_output |> set_err);
   Options.Output.(of_filename mdl_output |> set_mdl);
   `Ok()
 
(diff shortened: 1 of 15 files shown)


Halbaroth:
It is ready :)

src/bin/common/parse_command.ml
```diff
@@ -413,6 +388,8 @@ let mk_limit_opt age_bound fm_cross_limit timelimit_interpretation
 let mk_models_opt b =
   if b then begin
     set_interpretation ILast;
+    (* TODO: The generation of models is supported only with the SAT solver
```
bclement-ocp, line 391: Leaving a pointer for #553

src/bin/common/parse_command.ml
```diff
@@ -1231,32 +1207,40 @@ let parse_fmt_opt =
...
+  let std_output =
+    let doc =
+      Format.sprintf
+        "Set the standard output used by default to print the results,
     models and unsat cores. Possible values are %s."
-        (Arg.doc_alts ["stdout"; "stderr"; "<filename>"]) in
-    Arg.(value & opt formatter_conv Stdout & info ["std-formatter"] ~docs ~doc)
+        (Arg.doc_alts ["stdout"; "stderr"; "<filename>"])
+    in
+    Arg.(value & opt string "stdout" & info ["std-output"] ~docv:"CHANNEL"
```
bclement-ocp: Can we keep the old option name (or optionally both names) so as to avoid potentially breaking existing code that uses the option?

src/bin/common/parse_command.ml
```diff
@@ -1231,32 +1207,40 @@ let parse_fmt_opt =
...
+  let err_output =
+    let doc =
+      Format.sprintf
+        "Set the error output used by default to print error, debug and
          warning informations. Possible values are %s."
-        (Arg.doc_alts ["stdout"; "stderr"; "<filename>"]) in
-    Arg.(value & opt formatter_conv Stderr & info ["err-formatter"] ~docs ~doc)
+        (Arg.doc_alts ["stdout"; "stderr"; "<filename>"])
+    in
+    Arg.(value & opt string "stderr" & info ["err-output"] ~docv:"CHANNEL"
```
bclement-ocp: Same as above, please keep `--err-formatter` working

src/lib/util/options.ml
```diff
@@ -28,38 +28,90 @@
...
+
+  let std_output = ref Stdout
+  let err_output = ref Stderr
+  let wrn_output = ref Stderr
+  let dbg_output = ref Stderr
+  let mdl_output = ref Stdout
+  let usc_output = ref Stdout
+
+  let contents = function
+    | Stdout | Stderr | Channel _ | Invalid -> ""
```
bclement-ocp: This is not a very good idea and can only cause confusion if that function ever ends up called on a non-buffer output. At the very least this should raise an exception, but a better API design would be:

 - Replace `Buffer of Buffer.t * Format.formatter` with `Formatter of Format.formatter`
 - Remove `create_buffer` and replace it with `let of_formatter fmt = Formatter fmt`
 - Remove `contents` entirely
 - Users of `create_buffer` (i.e. the js_worker) are responsible for calling `formatter_of_buffer` and `Buffer.contents` directly (and maybe there should be a `pp_print_flush` somewhere?)

src/lib/util/options.mli
```diff
@@ -1051,60 +1051,93 @@ val get_is_gui : unit -> bool
...
+
+  val create_channel : string -> t
+  (** [create_channel filename] create an out channel on the file [filename].
+      If the file does not exist, the procedure creates it. An existant file
+      is truncated to zero length. *)
+
+  val contents : t -> string
+  (** Retrieve the content of the underlying buffer if any. *)
+
+  val at_exit : unit -> unit
```
bclement-ocp: It would be better to call this something like `close_all` instead of `at_exit`. Generally it is better to name things based on what they do rather when possible.

src/lib/util/options.mli
```diff
@@ -1051,60 +1051,93 @@ val get_is_gui : unit -> bool
...
+    | Stdout
+    | Stderr
+    | Channel of out_channel * Format.formatter
+    | Buffer of Buffer.t * Format.formatter
+    | Invalid
+
+  val create_buffer : unit -> t
+  (** Create a new buffer. *)
+
+  val create_channel : string -> t
```
bclement-ocp, line 1072: `of_path` or `of_filename` would be better names and make it clearer what the argument means.

src/lib/util/options.mli
```diff
@@ -1051,60 +1051,93 @@ val get_is_gui : unit -> bool
...
+
+  val set_mdl : t -> unit
+  (** Set [fmt_mdl] accessible with {!val:get_fmt_mdl} *)
+
+  val set_usc : t -> unit
+  (** Set [fmt_usc] accessible with {!val:get_fmt_usc} *)
+end
+
+(** Print message as comment in the corresponding output format *)
+val print_output_format: Format.formatter -> string -> unit
```
bclement-ocp: I know this function is not new in the PR, but if "Print a message as comment in the corresponding output format" is an accurate description of what is doesn, maybe we could take the opportunity to rename it into `pp_comment`?

bclement-ocp (changes requested):
Looks good, just a couple minor naming concerns.

src/bin/common/parse_command.ml
Halbaroth: We can also deprecate the old names?

bclement-ocp: I am fine with deprecating the old names, as long as they keep working.

src/lib/util/options.ml
```diff
@@ -28,38 +28,83 @@
...
+    | Stderr
+    | Channel of out_channel * Format.formatter
+    | Fmt of Format.formatter
+    | Invalid
+
+  let of_formatter fmt = Fmt fmt
+
+  let of_filename = function
+    | "stdout" -> Stdout
+    | "stderr" -> Stderr
```
bclement-ocp: Ah, sorry, I suggested `of_filename` instead of `create_channel` based on the documentation on the `.mli` but forgot about the special cases for stdout and stderr — `create_channel` with appropriate mention in the doc of the special cases for stdout and stderr is better in that case.

src/lib/util/options.ml
Halbaroth: I only flush the formatter. According to the documentation:

formatter_of_buffer b returns a new formatter writing to buffer b. At the end of pretty-printing, the formatter must be flushed using [Format.pp_print_flush](https://v2.ocaml.org/api/Format.html#VALpp_print_flush) or [Format.pp_print_newline](https://v2.ocaml.org/api/Format.html#VALpp_print_newline), to print all the pending material into the buffer.

it should be fine :)

src/lib/util/options.ml
Halbaroth: Ok :)

src/bin/common/parse_command.ml
```diff
@@ -1230,33 +1210,45 @@ let parse_theory_opt =
...
+  let std_output, deprecated_std_output =
+    let doc =
+      Format.sprintf
+        "Set the standard output used by default to print the results,
     models and unsat cores. Possible values are %s."
-        (Arg.doc_alts ["stdout"; "stderr"; "<filename>"]) in
-    Arg.(value & opt formatter_conv Stdout & info ["std-formatter"] ~docs ~doc)
+        (Arg.doc_alts ["stdout"; "stderr"; "<filename>"])
+    in
+    let deprecated = "this option is depreciated. Please use --std-output." in
```
bclement-ocp, line 1225: Does `cmdliner` displays what option this refers to, or should this be `--std-formatter is deprecated, please use --std-output instead`? The information about deprecation should also be present in `--help`, I am not familiar with `~deprecated` so not sure if it is.

src/bin/common/parse_command.ml
```diff
@@ -1230,33 +1210,45 @@ let parse_theory_opt =
...
+    Arg.(value & opt string "stdout" & info ["std-output"] ~docv ~docs ~doc),
+    Arg.(value & opt string "stdout" & info ~deprecated ["std-formatter"]
```
bclement-ocp: I would prefer something like this:

```ocaml
let merge_formatters default preferred deprecated =
  match preferred, deprecated with
  | Some fmt, _ -> fmt
  | None, Some fmt -> fmt
  | None, None -> default

let std_formatter = 
  ...
  let std_output = Arg.(value & opt (some' formatter_conv) None & info ["std-output"] ~docs ~doc ~docv in
  let std_formatter = Arg.(value & opt (some' formatter_conv) None & info ["std-formatter"] ~deprecated in
  Term.(const (merge_formatters Stdout) $ std_output $ std_formatter)
```

so that the existence of deprecated arguments stays within the CLI API.

src/bin/common/parse_command.ml
```diff
@@ -603,11 +581,13 @@ let mk_opts file () () debug_flags ddebug_flags dddebug_flags rule () halt_opt
...
-  set_std_fmt (value_of_fmt std_fmt);
-  set_err_fmt (value_of_fmt err_fmt);
-  set_fmt_mdl (value_of_fmt mdl_fmt);
+let mk_output_channel_opt std_output deprecated_std_output err_output
+    deprecated_err_output mdl_output =
+  Options.Output.(create_channel std_output |> set_std);
+  Options.Output.(create_channel deprecated_std_output |> set_std);
+  Options.Output.(create_channel err_output |> set_err);
+  Options.Output.(create_channel deprecated_err_output |> set_err);
+  Options.Output.(create_channel mdl_output |> set_mdl);
```
bclement-ocp: I don't think this works? Because all options return formatter names (and not options), if say `--std-output` is provided, we will set the standard channel to the value given to `--std-output` and then immediately replace that with the value from `--std-formatter` (or its default value of stdout), rendering `--std-output` useless.

bclement-ocp (changes requested):

src/lib/util/options.ml
bclement-ocp: The documentation you quoted states that the formatter must be flushed but I didn't see where this actually happens in the code?

src/bin/common/parse_command.ml
Halbaroth: Yes, cmdliner displays the option in the warning message and there is a deprecated tag in the man page also.

src/bin/common/parse_command.ml
Halbaroth: Looks better indeed :)

src/lib/util/options.ml
Halbaroth: It should happen in the finalizer `close_all` but I forgot to add it in the js_worker. I will do it right now. Using `set-option` calls systematically the function `Output.close` on the old output channel.

Halbaroth: Is it ok?

bclement-ocp: Calling `close_all` on `at_exit` in the JS worker is good, but not sufficient to ensure we get the proper content of the buffer. The corresponding formatter must be flushed with `pp_print_flush` before calling `Buffer.contents`.

Halbaroth: ooh I thought `Buffer.contents` calls the flush procedure!

bclement-ocp (approved):
I think we could have simply propagated the formatter from `create_buffer` but it looks good to me now, let's merge this :)

PR #609, opened 2023-05-26
Fix model generation

Halbaroth:
I forgot a case in the comparison funciton after removing all the polymorphic comparison in #599. The model generation was completely broken. A good argument to use ppx_compare as soon as possible.

### DIFF
--- a/src/lib/util/options.ml
+++ b/src/lib/util/options.ml
@@ -316,25 +316,29 @@ let set_unsat_core b = unsat_core := b
 
 let equal_mode a b =
   match a, b with
-  | INone, INone | IFirst, IFirst | IEvery, IEvery | ILast, ILast -> true
-  | INone, (IFirst | IEvery | ILast)
-  | IFirst, (INone | IEvery | ILast)
-  | IEvery, (INone | IFirst | ILast)
-  | ILast, (INone | IFirst | IEvery) -> false
+  | INone, INone -> true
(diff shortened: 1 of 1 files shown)


src/lib/util/options.ml
```diff
@@ -316,8 +316,11 @@ let set_unsat_core b = unsat_core := b
 
 let equal_mode a b =
   match a, b with
-  | INone, INone | IFirst, IFirst | IEvery, IEvery -> true
-  | _ -> false
+  | INone, INone | IFirst, IFirst | IEvery, IEvery | ILast, ILast -> true
+  | INone, (IFirst | IEvery | ILast)
+  | IFirst, (INone | IEvery | ILast)
+  | IEvery, (INone | IFirst | ILast)
+  | ILast, (INone | IFirst | IEvery) -> false
```
bclement-ocp: I find that the least error-prone and most future-proof way to write such functions is:

```ocaml
match a, b with
| INone, INone -> true
| INone, _ | _, INone -> false
| IFirst, IFirst -> true
| IFirst, _ | _, IFirst -> false
| IEvery, IEvery -> true
| IEvery, _ | _, IEvery -> false
| ILast, ILast -> true
```
```

Halbaroth: You're right. Looks better :)

bclement-ocp (approved):

PR #613, opened 2023-06-05
Do not print warning with dolmen frontend

bclement-ocp:
On automatically generated files coming from provers, it is not rare for there to be unused arguments to functions and other "bad practice" that usually warrants a warning in programming languages.

However, when running an automated prover on these files, it does not make much sense to print warnings, because they will probably not be read by the user — and so they just slow us down.

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -235,7 +235,7 @@ let main () =
   let mk_state ?(debug = false) ?(report_style = State.Contextual)
       ?(reports =
         Dolmen_loop.Report.Conf.mk
-          ~default:Dolmen_loop.Report.Warning.Status.Enabled)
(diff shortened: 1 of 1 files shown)


Halbaroth:
We could display them when the verbosity flag is set up?

bclement-ocp:
The verbosity flag is set when we want information about what Alt-Ergo is doing, rather than information about the problem file, so I'd be against it. Most of the time users (including us) can't really change the files they are working with (because they are either benches written by other people, or were generated by automated tools they don't control) --- and so the warnings are pure noise.

I'd be fine adding a different flag (like `-W`) sometime, but I don't think it's very urgent because I don't think it will get used (I won't certainly be using it, at least).

Halbaroth:
Ok :)

Halbaroth (approved):

PR #614, opened 2023-06-05
Refactoring models invocation

Halbaroth:
This PR adds the support for the instruction `get-model`.
The command is only supported through the new frontend Dolmen. It means you cannot use the statement `get-model` without the option `--frontend dolmen`. 

There is no support for the `get-model` statement in the native language of AE.


Halbaroth:
It is ready

src/bin/common/parse_command.ml
```diff
@@ -391,7 +391,10 @@ let mk_models_opt b =
     (* TODO: The generation of models is supported only with the SAT solver
        Tableaux. Remove this line after merging the OptimAE PR.
        See https://github.com/OCamlPro/alt-ergo/pull/553 *)
-    set_sat_solver Tableaux
+    set_sat_solver Tableaux;
+    (* The command line option overrides the behavior of the get-model
+       commands. *)
```
bclement-ocp, line 396: I believe we discussed this multiple times and agreed that command line options should not override the behavior of in-band commands. If you use the option to get a model on each `check_sat` / `check-sat`, you should get that. If you *also* use `get-model`, you should get two models (or more, depending on the number of times `get-model` is called).

Halbaroth:
AE still displays the status after printing the model, even if you use the native language.

src/bin/common/solving_loop.ml
```diff
@@ -96,14 +105,16 @@ let main () =
           state.solver_ctx.ctx
         in
         let cnf = List.rev @@ Cnf.make l td in
-        let () = solve all_context (cnf, name) in
+        let sat_env = solve state.solver_ctx.sat_env all_context (cnf, name) in
```
bclement-ocp: I believe this is incorrect as it threads the SAT state across multiple `goal`s.

`typed_loop` is only used by the legacy typer, it shouldn't be changed by this PR.

src/bin/common/solving_loop.ml
```diff
@@ -175,7 +186,7 @@ let main () =
       exit 142
   in
 
-  let solver_ctx_key: solver_ctx State.key =
+  let solver_ctx_key: SAT.t solver_ctx State.key =
```
bclement-ocp: Instead of parameterizing the `solver_ctx` by a `SAT.t`, we should create two different keys for the `solver_ctx` and the `sat_ctx`.

```ocaml
let sat_state_key : SAT.t State.key = State.create_key ~pipe:"" "sat_state" in
```

src/bin/common/solving_loop.ml
```diff
@@ -438,7 +448,8 @@ let main () =
             }
         in
         let cnf = List.rev rev_cnf in
-        let () = solve all_context (cnf, goal_name) in
+        let solver_ctx = State.get solver_ctx_key st in
+        let sat_env = solve solver_ctx.sat_env all_context (cnf, goal_name) in
```
bclement-ocp: I believe this is incorrect; solves should not reuse the previous SAT state.

src/lib/util/options.ml
```diff
@@ -346,13 +346,15 @@ let get_timelimit_per_goal () = !timelimit_per_goal
 (** Output options *)
 
 let interpretation = ref INone
+let print_all_interpretations = ref false
```
bclement-ocp: Since we are now using the more standard "model" instead of "interpretations", this probably should be `print_all_models` instead.

(As an aside, I believe that both CVC5 and Z3 call this option `dump_models` and maybe we should do the same)

bclement-ocp (changes requested):

bclement-ocp:
> AE still displays the status after printing the model, even if you use the native language.

I assume you mean before printing the model? But that's good, thanks for checking!

Halbaroth:
> > AE still displays the status after printing the model, even if you use the native language.
> 
> I assume you mean before printing the model? But that's good, thanks for checking!

Alas, I didn't. As I explained on Zulip, AE prints its models in a exception handler. After printing it, the exception is raised again and a second handler prints the status.

src/bin/common/solving_loop.ml
Halbaroth: It is almost correct. The field `sat_env` in the state contents the last environment produced by the function solve. It's not used by the next call to the SAT solver. But I should replace the field for the sat environment by an option since the SAT solver can timeout. In this case, the current code keeps the last produced model, which is counterintuitive.

bclement-ocp: ah, right, `solve` takes the `sat_env` just to return it in case of timeout -- indeed using an option is better

I suggest using a separate state key for the sat env in dolmen and simply ignoring the new result of `solve` (`let _ = solve ...`) in the `typed_loop` since AFAIK the legacy frontend has no way to display the models anyways.

Halbaroth: Ok, I will ignore the result in the legacy frontend.

Halbaroth:
I think we can fix the issue with the order between the status and the model in another PR. Modifying the way AE catches exceptions can be tricky :)

bclement-ocp:
I don't think we need to modify the way we catch exceptions, we should get away with simply adding the model as payload to `I_dont_know`.

But I agree, this should be a separate PR.

src/bin/common/solving_loop.ml
```diff
@@ -59,6 +58,13 @@ let main () =
 
   let module FE = Frontend.Make (SAT) in
 
+  let empty_solver_ctx = {
+    ctx = [];
+    local = [];
+    global = [];
+  }
+  in
```
bclement-ocp: No longer needs to be moved, keep it where it was

src/lib/reasoners/fun_sat.ml
```diff
@@ -1924,6 +1925,18 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+    if not @@ Options.get_dump_models () then
+      let env = compute_concrete_model env true in
+      Options.Time.unset_timeout ~is_gui:(Options.get_is_gui());
+      let prop_model = extract_prop_model ~complete_model:true env in
+      Th.output_concrete_model (Options.Output.get_fmt_mdl ()) ~prop_model
+        env.tbox;
+      terminated_normally := true
+    else
+      Printer.print_wrn "get-model commands are ignored in presence of \
+                         --model option."
```
bclement-ocp: We should not ignore `get-model` here, see above.

src/bin/common/solving_loop.ml
```diff
@@ -470,12 +490,27 @@ let main () =
         handle_option loc name value;
         st
 
+      | {contents = `Get_model; _ } ->
+        if Options.get_interpretation () then
+          match State.get sat_env_key st with
+          | Some sat_env -> SAT.get_model sat_env; st
+          | None ->
+            (* TODO: add the location of the statement. *)
+            Printer.print_wrn "No model produced.";
```
bclement-ocp: This is a response to a command; it should be `(error "No model produced.")` according to the smtlib standard. It's not very important, but we should open an issue.

src/bin/common/solving_loop.ml
```diff
@@ -470,12 +490,27 @@ let main () =
...
+            st
+        else
+          begin
+            (* TODO: add the location of the statement. *)
+            Printer.print_wrn "You have to set the option :produce-models \
+                               before using the statement (get-model).";
```
bclement-ocp: Ditto; should be an output on the model formatter `(error "You have to set the option :produce-models")` etc.

(Actually would probably be better to refer to the command-line option for `:produce-models` which I don't remember the name of currently)

src/bin/common/solving_loop.ml
```diff
@@ -38,11 +38,10 @@ type solver_ctx = {
   global : Commands.sat_tdecl list;
 }
 
-let empty_solver_ctx = { ctx = []; local = []; global = []; }
-
 (* Internal state while iterating over input statements *)
-type 'a state = {
+type ('a, 'b) state = {
   env : 'a;
+  sat_env : 'b;
```
bclement-ocp: This is never used and hence not needed.

src/bin/common/solving_loop.ml
```diff
@@ -96,15 +104,17 @@ let main () =
           state.solver_ctx.ctx
         in
         let cnf = List.rev @@ Cnf.make l td in
-        let () = solve all_context (cnf, name) in
+        let sat_env = solve all_context (cnf, name) in
```
bclement-ocp: Should just be `let _ = solve all_context (cnf, name) in` here.

bclement-ocp (changes requested):

src/lib/reasoners/fun_sat.ml
Halbaroth: Actually, the command is not ignored but useless in presence of the flag `--model`. We can print nothing.

bclement-ocp: What do you mean by "print nothing"? We should print the model.

Halbaroth:
> I don't think we need to modify the way we catch exceptions, we should get away with simply adding the model as payload to `I_dont_know`.

Sounds good :) I will do it quickly :D

src/bin/common/parse_command.ml
bclement-ocp: I believe that comment should now be removed?

src/bin/common/solving_loop.ml
```diff
@@ -70,7 +74,7 @@ let main () =
           Options.Time.set_timeout ~is_gui:false (Options.get_timelimit ());
         end;
       SAT.reset_refs ();
-      let _ =
+      let sat_env, _, _ =
```
bclement-ocp, line 77: Thinking about it over lunch: `sat_env` is a really confusing name. `partial_model` would better describe what this actually is.

src/bin/common/solving_loop.ml
```diff
@@ -373,7 +386,8 @@ let main () =
         in
         let rev_cnf = D_cnf.make (State.get State.logic_file st).loc l td in
         let cnf = List.rev rev_cnf in
-        let () = solve all_context (cnf, name) in
+        let solver_ctx = State.get solver_ctx_key st in
```
bclement-ocp, line 389: This is somewhat of a nit and I am fine with leaving this as is but for future reference: it is good software engineering to keep transient variable definitions as close to their use as possible. There are two related reasons:

 - It makes the intention (here: to update the solver context in the state) clearer. With this change, the reader needs to figure out whether the calls to `Solver.set` below may be undoing another `State.set` done by a function call or potential side effects between line 389 and line 405 (it doesn't, but the question still needs to be asked). Previously it was clear that no such things was possible, because the call to `State.set` was immediately after the call to `State.get`.
 - If later we *do* need to perform another `State.set` in the middle, having the definition and use of `solver_ctx` far apart increases the risk that we do introduce a bug by unintentionally having the new change reverted.

src/lib/util/options.mli
```diff
@@ -212,6 +212,9 @@ val set_input_format : input_format -> unit
 *)
 val set_interpretation : interpretation -> unit
 
+(** [all_interpretations] accessible with {!val:get_all_interpretations}. *)
```
bclement-ocp, line 215: Docstring was not updated

bclement-ocp (approved):
LGTM; please fix the comments and maybe change the name of `sat_env` if you are motivated, then we are good to merge.

src/bin/common/parse_command.ml
Halbaroth: Don't worry :) I didn't forget it

bclement-ocp: Apologies for misunderstanding, i thought you were done!

src/bin/common/solving_loop.ml
Halbaroth: I agree :)

src/bin/common/parse_command.ml
bclement-ocp: You didn't forget but you still merged with the comment in there…

src/lib/util/options.mli
bclement-ocp: This was merged without updating the docstring.

PR #615, opened 2023-06-05
Add timelimit support in dolmen frontend

bclement-ocp:
This patch adds timelimit support to the dolmen frontend, which was previously ignoring the `--timelimit` option when `--timelimit-per-goal` was not set.


Halbaroth (approved):
LGTM

PR #617, opened 2023-06-05
Do not sort fields according to the first appearance of their name

bclement-ocp:
Currently if you were to write the following code:

```
predicate sneaky(b: int, c: int) = true
type pair = { a: int ; b: bool ; c : bool }
logic p : pair
axiom p_def : p = { a = 0 ; b = false ; c = true }
```

it would be incorrectly treated by the Dolmen frontend as a "type-safe" (quotes for obvious reasons) version of:

```
predicate sneaky(b: int, c: int) = true
type pair = { a: int ; b: bool ; c : bool }
logic p : pair
axiom p_def : p = { b = 0 ; c = false ; a = true }
```

What?? That is because `Ty.trecord` when called without a `record_constr` argument re-orders the field name according to their weak-table order (i.e., roughly, the order in which they appear in the program source) and d_cnf (sanely) doesn't properly account for that.

This patch makes it so that `Ty.trecord` only performs this reordering when explicitly asked, and the only place where we explicitly ask for this reordering is in the old typechecker. We *probably* could get away with not doing it at all, but I am scared it will break something else at this point, so I err on the side of safety.

There is also an actual error message in `records.ml` instead of an [assert false].

### DIFF
--- a/src/lib/reasoners/records.ml
+++ b/src/lib/reasoners/records.ml
@@ -256,7 +256,9 @@ module Shostak (X : ALIEN) = struct
           in
           let record = is_mine (Record (flds, tyr)) in
           record, (left_abs_xe2, record) :: acc
-        | _ -> assert false
+        | ty ->
+          Util.failwith
+            "Not a record type: `%a" Ty.print_full ty
     in
     let abs_access = normalize (Access (field, embed abs_right_xe, ty)) in
     is_mine abs_access, acc
--- a/src/lib/frontend/d_cnf.ml
+++ b/src/lib/frontend/d_cnf.ml
@@ -255,7 +255,7 @@ and handle_ty_app ?(update = false) ty_c l =
 let mk_ty_decl (ty_c: DE.ty_cst) =
   match DT.definition ty_c with
   | Some (
-      Adt { cases = [| { cstr = { id_ty; _ }; dstrs; _ } |]; _ }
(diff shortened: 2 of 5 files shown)


src/lib/reasoners/records.ml
```diff
@@ -256,7 +256,9 @@ module Shostak (X : ALIEN) = struct
           in
           let record = is_mine (Record (flds, tyr)) in
           record, (left_abs_xe2, record) :: acc
-        | _ -> assert false
+        | ty ->
+          Fmt.failwith
+            "Not a record type: `%a" Ty.print_full ty
```
Halbaroth: We have a similar function in the Util module. We should use it to avoid adding an extra dependency. Strangely, we have access to the Fmt module even if there is no mention of it in any dune files?

Halbaroth: Ping

bclement-ocp:
Updated to use `Util.failwith` instead of `Fmt.failwith`.

Halbaroth (approved):

src/lib/reasoners/records.ml
bclement-ocp: Sorry, I didn't see your messages, did you maybe forget to click "Request changes"?

We have a transitive dependency on fmt  through dolmen (and maybe cmdliner)

PR #601, opened 2023-05-22
Remove GUI

Halbaroth:
This PR removes completely the old GUI.

PR #618, opened 2023-06-05
Remove useless `id` arg from multiple expr.ml funs

hra687261:


bclement-ocp:
I believe @Halbaroth explained to me once that these were used by the GUI, but this PR doesn't look like it touches the GUI. Was it incorrect, or did you leave some entry points for the GUI?

If so, we should remove them as well once #601 is merged.

Halbaroth:
I have a PR that does exactly the same thing. I don't publish it yet because I am waiting the review for the GUI PR :)

hra687261:
> I believe @Halbaroth explained to me once that these were used by the GUI, but this PR doesn't look like it touches the GUI. Was it incorrect, or did you leave some entry points for the GUI?

I'm pretty sure the GUI builds a typed AST which is then translated to exprs and all the entry-points are from `cnf.ml`, `d_cnf.ml` and a couple other "backend" modules. So unless it's done indirectly somehow, I'm not seeing it.

> I have a PR that does exactly the same thing. I don't publish it yet because I am waiting the review for the GUI PR :)

We can close this PR if that is the case.
Do you mind explaining how the GUI uses these ids?

hra687261:
Well since the GUI was removed, I guess this one can be merged then?

bclement-ocp:
should we remove the `id` parameter of `mk_form` as well?

bclement-ocp:
Also can we remove the `Expr.id` function now that it is meaningless?

hra687261:
Will do!

hra687261:
Done!

bclement-ocp (approved):
:+1:

PR #626, opened 2023-06-06
Restore floating-point prelude support

bclement-ocp:
Adding int parameters to `Sy.Float` in #596 involuntarily broke the floating-point prelude. This is because when we build the expressions, the parameters of the floating-point constructor may be variables, which then get instantiated by actual constants in a later phase.

This restores the old behavior, where we store the precision/exponent/mode arguments as expressions, and convert them into constants lazily post-instantiation.

### DIFF
--- a/src/lib/reasoners/arith.ml
+++ b/src/lib/reasoners/arith.ml
@@ -115,7 +115,7 @@ module Shostak
     match sy with
     | Int _ | Real _ -> true
     | Op (Plus | Minus | Mult | Div | Modulo
-         | Float _ | Fixed | Abs_int | Abs_real | Sqrt_real
+         | Float | Fixed | Abs_int | Abs_real | Sqrt_real
          | Sqrt_real_default | Sqrt_real_excess
          | Real_of_int | Int_floor | Int_ceil
          | Max_int | Max_real | Min_int | Min_real
@@ -295,7 +295,15 @@ module Shostak
         P.add p (P.mult_const coef p3), ctx
 
     (*** <begin>: partial handling of some arith/FPA operators **)
-    | Sy.Op Sy.Float (prec, exp), [mode; x] ->
+    | Sy.Op Sy.Float, [prec; exp; mode; x] ->
+      let int_of_term t =
+        try
+          E.int_view t (* ! may be negative or null *)
+        with Failure e ->
+          Printer.print_err "%s" e;
+          assert false
+      in
+      let prec = int_of_term prec and exp = int_of_term exp in
       let aux_func e =
         let res, _, _ = Fpa_rounding.float_of_rational prec exp mode e in
         res
--- a/src/lib/reasoners/arith.ml
+++ b/src/lib/reasoners/arith.ml
@@ -296,14 +296,7 @@ module Shostak
 
     (*** <begin>: partial handling of some arith/FPA operators **)
     | Sy.Op Sy.Float, [prec; exp; mode; x] ->
-      let int_of_term t =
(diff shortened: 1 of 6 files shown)


Gbury:
One question: the fpa prelude is broken currently, but we still have successful CI runs, so  do we not test the fpa prelude in the CI ? If that's the case, then we should definitely add some tests that use the fpa prelude.

bclement-ocp:
That is correct. And yeah, we definitely should have fpa tests — in fact we should at least test that all preludes we ship are parsed correctly. I will make a follow-up PR for that.

src/lib/reasoners/arith.ml
```diff
@@ -295,7 +295,15 @@ module Shostak
...
-    | Sy.Op Sy.Float (prec, exp), [mode; x] ->
+    | Sy.Op Sy.Float, [prec; exp; mode; x] ->
+      let int_of_term t =
+        try
+          E.int_view t (* ! may be negative or null *)
+        with Failure e ->
+          Printer.print_err "%s" e;
+          assert false
+      in
+      let prec = int_of_term prec and exp = int_of_term exp in
```
hra687261: Why not just:
```suggestion
      let prec = E.int_view prec and exp = E.int_view exp in
```

src/lib/structures/fpa_rounding.ml
```diff
@@ -286,20 +269,22 @@ let round_to_integer mode q =
 let make_adequate_app s l ty =
   match s with
   | Sy.Name (hs, Sy.Other) when Options.get_use_fpa() ->
+    let ei i = E.int (string_of_int i) in
+    let float prec exp ?(mode = _NearestTiesToEven__rounding_mode) x =
+      Sy.(Op Float), [ei prec; ei exp; mode; x]
+    in
+    let float32 = float 24 149 in
+    let float64 = float 53 1074 in
```
hra687261: A bit pedantic but why not:
```suggestion
    let float prec exp ?(mode = _NearestTiesToEven__rounding_mode) x =
      Sy.(Op Float), [E.int prec; E.int exp; mode; x]
    in
    let float32 = float "24" "149" in
    let float64 = float "53" "1074" in
```

hra687261 (approved):
LGTM, altough I added a couple of inconsequential suggestions.

hra687261 (approved):
LGTM, altough I added a couple of inconsequential suggestions.

src/lib/reasoners/arith.ml
bclement-ocp: I want to reply that it is not really nice to expose a raw exception failure to the user, but on the other hand we fail with `assert false` in the other case so that wouldn't be a very compelling argument. I am OK with your suggested change.

src/lib/structures/fpa_rounding.ml
bclement-ocp: This is trading a tiny bit of typing for a tiny bit of performance, and I like types. If you find it too convoluted I can use your version (I knew I should have gone all the way and defined `int' n = int (string_of_int n)` in `expr.ml` (: ).

hra687261: I agree that using the right types is more elegant :) no need to change

bclement-ocp:
The JS and Make build fail because we are pinning dolmen to its git version and https://github.com/Gbury/dolmen/pull/157 moved types around, but that is not related to the PR.

Not sure why macOS fails to build Menhir, but that also seems unrelated, I will force the merge.

PR #627, opened 2023-06-07
Add CI tests that we can correctly parse the prelude

bclement-ocp:
As discussed in #626 (note that these are not expected to pass until #626 is merged).

This does not yet add actual floating-point tests because I am not sure where to find such tests publicly. We probably can import why3-generated tests, which I will look into.


bclement-ocp:
Looks like the CI is complaining about the prelude files not existing, which they should. I'm not sure what this is about…

bclement-ocp:
OK, it works better as Cram tests (the tests are still failing but now they are failing because of #626 )

Halbaroth (approved):

PR #628, opened 2023-06-07
Better pre-commit hook

bclement-ocp:
The current pre-commit hook in rsc/extra/pre-commit--git-hook is fairly frustrating to use for multiple reasons:

 - It always checks *all* the files in the repository, making it slow;
 - It prints messages even when everything is fine (if I remember correctly, not been using it for a long time)
 - It always check against the working copy, not the index, meaning that trying to commit twice (without adding the changes in between) always work.

This patch provides an alternative implementation of a pre-commit hook that:

 - Only checks the files that will be modified by the tentative commit
 - Is silent when all checks pass
 - Checks against the index, ensuring that it will never let you commit ill-indented files (well, unless you use `--no-verify` of course).

For indentation errors, if there are any, the script re-indents the corresponding file(s) using `ocp-indent`, and tells you about the changes. You must add (or not) the corresponding changes manually. This is required to be compatible with workflows based on partial commits (using `git add -p`).

For whitespace errors, the script simply checks them, and does not let you commit if there are any, because that is provided for free by `git`. Whitespace errors are easy to avoid with proper editor configuration, and as an alternative a `.editorconfig` file is provided to help with editor configuration.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[Makefile]
+indent_style = tab
--- a/rsc/extra/pre-commit--git-hook
+++ b/rsc/extra/pre-commit--git-hook
@@ -1,14 +1,32 @@
 #!/bin/sh
 
-git_repo=`git rev-parse --show-toplevel`
-cd $git_repo/rsc/extra
-
-res=`./check_style.sh`
-echo $res
-
-if [ "$res" = "All files have correct style" ] ; then
-	./check_indentation.sh
-	exit $?
+if git rev-parse --verify HEAD >/dev/null 2>&1
+then
+	against=HEAD
 else
-  exit 1
+	# Initial commit: diff against an empty tree object
+	against=$(git hash-object -t tree /dev/null)
 fi
+
+# Redirect output to stderr.
+exec 1>&2
+
+# Reindent files using ocp-indent
+errors=0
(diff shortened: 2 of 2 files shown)


Halbaroth:
Are `check_indentation.sh` and `check_style.sh` still necessary? We may remove them :)

bclement-ocp:
No, if we go forward with this, I believe they are not used anymore, and neither is `ocpchecker`. I did not remove them because I figured some people may want to keep the old pre-commit hook working (and some pepole may be running them manually), but I would be happy to. Do you want me to do it in this PR?

Halbaroth:
We have a copy of them in the repository. I believe no one will miss them.

bclement-ocp:
Actually, both `check_indentation.sh` and `check_style.sh` are run on the CI (:

Halbaroth:
Maybe we can replace them by your script? I remembered there were used at some point but I didn't find where. I considered to remove them months ago.

bclement-ocp:
The new pre-commit hook is written specifically as a pre-commit hook, and integrates with git for that purposes. Notably, it only considers presently modified files, it is not written to check the whole repository as must be done on the CI.

We can clean up the indentation/style checking scripts if necessary (but really, we should just switch to ocamlformat), but i think that is out of scope for this PR, which is concerned about the local user experience.

Halbaroth:
I agree. We will clean up the scripts for CI later and at some point we should switch to ocamlformat. I prefer switching to ocamlformat after merging OptimAE.

Halbaroth (approved):

Halbaroth:
I approved your PR but Github asks me to update your branch before merging on next. I won't touch it ;)

PR #594, opened 2023-04-24
Support for the check_sat statement in Dolmen frontend

Halbaroth:
The model PR add a new statement in the native language that is not present in Dolmen. 

Do not merge before the PR https://github.com/Gbury/dolmen/pull/140 is released.

PR #630, opened 2023-06-08
Support dolmen 0.9

### LINKED ISSUE #629: Adapt code to Dolmen interface changes
The CI started failing after https://github.com/Gbury/dolmen/pull/157 because we pin Dolmen to the git version. We need to update our use of Dolmen APIs.

bclement-ocp:
Fixes #629


bclement-ocp:
Ah, great, we have some CI jobs that pin dolmen to the dev version and some that don't, so we are in a sort of deadlock situation.

The "Temporary fix" is no longer needed because the code we needed was shipped in Dolmen 0.8.1, so I will make everyone use the release version of Dolmen. This should only be merged once dolmen 0.9 is released.

Halbaroth:
Great :)

bclement-ocp:
Alright, I fought the GitHub Actions until the build passed, but now the tests don't pass… Because the outputs are empty? I am not sure what is going on, everything works locally (but not with 4.10.1). I am at a bit of a loss but will dig deeper…

bclement-ocp:
Right. It turns out that with old OCaml versions the outputs *ARE* indeed empty.

bclement-ocp:
Actually on all versions; I was using cached results — and of course the culprit is #594 not being merged yet

bclement-ocp:
Alright, looks like this works. There has been some shuffling in the GitHub Actions because I thought I needed to get the pinning working but actually didn't, but I think it's easier to deal with if all the build actions are in the same workflow so I left it that way.

Adding `pin-depends` to the `.opam.template` is the only way that I got it working reliably both locally and on the CI.

@Halbaroth can you take a look? This should work now even before dolmen 0.9 is released, which is good because we have multiple stuff that depends on it going now.

Halbaroth (approved):

bclement-ocp:
This was waiting forever on build with OCaml 4.10.0 to merge because we are now building with 4.10.1 :sweat_smile:

PR #639, opened 2023-06-08
Get rid of format shims

### LINKED ISSUE #636: Get rid of `Format_shims.ml`
The module `Format_shims.ml` contains some code to support the OCaml compiler before 4.08. Since the oldest supported version by Dolmen is 4.08, there is no reason to keep this code and the dune rules used to generate it. The relevant code for newer compilers can be moved in `Printer.ml`.

Halbaroth:
This PR solves #636 

I increased the minimal OCaml version to 4.08.


bclement-ocp (approved):
LGTM

bclement-ocp:
Why is this tagged 2.6.0? It is an innocuous change, and we already effectively will require 4.08 through dolmen for 2.5.0, so we should merge now.

Halbaroth:
Ok :)

PR #644, opened 2023-06-12
Ignore type definitions in d_cnf

hra687261:
Otherwise, if a type definition is encountered then the assertion failure in https://github.com/OCamlPro/alt-ergo/blob/959572e2711230d9ca29023792d5b9c0c2d2e5e0/src/lib/frontend/d_cnf.ml#L1380 is raised


src/lib/frontend/d_cnf.ml
```diff
@@ -1293,7 +1293,7 @@ let make dloc_file acc stmt =
       C.{ st_decl; st_loc } :: acc
 
     (* Function and predicate definitions *)
-    | { contents = `Defs defs; loc; _ } ->
+    | { contents = `Defs (`Term_def _ :: _ as defs); loc; _ } ->
```
bclement-ocp: What if there is a type definition followed by a term definition in the list? I don't think that can happen currently, but we shouldn't silently fail in that case. What about using `List.filter_map` to explicitly ignore `Type_alias` and `Instanceof` on line 1380?

bclement-ocp: Also does this mean there are no type definitions in the test suite? If so we should add one

hra687261: > What about using List.filter_map to explicitly ignore Type_alias and Instanceof on line 1380?

Will do (Although I also think that type definitions and term definitions are bundled together and can't be mixed AFAIK)

> Also does this mean there are no type definitions in the test suite? If so we should add one

I'm pretty sure that there are, it's just that type definitions are handled by Dolmen, we don't need to translate them to anything or do anything with them. We can get the info of a type from dolmen with it's name only.

bclement-ocp: > I'm pretty sure that there are, it's just that type definitions are handled by Dolmen, we don't need to translate them to anything or do anything with them. We can get the info of a type from dolmen with it's name only.

If there is none, we should understand why we are not hitting this `assert false` on the test suite. That being said, from a cursory look of Dolmen's API, I think only type aliases can end up here, and it looks like there is indeed no type aliases in the test suite.

Gbury: Note: indeed, type aliases and type definitions should be ignored in alt-ergo, since dolmen already does everything needed, but `Instanceof` should not be ignored: rather if alt-ergo encounters one, that should raise an error (these statements are only used in models when completing a polymorphic partially-defined builtin (think e.g. adt destructors/projections) by defining it on a instance (e.g. defining the behaviour of `List.tl` on an `int list`)).

hra687261: With the API changes it's a bit confusing, in 0.8.1, what we have is `Type_def` and `Term_def`, no `Type_alias` and no `Instanceof`.
@Gbury can you confirm that `Type_def` and `Term_def` cannot be mixed in the same list? (Should there be two separate `def` types in that case?)
If that is the case I am not convinced that we want to iterate over lists of type definitions that we will ignore anyway.
When it comes to silently failing, I can print a warning mentioning the ignored statements.

bclement-ocp: If Dolmen's API allows them to be mixed in the same list, I'd prefer not to depend on the fact that it actually never happens, especially when it is so cheap to do so. If we want to ignore them, it should be split into two constructors at the Dolmen level.

Also, there should be no warning about ignoring type definitions, it is expected that we ignore them (also it is just as expensive to iterate on the type definition to print the warning as it is to iterate them to ignore them).

Re: `Type_def` vs `Type_alias` — maybe we should wait on 0.9 then…

Gbury: `Type_def` was renamed to `Type_alias` to be more clear: such statements do not actually create a new type, just an alias that can be used for convenience.

Else, it's as @bclement-ocp said: as long as the type do not forbid it, it's not a good idea to try and rely on assumptions about what can happen or not, especially when it's really not that costly to ignore some statements in a list.

hra687261: Alright, did I miss anything?

bclement-ocp (approved):
LGTM!

PR #647, opened 2023-06-13
Make floating-point primitives builtin

bclement-ocp:
This patch makes it so that:

 - The primitive definitions are no longer needed in the prelude (because they are builtins)

 - The `--use-fpa` flag is no longer needed (because the builtins are known as builtins, and no longer depend on hijacking type/function definitions based on their names)

The crux of the changes are:

 - The `Fpa_rounding` module now only defines the fpa rounding modes and related functions, it does not deal with symbols or expressions anymore
 - There are new constant operators for `Fpa_rounding.fpa_rounding` instead of relying on manually constructed constructors
 - There are new function operators for the various builtins (e.g. `Is_a_theory_constant`)

The rest of the changes are concerned with:

 - Adding knowledge of the new operators to both typers (the legacy typer and d_cnf)
 - Adding builtin support to the legacy typer. The changes to the legacy typer to add builtin support are fairly simple and relatively well separated from the floating-point changes, so I did not make a separate PR, but I can do so if needed.
 - Use of the operators instead of function names (for `"is_a_theory_constant"` and co)
 - Removal of the definitions from the preludes

Futher things to note:

 - In this PR the fpa primitives are added in both alt-ergo and smtlib2 files with the Dolmen frontend (but not the legacy frontend). This is because Dolmen does not allow us to depend on the language for builtins, but it will in the next release. so there is some temporary code that will deal with it.
 - There is a new builtin for checking if a real is an integer. This is for (future) compatibility with the smtlib2 Reals_ints theory. The operator is currently completely abstract but it will be made concrete through prelude changes in another PR (I did not include it in this PR because the other PR also changes other tihngs to the prelude, and I prefer to make the prelude changes in this PR minimal)
 - I suggest having the fpa theory prelude file open on the side when reviewing the changes to the typechecker and to d_cnf :)


Halbaroth (approved):

Halbaroth:
Looks great :) I have nothing to say about it.

bclement-ocp:
The build is failing b/c Gbury/dolmen#160 got merged and the types are no longer correct, I wlil fix them and merge (actually there will be another PR for the dolmen changes)

bclement-ocp:
Tests were failing due to this:

```
   $ alt-ergo --use-fpa ../fpa-theory-2017-01-04-16h00.ae
+  alt-ergo: option '--use-fpa': this option is always enabled
```

:)

PR #650, opened 2023-06-14
Incorporate the FPA prelude as a builtin

bclement-ocp:
This patch adds support for "builtin" theory preludes, i.e. preludes
that are embedded in the Alt-Ergo binary (actually, library) rather than
being bundled as separate files.  The FPA prelude is moved from an
"old-style" prelude to such a bundled prelude -- in fact, to three
separate preludes (RIA, NRA, and FPA) to better reflect the fact that
some of the axioms within are not specific to floating point arithmetic
at all.

This new style of preludes reflects the idea that the fact that
floating-point support is implemented using a prelude is just that: an
implementation detail, and is otherwise irrelevant. It could as well be
a builtin theory, and there should be no difference to the end user.

The `--prelude` option is preserved, both for backwards compatibility
but also because the concept of user-defined preludes is useful. The new
builtin preludes are controlled with the `--enable-preludes`,
`--disable-preludes`, and `--disable-default-preludes` option.

Note that, keeping in sync with the remark above, we may want to call
these new options `--enable-theories` / `--disable-theories` (although
we can't really have `--disable-default-theories`), but if we do so, we
should consolidate somehow to expose consistent options with the
existing `--disable-adts` and co (I think having a single option to
enable/disable the theories selectively is better than adding
`--disable-fpa`, `--disable-ria` and `--disable-nra`, but I could be
convinced otherwise).

Note: This PR includes #647 and should not be merged as-is, but only after #647 are merged.
It also will not build on the CI, as it depends on Gbury/dolmen#160 being merged first.

The [last commit](https://github.com/OCamlPro/alt-ergo/commit/f90f9e94aaa71799152a87a52fcb90ca195dc9c6) can be reviewed independently, however, the API changes in Dolmen from that PR are fairly minor.

### DIFF
--- a/src/bin/common/parse_command.ml
+++ b/src/bin/common/parse_command.ml
@@ -703,13 +703,13 @@ let parse_execution_opt =
       else
         let p' = Filename.concat Config.preludesdir p in
         if Sys.file_exists p' then begin
-          begin if String.starts_with ~prefix:"b-set-theory" p then
+          begin if Compat.String.starts_with ~prefix:"b-set-theory" p then
               Printer.print_wrn ~header:true
                 "Support for the B set theory is deprecated since version \
                  2.5.0 and may be removed in a future version. If you are \
                  actively using it, please make yourself known to the Alt-Ergo \
                  developers by writing to <alt-ergo@ocamlpro.com>."
-            else if String.starts_with ~prefix:"fpa-theory" p then
+            else if Compat.String.starts_with ~prefix:"fpa-theory" p then
               Printer.print_wrn ~header:true
                 "@[Support for the FPA theory has been integrated as a builtin \
                  theory prelude in version 2.5.0 and is enabled by default. \
--- a/src/bin/common/parse_command.ml
+++ b/src/bin/common/parse_command.ml
@@ -325,7 +325,7 @@ let mk_execution_opt frontend input_format parse_only parsers
   let output_with_formatting = (not no_formatting_in_output) || pretty_output in
   let output_with_forced_flush =
     (not no_forced_flush_in_output) && (not pretty_output) in
-  set_infer_input_format input_format;
+  set_infer_input_format (Option.is_none input_format);
   let input_format = match input_format with
     | None -> Native
     | Some fmt -> fmt
@@ -400,7 +400,7 @@ let mk_output_opt
     interpretation use_underscore unsat_core output_format model_type models
   =
   let `Ok () = mk_models_opt models in
-  set_infer_output_format output_format;
+  set_infer_output_format (Option.is_none output_format);
   let output_format = match output_format with
     | None -> Native
     | Some fmt -> fmt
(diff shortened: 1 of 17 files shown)


src/bin/common/parse_command.ml
```diff
@@ -1216,13 +1256,81 @@ let parse_theory_opt =
...
+      let doc =
+        Format.asprintf "Enable theory preludes, multiple comma-separated values
+        are supported. $(docv) must be %s."
+          (Arg.doc_alts_enum theory_enum)
+      in
+      let docv = "THEORY" in
+      Arg.(
+        value & opt (list theory) [] & info ["enable-preludes"] ~docs ~doc ~docv
+      )
+    and disable_preludes =
```
Halbaroth, line 1281: In my opinion, it would nicer to adda common option to disable theory. We could type --disable-theory nameofprelude, adt, ...

src/bin/common/parse_command.ml
```diff
@@ -693,11 +693,51 @@ let parse_execution_opt =
...
+            else if String.starts_with ~prefix:"fpa-theory" p then
+              Printer.print_wrn ~header:true
+                "@[Support for the FPA theory has been integrated as a builtin @ \
+                 theory prelude in version 2.5.0 and is enabled by default. This @ \
+                 option is no longer needed, and the '%s'@ prelude will be removed@ \
+                 in a later version.@]" p
+          end;
+
+          if Sys.is_directory p' then
+            Fmt.error "'%s' is a directory" p
```
Halbaroth, line 721: We have a printer for error in Printers

Halbaroth (approved):

src/bin/common/parse_command.ml
bclement-ocp: Agreed but I think it is out of scope for this PR (and probably for 2.5.0 ? I opened #657 ; but not sure we should add it to 2.5.0 at that point)

What I could do is rename this option `--disable-theories` and `--enable-theories` etc. but "lie" and only include the prelude theories in there. Then we can move the other options to the list later. Does that sound like an acceptable compromise?

src/bin/common/parse_command.ml
bclement-ocp: I don't think we do; this is an error occuring during argument parsing/setup, it should use cmdliner error mechanism rather than Alt-Ergo's

`Fmt.error` returns a `('a, string) result` which is used by cmdliner

src/bin/common/parse_command.ml
Halbaroth: I agree. Let's merge your solution for the 2.5.0 :)

bclement-ocp:
Fixed the style issues

src/bin/common/parse_command.ml
bclement-ocp: Should I name it `--enable-preludes` or `--enable-theories`? I think if we plan on going ahead with renaming to `--enable-theories` it is better to call it `--enable-theories` immediately, would avoid having to deprecate one more option, even though I feel a bit icky about the name if it only affects prelude theories

bclement-ocp:
Merging for now, we will figure out the name of the option later

PR #651, opened 2023-06-14
Move changelog to CHANGES.md

bclement-ocp:
This patch moves the changelog from its old location in docs/sphinx_docs to the toplevel of the repository, in CHANGES.md.

The Markdown parser used by Sphinx is changed to MyST since recommonmark which we were using is deprecated, and does not permit easily including a file outside of the sphinx_docs directory.


bclement-ocp:
Finally, it builds!!! Should be ready for review.

Halbaroth (approved):

PR #652, opened 2023-06-14
Semantic triggers

Halbaroth:
See https://github.com/Gbury/dolmen/pull/162


bclement-ocp:
These changes are included in #662

PR #653, opened 2023-06-14
Restore --inequalities-plugin functionality

bclement-ocp:
The plugins loaded with `--inequalities-plugin` were never loaded because the `IntervalCalculus` module was calling
`Inequalities.get_current` upon being initialized, which happens before command line options are parsed.

With this patch:

 - The plugin is loaded early, at command line parsing time

 - It is only used lazily on first use by the arithmetic reasoner

 - There is a mechanism in place to error out if someone tries to change the arithmetic plugin after the arithmetic reasoner has started using it

Arguably the 3rd point is not necessary, but I think it is a good safety feature to avoid weird surprises in the future. If someone is unhappy about that (e.g. because they use Alt-Ergo as a library and want to dynamically switch the plugin), we can revisit.

The code that tried (but failed, due to being called too early) to do the lazy initialization in `Inequalities` is not useful anymore, so it is removed to avoid further confusions in the future.


bclement-ocp:
Using lazy is probably the right thing to do here. I do not know what happened in `tools/gentest`… Will revert

bclement-ocp:
What happened is that I have edited the file and so `ocp-indent` was run on it :)

Halbaroth (approved):
LGTM

PR #654, opened 2023-06-15
Cleanup release process in preparation of 2.5.0

bclement-ocp:
This patch contains several changes of relatively limited scope in preparation of the 2.5.0 release. Each commit is documented and can be reviewed independently.

Essentially, the changes are concerned with removing the old `configure.ml` and replace the plugins and prelude loading mechanism with the `dune-site` mechanism bulit into `dune`.

In addition, the AB-Why3 plugin is extracted to a separate package since it has different license.

Note: This includes #630  and #653 for simplicity and should not be merged as is, only after these have been merged.


bclement-ocp:
Added the b-set-theory prelude to the AB-Why3 plugin.

bclement-ocp:
Rebased on current next

Halbaroth (approved):
LGTM

PR #658, opened 2023-06-15
Cleanup CLI options for models

bclement-ocp:
The current `--models` and `--interpretation` options are relatively obtuse and unintuitive. This patch makes the following change:

 - The `--models` option is removed (it was never released), and replaced with the equivalent `--dump-models` option. The option name is taken to be identical to the equivalent option in other SMT solvers (at least CVC5 and Z3). The option sets `--interpretation last` by default, but can be overriden by an explicit `--interpretation` flag.

 - The new `--produce-models` option is equivalent to `--interpretation last` and enables model generation (non-none `--interpretation` values override it). The name is taken to be identical to the corresponding `set-option`, `:produce-models`, and to the corresponding flag in other SMT solvers (at least CVC5).

In addition, the argument parsing is refactored to ensure a better user experience in the presence of those flags:

 - The sat solver is set to "Tableaux" *only* if the `--sat-solver` option is not present. If `--sat-solver` is explicitely given by the user, there is an error instead of overriding the option.

 - If the user tries to do a `(get-model)` using the Psmt2 frontend, we redirect them to the Dolmen frontend. Probably not necessary, but it bugged me.


bclement-ocp:
Note that Alt-Ergo on `next` crashes when given `--interpretation last` because it tries to use model generation with the satML solver.

Halbaroth:
Is it for the release 2.5.0? It should be since we never released the option `--models`.

bclement-ocp:
Yes, I forgot to add the milestone — will rebase also

bclement-ocp:
Rebased on current next

src/bin/common/parse_command.ml
```diff
@@ -895,32 +853,171 @@ let parse_output_opt =
...
+          | "tableaux-like" | "Tableaux-like" ->
+            Ok Util.Tableaux
+          | "tableaux-cdcl" | "Tableaux-CDCL"
+          | "tableaux-CDCL" | "Tableaux-cdcl" ->
+            Ok Util.Tableaux_CDCL
+          | sat_solver ->
+            Error ("Args parsing error: unkown SAT solver " ^ sat_solver)
+
+        in
+        Arg.(conv' (parse, Util.pp_sat_solver))
```
Halbaroth, line 936: You can use the predefined converter `enum`. Uhm but we cannot manage alternative spellings with the predefined converter. In my opinion we should allow the capitalized versions.

Halbaroth (approved):

src/bin/common/parse_command.ml
bclement-ocp: Yeah, I didn't use `enum` because I wanted to preserve the old lax behavior. Did you mean "we should *not* allow the capitalized version"?

One intermediate solution would be to keep the current code but choose a single name for each solver and emit a deprecation warning if other cases are used, then switch to `enum` and remove the deprecated cases in the next release.

Halbaroth: Yes, I meant we shouldn't allow them. Let's deprecated them.

bclement-ocp: OK, I'll merge this and make a separate PR with the deprecation!

PR #662, opened 2023-06-19
[dolmen] Add support for theory extensions

bclement-ocp:
This patch adds support for theory extensions in the Dolmen frontend.

It includes https://github.com/OCamlPro/alt-ergo/pull/660 because it touches similar parts of the code and
there would be conflicts otherwise.


Halbaroth (approved):
LGTM

PR #663, opened 2023-06-19
Add a debugging option to display the commands sent to the solver

bclement-ocp:
Use `-d commands` to display the commnds before sending them to the solver.  This is useful to debug parsing issues and differences between frontends.


hra687261 (approved):
LGTM!

PR #665, opened 2023-06-19
[dolmen] Make a roundtrip through [Q] to normalize reals

bclement-ocp:
When using the Dolmen frontend, we are not properly normalizing the representation of reals, which means that we can have multiple representations of the same underlying real value (e.g. [2] and [0x1.0p1]). This means that the Dolmen frontend misses some early optimization opportunities compared to the legacy frontend.

This was found using the `-d commands` option from #663.


hra687261 (approved):
LGMT!

bclement-ocp:
By the way we should maybe consider using `Q.t` directly rather than a string, since most real constants will become `Q.t` when ingested by the arith module anyways.

PR #667, opened 2023-06-19
[frontend:legacy] Explicitely disable polymorphic enums

### LINKED ISSUE #517: Error when dealing with polymorphic algebraic datatypes
The error message "alt-ergo: algebraic datatype are not supported" does not come from Alt-Ergo but from Why3 (here: https://gitlab.inria.fr/why3/why3/-/blob/master/src/printer/alt_ergo.ml#L366).

But ```omg``` still doesn't pass the type checking in Alt-Ergo:
```
type 'a omg = Omg
logic a: bool omg3
goal g:
    match a with
    | Omg -> true
    end
```
with the error: 
```
[Error]File "try-alt-ergo-file", line 1, characters 1-18:
Typing Error: Wrong number of arguments when applying omg
```

while ```omg2``` passes the type checking even thought there is an unused type parameter here as well:
```
type ('a, 'b) omg2 = Omg2 of {x2 : 'a}
logic a2:(bool, bool) omg2
goal g2: 
    match a2 with
    | Omg2 (_) -> true 
    end
```

So I think it is in fact a bug that should be fixed, but the error message in Why3 is also incorrect and should be corrected as well since Alt-Ergo does support algebraic data types, including in version 2.4.0 which is the version that is used in Why3 if I'm not mistaken.

_Originally posted by @hra687261 in https://github.com/OCamlPro/alt-ergo/issues/486#issuecomment-940712854_

bclement-ocp:
Polymorphic enums are not supported by the legacy typechecker (enums aka sum types are always monomorphic), but they fail with a cryptic error message (see #517).

The dolmen frontend deals with this properly, and we plan on deprecating the legacy typechecker in the future, so it's probably not worth fixing -- but at the very least we should have a more informative error message.

Fixes #517

### DIFF
--- a/src/lib/frontend/typechecker.ml
+++ b/src/lib/frontend/typechecker.ml
@@ -147,7 +147,7 @@ module Types = struct
       let ty = Ty.text ty_vars id in
       ty, { env with to_ty = MString.add id ty env.to_ty }
     | Enum lc ->
-      if Stdlib.(ty_vars <> []) then
+      if not (Lists.is_empty ty_vars) then
         Errors.typing_error (PolymorphicEnum id) loc;
       let ty = Ty.tsum id lc in
       ty, { env with to_ty = MString.add id ty env.to_ty }
--- a/src/lib/frontend/typechecker.ml
+++ b/src/lib/frontend/typechecker.ml
@@ -147,6 +147,8 @@ module Types = struct
       let ty = Ty.text ty_vars id in
       ty, { env with to_ty = MString.add id ty env.to_ty }
     | Enum lc ->
+      if Stdlib.(ty_vars <> []) then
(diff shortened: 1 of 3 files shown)


src/lib/frontend/typechecker.ml
```diff
@@ -147,6 +147,8 @@ module Types = struct
       let ty = Ty.text ty_vars id in
       ty, { env with to_ty = MString.add id ty env.to_ty }
     | Enum lc ->
+      if Stdlib.(ty_vars <> []) then
```
hra687261:
```suggestion
      if not (Lists.is_empty ty_vars) then
```

hra687261 (approved):
LGTM!

Issue #625, opened 2023-06-06
BV primitives


bclement-ocp:
Add support for the SMT-LIB2 Bit-Vector primitives.
Features:
- [x] Support for the all the SMT-LIB2 Bit-Vector primitives
- [x] Fix the bugs that are present in older versions: 
  - #645
  - #649 
  - #664
- [x] Don't fully expand `bv2nat` (maybe `nat2bv` too?) right away
- [x] Remove the Canonizer in "bitv.ml", build simple-terms straight away and use only simple terms internally.
- [ ] Update website documentation #648

bclement-ocp:
Regarding the 3rd point, I am not sure we want to do this, at least for 2.5.0

I believe going through the canonizer allows to avoid work in some complex cases; for instance consider `(complex_expression @ [|0|]) ^ {1, 1}` where `complex_expression` has a lot of concat / extract etc. If we go through the canonizer, we can avoid some of the work required to parse the `complex_expression`. If we remove the canonizer and only use simple terms, we will do a lot of computation and simplifications inside the `complex_expression` that will ultimately be thrown away.

Note that we don't avoid *all* the work required to parse the `complex_expression` (because we still make the term → canonizer translation). We could avoid this by making the canonizer lazy somehow (either with a constructor `I_Term of E.t` that gets "forced" when needed — or with actual `Lazy.t`).

hra687261:
> I believe going through the canonizer allows to avoid work in some complex cases

From what I understand of the code, the `make_rec` function in the Canonizer goes through the entire `complex_expression` building the entire combination of `I_Comp` and `I_Ext` which will be only applied once the `sigma` function is called.
By removing the canonizer and producing simple_terms right away, that intermediary state with `I_Comp`s and `I_Ext`s will be skipped and we will produce simple terms right away.
With `(complex_expression @ [|0|]) ^ {0, 0}`:
Instead of producing `I_Ext (I_Comp (make_rec complex_expression, make_rec [|0|]), 0, 0, 1)` then applying the `sigma` function on it (which will go through the entire thing to apply the concatenations and extractions), we would just produce `make_rec [|0|]` because the extraction will be applied right away to produce simple terms.
The bounds of extractions will be passed down and applied as soon as possible, the only case in which they won't and in which they'll produce simple terms (`Ext _`) is when they come across a variable over which the extraction can't be applied anyway.

bclement-ocp:
[...]

PR #669, opened 2023-06-20
[Bitv] Add support for the SMT-LIB2's BV primitives

hra687261:
See #625


src/lib/frontend/d_cnf.ml
```diff
@@ -786,7 +786,130 @@ let mk_add translate sy ty l =
     | [] -> []
   in
   let args = aux_mk_add l in
-  E.mk_term sy args ty
+  E.mk_term (Sy.Op Sy.Plus) (List.fast_sort E.compare args) ty
```
bclement-ocp: Isn't this re-introducing the issues we had with #505? Why is it OK to sort the arguments to `+` here?

src/lib/frontend/d_cnf.ml
```diff
@@ -786,7 +786,130 @@ let mk_add translate sy ty l =
...
+    let c2 = E.mk_and (E.mk_eq ~iff:false xsign ysign) (mk_bvult x y) false in
+    E.mk_or c1 c2 false
+
+let mk_bvsle n x y =
+  if n = 0 then E.faux else
+    let xsign = mk_bvsign n x in
+    let ysign = mk_bvsign n y in
+    let c1 = E.mk_and (bvsign_is_pos xsign) (bvsign_is_neg ysign) false in
+    let c2 = E.mk_and (E.mk_eq ~iff:false xsign ysign) (mk_bvule x y) false in
+    E.mk_or c1 c2 false
```
bclement-ocp: I understand that these are only used in the Dolmen frontend, but they probably should be in `expr.ml` instead?

Although I am not sure how I feel about eagerly expanding all these; is there a reason these are not constructors with an interpretation? (If the reason is "it is simper this way" it is fine, but we probably should make them constructors later)

src/lib/frontend/d_cnf.ml
```diff
@@ -918,11 +1040,210 @@ let rec mk_expr ?(loc = Loc.dummy) ?(name_base = "")
...
+          | B.Bitv_neg n, [t] ->
+            mk_bvneg n (aux_mk_expr t)
+
+          | B.Bitv_not _, [{
+              term_descr = App
+                  ({ term_descr = Cst {builtin = B.Bitv_not _; _ }; _ }, _, [t])
+            ; _ }] ->
+            (* TODO: do the simplification in a smarter smarter way
+               should (bvnot (bvconcat(t1, bvnot t2)))
+               be translated to (bvconcat(bvnot t1, t2))? *)
```
bclement-ocp: Is the frontend really the right place to do these simplifications rather than the bitvector theory?

src/lib/reasoners/bitv.ml
```diff
@@ -88,9 +81,19 @@ module Shostak(X : ALIEN) = struct
 
   let name = "bitv"
 
+  let compare_tvar v1 v2 =
+    if v1 == v2 then 0
+    else
+      let c = X.str_cmp v1.var v2.var in
```
bclement-ocp: Is `str_cmp` required here? If so, the function should be called `str_cmp_tvar` for clarity.

src/lib/reasoners/bitv.ml
```diff
@@ -729,51 +1154,96 @@ module Shostak(X : ALIEN) = struct
 
   let print = Debug.print_C_ast
 
+  let nat_of_bv term bv m =
```
bclement-ocp, line 1678: Is it clear that the fully-expanded version of `BV2Nat` is what we want rather than only converting fully interpreted bvs?

src/lib/frontend/d_cnf.ml
hra687261: Not intentional, probably a rebase error, I'll fix it.

src/lib/frontend/d_cnf.ml
hra687261: > I understand that these are only used in the Dolmen frontend, but they probably should be in expr.ml instead?

I'm ok with moving them to `expr.ml`

> If the reason is "it is simper this way"

Yes, that's the only reason. I'll add some constructors and move their simplification to bitv.ml

src/lib/frontend/d_cnf.ml
hra687261: Outdated comment, the simplification are indeed done in the theory.

src/lib/reasoners/bitv.ml
hra687261: It is not, I'll fix this.

src/lib/reasoners/bitv.ml
hra687261: Not sure, it's likely that `hash_cmp` is sufficient.

src/lib/reasoners/bitv.ml
hra687261: It will be done in a separate PR

src/lib/structures/symbols.mli
```diff
@@ -46,6 +46,14 @@ type operator =
   (* BV *)
   | Concat
   | Extract of int * int (* lower bound * upper bound *)
+  | BV2Nat | Nat2BV of int
+  | BVExtend of bool * int
+  (* (sign, n) if sign = true then extend sign otherwise extend with zeros *)
+  | BV_repeat of int
+  | BV_rotate of int * bool (* true for right, false for left *)
```
bclement-ocp: Why not use a signed rotation index rather than an extra boolean?

src/lib/structures/symbols.ml
```diff
@@ -319,6 +341,32 @@ let to_string ?(show_vars=true) x = match x with
   | Op Extract (i, j) -> Format.sprintf "^{%d; %d}" i j
   | Op Tite -> "ite"
   | Op Reach -> assert false
+  | Op BV2Nat -> "bv2nat"
+  | Op Nat2BV m -> Format.sprintf "nat2bv[%d]" m
+  | Op BVExtend (true, n) -> Format.sprintf "bv_sign_extend[%d]" n
+  | Op BVExtend (false, n) -> Format.sprintf "bv_zero_extend[%d]" n
+  | Op BV_repeat n -> Format.sprintf "bv_repeat[%d]" n
+  | Op BV_rotate (n, true) -> Format.sprintf "bv_rotate_right[%d]" n
+  | Op BV_rotate (n, false) -> Format.sprintf "bv_rotate_false[%d]" n
```
bclement-ocp:
```suggestion
  | Op BV_rotate (n, false) -> Format.sprintf "bv_rotate_left[%d]" n
```

src/lib/structures/symbols.mli
```diff
@@ -46,6 +46,14 @@ type operator =
   (* BV *)
   | Concat
   | Extract of int * int (* lower bound * upper bound *)
+  | BV2Nat | Nat2BV of int
+  | BVExtend of bool * int
+  (* (sign, n) if sign = true then extend sign otherwise extend with zeros *)
```
bclement-ocp: We should avoid magic booleans if possible, what about either `type extend = Zero_extend | Sign_extend` or polymorphic variants for clarity?

(Alternatively, `BVExtend of { sign : bool ; n : int }` would be less magical)

src/lib/structures/expr.ml
```diff
@@ -1218,6 +1229,148 @@ let mk_builtin ~is_pos n l =
   if is_pos then pos else neg pos
 
 
+(** smart constructors for bit-vector terms *)
+
+let mk_bitv_concat t1 t2 sz =
+  mk_term (Sy.Op Sy.Concat) [t1; t2] (Ty.Tbitv sz)
+
+let mk_bitv_extract i j t sz =
+  mk_term (Sy.Op (Sy.Extract (i, j))) [t] (Ty.Tbitv sz)
```
bclement-ocp: Nit: these probably should be called `mk_bvconcat` and `mk_bvextract` for consistency with the other fns

bclement-ocp: BTW `mk_term (Op (Extract (i, j))) [t] (Tbitv sz)` should work, thanks to type-directed disambiguation.

src/lib/structures/expr.ml
```diff
@@ -1218,6 +1229,148 @@ let mk_builtin ~is_pos n l =
...
+
+let mk_bvnot n t =
+  mk_term (Sy.Op Sy.BVnot) [t] (Ty.Tbitv n)
+
+let mk_bvsign n t =
+  let t1 = mk_bitv_extract (n-1) (n-1) t n in
+  mk_eq ~iff:false t1 (bitv "1" (Ty.Tbitv 1))
```
```suggestion
  mk_eq ~iff:false t1 bvone
```

src/lib/structures/expr.ml
```diff
@@ -1218,6 +1229,148 @@ let mk_builtin ~is_pos n l =
...
+let mk_bv2nat t =
+  mk_term (Sy.Op Sy.BV2Nat) [t] Ty.Tint
+
+let mk_nat2bv n t =
+  mk_term (Sy.Op (Sy.Nat2BV n)) [t] (Ty.Tbitv n)
+
+let mk_bvneg n x =
+  mk_nat2bv n (
+    mk_term (Sy.Op Sy.Minus) [
+      mk_term (Sy.Op Sy.Pow) [ itwo; int (string_of_int n) ] Ty.Tint;
```
```suggestion
      int Z.(pow ~$i n |> to_string) ;
```

src/lib/structures/expr.ml
```diff
@@ -1218,6 +1229,148 @@ let mk_builtin ~is_pos n l =
...
+  let y' = mk_term (Sy.Op Sy.BV2Nat) [y] Ty.Tint in
+  let natres = mk_term (Sy.Op Sy.Mult) [x'; y'] Ty.Tint in
+  mk_term (Sy.Op (Sy.Nat2BV n)) [natres] (Ty.Tbitv n)
+
+let mk_bvudiv n x y =
+  (* Assuming that if bv2nat(y) = 0 then bv2nat(x)/bv2nat(y) is
+     undefined and that it makes nat2bv[n](bv2nat(x)/bv2nat(y)) also
+     undefined. *)
+  mk_nat2bv
+    n (mk_term (Sy.Op Sy.Div) [mk_bv2nat x; mk_bv2nat y] Ty.Tint)
```
bclement-ocp: https://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml explicitly defines `(bvudiv x 0)` to be `(1)` (repeated `n` times where `n` is the size of x)

src/lib/structures/expr.ml
```diff
@@ -1218,6 +1229,148 @@ let mk_builtin ~is_pos n l =
...
+
+let mk_bvurem n x y =
+  (* Assuming that if bv2nat(y) = 0 then bv2nat(x)%bv2nat(y) is
+     undefined and that it makes nat2bv[n](bv2nat(x)%bv2nat(y)) also
+     undefined. *)
```
bclement-ocp: https://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml explicitely defines `(bvurem x 0)` to be `x`

src/lib/frontend/d_cnf.ml
```diff
@@ -924,11 +923,144 @@ let rec mk_expr ?(loc = Loc.dummy) ?(name_base = "")
...
+
+          (* BV operators that produce ITEs or booleans are partially expanded
+             here instead of in the theory. *)
+
+          | B.Bitv_ashr n, [ x; y ] ->
+            E.mk_bvashr n (aux_mk_expr x) (aux_mk_expr y)
+
+          | B.Bitv_ult n, [ x; y ] ->
+            if n = 0 then E.faux else
+              E.mk_bvult (aux_mk_expr x) (aux_mk_expr y)
```
bclement-ocp: The case `n = 0` should be handled by `E.mk_bvult` (same for other functions below)

src/lib/frontend/d_cnf.ml
```diff
@@ -924,11 +923,144 @@ let rec mk_expr ?(loc = Loc.dummy) ?(name_base = "")
...
+
+          | B.Bitv_shl n, [ x; y ] ->
+            E.mk_term
+              (Sy.Op Sy.BVshl) [aux_mk_expr x; aux_mk_expr y] (Ty.Tbitv n)
+
+          | B.Bitv_lshr n, [ x; y ] ->
+            E.mk_bvlshr n (aux_mk_expr x) (aux_mk_expr y)
+
+          (* BV operators that produce ITEs or booleans are partially expanded
+             here instead of in the theory. *)
```
bclement-ocp: Why? I had a hunch it might break something otherwise, is it the case? The rationale should be explained here.

src/lib/frontend/d_cnf.ml
```diff
@@ -924,11 +923,144 @@ let rec mk_expr ?(loc = Loc.dummy) ?(name_base = "")
               | _ -> assert false
             end
 
+          | B.Bitv_repeat { n; k }, [t] ->
+            E.mk_term (Sy.Op (BV_repeat k)) [aux_mk_expr t] (Ty.Tbitv (n*k))
+
+          | B.Bitv_zero_extend { n; k }, [t] ->
+            let t' = aux_mk_expr t in
+            if k = 0 then t' else
+              E.mk_term (Sy.Op (BVExtend (false, k))) [t'] (Ty.Tbitv (n + k))
```
bclement-ocp: why not `mk_bvextend` (which should deal with this case)?

src/lib/frontend/d_cnf.ml
```diff
@@ -783,7 +783,7 @@ let mk_gt translate ty x y =
   else
     E.mk_builtin ~is_pos:true Sy.LT [translate y; translate x]
 
-let mk_add translate sy ty l =
+let mk_add translate ty l =
```
bclement-ocp: Nit: this could probably be renamed `mk_chainable` and used for `Sy.Sub`, `Sy.Mul`, etc.  Or the call to `mk_add` could be removed — I am not sure why `Sy.Plus` is singled out here? Has nothing to do with the rest of the PR, so feel free to ignore, just pointing it out.

src/lib/reasoners/bitv.ml
```diff
@@ -139,7 +139,14 @@ module Shostak(X : ALIEN) = struct
 
   let is_mine_symb sy _ =
     match sy with
-    | Sy.Bitv _ | Sy.Op (Sy.Concat | Sy.Extract _)  -> true
+    | Sy.Bitv _
+    | Sy.Op (
+        Concat | Extract _ | BVExtend _ | BV2Nat | Nat2BV _
```
bclement-ocp, line 178: I am wondering if having `BV2Nat` (which returns values of type `int`) as a theory constant for `bitv` is going to cause Shostak issues. I think it is fine because it should only make it so that we call `X3.make` instead of `X1.make` and then for the rest it behaves as a `Tint` expr, but just want to make sure this is something that was thought about.

src/lib/reasoners/bitv.ml
```diff
@@ -289,40 +315,557 @@ module Shostak(X : ALIEN) = struct
...
+      if n = 1
+      then mk_bitv_xnor 1 t1 t2
+      else
+        mk_bitv_and 1
+          (mk_bitv_xnor 1
+             (E.mk_bitv_extract (n-1) (n-1) t1 1)
+             (E.mk_bitv_extract (n-1) (n-1) t2 1))
+          (mk_bitv_comp (n-1)
+             (E.mk_bitv_extract 0 (n-2) t1 (n-1))
+             (E.mk_bitv_extract 0 (n-2) t2 (n-1)))
```
bclement-ocp: What is `bvcomp`? I don't see it in the smtlib bv theory?

src/lib/reasoners/bitv.ml
```diff
@@ -289,40 +315,557 @@ module Shostak(X : ALIEN) = struct
...
+        )
+
+    let rec mk_rotate_right n i t =
+      if i = 0 then t else
+        mk_rotate_right n (i - 1) (
+          E.mk_bitv_concat
+            (E.mk_bitv_extract 0 0 t 1)
+            (E.mk_bitv_extract 1 (n - 1) t (n - 1))
+            n
+        )
```
bclement-ocp: Where are these from? I don't think they are in the smtlib2 bv theory — are they custom addition? If so can we clearly delineate what is from smtlib2 and what is custom, and the reasoning for needing additional custom primitives?

(Knowing that if they are not standard, most tools won't generate them anyways)

By the way why do we rotate one bit at a time, is there something wrong with computing the split point and swapping the corresponding parts?

src/lib/reasoners/bitv.ml
```diff
@@ -1191,31 +1780,36 @@ module Shostak(X : ALIEN) = struct
...
-      type t = (X.r simple_term) list
-      let compare = compare_mine
-    end)
-*)
   let fully_interpreted _ = true
 
-  let term_extract _ = None, false
+  let term_extract r =
+    match X.extract r with
+    | None -> X.term_extract r
```
bclement-ocp: Isn't this going to cause infinite loops?

src/lib/reasoners/bitv.ml
```diff
@@ -1191,31 +1780,36 @@ module Shostak(X : ALIEN) = struct
...
+                match r with
+                | { bv = Cte b; sz; } ->
+                  let nsz =  n + sz in
+                  nsz,
+                  E.mk_bitv_concat acc (mk_bv b sz) nsz
+                | _ -> raise Exit
+            ) (sz, mk_bv b sz) tl
+          in
+          Some t, false
+        | _ -> None, false
```
bclement-ocp: I am not sure why/where we need this, can you clarify? Is it for interaction with the arith theory?

src/lib/reasoners/bitv.ml
```diff
@@ -1113,35 +1656,82 @@ module Shostak(X : ALIEN) = struct
 
   let print = Debug.print_C_ast
 
+  let nat_of_bv term bv m =
+    let mk_2pow n =
+      if n > 1 then
+        E.mk_term (Sy.Op Sy.Pow) [E.itwo; E.int (Int.to_string n)] Ty.Tint
```
bclement-ocp:
```suggestion
        E.int Z.(pow ~$2 n |> to_string)
```

src/lib/reasoners/bitv.ml
```diff
@@ -1113,35 +1656,82 @@ module Shostak(X : ALIEN) = struct
...
+      else if n = 1 then E.itwo else E.ione
+    in
+    let mk_ite term n pow =
+      let nthbv = E.mk_bitv_extract n n term 1 in
```
bclement-ocp: Is the shadowing of `n` intended here?

src/lib/reasoners/bitv.ml
```diff
@@ -1113,35 +1656,82 @@ module Shostak(X : ALIEN) = struct
...
+              cnt - sz, l @ acc
+            | { bv = Other r; sz } ->
+              begin match X.term_extract r with
+                | Some t, _ ->
+                  let l =
+                    List.init sz (fun i -> mk_ite t (sz - i - 1) (cnt - i))
+                  in
+                  cnt - sz, l @ acc
+                | None, _ -> raise Exit
+              end
```
bclement-ocp, line 1710: Just to clarify — the `raise Exit` here is because we can't make semantic values in an if-then-else but only terms?

src/lib/reasoners/bitv.ml
```diff
@@ -1113,35 +1656,82 @@ module Shostak(X : ALIEN) = struct
...
+      let nthbv = E.mk_bitv_extract n n term 1 in
+      let cond = E.mk_eq ~iff:false E.bvone nthbv in
+      E.mk_ite cond (mk_2pow pow) E.izero
+    in
+    let bvr, ctx = Canonizer.make_aux bv [] in
+    let r = Canonizer.sigma bvr in
+    let _, natl =
+      List.fold_left (
+        fun (cnt, acc) simple_t ->
+          try match simple_t with
```
bclement-ocp, line 1694: Is it really useful to do all these extraction shenanigans here? It looks like it would be simpler to do it in `mk_ite` directly (but maybe that doesn't work because `ite` must take a term as argument, actually).

src/lib/reasoners/bitv.ml
```diff
@@ -289,40 +315,557 @@ module Shostak(X : ALIEN) = struct
...
+      bitv_to_icomp (List.hd res) (List.tl res), ctx
+
+    let bit_of_expr t =
+      match E.term_view t with
+      | { f = Sy.Int s; _ } ->
+        begin match Hstring.view s with
+          | "1" -> Some true
+          | "0" -> Some false
+          | _ ->
+            failwith (
```
bclement-ocp: Use `Util.failwith` instead of `failwith` + `asprintf`

src/lib/reasoners/bitv.ml
```diff
@@ -289,40 +315,557 @@ module Shostak(X : ALIEN) = struct
...
+    let mk_nthbv_eq bv_t n e =
+      E.mk_eq ~iff:false (E.mk_bitv_extract n n bv_t 1) e
+
+    let get_nth_eq_b bv_t cnt t_b =
+      mk_nthbv_eq bv_t cnt (bv_of_bool t_b)
+
+    let bitv_of_nat ?(neg = false) bv_t nat_t size ctx =
+      let bitf = if neg then (fun b -> not b) else (fun b -> b) in
+      let pow n =
+        E.mk_term (Sy.Op Sy.Pow) [E.itwo; E.int (Int.to_string n)] Ty.Tint
```
bclement-ocp:
```suggestion
        E.int Z.(pow ~$2 n |> to_string)
```

src/lib/reasoners/bitv.ml
```diff
@@ -289,40 +315,557 @@ module Shostak(X : ALIEN) = struct
...
+                let t2, nctx =
+                  aux ncnt 0 (mk_nthbv_eq bv_t cnt rv_t :: nctx)
+                in
+                let t = mk_i_comp t2 t1 (t2.sz + 1) in
+                t, nctx
+          end
+      in
+      aux 0 0 ctx
+
+    let bvor_ctes ~is_and b1 sz1 b2 sz2 =
```
bclement-ocp: Having a function called `bvor` that takes an `is_and` flag is slightly confusing. I'd prefer `bvbin_ctes` with an argument of type `type binop = Bin_and | Bin_or` but we already do this elsewhere so I guess it's OK

src/lib/reasoners/bitv.ml
```diff
@@ -289,40 +315,557 @@ module Shostak(X : ALIEN) = struct
...
+      let bres = if is_and then b1 && b2 else b1 || b2 in
+      if sz1 > sz2 then
+        {bv = Cte bres; sz = sz2}, [{ bv = Cte b1; sz = sz1 - sz2 }], []
+      else if sz1 < sz2 then
+        {bv = Cte bres; sz = sz1}, [], [{ bv = Cte b1; sz = sz2 - sz1 }]
+      else {bv = Cte bres; sz = sz1}, [], []
```
bclement-ocp: Why do we need to support cases where the sizes don't match?

src/lib/reasoners/bitv.ml
```diff
@@ -289,40 +315,557 @@ module Shostak(X : ALIEN) = struct
...
-          let r1, ctx = make_rec t1 ctx in
-          { sz = j - i + 1 ; bv = I_Ext (r1,i,j)}, ctx
-
-        | { E.ty = Ty.Tbitv n; _ } ->
+      let ctx = List.rev_append !nctx ctx in
+      bitv_to_icomp (List.hd res) (List.tl res), ctx
+
+    let bit_of_expr t =
+      match E.term_view t with
+      | { f = Sy.Int s; _ } ->
```
bclement-ocp, line 381: I understand the reasoning behind this, I am a bit worried about theory interactions and "breaking the veil" of theory combination in potentially unsound ways. I think it is probably OK, but would like to hear your thought on it, if any.

src/lib/reasoners/bitv.ml
```diff
@@ -289,40 +315,557 @@ module Shostak(X : ALIEN) = struct
...
+    let mk_bvget_and_eq nt cnt t1 cnt1 t2 cnt2 =
+      E.mk_eq ~iff:false (E.mk_bitv_extract cnt cnt nt 1) (
+        E.mk_ite
+          (E.mk_eq ~iff:false (E.mk_bitv_extract cnt1 cnt1 t1 1) E.bvone)
+          (E.mk_bitv_extract cnt2 cnt2 t2 1)
+          E.bvzero
+      )
+
+    (** turn into a dispatcher function  *)
+    let mk_o_eqs =
```
bclement-ocp: Can you add a comment clarifying what this is doing?

bclement-ocp (commented):
Alright, I made a first pass with a few comments/questions. There are still things I don't understand, I will do a second pass tomorrow.

src/lib/structures/symbols.mli
hra687261: Done

src/lib/structures/symbols.mli
hra687261: Done (with `BVExtend of { sign : bool ; n : int }`)

src/lib/structures/expr.ml
hra687261: Done (Switching to ocamlformat will probably remove most of the unnecessary `Ty`s and `Sy`s, I think)

src/lib/reasoners/bitv.ml
hra687261: Well, sizes don't necessarily match: ` bvor ([|00|] @ x[2]) [|1111|]` will be treated in two iterations, the first one for `bvor [|00|]  [|11|]` then `bvor x[2] [|11|]`

src/lib/reasoners/bitv.ml
hra687261: It's true that it is not pretty, but since this interaction is pretty minimal and only done on the level of literal values, I'm not seeing what could be a source of unsoundness in this case in particular. But there's probably a proper way of doing this or avoiding to do it.

src/lib/reasoners/bitv.ml
```diff
@@ -289,40 +315,560 @@ module Shostak(X : ALIEN) = struct
     let bitv_to_icomp =
       List.fold_left (fun ac bt ->{ bv = I_Comp (ac,bt) ; sz = bt.sz + ac.sz })
 
-    let string_to_bitv s =
+    let string_to_bitv ?(neg = false) s ctx =
+      let bitf = if neg then (fun b -> not b) else (fun b -> b) in
       let tmp = ref[] in
-      String.iter(fun car -> tmp := (not @@ Char.equal car '0', 1)::(!tmp)) s;
+      let nctx = ref[] in
```
Halbaroth: Where do you use this list?

src/lib/frontend/d_cnf.ml
hra687261: Done

src/lib/frontend/d_cnf.ml
hra687261: Done

src/lib/reasoners/bitv.ml
hra687261: I don't see the shadowing?

src/lib/reasoners/bitv.ml
hra687261: Yes

src/lib/reasoners/bitv.ml
hra687261: Done

src/lib/reasoners/bitv.ml
hra687261: Done

src/lib/reasoners/bitv.ml
hra687261: Good catch, I don't!

src/lib/frontend/d_cnf.ml
hra687261: True, it has nothing to do the PR, so I'll remove the change.
(But I think there were issues with the order of creating the expressions, so I'm not sure that the same function can be used for all the operations)

src/lib/reasoners/bitv.ml
hra687261: In "QF_BV": https://smtlib.cs.uiowa.edu/logics-all.shtml  Probably an extension of the theory used by other provers.

src/lib/reasoners/bitv.ml
bclement-ocp: I would have expected
```ocaml
let mk_ite term m pow =
  let nthbv = E.mk_bitv_extract n m term 1 in
  ...
```
here.

src/lib/reasoners/bitv.ml
bclement-ocp: Not sure I understand, all the pairs of arguments to `bvor` have the same size here?

Ah wait this is the case that performs the split for `or`, nevermind.

Hm. I have not dug deep enough into the details, but would it be possible to extract the part of the function responsible for extraction, so that we can write something like

```
let h1, tl1 = extract_first_n bv1 (min sz1 sz2) in
let h2, tl2 = extract_first_n bv2 (min sz1 sz2) in
let h, ctx = bvor_uniform h1 h2 in
let t, ctx = aux tl1 tl2 ctx in
h :: t, ctx
```

and isolate the repeated `if-then-else` patterns? This would make it more readable & less error-prone.

src/lib/reasoners/bitv.ml
hra687261: Yes, that's the rational behind it, it shouldn't cause any issue AFAIK

src/lib/reasoners/bitv.ml
hra687261: Dolmen defines a few extensions of the fixed size bit-vector theory which are used by other theories/provers which why I added them as well. ([Z3](https://microsoft.github.io/z3guide/docs/theories/Bitvectors) seems to have them too).

> By the way why do we rotate one bit at a time, is there something wrong with computing the split point and swapping the corresponding parts?

Nope, just simpler/more straightforward, I'll fix it.

src/lib/structures/expr.ml
hra687261: Fixed

src/lib/structures/expr.ml
hra687261: Fixed

src/lib/frontend/d_cnf.ml
hra687261: Shostak seemed to not like receving ITEs as the output of theorie's `make` function.
But I have to check, there might be something I missed or it might be due to the split of the make function into two `make` and `make_aux`/`make_rec`, one of which produces `r` terms and thte other produces a list of `r simple_term alpha_term`. But yeah, needs a bit more investigation.

src/lib/reasoners/bitv.ml
hra687261: True that it could, but it's outdated code that will be removed

src/lib/reasoners/bitv.ml
hra687261: We don't anymore, I added at a time in which `bv2nat` was a symbol of the arithmetic theory.

src/lib/reasoners/bitv.ml
hra687261: Exactly, `ite` must take a term as an argument, so I think they are. (Although there might a way to avoid it or do it in a smarter way, I'll give it a think)

src/lib/reasoners/bitv.ml
hra687261: Done!

src/lib/reasoners/bitv.ml
hra687261: True, it's a lot simpler and more elegant, thanks! it's done

bclement-ocp:
@hra687261 you mentioned crashes, can you add the crashing test cases here?

hra687261:
Done the crashing tests are under `tests/dolmen/bitv/errors` and goals that should be proven but aren't are under `tests/dolmen/bitv/todo`
(Although there might be more since I haven't ran the benches in a while)

src/lib/structures/expr.ml
```diff
@@ -1218,6 +1229,173 @@ let mk_builtin ~is_pos n l =
...
+      mk_bv2nat x
+    ] Ty.Tint
+  )
+
+let mk_bvult n x y =
+  if n = 0 then faux else
+    mk_builtin ~is_pos:true Sy.LT [ mk_bv2nat x; mk_bv2nat y ]
+
+let mk_bvule n x y =
+  if n = 0 then faux else
```
bclement-ocp: I'd assume that `ule` is true for bv of size 0, not false?

hra687261: Intuitively yes, but, after testing, seems like both CVC5 and Z3 reject bit-vectors of size 0 at the parsing/typing phase, dunno if I missed it, but I don't remember reading this in the doc. Maybe dolmen need's to reject them too? (cc @Gbury)

hra687261: Seems like it is the case: https://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml
```
 :sorts_description "
    All sort symbols of the form (_ BitVec m)
    where m is a numeral greater than 0.
 "
```

Gbury: Indeed, bitvectors of size `0` should be refused. I'll open a PR in dolmen to fix that.

hra687261: In that case should the test `n = 0` be simply removed or be replaced by an assertion? Or we might want in some cases, internally to have 0-sized bit-vectors although I can't come up with a case in which we'd want that

bclement-ocp: For now we should reject them when converting to Expr I'd say

Halbaroth:
Is it ready for review?

bclement-ocp:
Sorry, this one should be closed, it has been split off into #733 and #730

PR #670, opened 2023-06-20
Use the dolmen frontend for some tests

hra687261:
This separates the testing commands into two lists, `lcmds` for the one that are meant to be ran with the legacy frontend and `dcmds` for the ones that are meant to be ran with the dolmen frontend.
When the gentests scripts encounters a repo named "dolmen". It ignores the commands in `dcmds` and runs all the commands in `lcmds` using the dolmen front-end.
This will make it possible to write non-regression tests that only work with the dolmen frontend, like the ones in #669.
This is probably not the smartest/cleanest way to do it, but I just wanted a quick fix, I am open to suggestion on how to make it cleaner/safer/prettier.


bclement-ocp:
I think the best way to do it would be to have the frontends in a list and filter it out when we get to nested dirs. I will propose an alternative patch in a bit

hra687261:
Great, feel free to close this one

bclement-ocp:
#671

PR #671, opened 2023-06-20
Run all test with dolmen frontend in dolmen directory

bclement-ocp:
Less intrusive version of #670


hra687261 (commented):
The only issue I see, is that the commands that already have `--frontend dolmen` in their body will have `--frontend dolmen` added to them if they are under a repository called "dolmen".
Not sure if Cmdliner will complain about it.

bclement-ocp:
They don't, `add_arg` explicitly checks for this case.

(However we will run some tests twice in that case, which I think is OK, but we could remove them if needed)

hra687261 (approved):

PR #672, opened 2023-06-20
Make fpa_rounding_mode an ADT again

bclement-ocp:
In #647 the fpa_rounding_mode was converted from an ADT to a builtin type.  This broke the use of [is_theory_constant] on the fpa_rounding_mode type, because it is no longer a theory constant (it previously was a theory constant for the ADT theory, but being a builtin it no longer has an associated theory).

This patch makes the fpa_rounding_mode type an ADT again.


Halbaroth (approved):
LGTM

PR #674, opened 2023-06-20
Clear preludes in options when `--disable-default-preludes` is set

hra687261:
Otherwise the option is pointless since they are loaded anyways, at least with the dolmen frontend.

### DIFF
--- a/src/bin/common/parse_command.ml
+++ b/src/bin/common/parse_command.ml
@@ -1303,7 +1303,7 @@ let parse_theory_opt =
     let preludes enable_preludes disable_preludes disable_builtin_preludes =
       let preludes =
         if disable_builtin_preludes then
-          []
+          (Options.set_preludes []; [])
         else
           Preludes.default
       in


bclement-ocp:
`--disable-default-preludes` should impact the builtin preludes which are available in `Options.get_theory_preludes` and set with `Options.set_theory_preludes`, it has nothing to do with the user-define preludes that are set by `Options.set_preludes`.

The option `--disable-default-preludes` works on my machine with the dolmen frontend:

```shell
[nix-shell:~/Code/alt-ergo]$ echo '(check-sat)' | dune exec -- alt-ergo --frontend dolmen -i smtlib2 -o smtlib2
Done: 99% (879/881, 2 left) (jobs: 1)File "<builtins>/fpa.ae", line 28, character 8-19:
; 28 |         x in [i, j],
;              ^^^^^^^^^^^
; Error The term: `x` has type `real` but was expected to be of type `int`
; 
[nix-shell:~/Code/alt-ergo]$ echo '(check-sat)' | dune exec -- alt-ergo --frontend dolmen -i smtlib2 -o smtlib2 --disable-default-preludes
Done: 99% (879/881, 2 left) (jobs: 1)  
; File "", line 1, characters 1-12: I don't know (0.5029) (0 steps) (goal g_1)
; 
unknown
```

Can you share the CLI invocation that doesn't work?

hra687261:
True, the error was probably due to not regenerating tests scripts

PR #676, opened 2023-06-21
Add dependency to the alt-ergo binary for the tests

### LINKED ISSUE #638: The non-regression tests don't run again after modifying the binary
The commands `make runtest` or `dune runtest` don't always run again the non-regression tests after
some modifications of the binary. I have to clean the directory to get the expected behavior.

I guess the issue comes from my PR #559.

bclement-ocp:
Currently the `gentest` script only adds the input file as a dependency for the test. This means that if the input file has not changed, dune will not re-run the test, which is clearly not what we want.

This is cause of issues because tests that shouldn't pass, look like they pass --- but they actually just aren't run! This affects the CI as well, because we keep a dune cache across CI runs.

This patch fixes the issue by adding a dependency to the alt-ergo binary, forcing the tests to re-run whenever the binary changes (after recompilation).

Fixes #638

### DIFF
--- a/tools/gentest.ml
+++ b/tools/gentest.ml
@@ -149,7 +149,7 @@ end = struct
 @[<v 1>\
 (rule@,\
 (target %a)@,\
-(deps (:input %s))@,\
+(deps (:input %s) %%{bin:alt-ergo})@,\
 (package alt-ergo)@,\
 @[<v 1>(action@,\
 @[<v 1>(no-infer@,\


bclement-ocp:
Note that the test failures are expected; they are actual test failures on `next` that are not caught due to the bug that this PR fixes.

The test failures are due to Dolmen currently not working with the builtin preludes, see #677

hra687261 (approved):
LGTM!

bclement-ocp:
Actually, dune's rules should be able to infer the dependency automatically here if we use `%{bin:alt-ergo}` instead of `alt-ergo` as a command.

bclement-ocp:
Superseded by #684

PR #677, opened 2023-06-21
Temporarily disable builtin preludes

bclement-ocp:
They must be explicitly enabled with `--enable-preludes` for now. This is because the fpa preludes are broken with Dolmen currently, and so we the tests are failing when they are enabled by default.


hra687261 (approved):
LGTM!

PR #679, opened 2023-06-21
Represent bitvector variables using terms

### LINKED ISSUE #350: [Bug] Assertion failure in lib/reasoners/shostak.ml when reasoning on bitv

### LINKED ISSUE #664: Unsoundness with bit-vectors.
Given the following file: 
```
logic T: bitv[2]

goal g:
    ( T = [|00|] )
  or
    ( [|00|] <> ( [|0|] @ T^{0,0}) )
```
or
```
(set-logic QF_BV)
(declare-fun T () (_ BitVec 2))
(assert
  (and
    (distinct T (_ bv0 2))
    (= (_ bv0 2) (concat (_ bv0 1) ((_ extract 0 0) T)))))
(check-sat)
```
AE responds with `Valid` (`unsat`) while it's obvious that the goal is sat for `T = [|01|]` for example.

bclement-ocp:
The reasoning for this patch is explained
[here](https://github.com/OCamlPro/alt-ergo/pull/472#issuecomment-1498710402):

> The issue comes from the way Alt-Ergo manages `uninterpreted semantic values` in the Shostak module. By `uninterpreted semantic value`, I mean a semantic value whose the top symbol is uninterpreted. For instance:
>
>     * `x`
>
>     * `f(34)`
>
>     * `f(g(5)+5)`
>
>
> But `x+y` doesn't fall into this category since its top symbol is interpreted in the `Arith` theory (even if, it is not interpreted in the usual sense since `x` and `y` are variables).
>
> Now, you have two important functions defined in the Shostak module: `solve` and `leaves`. The `leaves` function returns the direct uninterpreted subvalues of a semantic value. For instance:
>
>     * `leaves(x) = []`
>
>     * `leaves(f(g(5)+5)) = [g(5)]`
>
>
> The function `solve` tries to solve a system of equations between semantic values and in case of success, it returns a sequence of substitutions. More precisely, these substitutions replace some leaves in the members of the equalities by semantic values. So the substitutions are of the form : ` r1 --> R1, r2 --> R2, ...` where `r1, r2, ...` are _uninterpreted_ semantic value.
>
> Unfortunately, the Bitvector theory produces wrong substitutions. The module produces substitutions whose the left member is interpreted. By miracle, these terms are not necessary to solve the equations, so if you break the invariant above, AE successes.
>
> Actually, it makes sense that Bitv produces such substitutions. The module has it own variable type. For the global theory, a variable produces in Bitv looks like an interpreted semantic value of Bitv and the Shostak module has no way to know that it´s actually a variable since the type of semantic value of Bitv is abstract.
>
> I had notice this strange design before. I am thinking now how to change it properly.

This patch associates with each bitvector variable a term, which is a skolem constant representing the variable (note that we must use a `Symbols.Name`, not a `Symbols.Var` here, even though we are representing a "variable" for the bitvector theory, because `Symbols.Var` is used *only* for the bound variables of quantifiers). This skolem constant is the one that is used to communicate with the Shostak theory: we advertise it as a leaf (previously, we were returning bitvector-theory-level terms as leaves, which is incorrect, because leaves *MUST* only be terms with an uninterpreted top symbol), and we accept it as input of a substitution.

The uid of the term is used for comparisons instead of the old variable counter. Although the uid is not identical (there are many terms that are not bitvector variables), the uids for fresh variables are guaranteed to be strictly increasing (a fresh variable can't, by definition, be already present in the weak table), and so the existing logic about variable ordering should not change.

Note that there may be some cases where we treat a bitvector variable as an alien, and that could cause a slight change in behavior; however, I don't believe this can actually happen because the term representatives can only appear as leaves to a bitvector expression, and so they should not be manipulated by the Shostak combiner without going through us.

Fixes #350
Fixes #664


bclement-ocp:
After doing a deeper dive into the guts of the bitvector solver in #683 I now think that this patch is overkill and bad for performance. Most use of the variables from the bitvector solver are internal, and it looks like an oversight that they are actually exposed outside of the bitvector theory.

Given my current understanding of the bitvector solver, only the variables that are actually "exported" by `build_solution` need to be associated with actual term variables. This is because `sys_solve` never places pre-existing variable on the right-hand side of equations, but always create fresh variables (even for equalities involving two variables, see `other_vs_other`).

In fact, I now believe that the proper fix is to eliminate `xterm` from `simple_term_aux` and only use the bitvector variables inside the solver, since they are only ever relevant within a *single* call to `solve` (at least that I can tell).

PR #680, opened 2023-06-21
Proper C-bitvector substitutions

### LINKED ISSUE #649: Assertion failure in "src/lib/reasoners/bitv.ml", line 569
On the following test:
```
logic x: bitv[2]

goal g:
  forall y: bitv[4].
    (y^{3,3} @ y^{3,3}) @ y = ([|00|] @ x) @ [|00|]
```
Or
```
(set-logic BV)
(declare-const x (_ BitVec 2))
(assert (
    exists
      ((y (_ BitVec 4)))
      (distinct
        (concat (concat ((_ extract 3 3) y) ((_ extract 3 3) y)) y)
        (concat (concat (_ bv0 2) x) (_ bv0 2))
)))
(check-sat)
```
The smt2 version only works with the dolmen frontend because apparently, the native SMT2 frontend doesn't recognize `BitVec`.
Could be linked to #645 but it's not clear as it seems that this issue stems from the quantified variable.

bclement-ocp:
When solving bitvector equalities, if there is a C-substitution that is found, applying that substitution will always result in an assertion failure, because we are iterating over a list recursively and have an unconditional `assert false` in the empty case.

I believe that returning an empty list here is the correct thing, and makes the function an actual substitution.

Note that in many cases we don't hit the assertion failure because if there is no substitution at all, the `apply_subs` function is not called in `equations_slice` (if I had to wager a guess, I would say that the `assert false` in `apply_subs` came from trying to ensure that the `subs` was not empty -- except it was not applied to the right variable).

Fixes #649

### DIFF
--- a/src/lib/reasoners/bitv.ml
+++ b/src/lib/reasoners/bitv.ml
@@ -566,9 +566,11 @@ module Shostak(X : ALIEN) = struct
 
     let apply_subs subs sys =
       let rec f_aux = function
-        |[] -> assert false
-        |v::r -> try let (v1,v2) = List.assoc v subs in v1::v2::(f_aux r)
-          with _ -> v::(f_aux r)
+        |[] -> []
+        |v::r ->
+          match List.assoc v subs with
+          | (v1, v2) -> v1 :: v2 :: f_aux r
+          | exception Not_found -> v :: f_aux r
       in List.map (fun (t,vls) ->(t,List.map f_aux vls))sys
 
     let equations_slice parts =
--- a/tests/issues/649/649.ae
+++ b/tests/issues/649/649.ae
@@ -1,0 +1,5 @@
+logic x: bitv[2]
+
+goal g:
+  forall y: bitv[4].
+    (y^{3,3} @ y^{3,3}) @ y = ([|00|] @ x) @ [|00|]
--- a/tests/issues/649/649.expected
+++ b/tests/issues/649/649.expected
@@ -1,0 +1,2 @@
+
+unknown
--- a/tests/issues/649/dolmen/649.expected
+++ b/tests/issues/649/dolmen/649.expected
@@ -1,0 +1,2 @@
+
+unknown
--- a/tests/issues/649/dolmen/649.smt2
+++ b/tests/issues/649/dolmen/649.smt2
@@ -1,0 +1,10 @@
+(set-logic BV)
+(declare-const x (_ BitVec 2))
+(assert (
+    exists
+      ((y (_ BitVec 4)))
+      (distinct
+        (concat (concat ((_ extract 3 3) y) ((_ extract 3 3) y)) y)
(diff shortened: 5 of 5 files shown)


bclement-ocp:
The tests failures are not related to this PR, see #677

hra687261 (approved):
LGTM!

bclement-ocp:
Ah, I forgot about this one.

After writing #683 I think that the fix here is actually not correct.

More specifically, I believe we must do `f_aux (v1 :: v2 :: r)` rather than `v1 :: v2 :: f_aux r`. This is because when we have a large C variable it can get split multiple times. Consider for instance a C variable of size 3, and the pattern `[1; 1; 1]`. Then the C variable will be split with size `(1, 2)` and then the variable of size 2 will be split again into two variables of size 1, and so we must apply the substitution recursively to the result of the substitution.

(Actually now that I write it out — `v1 :: f_aux (v2 :: r)` is probably sufficient but maybe that's pushing it)

bclement-ocp:
Sorry, rather, I think the fix is *correct* but not enough. Let's merge this and discuss changing the `f_aux` call somewhere else (what I described above is one of the issues preventing the equivalence that I mentioned in #683 )

hra687261:
I guess that the substitution could map vars to lists of vars instead of pairs of vars?
True, you did mention the recursivity of the C-substitutions in the doc. 
Are we sure that in `v1 :: f_aux (v2 :: r)`, `v1` was not substituted as well?

bclement-ocp:
Mapping vars to lists of vars doesn't really help because when we create the substitution, we always create a substitution of one var to two vars (it happens in `slice_composition`), so we would have to do some work there (i.e. eagerly apply the substitution to itself).

My understanding is that the substitutions are created in `slice_composition` when we have a pattern of size `n` and a var of size `sz > n` that gets split into `st_n` and `res` and recorded as `(st_n, res)`. Then, `st_n` is added as is to the output composition and won't be split again, but `res` is added back the the composition we are slicing, and so can get split again. I don't think there are other places we create substitutions?

(Keeping in mind the invariant that the substitutions are created for a single multi-equation and C variables can only appear once in each multi-equations to begin with)

hra687261:
What I mean was that, when we create the substitution, we could create a substitution of a var to a list.

AFAIK, yes, the substitutions are created in `slice_composition` when the size of the variable is bigger than the pattern. What I meant above is, instead of just looking at a variable `v` and a pattern `p`, we can have something that looks into the next patterns and check if the newly produced variable (the one on the right) doesn't need splitting as well? And if it is the case, the created substitution will map the original variables to all the produced splits.

bclement-ocp:
Hmm we could have a more powerful version of `slice_var` that does all the splits in one go I suppose, yeah, that sounds like a good idea actually. I have some more pending code to push for review on `slice_composition`, I will make that change separately.

PR #681, opened 2023-06-21
[dolmen] Add support for semantic triggers

bclement-ocp:
This patch adds support for semantic triggers in the dolmen frontend after Gbury/dolmen#162

This is an alternate fix for #677 as it allows the builtin preludes to be parsed with Dolmen.


bclement-ocp:
Note that the tests require #662 to actually pass.

bclement-ocp:
@Halbaroth is there something blocking this?

Halbaroth (approved):

PR #683, opened 2023-06-22
Document the bitvector solver

bclement-ocp:
This patch adds documentation to the bitvector solver that explain my understanding of its inner working.

I believe it is correct, and have outlined the parts where I am not sure what is going on, it can be completed at a later time.

Note that this deep dive into the solver revealed a couple potential bugs (mentioned in passing in the documentation). This patch explicitely is only about *documenting* the current state of the solver, so the bugs will be fixed in separate patches.

### DIFF
--- a/src/lib/reasoners/bitv.ml
+++ b/src/lib/reasoners/bitv.ml
@@ -32,6 +32,32 @@ module Sy = Symbols
 module E = Expr
 
 type sort_var = A | B | C
+(* The variables used by the bitvector solver can be split into three
+   categories that have associated invariants.
+
+   These invariants are recalled in the functions that make use of them, but the
+   outline is as follows. The solver works with *partitioned multi-equations*.
+   A partitioned multi-equation is an equality: X == e1 = e2 = ... = en
+   where X is a term (or an alien) and e1, e2, ..., en are terms in the
+   bitvector algebra, that may contain variables.
+
+   The solver maintains a system of partitioned multi-equations. In the system,
+   there is at most one multi-equation for each variable X (while building the
+   system, the multi-equations are initially split into binary equations [X = e]
+   where this restriction does not apply technically; in this representation,
+   all the right-hand side of equalities involving the same left-hand side
+   should be treated as part of the same multi-equation).
+
+   Then, we maintain the invariants that, within the system:
+
+   - Each A variable appears at most once (this is used to represent portions
+      of a term that are irrelevant to a particular constraint)
+
+   - Each B variable appears in at most one *equation* (i.e. a single term
+      [ei]), but appears *multiple times* in that equation (never only once)
+
+   - Each C variable appears at most once in each multi-equation, but appears
+      in several (more precisely, two) distinct multi-equations. *)
 
 let compare_sort s1 s2 =
   match s1, s2 with
@@ -470,13 +496,8 @@ module Shostak(X : ALIEN) = struct
 
         [st1 = c] and [st2 = c]
 
-       (Note: I am not sure if we could orient [st1 = st2] or [st2 = st1] when
-       one of them is already a variable, or maybe just when it is a
-       C-variable. I think we don't, because we must generate fresh
-       substitutions for all the original variables -- I think, see
-       [other_vs_other] below)
-
        Requires: [size st1 = size st2]
+       Requires: [st1] and [st2] are distinct (for the C variables invariant)
     *)
     let other_vs_other st1 st2 =
       let c = fresh_bitv C st1.sz in [ (st1,c) ; (st2,c) ]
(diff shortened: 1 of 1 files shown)


src/lib/reasoners/bitv.ml
```diff
@@ -452,6 +526,38 @@ module Shostak(X : ALIEN) = struct
...
+          The [b] vector is then split into two parts to properly align the
+          repetition. I won't go into the details of the modulo arithmetic, but
+          here is an example ([u] and [vv] are two [1] and [2] size B-variables)
+
+             _ nn_overl = 3
+            / \
+            xxxxxxx???     uvvuvvuvvu???
+            ???yyyyyyy  => ???uvvuvvuvvu
+            \________/
+              b_box = 10
```
hra687261: My understanding is the following:
```
  sz_b1 = b_box % nn_overl (size of b1)
  sz_b2 = nn_overl - sz_b1 (size of b2)
  a1[i] @ ((b1 @ b2) * (b_box/nn_overl)) @ b1 @ a2[siz - tai - i2]
```
So in the example you gave, I would expect the solution to be something like:
```
  xxxxxxx???
  ???yyyyyyy
  uvvuvvuvvu
```
I think there was extra iteration of the while loop in the result of the example you gave.

src/lib/reasoners/bitv.ml
```diff
@@ -478,6 +584,30 @@ module Shostak(X : ALIEN) = struct
         ({ bv = Other xt ; sz = siz } , a3 @ (!acc) @ a1)
       end
 
+    (* [sys_solve sys] orients a system of equations between simple terms.
+
+       The resulting system contains equations between a simple term on the
+       left, and a solver_simple_term on the right.  The solver_simple_term only
+       involved *fresh* A, B, and C variables.
```
```suggestion
       involves *fresh* A, B, and C variables.
```

src/lib/reasoners/bitv.ml
```diff
@@ -478,6 +584,30 @@ module Shostak(X : ALIEN) = struct
...
+
+       In a single equation of the resulting system:
+
+       - The A variables only refer to parts of the variable that were
+          unconstrained in the original equation. Each A variable appears
+          *exactly once* in the oriented system.
+       - The B variables appear due to equalities involving multiple
+          extractions of the same uninterpreted term. They appear in a single
+          equation, but can appear *multiple times in a single equation*.
+       - The C variables appear to due equalities involving multiple
```
```suggestion
       - The C variables appear due to equalities involving multiple
```

src/lib/reasoners/bitv.ml
hra687261: or to be clearer: 
```
  $$$xxxxxxx???###
  $$$???yyyyyyy###
  \\\uvvuvvuvvu///
where a1 = \\\, b1 = u, b2 = vv, a2 = ///
```

hra687261 (approved):
Thanks for taking the time to document the solver, it is very helpful!
Do you think it would make sens to move the definitions of the kinds of the variables (A, B and C) to the type declaration? It would avoid duplicating it, but it make reading the doc less easy.

hra687261:
> Note that this deep dive into the solver revealed a couple potential bugs (mentioned in passing in the documentation). This patch explicitely is only about documenting the current state of the solver, so the bugs will be fixed in separate patches.

I assume that you are refering to the comment on the lines 770-771.
Do you have any specific bugs in mind? Can you create issues for them?

bclement-ocp:
> > Note that this deep dive into the solver revealed a couple potential bugs (mentioned in passing in the documentation). This patch explicitely is only about documenting the current state of the solver, so the bugs will be fixed in separate patches.
> 
> I assume that you are refering to the comment on the lines 770-771. Do you have any specific bugs in mind? Can you create issues for them?

I have fixes coming up in PRs soon, actually :)

src/lib/reasoners/bitv.ml
bclement-ocp: Yeah, I messed up the number of uvv — I think your way of writing it is clearer, and I will also include your formula. Thanks!

hra687261:
That's even better! :grin:

bclement-ocp:
> Thanks for taking the time to document the solver, it is very helpful! Do you think it would make sens to move the definitions of the kinds of the variables (A, B and C) to the type declaration? It would avoid duplicating it, but it make reading the doc less easy.

I thought about it, but then decided against it — I think that repetition is useful here, because while the invariants are the same everywhere, the types involved change (sometimes we have a full system / a partitioned system / a single multi-equation) and I think it improves readability having each function recall the invariants in its own context. Optimizing for readability is best here in my opinion. It probably makes sense to add the invariants *also* to the type level however, I will do that after my coffee.

bclement-ocp:
I have forced the merge; given that this only changes comments running the CI is not very useful.

PR #685, opened 2023-06-23
Associate syntactic terms to bitvector variables

### LINKED ISSUE #350: [Bug] Assertion failure in lib/reasoners/shostak.ml when reasoning on bitv

### LINKED ISSUE #664: Unsoundness with bit-vectors.

bclement-ocp:
This is a new version of #679, which itself was a new version of #472. A similar fix is also in #553 as bb554381a142600b15bde96adcc3a9c0813edd34

The crux of it is that the bitvector theory returns *interpreted* values as leaves, where the rest of the solver requires leaves to be *uninterpreted terms*.

The main difference of this patch with the other patches above is that here we only create the fresh terms associated with variables *after* solving, so we only create fresh terms that we do expose. On the other hand, the solver may create many variables internally that gets split or otherwise merged and whose associated terms would never get used.

Fixes #350
Fixes #664


src/lib/reasoners/bitv.ml
```diff
@@ -124,38 +123,26 @@ module Shostak(X : ALIEN) = struct
...
-    | Alien _, Var _ -> -1
-
   let compare_simple_term = compare_alpha_term (fun st1 st2 ->
       match st1, st2 with
       | Cte b1, Cte b2 -> Bool.compare b1 b2
       | Cte false , _ | _ , Cte true -> -1
       | _ , Cte false | Cte true,_ -> 1
 
-      | Other t1 , Other t2 -> compare_xterm t1 t2
+      | Other t1 , Other t2 -> X.str_cmp t1 t2
```
hra687261, line 140: I wonder if `X.hash_cmp` would be enough in these cases? Instead of the structural comparison, AFAIK order here should cause too many issues (Like it does in expressions)

src/lib/reasoners/bitv.ml
```diff
@@ -315,33 +302,17 @@ module Shostak(X : ALIEN) = struct
...
-
-    (* unused
-       open Canonizer
-       let rec print_I_ast fmt ast = match ast.bv with
-       | I_Cte b -> fprintf fmt "%d[%d]@?" (if b then 1 else 0) ast.sz
-       | I_Other (Alien t) -> fprintf fmt "%a[%d]@?" X.print t ast.sz
-       | I_Other (Var tv) -> fprintf fmt "%a@?" print_tvar (tv,ast.sz)
-       | I_Ext (u,i,j) -> fprintf fmt "%a<%d,%d>@?" print_I_ast u i j
-       | I_Comp(u,v) -> fprintf fmt "@[(%a * %a)@]" print_I_ast u print_I_ast v
-    *)
```
hra687261, line 331: Why not keep a printer for terms? (Ok, it is commented and not used) but I think it would be good to have it for debugging purposes and to avoid having to write it if it is needed at some point.

src/lib/reasoners/bitv.ml
```diff
@@ -1090,12 +1048,9 @@ module Shostak(X : ALIEN) = struct
...
+    match biv.bv , x with
     | Canonizer.I_Cte _ , _ -> biv
-    | Canonizer.I_Other (Var y), Var z when compare_tvar y z = 0 ->
-      extract subs biv.sz
-    | Canonizer.I_Other (Var _) , _ -> biv
-    | Canonizer.I_Other (Alien tt) , _ ->
+    | Canonizer.I_Other tt , _ ->
       if X.equal x tt then
         extract subs biv.sz
       else extract (X.subst x subs tt) biv.sz
```
hra687261: Since `x` is no longer matched, it can be removed from the pattern matching.

tests/issues/350.ae
```diff
@@ -0,0 +1,31 @@
...
+
+(* Not *)
+function b2not(x: bitv[2]): bitv[2] =
+    bnot(x^{1,1}) @ bnot(x^{0,0})
+function b4not(x: bitv[4]): bitv[4] =
+    b2not(x^{2,3}) @ b2not(x^{0,1})
+function b8not(x: bitv[8]): bitv[8] =
+    b4not(x^{4,7}) @ b4not(x^{0,3})
+
+goal g: b8not([|10001111|]) = [|01110000|]
```
hra687261:
```suggestion
goal g: b8not([|10001111|]) = [|01110000|]
goal gn:   bnot([|1|]) = [|0|]
goal gn2:  not(b2not([|10|])=[|00|])
goal gn4:  b4not([|1001|])=[|0110|]
(* Start bugging at this line *)
goal gn8:  b8not([|10001111|])=[|01110000|]
goal gn16: b16not([|1000100011111111|])=[|0111011100000000|]
```
Could be good to have the full test-file (needs promotion of the `.expected` file)

hra687261 (approved):
LGTM

tests/issues/350.ae
bclement-ocp: Yes, I am not sure why it got truncated…

src/lib/reasoners/bitv.ml
bclement-ocp: True! Will fix

src/lib/reasoners/bitv.ml
bclement-ocp: OK, but then let's keep it uncommented so that it fails if we change things.

src/lib/reasoners/bitv.ml
bclement-ocp: No, `str_cmp` should be required here because this is used in `compare` which, in turn, is called by `str_cmp`. I will add a comment to clarify this.

PR #686, opened 2023-06-23
[bitv] Expand the pattern after B-variable splitting

### LINKED ISSUE #645: Assertion failure in "src/lib/reasoners/bitv.ml", line 538

bclement-ocp:
This patch rewrites the `slice_composition` function to take care of two edge cases:

 - First, it may happen that spliting a B variable causes another instance of the same B variable to become smaller than its expected size in the pattern. When that happens, we record the information, recompute the pattern, then re-slice the whole system.

 - Second, it may happen that we split the *second* (or third or...) occurence of a B variable. In this case, `slice_composition` is currently only substituting on the remainder of the composition, and so the first occurence is not substituted. To fix that case, we keep the substitution of B variables around, and apply it to the whole equation once it has been sliced. This can again change the pattern and cause re-slicing.

Note that this keeps the "one-var-to-two-vars" implementation of substitutions. As discussed in #680, we should switch to a non-recursive version, but this is unrelated to the issues fixed by this patch.

Fixes #645


bclement-ocp:
I messed up the rebase and forgot the tests, should be good now @hra687261

bclement-ocp:
The build failure is due to Gbury/dolmen#162 being merged. I wlil make a PR shortly to remove `B.In_interval` in the short-term; the long-term fix is #652

bclement-ocp:
> the long-term fix is #652

I meant #681

src/lib/reasoners/bitv.ml
```diff
@@ -699,28 +711,117 @@ module Shostak(X : ALIEN) = struct
...
+
+       If the composition [eq] does not match the pattern in [pat]:
+       - When variables in the composition are bigger than the expected pattern
+          size, the variables are split into two parts, the first of which has
+          the size expected by the pattern.
+
+          Splits involving A variables are not recorded, because A variables are
+          guaranteed to have unique occurences.
+
+          Splits involving B and C variables are accumulated in [subs].
```
hra687261: Hm, difference will probably be insignificant, but would there be a point in splitting `subs` into `b_subs` and `c_subs`? Since B variables can appear multiple times but only in a single composition, while C variables can appear multiple time but not in the same composition, I don't think that it is needed to have both substitutions in the same place.

hra687261 (approved):
LGTM

src/lib/reasoners/bitv.ml
bclement-ocp: Unless I am mistaken/misremembering, `subs` should be a pair `(b_subs, c_subs)` for this exact reason. I could clarify the comment.

hra687261: Right, line 788

PR #696, opened 2023-06-26
[Bitv] slice vars to lists instead of pairs

### LINKED ISSUE #689: [bitv] Make B and C substitutions be nonrecursive
What I mean was that, when we create the substitution, we could create a substitution of a var to a list.

AFAIK, yes, the substitutions are created in `slice_composition` when the size of the variable is bigger than the pattern. What I meant above is, instead of just looking at a variable `v` and a pattern `p`, we can have something that looks into the next patterns and check if the newly produced variable (the one on the right) doesn't need splitting as well? And if it is the case, the created substitution will map the original variables to all the produced splits.

_Originally posted by @hra687261 in https://github.com/OCamlPro/alt-ergo/issues/680#issuecomment-1603883867_

hra687261:
An attempt to fix #689
This doesn't remove all recursive substitutions, but should decrease their number, especially when a substitution comes from splitting a variable that is the result of another split variable. (when the resulting variable on the right also doesn't fit into its pattern)


src/lib/reasoners/bitv.ml
```diff
@@ -692,21 +692,39 @@ module Shostak(X : ALIEN) = struct
...
+    let slice_var var pat_hd pat_tl =
+      let mk, vsz, tr =
+        match var.bv with
+        | S_Cte _ -> (fun sz -> { var with sz }), var.sz, None
+        | S_Var { sorte = A; _ } ->
+          (fun sz -> { bv = S_Var (fresh_var A); sz }), var.sz, Some A
+        | S_Var { sorte = B; _ } ->
+          (fun sz -> { bv = S_Var (fresh_var B); sz }), var.sz, Some B
+        | S_Var { sorte = C; _ } ->
+          (fun sz -> { bv = S_Var (fresh_var C); sz }), var.sz, Some C
```
bclement-ocp: `vsz` is not needed in the match, it is always equal to `var.sz`

src/lib/reasoners/bitv.ml
```diff
@@ -784,36 +802,38 @@ module Shostak(X : ALIEN) = struct
             let c, eq, pat = apply_sub_rev bsub c [] [] req in
             eq, csub, if c then Some pat else None
         end
-      | st :: eq, n :: pt when st.sz < n ->
+      | { bv = S_Var { sorte = B; _ }; _ } as st :: eq,
+        n :: pt when st.sz < n ->
```
bclement-ocp: Why can't we use `slice_var` here?

bclement-ocp: Also why doesn't `slice_var` return the `changed` boolean instead of having a special case for B-vars?

src/lib/reasoners/bitv.ml
```diff
@@ -784,36 +802,38 @@ module Shostak(X : ALIEN) = struct
...
         (* Since we start with a [slicing_pattern], this should only occur when
            a B-variable has been split into parts below. *)
         slice_vars eq (n - st.sz :: pt) (true, st :: req, subs)
       | st :: eq, n :: pt when st.sz = n ->
         slice_vars eq pt (c, st :: req, subs)
       | st :: eq, n :: pt ->
-        let (st_n, rst, flag) = slice_var st n in
+        let (nvar_list, pat_tail, flag) = slice_var st n pt in
```
bclement-ocp: Please keep the name `pt` instead of `pat_tail`. Yes `pt` may not be a good name, but it's better to have one name to refer to a thing rather than multiple names.

bclement-ocp (commented):
I need to make a second pass to make sure that the order is correct, but I think it is.

Can you clarify the cases where recursive substitutions still get created? I would think only B vars are concerned?

src/lib/reasoners/bitv.ml
bclement-ocp: Also I think this can be simplified:

```ocaml
let mk, tr =
  match var.bv with
  | S_Cte _ -> (fun sz -> { var with sz }), None
  | S_Var { sorte; _ } -> (fun sz -> { bv = S_Var (fresh_var sorte); sz }), Some sorte
in
...
```

hra687261: Done

src/lib/reasoners/bitv.ml
hra687261: Hm, I changed it because it is not the same one and I wanted to avoid shadowing it, `pt'` should be better?

src/lib/reasoners/bitv.ml
hra687261: > Why can't we use slice_var here?

Because we're not slicing variables, this is the result of a previous slicing and substitution of a B variable.

> Also why doesn't slice_var return the changed boolean instead of having a special case for B-vars?

It does make sens to make `slice_var` return `changed`, but since `apply_sub_rev` also returns `changed` anyway, though it wouldn't be necessary, since `apply_sub_rev` would do it anyway.
The special case for B is due to the fact that that case can only be reached with B variable that was the result of a slicing and a substitution. (This also returns `changed = true` which is weird, correct but probably unnecessary, I think it would be "prettier" to have only `apply_sub_rev` return `changed` that way the argument can be removed since `apply_sub_rev` is only applied at the end)

hra687261:
> Can you clarify the cases where recursive substitutions still get created? I would think only B vars are concerned?

Yes, only B variables are concerned. If a B variable is split into `n` new variables, then a second instance of that B variable is replaced with those `n` new ones, I guess it's possible for one of them to need splitting which would lead to a split/substitution inside a split? Tbh I'm not super sure it can happen, I'll think about it and try to reproduce it

src/lib/reasoners/bitv.ml
bclement-ocp: Ah right, I got confused by the change that added the `sorte = B` case, thanks for the clarification. We should either perform the splitting (as was done before) or have an `assert false` here if the we get `st.sz < n` and the `sorte` is not `B` rather than fall through to the slicing case below.

Regarding returning `changed = true` here — you are right it is probably unnecessary because the `bsub` must be non-empty if we reach that case (when I wrote the code, I initially planned on removing `slicing_pattern`, in which case `changed = true` here would have been necessary). I am OK with not setting `changed` here but we probably should assert that `bsub` is nonempty in that case, to ensure that `apply_sub_rev` actually gets called.

src/lib/reasoners/bitv.ml
bclement-ocp: I think it would be an error to keep using the old `pt` here, and so shadowing is fine. Conceptually, `pt` refers to "the rest of the pattern" and after `slice_var` is called, it returns "the rest of the pattern". But if you insist on not shadowing, then yes, `pt'` is lower cognitive complexity because it's clearer that it is "the same kind of thing" as `pt'`.

bclement-ocp:
Yeah, I am positive that can happen. My point was more that if there can't be recursive substs for C vars (which I agree with you that I think there can't be) then we could remove the recursive call in `apply_subs`.

BTW, it would probably be good to have tests for:

 - Multi-splitting of a B variable (forward and backward)
 - Multi-splitting of a C variable

It is fairly easy to trigger the behavior (write equalities between concatenations of extractions of distinct variables for C vars, write equalities between concatenations of extractions of the same variable for B vars); I am not entirely sure how to check that it is correct. Assert the negation of the implied equalities by hand, probably?

`x[4] @ y[4] = a[1] @ b[1] @ c[1] @ d[1] @ x[4]` should cause four-way splitting of a C var (if I am not mistaken) and we should check that it implies `y^{0,0} = d`, `y^{1,1} = c`, etc.

I had an example to cause forward splitting of B-vars in #675 which we can reuse.

(I don't think it's *necessary*, but it would help make sure we don't mess up the orders in subtle ways)

src/lib/reasoners/bitv.ml
hra687261: I guess intentional shadowing is fine, as long as we know what we are doing. But I still think that out of clarity, it's better to keep `pt'` as a way to say that it is a modified version of `pt`.

src/lib/reasoners/bitv.ml
hra687261: Right, It's true that the slicing case doesn't have a `when st.sz > n`, so yeah, an assertion is better. Done

src/lib/reasoners/bitv.ml
bclement-ocp: Interesting, I would say that `pt` is clearer to me because the fact that there is an assignment at all indicates that it is a modified version. Anyways, `pt'` it is here

src/lib/reasoners/bitv.ml
```diff
@@ -785,35 +799,40 @@ module Shostak(X : ALIEN) = struct
             eq, csub, if c then Some pat else None
         end
       | st :: eq, n :: pt when st.sz < n ->
+        assert (not (Lists.is_empty (fst subs)));
+        assert (match st.bv with S_Var { sorte = B; _ } -> true | _ -> false);
```
bclement-ocp: The `assert` should go after the explanatory comment rather than before, since the comment explains the rationale for the asserts.

hra687261: Done.

bclement-ocp (approved):

hra687261:
For some reason the last commit doesn't appear (yet?). Could be a GH problem, but it removes the recursive call in `apply_subs` and replaces `f_aux (vl @ r)` with `vl @ f_aux r`.
Also added the test you suggested and updated the test of the issue 645 with the other cases (just in case) but I can remove them if you think it is out of the score of the PR or unnecessary.

Edit: not sure I understand why the commit doesn't appear here, but here's the change:
https://github.com/hra687261/alt-ergo-fuzz/blob/9f4c391cb3812e1754daa8ed5b71fff84dc7d528/src/lib/reasoners/bitv.ml#L897

bclement-ocp:
Sounds good! You should wait for the commit to appear in the GH UI before merging or it won't be picked up, we had the issue with @Halbaroth on another PR.

PR #698, opened 2023-06-27
Add support for (check-sat-assuming) in Dolmen frontend

### LINKED ISSUE #695: [dolmen] Support (check-sat-assuming)
The legacy psmt2 frontend supports `(check-sat-assuming)`, but the Dolmen frontend currently doesn't.

Or rather, it partially does — it seems to *ignore* (!!) the assumptions, see this example:

```smtlib2
(set-logic ALL)

(define-fun b () Bool false)
(check-sat-assuming (b))
```

The legacy frontend correctly answers `sat`, but the dolmen frontend answers `unknown`.

bclement-ocp:
This changes the way we parse the `` `Solve`` constructor from Dolmen.

Currently, we accept the following forms:

 - `` `Solve ([], [t])``, i.e. `goal g: t` in the native format
 - `` `Solve ([t], [])``, i.e. `check_sat t` in the native format, and `(check-sat-assuming (t))` in the smtlib2 format
 - `` `Solve ([], [])``, i.e. `(check-sat)` or `(check-sat-assuming ())` in the smtlib2 format

This patch adds support for the more general case of `(check-sat-assuming)` by handling `` `Solve (ts, [])``.

This is done by making the `` `Check`` constructor take a list of term (rather than a single term), and generalizing the `pp_query` preprocessing helper in `D_cnf` to accept an arbitrary list of hypotheses.

(Note that this means `(check-sat)` now has `Sat` as a `goal_sort` rather than `Thm`, which is conceptually more correct but doesn't really make a difference in practice)

Fixes #695

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -442,9 +442,8 @@ let main () =
         in
         let contents =
           match contents with
-          | `Solve ([], []) -> `Goal (DStd.Expr.Term.(of_cst Const._false))
+          | `Solve (hyps, []) -> `Check hyps
           | `Solve ([], [t]) -> `Goal t
-          | `Solve ([t], []) -> `Check t
           | _ ->
             let loc = DStd.Loc.loc file_loc loc in
             Util.failwith "%a: internal error: unknown statement"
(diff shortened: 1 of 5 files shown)


Halbaroth (approved):
LGTM

PR #703, opened 2023-06-27
[models] Add tests for model generation

bclement-ocp:
This updates the [gentest] tool to be able to add tests for model generation.

Note that while writing this I noticed:

 - Model generation doesn't work for `check_sat` in the Alt-Ergo language (the models are not printed, unless `--dump-model` is provided)
 - When using `check_sat` with Alt-Ergo or `(check-sat-assuming)` for smtlib2, we are not actually taking into consideration the assumptions when printing the model and print the pre-solve model. Or something similar (the gist of it is that `process_decl` discards the generated model). I will make separate PRs for these.


Halbaroth (commented):
LGTM

bclement-ocp:
Uh, looks like some of the expected files are empty, not sure what's going on

Halbaroth:
Did you promote them with `make promote`?

bclement-ocp:
No I just messed up a rule in `gentest` and the tests were not actually being run :grin: It should be good now if you can re-review!

Halbaroth (approved):

PR #711, opened 2023-06-27
Documentation of the ite relations

Halbaroth:
This PR add a documentation of the implementation of `ite_rel`.
As Basile explained this theory simplify an ite `t` of the form `ite(cond, b1, b2)` and whose the condition `cond` is assumed in the current context by adding equation `t = b1` or `t = b2` according to the truth value of `cond`.

I noticed that `ite_rel` does this simplification only for predicate condition. I don't know if there is actually other cases.

### DIFF
--- a/src/lib/reasoners/ite_rel.ml
+++ b/src/lib/reasoners/ite_rel.ml
@@ -56,12 +56,12 @@ module TB =
 
 (* The present theory simplifies the ite t of the form
     ite(pred, t1, t2)
-   where pred is an assumed predicate by adding the equation
+   where pred is an assumed predicate by introducing the equation
    t = t1 or t = t2 according to the truth value of pred. *)
 
 type t = {
   pending_deds      : Ex.t ME2.t;
-  (* Map of pending deducations to their explanation. A deduction is just
+  (* Map of pending deductions to their explanation. A deduction is just
      an equation of the form t = b where t is an ite and b is one of its
      branches. *)
 
--- a/src/lib/reasoners/ite_rel.ml
+++ b/src/lib/reasoners/ite_rel.ml
@@ -97,7 +97,7 @@ let add_to_guarded p s t mp =
   let st = try ME.find p mp with Not_found -> SE2.empty in
   ME.add p (SE2.add (s, t) st) mp
 
-(* Check if the condition of the ite t is an assumed predicate. If do so,
+(* Check if the condition of the ite t is an assumed predicate. If so,
    select the appropriate branch b of the ite and produce the deduction t = b.
    Otherwise save the if and else branches of t in order to retrieve them
    quickly in the assume function. *)
--- a/src/lib/reasoners/ite_rel.ml
+++ b/src/lib/reasoners/ite_rel.ml
@@ -54,28 +54,37 @@ module TB =
         if c <> 0 then c else Stdlib.compare b1 b2
     end)
 
-(* The present theory simplifies the ite t of the form
+(* The present theory simplifies the ite terms t of the form
     ite(pred, t1, t2)
    where pred is an assumed predicate by introducing the equation
    t = t1 or t = t2 according to the truth value of pred. *)
 
 type t = {
   pending_deds      : Ex.t ME2.t;
-  (* Map of pending deductions to their explanation. A deduction is just
-     an equation of the form t = b where t is an ite and b is one of its
-     branches. *)
+  (* Map of pending deductions to their explanation. A deduction is an equation
+     of the form t = b where t is an ite term and b is one of its branches.
+     A deduction is added to pending_deds if the condition of the ite term
+     is entailed by the current state of the solver. If so, the appropriate
+     branch is selected according to the truth value of the condition. *)
 
   guarded_pos_deds  : SE2.t ME.t;
-  (* Map of the conditions to the if branches. *)
+  (* Map of the condition of ite terms to its if branch.
+     This map contains only condition that was not entailed yet by the
+     current state of the solver during the registration of the ite term. *)
 
   guarded_neg_deds  : SE2.t ME.t;
-  (* Map of the conditions to the else branches. *)
+  (* Map of the condition of ite terms to its else branch.
+     This map contains only condition whose the negation was not entailed yet
+     by the current state of the solver during the registration of the ite
+     term. *)
 
   assumed_pos_preds : Ex.t ME.t;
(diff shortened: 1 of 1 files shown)


src/lib/reasoners/ite_rel.ml
```diff
@@ -54,21 +54,37 @@ module TB =
...
-    assumed_neg_preds = ME.empty;
-  }
+(* The present theory simplifies the ite t of the form
+    ite(pred, t1, t2)
+   where pred is an assumed predicate by adding the equation
+   t = t1 or t = t2 according to the truth value of pred. *)
+
+type t = {
+  pending_deds      : Ex.t ME2.t;
+  (* Map of pending deducations to their explanation. A deduction is just
```
bclement-ocp: What does "pending" mean? My intuition is that this a deduction that is true under the corresponding explanation, and the explanation is assumed to be true in the current context, is that accurate?

bclement-ocp:
```suggestion
  (* Map of pending deductions to their explanation. A deduction is just
```

src/lib/reasoners/ite_rel.ml
```diff
@@ -54,21 +54,37 @@ module TB =
...
+     an equation of the form t = b where t is an ite and b is one of its
+     branches. *)
+
+  guarded_pos_deds  : SE2.t ME.t;
+  (* Map of the conditions to the if branches. *)
+
+  guarded_neg_deds  : SE2.t ME.t;
+  (* Map of the conditions to the else branches. *)
```
bclement-ocp: Do these contain `condition → if` and `condition → else` mappings for all the ite that we have seen, or only for the ites that are currently true?

I see the range is `SE2.t`; is the other component of the pair the corresponding ite (so that `c` maps to the pair `((ite c t e), t)` in `guarded_pos_deds` and to the pair `((ite c t e), e)` in `guarded_neg_deds`?

src/lib/reasoners/ite_rel.ml
```diff
@@ -81,6 +97,10 @@ let add_to_guarded p s t mp =
   let st = try ME.find p mp with Not_found -> SE2.empty in
   ME.add p (SE2.add (s, t) st) mp
 
+(* Check if the condition of the ite t is an assumed predicate. If do so,
```
bclement-ocp: What does "an assumed predicate" here? Is it "a predicate entailed by the current state"?

bclement-ocp:
```suggestion
(* Check if the condition of the ite t is an assumed predicate. If so,
```

src/lib/reasoners/ite_rel.ml
Halbaroth: Some deductions are produced when we add the ite with `add`. This happened if we know its condition is true or false because we assumed it before with the function `assume`. 
All the pending deductions are returned by the function `assume`.

The explanation `ex` comes from the explanation introduced by the predicate in `assume`.

We can rename this field. I agree the name isn't appropriate since the pending map is never flushed.

src/lib/reasoners/ite_rel.ml
Halbaroth: I am not totally sure. We should check what is the semantic of the assume function in `Sig_rel.RELATION`. By assumed predicate, I mean predicate that appears once in the argument of the assume function in `Ite_rel`.

src/lib/reasoners/ite_rel.ml
Halbaroth: `guarded_pos_deds` contains all the if branch of ite terms whose the truth value of the condition was unknown when we registered the ite with the function `add`.

Strangely, `assume` doesn't remove the branches in these caches when the condition is finally assumed.

src/lib/reasoners/ite_rel.ml
bclement-ocp: This is a documentation PR, it should not change any names.

From what you are saying, I understand that a "pending deduction" is a deduction that is true in the current model, associated with the explanation that made it true? If so we should add the info to the doc of `pending_deds` (saying that "`pending_deds` is the map of pending deductions to their explanations" is not very useful to understand the code if we don't explain what a "pending deduction" is).

src/lib/reasoners/ite_rel.ml
bclement-ocp: OK, yeah, I *think* that is the case then but not 100% sure. Maybe add the info that "assumed predicate" means "predicate that appears in the argument of `assume`", that is already useful, and we will understand when `assume` is called later then.

src/lib/reasoners/ite_rel.ml
bclement-ocp: Thanks for the clarification We should add that info to the doc then :)

bclement-ocp:
> I noticed that ite_rel does this simplification only for predicate condition. I don't know if there is actually other cases.

Can you clarify what you mean by this?

src/lib/reasoners/ite_rel.ml
bclement-ocp: More precisely I believe that the predicates that `assume` gets called on are predicates entailed by the current state, but I do not know whether all predicates that have been either asserted or learnt result in a call to `assume`.

src/lib/reasoners/ite_rel.ml
Halbaroth: Ok ;)

Halbaroth:
I improved the documentation. Should be better now :)

Halbaroth:
> > I noticed that ite_rel does this simplification only for predicate condition. I don't know if there is actually other cases.
> 
> Can you clarify what you mean by this?

Sure :) I meant the module considers only ite terms whose the condition is a predicate. I don't know if we allow a condition to be something else. Since this module is currently dead code, we cannot know if there were ite terms whose the condition was a boolean term but wasn't a predicate in the previous codebase.

bclement-ocp (approved):
Looks good, thanks for going through the code!

I think that boolean terms and predicates are the same type now? But it does look like equalities/disequalities (i.e. non-`Pred` `lit_view`) are not handled by the theory rn.

Halbaroth:
> Looks good, thanks for going through the code!
> 
> I think that boolean terms and predicates are the same type now? But it does look like equalities/disequalities (i.e. non-`Pred` `lit_view`) are not handled by the theory rn.

It should be but anyway :)

PR #712, opened 2023-06-28
Make Dolmen ignore the "shawoding" warning

hra687261:
Many files from the SMT-LIB benchmarks have (intentional or not) shadowing.
Allowing the warning pollutes the output.


hra687261:
The CI failure is not caused by the changes of the PR

bclement-ocp (commented):
If I remember correctly, I left it here because the legacy frontend had the same warnings, but I'm fine with removing them. I have already made my position known on "cleanliness" warnings for what is often generated code.

bclement-ocp (approved):

PR #714, opened 2023-06-29
Create timers

### LINKED ISSUE #709: Wrong timer in ite_rel

Halbaroth:
This PR adds a separated timer for the ite theory. Since the ite_rel is actually deadcode, since modification has no effect on the behaviour of Alt-Ergo.


hra687261 (approved):

bclement-ocp:
Note: "the ite_rel is actually deadcode" is what we supposed at the time but turned out to not actually be true.

PR #717, opened 2023-06-30
Use --enable-theories instead of --enable-preludes

### LINKED ISSUE #673: Decide between `--enable-preludes` and `--enable-theories` for 2.5.0
Should I name it `--enable-preludes` or `--enable-theories`? I think if we plan on going ahead with renaming to `--enable-theories` it is better to call it `--enable-theories` immediately, would avoid having to deprecate one more option, even though I feel a bit icky about the name if it only affects prelude theories

_Originally posted by @bclement-ocp in https://github.com/OCamlPro/alt-ergo/pull/650#discussion_r1231300706_

bclement-ocp:
This patch removes the `--enable-preludes` and `--disable-preludes` flags that leak implementation details and replaces them with `--enable-theories` and `--disable-theories` flags that are more user-friendly.

The flags `--no-ac` and `--disable-adts` are now deprecated aliases for `--disable-theory ac` and `--disable-theory adt`, respectively.

The `--disable-default-preludes` flag has been removed, because it would be confusing for users to have a `--disable-default-theories` flag that only disables the prelude theories, ADT, and AC but not other theories such as bitvectors. In the future, we should migrate all the theories to the `--enable-theory` / `--disable-theory` flags and then we can reintroduce the flag, which would behave similarly to `--restricted` mode.

Fix #673


Halbaroth (approved):
LGTM

PR #718, opened 2023-06-30
Add knowledge about floor/ceil bounds

bclement-ocp:
In the RIA prelude, we have axioms with semantic triggers to propagate bounds on reals to bounds on integers using `real_of_int` (`real_of_int_to_int_1` and `real_of_int_to_int_2`).

These state respectively that:

 - If `real_of_int(i) ≤ y` holds in the reals, then `i ≤ int_floor(y)` holds in the integers.

 - If `y ≤ real_of_int(i)` holds in the reals, then `int_ceil(y) ≤ i` holds in the integers.

This patch adds the corresponding propagations from bounds on integers to bounds on reals. More precisely it adds the propagations:

 - If `int_floor(x) ≤ i` holds in the integers, `x < real_of_int(i + 1)` holds in the reals

 - If `i ≤ int_floor(x)` holds in the integers, `real_of_int(i) ≤ x` holds in the reals

 - If `int_ceil(x) ≤ i` holds in the integers, `x ≤ real_of_int(i)` holds in the reals

 - If `i ≤ int_ceil(x)` holds in the integers, `real_of_int(i - 1) < x` holds in the reals

(Note that the test has "unknown" expected but it should be SAT -- this is because the ria prelude is disabled currently)

### DIFF
--- a/src/preludes/ria.ae
+++ b/src/preludes/ria.ae
@@ -56,6 +56,34 @@ theory Real_of_Int extends RIA =
      {i <= real_of_int(x)}.
      k <= x
 
+  (* floor(x) ≤ i iff x < i + 1 *)
+  axiom int_floor_ub:
+    forall x, y : real, i : int
+    [ int_floor(x), not_theory_constant(x), int_floor(x) in ]?, i], y |-> real_of_int(i + 1) ]
+    { int_floor(x) <= i }.
+    x < y
+
+  (* i <= floor(x) iff i <= x *)
+  axiom int_floor_lb:
+    forall x, y : real, i : int
+    [ int_floor(x), not_theory_constant(x), int_floor(x) in [i, ?[, y |-> real_of_int(i) ]
+    { i <= int_floor(x) }.
+    y <= x
+
+  (* ceil(x) ≤ i iff x ≤ i *)
+  axiom int_ceil_ub:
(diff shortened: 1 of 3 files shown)


bclement-ocp:
Build failure on macOS is spurious and not related to the PR.

src/preludes/ria.ae
```diff
@@ -56,6 +56,34 @@ theory Real_of_Int extends RIA =
      {i <= real_of_int(x)}.
      k <= x
 
+  (* floor(x) ≤ i iff x < i + 1 *)
```
Halbaroth, line 59: I guess you meant `floor(x) <= i implies x < i+1` instead of `floor(x) <= i iff x < i+1`. 
Besides, I don't understand why you need to use the semantic trigger `not_theory_constant` here?

Halbaroth (approved):

src/preludes/ria.ae
bclement-ocp: `floor(x) <= i iff x < i + 1` is correct, unless I am mistaken? Even though we only implement the implication.

`not_theory_constant` is here to avoid adding useless axioms, e.g. `3 < 4` if we know that `floor(3.5) <= 3`

bclement-ocp:
Note: waiting on #720 to make sure the test pass :)

PR #725, opened 2023-07-07
Differentiate between Dolmen exceptions and others

hra687261:


bclement-ocp:
Do you have an example case where this is relevant? I thought dolmen was supposed to be printing its own errors, and we should see exceptions from Dolmen. Is that inaccurate?

hra687261:
Not if the exception is not a "Dolmen exception" but a runtime exception. It would be useful to know if a stack overflow exception is raised from Dolmen or from the solver. But this is wrong anyway because `handle_exn` and `finally` catch all exceptions that are raised in the pipeline.

PR #726, opened 2023-07-07
[bitv] Add support for (bvnot) in the solver

bclement-ocp:
This patch adds support for the negation (bvnot) operator in the bitvector solver.

There are two components to this support:

 - First, the union-find data structure used by the solver is replaced with an actual union-find data structure implemented using Tarjan's efficient algorithm, rather than using an implementation with sets and maps.

 - Second, the new union-find data structure is augmented with a link between a class representative and the representative of its negated class. When merging two classes, their negated classes are merged as well (and, in particular, if a class gets forced to all ones or all zeroes, its negated classes gets forced to the other bit value). If a class is ever merged with its negated class, the problem is unsolvable.

The implementation is not necessarily the most efficient (e.g. we iterate over lists to negate them rather than directly creating the negated variables at several points), but it should be fairly simple and requires few changes to the existing solver infrastructure. The performance can be improved later if necessary.

This extracted and simplified from part #669 and works towards better support for #625.

Note that I am not sure whether we want this for 2.5.0 or not. It would be nice to have it, but is not necessary, strictly speaking.


bclement-ocp:
I am going to add this as part of a different PR set, disregard for now.

PR #729, opened 2023-07-10
Use dune's include stanza for generated tests

bclement-ocp:
Currently, the dune files for tests are generated using the Makefile, and not included in the repository. There are often issues when changing branches that add new tests, because the dune files easily get out of sync, and regularly require manual intervention to remove them when new test directories are added.

This patch uses dune's auto-generated files and include mechanism for test generation instead. More precisely:

 - The test generation script (gentest.ml) is modified to write all the test cases on the standard output as a single file, rather than creating the dune test files by itself. The generated file makes use of the `subdir` stanza to locate the targets appropriately.

 - There is a new `dune.inc` file that is included by the `dune` file in `tests`, and a `diff` rule that generates the content of that `dune.inc` file using the `gentest.ml` script.  This is the way rules generation must work in dune, and it means that the `dune.inc` file must be commited to the repository --- which may not be such a bad thing, see above.

The rule generation is added to the `gentest`, `runtest`, `runtest-quick` and `runtest-ci` aliases through the `aliases` stanza, which requires bumping our dune dependency to 3.5.  If that is an issue, we can duplicate the rule instead.


hra687261 (approved):

hra687261:
> which requires bumping our dune dependency to 3.5. If that is an issue, we can duplicate the rule instead.

Fine by me, but could be worth getting a second opinion form @Halbaroth or @Gbury

bclement-ocp:
OK, I'll wait until either of them chimes in.

Halbaroth:
In my opinion, we shouldn't force the update to dune 3.5 for such a tiny benefit.

bclement-ocp:
OK, I will duplicate everything (athough I think most people already use a version of dune higher than 3.5 if they are using opam).

Halbaroth:
Could rebase this PR please?

bclement-ocp:
Looks like the CI doesn't see the test files for some reason, I will have to investigate...

PR #730, opened 2023-07-10
Add basic support for the SMT-LIB BV and QF_BV logics

bclement-ocp:
This is the first part in a reloaded version of #669.  It only includes parsing and typechecking support for the bit-vector primitives: bvnot, bvand, bvor, bv2nat and nat2bv (called int2bv because it also accepts negative inputs) are treated as uninterpreted functions; the other operators defined in the SMT-LIB are defined in terms of these primitives and of the existing `concat` and `extract` operators.

The support for the bit-vector functions will be improved in separate PRs focusing on the bv2nat/nat2bv interaction, so that we are at least able to solve simple arithmetic goals. In the future, we might want to consider additional strategies (e.g. bitblasting), but that is out of scope for now.

Following the discussion in #669, including the apparent restrictions around the building of "ite" in the theory and the lack of infrastructure for actually doing simplifications there (at least for now), the definition of most functions are moved back to `Expr.ml`. The PR also includes a few helper functions to help write the definition of the bitvector operators more concisely, which should help ensuring that they are correct.

### DIFF
--- a/src/lib/frontend/d_cnf.ml
+++ b/src/lib/frontend/d_cnf.ml
@@ -882,6 +882,7 @@ let rec mk_expr
   let name_tag = ref 0 in
   let rec aux_mk_expr ?(toplevel = false)
       (DE.{ term_descr; term_ty; term_tags = root_tags; _ } as term) =
+    let mk = aux_mk_expr in
     let res =
       match term_descr with
       | Cst ({ builtin; path; _ } as tcst) ->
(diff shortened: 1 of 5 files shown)


bclement-ocp:
I had mixed up `<` and `<=` for integers — it should be good now.

bclement-ocp:
This is +3295 on QF_BV just by being able to parse some otherwise trivial benches. There are 54 errors, all of which are stack overflows while typing (similar to #669, iirc).

hra687261 (approved):

PR #732, opened 2023-07-11
[bitv] Rewrite the canonizer

bclement-ocp:
This is a full rewrite of the Canonizer module in the bitvector solver. The new Canonizer (now called Canon) performs the same work as the old one, but does so slightly differently -- in particular, we now build terms lazily, which allows us to ignore parts of the tree for performance.

The main motivation is actually not performance, but the ability to integrate bitwise operators (bvnot, bvand, bvor) into the canonizer in a more natural way. This is essentially the same architecture as that of `make_aux` from #669, but lifted to use a simple custom type (`view`) to avoid building new terms during canonization and benefit from match exhaustiveness checks.

### DIFF
--- a/src/lib/reasoners/bitv.ml
+++ b/src/lib/reasoners/bitv.ml
@@ -192,137 +192,144 @@ module Shostak(X : ALIEN) = struct
       let compare = compare_solver_simple_term
     end)
 
-  module Canonizer = struct
+  module Canon = struct
+    type 'a view_descr =
+      | Vcte of string
+      | Vother of 'a
+      | Vextract of 'a * int * int
+      | Vconcat of 'a * 'a
 
-    type term_aux  =
-      | I_Cte of bool
-      | I_Other of X.r
-      | I_Ext of term * int * int
-      | I_Comp of term * term
+    and 'a view = { descr : 'a view_descr ; size : int }
 
-    and term = term_aux alpha_term
+    let view t =
+      match E.term_view t with
+      | { f = Bitv s; ty = Tbitv size; _ } -> { descr = Vcte s; size }
+      | { f = Op Concat; xs = [ t1; t2 ]; ty = Tbitv size; _ } ->
+        { descr = Vconcat (t1, t2); size }
+      | { f = Op Extract (i, j); xs = [ t' ]; ty = Tbitv size; _ } ->
+        assert (size = j - i + 1);
+        { descr = Vextract (t', i, j); size }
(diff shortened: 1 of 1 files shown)


bclement-ocp:
There are no regressions on the ~800 QF_BV benchmarks that we have the primitives for, and 6 benchmarks that used to timeout and now barely pass (on 4.14).

src/lib/reasoners/bitv.ml
```diff
@@ -192,137 +192,144 @@ module Shostak(X : ALIEN) = struct
...
+      f x ctx
+    let ret x ctx = x, ctx
+    let (and*) x y =
+      let* x = x in
+      let* y = y in
+      ret (x, y)
+    let (let+) x f ctx =
+      let x, ctx = x ctx in
+      f x, ctx
+    let (and+) = (and*)
```
hra687261, line 228: A bit weird to use `x` as both a function and the result of a function application, the argument `x` could be called `g` for example.

hra687261 (approved):
Not really what I had in mind for the canonizer but seems to be ok.

bclement-ocp:
> Not really what I had in mind for the canonizer but seems to be ok.

Happy to settle on another approach if you have something better to suggest, what did you have in mind?

src/lib/reasoners/bitv.ml
bclement-ocp: I think it is relatively idiomatic to have `x` be both the "stateful" `'a m = 'st -> 'a * 'st` and the "inner" `'a`, this is what happens e.g. when we do `let* x = x in ...` as well.

I'm happy to call it `mx` here, and will add a comment explaining that `(let*)` and `(let+)` implement `bind` and `fmap` for the state monad as well.

hra687261:
I'm referring to what was discussed in #625, although it doesn't really matter (for now at least), if that version turns out to be better I'll open a PR at some point for it.

bclement-ocp:
Are you referring to https://github.com/OCamlPro/alt-ergo/issues/625#issuecomment-1607452057 ? If so I believe the implementation in this PR does what you described?

With `(extract 0 0 (concat complex #b0))` we never recurse into `complex` (well technically we do look at the head constructor of `complex`, but that does not depend on the size of `complex`).

hra687261:
True, it does what is described. By "not what I had in mind", I was referring to the way in which to do it, what I had in mind was to completely remove the intermediary state (as it was said in the comment).
It's not clear that it would be better to remove the intermediary state though.

PR #733, opened 2023-07-12
Add support for ground bv2nat and int2bv reasoning

bclement-ocp:
This patch adds support for "combined" reasoning about the int2bv and bv2nat symbols in the arith and bitv reasoners.

More specifically:

 - The constructors for bv2nat and int2bv in `expr.ml` are now "smart" and know that they are related, so that we effectivly have the following rewrite rules:

    `(bv2nat ((_ int2bv n) x))` -> `(mod x (pow 2 n))`
    `((_ int2bv n) (bv2nat (as x (_ BitVec m))))`
      -> `(ite (> m n) (extract (- n 1) 0 x) (zero_extend (- m n) x))`

   These are added early in the bv2nat and int2bv constructors to ensure that we *never* actually build `((_ int2bv n) (bv2nat e))` or `(bv2nat ((_ int2bv n) e))` terms. This is actualy useful because the simplifications done for `bv2nat` (see below) are more powerful than what `int2bv` can recover, and there would be information loss otherwise e.g. for `((_ int2bv n) (bv2nat (concat e1 e2)))` which would become

     `((_ int2bv n) (+ (* (bv2nat e1) (pow 2 m)) (bv2nat e2)))`

   instead of

      `(concat e1 e2)`

   (assuming that `e2` is of type `(_ BitVec m)` and `e1` of type `(_ BitVec |n - m|)`).

 - The arith and bitv reasoners perform constant propagation of int2bv and bv2nat

 - The arith reasoner performs the `((_ int2bv n) (bv2nat x))` conversion more aggressivey, i.e. it performs the conversion for terms of the form `((_ int2bv n) e)` where the semantic value of `e` is of the form `(bv2nat e')`

   More complex transformations, such as transforming `(int2bv (* 2 (bv2nat e)))` into `(concat e #b0)`, are not supported.

 - The bitv reasoner also performs the `(bv2nat ((_ int2bv n) e))` conversion more aggressively, i.e. for terms that reduce to `((_ int2bv n) e)` after simplification.

 - The bitv reasoner performs additional simplifications on `bv2nat` terms after normalization. For instance, `(bv2nat (concat x y))` becomes `(+ (* (pow 2 n) (bv2nat x)) (bv2nat y))` when `y` is a bit-vector of size `n`, and `(bv2nat (extract j i x))` becomes `(mod (div (bv2nat x) (pow 2 i)) (pow 2 (+ (- j i) 1)))`

 - Finally, the bitv reasoner requires that the result of `bv2nat` for a bitvector of size `n` is in the range `0 ..< 2^n`

Combined, this allows exploiting the arith reasoner for a variety of problems mostly involving bitvector arithmetic with constants. In particular, we are able to handle many aritmetic expressions involving `bv2nat`.

On the other hand, bitvector expressions involving `int2bv` have more limited support. Indeed, not only does `int2bv` perform less simplifications, we are currently not able to exploit the relevant equivalence:

  `int2bv[n](x) = int2bv[n](y) iff x = y  (mod 2^n)`

### DIFF
--- a/src/lib/reasoners/arith.ml
+++ b/src/lib/reasoners/arith.ml
@@ -393,7 +393,7 @@ module Shostak
       begin match P.to_list p with
         | [], c ->
           let c = Q.to_z c in
-          let c = ZA.(erem c @@ pow ~$2 n) in
+          let c = ZA.(erem c @@ ~$1 lsl n) in
           let biv =
             String.init n (fun i ->
                 let i = n - i - 1 in
--- a/src/lib/reasoners/arith.ml
+++ b/src/lib/reasoners/arith.ml
@@ -31,6 +31,7 @@
 module Sy = Symbols
 module E = Expr
 
+module ZA = Z
 module Z = Numbers.Z
 module Q = Numbers.Q
 
@@ -119,7 +120,7 @@ module Shostak
          | Sqrt_real_default | Sqrt_real_excess
          | Real_of_int | Int_floor | Int_ceil
          | Max_int | Max_real | Min_int | Min_real
-         | Pow | Integer_log2
+         | Pow | Integer_log2 | Int2BV _
          | Integer_round) -> true
     | _ -> false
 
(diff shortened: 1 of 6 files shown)


hra687261 (approved):

bclement-ocp:
Note: I am re-running benches on QF_BV. There was an issue with the old canonizer where we could have bv terms as aliens; the new canonizer (from #732 ) no longer has that issue.

But the tests are still very slow b/c there are many timeouts… there are 42k tests, and if they all time out, if would take 4h to run all the tests. In practice it looks like it's more around half of them timing out?

bclement-ocp:
This is +1272-1843 which I assume is mostly due to #736 but still is a big drop.

Although there is also a relatively high number of segmentation faults (I tracked one to `Z.to_string` from ZArith on marvin, but couldn't reproduce locally) so… looks like something fishy is going on, maybe with the gmp version on marvin?

bclement-ocp:
> Although there is also a relatively high number of segmentation faults (I tracked one to Z.to_string from ZArith on marvin, but couldn't reproduce locally) so… looks like something fishy is going on, maybe with the gmp version on marvin?

Turns out the `Z.to_string` issue was spurious and the actual is a stack overflow because `bv2nat2 is not an bitvector but an integer and we were looping. Rerunning the tests.

bclement-ocp:
This is now `+1541-1693` (still on QF_BV) which is more acceptable, but we are still losing a lot. Maybe related to #736? On the other hand, we want to compute on ground problems!

(Note that, now that I think about it, this is probably a big part of the issue --- due to technical limitations we are currenty only able to simplify e.g. `(bvadd #b01 #b01)` into `#b10` but not `(bvadd x x)` into `#b10` in a context where we know `x = #b01`)

bclement-ocp:
> (Note that, now that I think about it, this is probably a big part of the issue --- due to technical limitations we are currenty only able to simplify e.g. `(bvadd #b01 #b01)` into `#b10` but not `(bvadd x x)` into `#b10` in a context where we know `x = #b01`)

Which I just figured out how to do, will try to make a PR soon.

bclement-ocp:
> > (Note that, now that I think about it, this is probably a big part of the issue --- due to technical limitations we are currenty only able to simplify e.g. `(bvadd #b01 #b01)` into `#b10` but not `(bvadd x x)` into `#b10` in a context where we know `x = #b01`)
> 
> Which I just figured out how to do, will try to make a PR soon.

Well, actually no, because my solution would break `solve`.

Halbaroth:
Can I merge this PR?

bclement-ocp:
Yes let's merge this after our discussion

PR #739, opened 2023-07-18
Fix issue 719

### LINKED ISSUE #719: get-model prints models even if the status of the goal is `unsat`
I was playing with Why3, trying to prove my specification of a quick sort. My PO looks like that:
```
(set-logic ALL)
(declare-const s Int)
(assert (and (<= 0 s) (<= s 10)))
(assert (forall ((k Int)) (=> (and (<= 0 k) (< k s)) (<= (select a k) (select a s)))))
(assert (forall ((k Int)) (=> (and (< s k) (<= k 10)) (<= (select a s) (select a k)))))
(assert (
  forall ((p Int) (q Int))
    (=> (and (<= 0 p) (< p q) (<= q (- s 1))) (<= (select a p) (select a q)))))
(assert (
  forall ((p Int) (q Int))
    (=> (and (<= (+ s 1) p) (< p q) (<= q 10)) (<= (select a p) (select a q)))))
(assert (not (
  forall ((p Int) (q Int))
    (=> (and (<= 0 p) (< p q) (<= q 10)) (<= (select a p) (select a q))))))
(check-sat)
(get-model)
```
Z3 and Alt-Ergo prove that the context is `unsat` but AE produces a model... 
```
unsat
(
  (define-fun s () Int 0)
  (define-fun a () (Array Int Int)
   (store (store (as @a (Array Int Int)) 1 0) 0 2))
)
```
I guess I did something wrong in my implementation of `get-model`. I will fix it for the release.

Halbaroth:
Please, do not merge this PR before #738.

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -358,9 +358,9 @@ let main () =
       if Stdlib.(Options.get_sat_solver () = Tableaux) then
         Options.set_interpretation ILast
       else Printer.print_wrn "%a The generation of models is not supported \
-                              for the current SAT solver. Please choose the \
+                              for the SAT solver %a. Please choose the \
                               SAT solver Tableaux."
-          Loc.report st_loc
+          Loc.report st_loc Util.pp_sat_solver (Options.get_sat_solver ())
     | ":produce-models", Symbol { name = Simple "false"; _ } ->
       Options.set_interpretation INone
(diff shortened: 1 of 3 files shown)


bclement-ocp (approved):
Oh, we already have a `consistent` flag! That is easier than I thought it'd be :)

Can we drop 27c220a and merge? It doesn't look like there are conflicts between the two PRs.

bclement-ocp (approved):

PR #741, opened 2023-07-19
Fix 715

### LINKED ISSUE #715: User-defined function doesn't agree with functions in printed models
If we define a function in the input file, the model should print exactly the same function. For instance the input file:
```
(set-logic ALL)
(declare-fun f (Int) Int)
(assert (forall ((x Int) (y Int)) (= (f (+ x y)) (+ (f x) (f y)))))
(assert (= (f 0) 0))
(define-fun g () Bool (= (f 1) 1))
(check-sat-assuming (g))
(get-model)
```
produces the model:
``` 
unknown
(
  
 ; Functions

(define-fun f ((arg_0 Int)) Int (ite (= arg_0 1) 1 0))

 ; Constants

(define-fun g () Bool true)

 ; Arrays not yet supported


)
```

The value for `f` is correct but AE gives a new definition for `g`.

I think Steven reported this bug, but didn't create an issue for it.

Halbaroth:
This PR fixes the issue #715. Basically, I add an extra field in the constructor `Name` of the module `Symbol` to 
keep track of user-defined symbols.

The bug is fixed for Dolmen and legacy frontend.

### DIFF
--- a/src/lib/frontend/d_cnf.ml
+++ b/src/lib/frontend/d_cnf.ml
@@ -1850,7 +1850,7 @@ let make dloc_file acc stmt =
           match def with
           | `Term_def (_, ({ path; _ } as tcst), _, _, _) ->
             let name_base = get_basename path in
-            let sy = Sy.name ~user_defined:true name_base in
+            let sy = Sy.name ~defined:true name_base in
             Cache.store_sy tcst sy
           | `Type_alias _ -> ()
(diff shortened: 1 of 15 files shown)


bclement-ocp (approved):
This looks good!

I would just call the flag `defined` (as in `define-fun`) and drop the `user` part, but I'm fine with the `user_defined` version.

PR #742, opened 2023-07-19
Remove plateform-dependent message in the cram test

### LINKED ISSUE #740: macOS build is broken
Because the error messages from Dynlink are different, yay :)

Not sure how to fix this properly, tbh. Maybe we should have different tests for linux/macOS, or maybe we should rewrite the test slightly.

(We shouldn't hide the message from dylink from the output anyways, because that is helpful information for users)

```
-            >> Failure message: error loading shared library: Dynlink.Error (Dynlink.Cannot_open_dll "Failure(\"$TESTCASE_ROOT/does-not-exist: cannot open shared object file: No such file or directory\")")
+            >> Failure message: error loading shared library: Dynlink.Error (Dynlink.Cannot_open_dll "Failure(\"dlopen($TESTCASE_ROOT/does-not-exist, 0x000A): tried: '$TESTCASE_ROOT/does-not-exist' (no such file)\")")
```

Halbaroth:
Fix the issue #740

### DIFF
--- a/tests/cram.t/run.t
+++ b/tests/cram.t/run.t
@@ -1,7 +1,7 @@
   $ export BUILD_PATH_PREFIX_MAP="SRC=$(dirname $(dirname $(dirname $(dirname $(which alt-ergo))))):$BUILD_PATH_PREFIX_MAP"
-  $ echo '(check-sat)' | alt-ergo --inequalities-plugin does-not-exist -i smtlib2 -o smtlib2
+  $ echo '(check-sat)' | alt-ergo --inequalities-plugin does-not-exist -i smtlib2 -o smtlib2 2> >(sed -E 's/\(\\".*\\"\)//' 1>&2)
   alt-ergo: ; Fatal Error: [Dynlink] Loading the 'inequalities' reasoner (FM module) plugin in "does-not-exist" failed!
-            >> Failure message: error loading shared library: Dynlink.Error (Dynlink.Cannot_open_dll "Failure(\"$TESTCASE_ROOT/does-not-exist: cannot open shared object file: No such file or directory\")")
+            >> Failure message: error loading shared library: Dynlink.Error (Dynlink.Cannot_open_dll "Failure")
   [125]
 
 Now we will have some tests for the models. Note that it is okay if the format
@@ -9,7 +9,7 @@ changes slightly, you should be concerned with ensuring the semantic is
 appropriate here.
 
   $ alt-ergo --frontend dolmen --produce-models model555.smt2 --no-forced-flush-in-output 2>/dev/null
-  
+
   unknown
   (
     (define-fun a1 () (Array Int Int)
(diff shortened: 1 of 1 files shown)


bclement-ocp:
It looks like there is an issue o Ubuntu now?

bclement-ocp:
I think this might be because you removed the spaces. Cram files are finicky about that sometimes.

(BTW, you need to include "fix #740" or "fixes #740" in the commit message so that GitHub closes the issue automatically I believe)

Halbaroth:
> I think this might be because you removed the spaces. Cram files are finicky about that sometimes.

I just run `make runtest` and `dune promote` and my laptop. It wants to remove these spaces...

bclement-ocp:
Weird, then it's probably not the issue

tests/cram.t/run.t
```diff
@@ -1,15 +1,15 @@
   $ export BUILD_PATH_PREFIX_MAP="SRC=$(dirname $(dirname $(dirname $(dirname $(which alt-ergo))))):$BUILD_PATH_PREFIX_MAP"
-  $ echo '(check-sat)' | alt-ergo --inequalities-plugin does-not-exist -i smtlib2 -o smtlib2
+  $ echo '(check-sat)' | alt-ergo --inequalities-plugin does-not-exist -i smtlib2 -o smtlib2 2> >(sed -E 's/\(\\".*\\"\)//' 1>&2)
```
bclement-ocp: I am confused, I am not sure what `2> >(sed ... 1>&2)` does.

Does a simple pipe `... | sed -E 's/\(\\".*\\"\)//'` not work?

bclement-ocp:
> > I think this might be because you removed the spaces. Cram files are finicky about that sometimes.
> 
> I just run `make runtest` and `dune promote` and my laptop. It wants to remove these spaces...

Hmmm that is weird because `dune promote` wants to add the spaces back on my laptop....

bclement-ocp:
Are we sure `sed` is available on the CI runner?

tests/cram.t/run.t
Halbaroth: I guess it works. My version preserve completely the stdout and apply the sed command only on stderr.

Halbaroth:
> Are we sure `sed` is available on the CI runner?

Nope. I don't know so I try ;)

tests/cram.t/run.t
bclement-ocp: Actually it doesn't because the error is printed on stderr and the pipe hides the return code, I think.

(I think I could parse what happens with your description -- `2> >(...))` redirects stderr to the process in `...`, then `sed` happens, and the output of sed is redirected back to `stderr` -- nice trick)

Halbaroth: The trick does not work on the CI. I will find another way.

Halbaroth:
The new fix works :) It seems my trick with subprocess didn't work on both Ubuntu and MacOS ci.

bclement-ocp (approved):
:tada:

PR #745, opened 2023-07-21
Partial support of the primitive bvnot

Halbaroth:
This PR add a partial support of the bitvector primitive `bvnot`. Basically, the canonizer of the bitvector theory pushes the not operator to the leaves of the semantic value as much as it can.

### DIFF
--- a/src/lib/reasoners/bitv.ml
+++ b/src/lib/reasoners/bitv.ml
@@ -199,7 +199,7 @@ module Shostak(X : ALIEN) = struct
       | Vextract of 'a * int * int
       | Vconcat of 'a * 'a
       | Vnot of 'a
-    [@@deriving show]
+      (*     [@@deriving show] *)
 
     type 'a view = { descr : 'a view_descr ; size : int }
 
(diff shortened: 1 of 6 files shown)


Halbaroth:
There is no regression :) It is ready for review ;)

src/lib/reasoners/bitv.ml
```diff
@@ -198,8 +198,10 @@ module Shostak(X : ALIEN) = struct
       | Vother of 'a
       | Vextract of 'a * int * int
       | Vconcat of 'a * 'a
+      | Vnot of 'a
+      (*     [@@deriving show] *)
```
bclement-ocp: Either remove this and do not add `ppx_deriving` as a dependency, or keep `ppx_deriving` as a dependency and uncomment this. I'm not too keen on adding `ppx_deriving` for `show` implementations (I find that the output is usually barely readable, due to printing "raw" OCaml ASTs), but at the very least let's not add an unused dependency :)

src/lib/reasoners/bitv.ml
```diff
@@ -227,11 +231,40 @@ module Shostak(X : ALIEN) = struct
...
+      List.for_all (fun { bv = st; _ } ->
+          match st with
+          | Cte _ -> true
+          | Other _ | Ext _ ->
+            (* After normalization, the semantic value in
+               the constructor Other and Ext cannot be a bitvector.
+               Thus, these values are always uninterpreted for the
+               Bitv theory. *)
+            false
+        ) r
```
bclement-ocp: Can we merge these two raising `Failure` if either `Other` or `Ext` are encountered, and catch it when we call `negate_abstract` instead?

src/lib/reasoners/bitv.ml
```diff
@@ -227,11 +231,40 @@ module Shostak(X : ALIEN) = struct
...
       match X.extract r with
-      | Some bv -> bv, ctx
+      | Some bv ->
+        if neg_mode then
+          if no_variable bv then
+            negate_abstract bv, ctx
+          else
+            [ { bv = Other (X.term_embed (E.BV.bvnot t)); sz } ], ctx
+        else
+          bv, ctx
```
bclement-ocp: See above, could write

```ocaml
if neg_mode then
  try negate_abstract bv, ctx with
  | Failure _ -> [ { bv = Other (x.term_embed (E.BV.bvnot t)); sz } ], ctx
else
  bv, ctx
```

bclement-ocp (changes requested):
I'm not very fond of `neg_mode` because the use of `mode` implies a more fundamental change of behavior to me. I'd rather call this just `neg` or maybe `negated` which I think is more clear that the function operates as usual, but computes the negation of the result.

But I don't really have a strong justification for this, so if you feel strongly about using `neg_mode`, it's fine by me.

src/lib/reasoners/bitv.ml
Halbaroth: Actually, I use this dependency and I plan to use it in the codebase later. In the present code, I don't keep the deriving attribute because it produces a warning if we don't use the print function. I used these print functions only for debugging.

bclement-ocp: What I mean is that if there is no *code* that actually uses the dependency, it shouldn't be a dependency. If the issue is that we want to use it but it generates "unused code" warning, let's silence warning 32 locally.

bclement-ocp (approved):

Issue #474, opened 2021-08-09
Assertion failure in "src/lib/reasoners/intervalCalculus.ml", line 1175, when using the CDCL solver


hra687261:
[c_262_0_ae_bis.ae.txt](https://github.com/OCamlPro/alt-ergo/files/6955011/c_262_0_ae_bis.ae.txt)

Running Alt-Ergo (built from the "next" branch) on the attached file with the command:
```
./alt-ergo --sat-solver CDCL c_262_0_ae_bis.ae
```
Causes the following error:
```
Fatal error: exception File "src/lib/reasoners/intervalCalculus.ml", line 1175, characters 4-10: Assertion failed
```

And when the other solvers are used (by replacing CDCL with Tableaux or Tableaux-CDCL or CDCL-Tableaux) then Alt-Ergo responds with "Valid".

Some additional interesting points:
 - if we remove the axioms ax_1 and ax_2, then the answer in the four cases is:"I don't know".
 - if we replace ```(ruqv0 * ruqv0) * (if true then ruqv0 else ur_3)``` in the line 14 with ```(ruqv0 * ruqv0) * ruqv0``` or ```(ruqv0 * ruqv0)```, then AE takes too long to answer (more than 3 minutes) when the CDCL solver is used. And answers with "Valid" in the other cases.
 - if we replace it with ```ruqv0``` then the answer is "Valid" in every case.
 - if we replace it with ```ur_3``` then AE says:"I Don't Know" when the CDCL solver is used, and "Valid" in the other cases.

bclement-ocp:
the test case in the issue is hidden by #731 but the underlying cause is probably still valid.

bclement-ocp:
The assertion failure looks to be related to confusion in `intervalCalculus` about class representatives. The `intervalCalculus` environment is not consistent in the way it represents e.g. `ur_0 * ur_0`: sometimes it is the AC-term `ur_0 * ur_0`, and sometimes it is the purified constant `!?__let2((ur_0 * ur_0))` introduced by AC(X), which is the "true" class representative.

More precisely, there is an inequality (`'*'` is the AC symbol for multiplication, `*` with no quotes is the polynomial multiplication):

```
-ur_0'*'ur_0'*'ur_0+ur_0+1<=0
```

and the normalized polynomial is tracked:

```
ur_0'*'ur_0'*'ur_0-ur_0 : [1;+inf[
```

[...]

PR #749, opened 2023-07-24
Propagate equalities within int2bv

bclement-ocp:
This patch adds some reasoning support in bitv_rel for equalities and disequalities between `int2bv` terms. More specifically, when an equality or disequality (~ stands for either `=` or `distinct`):

  `(~ ((_ int2bv n) t) ((_ int2bv n) u))`

is asserted, we also assert that `t` and `u` are congruent (resp. noncongruence) modulo `2^n`, i.e.

  `(~ (mod (- t u) (pow 2 n)) 0)`

This allows recovering some reasoning power involving bitvector terms containing `int2bv`, and in particular terms involving bitvector arithmetic (since bitvector arithmetic is currently implemented using `int2bv`).

### DIFF
--- a/src/lib/reasoners/bitv_rel.ml
+++ b/src/lib/reasoners/bitv_rel.ml
@@ -28,14 +28,70 @@
 (*                                                                        *)
 (**************************************************************************)
 
+module E = Expr
+module X = Shostak.Combine
+
 type t = unit
 
+let as_int2bv r =
+  match X.term_extract r with
+  | Some t, _ ->
+    begin match E.term_view t with
+      | { f = Op Int2BV n ; xs = [ t ] ; _ } -> Some (t, n)
+      | _ -> None
+    end
+  | _ -> None
+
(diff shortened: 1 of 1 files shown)


bclement-ocp:
Draft because I need to re-run benches on this.

bclement-ocp:
This is +40-0 on smtlib22/QF_BV. Not sure I want to nominate this for 2.5.0, it's nice to have but not necessary so we can postpone (but I had it laying around so might as well make the PR).

Halbaroth:
Let's postpone it.

bclement-ocp:
Following my investigations for #474 and #801 I have a better understanding of the type of equalities we see in relations, and in its current instantiation, this is not a general solution and will only propagate `int2bv` equalities in very specific situations. An implementation that correctly propagates `int2bv` equalities in the general case will be more involved and requires additional work, closing for now and opening an issue instead.

Issue #705, opened 2023-06-27
Give warning when a model is probably wrong


bclement-ocp:
e.g. with floats

PR #755, opened 2023-07-26
Issue 705

Halbaroth:
This PR fixes the issue #705.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,30 +1,5 @@
 ## unreleased
 
-### New features
-* add context reinitialisation (PR #490)
-* add Dolmen frontend (PR #491,#541,#545)
-* add model generation (PR #530)
-* support mutually recursive definitions in the native language (PR #549, #550)
-* support of some options of the SMT-LIB statement (set-option) (PR #608)
-* support for the (get-model) statement (required the Dolmen frontend) (PR #614)
-* support ground reasoning with some bitvector primitives (PR #730, #733, #745).
-* improve the ite reasoning (simplification of some ites) (PR #731)
-
-### Build
-* update to the new version of ocplib-simplex (0.5)
-* remove the support of the deprecated library num. Alt-Ergo only uses Zarith (PR #600)
-* remove the deprecated graphical interface (PR #601)
-* use Dune >= 3.5.
-
-### Bug fixes
-* issue 475 (PR #478)
-* issue 504 (PR #524)
-* issue 509 (PR #524)
-* fix infinite loop in Expr.SMTPrinter (PR #540)
-* remove flattening of arithmetic expressions in order to fix some regressions (PR #566)
-* fix a leak memory in the vector module (PR #606)
-* issue 737 (PR #738)
-
 ## v2.4.3 (2023-04-14)
 
 ### Build
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,30 @@
 ## unreleased
 
+### New features
+* add context reinitialisation (PR #490)
+* add Dolmen frontend (PR #491,#541,#545)
+* add model generation (PR #530)
+* support mutually recursive definitions in the native language (PR #549, #550)
+* support of some options of the SMT-LIB statement (set-option) (PR #608)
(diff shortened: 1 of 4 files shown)


src/lib/frontend/models.ml
```diff
@@ -481,6 +481,11 @@ let pp_constant ppf (_, t) =
   Format.fprintf ppf "%a" SmtlibCounterExample.pp_dummy_value_of_type t
 
 let output_concrete_model fmt props ~functions ~constants ~arrays =
+  if ModelMap.(is_suspicious_model functions || is_suspicious_model constants
+               || is_suspicious_model arrays) then
+    Format.fprintf fmt "This model is a best-effort. Please note tht it \
+                        could be wrong. @.";
```
bclement-ocp: Not a big fan of the formulation as it seems to imply that when the message is not there we claim that the model is correct.

What about something more along the lines of "Note: this model includes symbols for which model generation is known to be incomplete." which is more precise and also doesn't promise anything.

(BTW the model needs to be parseable, so this should be a Lisp comment i.e. the line must start with `;`)

src/lib/structures/modelMap.mli
```diff
@@ -53,3 +54,7 @@ val fold : (key -> elt -> 'acc -> 'acc) -> t -> 'acc -> 'acc
 val empty : t
 
 val is_empty : t -> bool
+
+val is_suspicious_model : t -> bool
```
bclement-ocp: This is in a file called `modelMap`, it can just be called `is_suspicious`.

src/lib/structures/modelMap.ml
```diff
@@ -80,24 +80,49 @@ module V = Set.Make
...
+  suspicious : bool;
+}
 
-let iter f =
+let is_suspicious_name hs =
+  match Hstring.view hs with
+  | "@/" | "@%" | "@*" -> true
+  | _ -> false
+
+(* The model generation is not ready for FPA and Bitvector theories. *)
```
bclement-ocp: s/not ready/known to be incomplete

CHANGES.md
```diff
@@ -1,5 +1,30 @@
...
+
+### Bug fixes
+* issue 475 (PR #478)
+* issue 504 (PR #524)
+* issue 509 (PR #524)
+* fix infinite loop in Expr.SMTPrinter (PR #540)
+* remove flattening of arithmetic expressions in order to fix some regressions (PR #566)
+* fix a leak memory in the vector module (PR #606)
+* issue 737 (PR #738)
+
```
bclement-ocp: This has nothing to do in this PR

bclement-ocp (changes requested):

Halbaroth:
Done

src/lib/frontend/models.ml
```diff
@@ -481,6 +481,11 @@ let pp_constant ppf (_, t) =
   Format.fprintf ppf "%a" SmtlibCounterExample.pp_dummy_value_of_type t
 
 let output_concrete_model fmt props ~functions ~constants ~arrays =
+  if ModelMap.(is_suspicious functions || is_suspicious constants
+               || is_suspicious arrays) then
+    Format.fprintf fmt "; This model is a best-effort. It includes symbols
+        for which model generation is known to be incomplete. @.";
```
bclement-ocp, line 487: I don't think we should be flushing the printer here, `@,` would be preferrable, but I am not sure we are in a box here, so let's leave it that way for now.

src/lib/structures/modelMap.ml
bclement-ocp: Please don't resolve conversations asking for a change without making the corresponding change.

bclement-ocp (approved):

src/lib/structures/modelMap.ml
Halbaroth: Sorry, I thought I did it.

bclement-ocp:
> This PR fixes the issue #705.

Note that for this to work, the text `fix #705` or `fixes #705` must be in the *commit message*. I have manually closed #705 as fixed.

PR #754, opened 2023-07-26
Finalize release 2.5.0

Halbaroth:
This PR contains the new change log for the release 2.5.0.

PR #756, opened 2023-07-26
Fix links in odoc mld files

bclement-ocp:
Nominating for 2.5.0 @Halbaroth


Halbaroth:
Can I merge it on next before merging the PR #754 ?

bclement-ocp:
Sure, it's probably easier -- but as soon as the PR exists I think we need to be careful with what we merge or not
(actually now that I think of it is is an issue of the process as described, the state between the creation of the PR and the merge of the PR is a bit fuzzy in practice... We should probably say that the branch is created off from the PR commit *before* the merge, so that things that get into `next` while the branch PR is under review don't go into the release by mistake)

Halbaroth:
Ok, I will merge the release PR first.

Halbaroth (approved):

bclement-ocp:
Hm not sure I follow -- either we tag after the merge and we should merge this (and #757 I suppose) before the 2.5.0 PR, or we tag before the merge and it doesn't matter, I think?

Halbaroth:
My plan is to merge #754, then create the tag and the branch of the release and merge on next and backport on 2.5.x this PR after.

PR #759, opened 2023-07-27
Update model documentation

Halbaroth:
This PR has to be backported on the release 2.5.0.

### DIFF
--- a/docs/sphinx_docs/Usage/index.md
+++ b/docs/sphinx_docs/Usage/index.md
@@ -40,8 +40,8 @@ Model generation is disabled by default. There are two recommended ways to enabl
   on demand using the statement `(get-model)`.
 
   Alternatively, you can enable model generation using the statement
-  `(set-option :produce-models true)` and the options `--sat-solver tableaux`
-  and `--frontend dolmen`.
+  `(set-option :produce-models true)`. This currently requires using the options
+  `--sat-solver tableaux` and `--frontend dolmen`.
 
 #### Examples
 
@@ -121,7 +121,7 @@ Model generation is disabled by default. There are two recommended ways to enabl
    (get-model)
 
   ```
-  and the command `alt-ergo --frontend model --sat-solver tableaux INPUT.smt2` produces
+  and the command `alt-ergo --frontend dolmen --sat-solver tableaux INPUT.smt2` produces
   the output model
   ```
   unknown
@@ -131,9 +131,10 @@ Model generation is disabled by default. There are two recommended ways to enabl
     (define-fun c () Int 0)
   )
   ```
-  *Note*: you need to select the SAT solver Tableaux as the model generation is not
-  supported yet by the other SAT solvers. The options `--dump-models` and `--produce-models`
-  select the right SAT solver for you.
+  *Note*: you need to select the Dolmen frontend and the SAT solver Tableaux as the
+  model generation is not supported yet by the other SAT solvers. The options
+  `--dump-models` and `--produce-models` select the right frontend and SAT solver
+  for you.
 
 ### Output
 The results of an Alt-ergo's execution have the following form :
--- a/docs/sphinx_docs/Usage/index.md
+++ b/docs/sphinx_docs/Usage/index.md
@@ -24,22 +24,101 @@ Alt-Ergo supports file extensions:
 See the [Input section] for more information about the format of the input files
 
 ### Generating models
-Since 2.5.0, Alt-Ergo also generates models in the case it concludes on the satisfiability of
-the formula.
-There is two ways to activate model generation:
+Alt-Ergo generates best-effort models in the case it cannot conclude the unsatisfiability of
+the context or the validity of a goal. By default, the model format is the SMT format.
 
-- with the `--model` option;
+By default the model generation is disable. There is three ways to activate it:
 
-- `with the --interpretation=VALUE`, where VALUE can be equal to:
-  * "none", and alt-ergo will not generate models (by default);
-  * "first", and alt-ergo will output the first model it finds;
-  * "every", alt alt-ergo will compute a model before each decision
-  * "last", and alt-ergo will output the last model it computes before returning 'unknown'.
-  Note that this mode only works with the option `--sat-solver tableaux`.
+- with the `--dump-models` option, Alt-Ergo will produce a model after each `goal` in the native language
+  or each `(check-sat)` in the SMT-LIB language. For instance the input file `INPUT.smt2`:
 
-NB: the `--model` option is equivalent to `--interpretation every --sat-solver tableaux`.
+  ```
+    (set-logic ALL)
+    (declare-fun a () Int)
+    (declare-fun b () Int)
+    (declare-fun c () Int)
 
-The default model format is the SMT format.
+    (assert (= a (+ b c)))
+    (check-sat)
+
+    (assert (distinct a b))
+    (check-sat)
+  ```
+  and the command `$ alt-ergo --dump-models INPUT.smt2` produces the output models
+
+  ```
+    (
+      (define-fun a () Int 0)
+      (define-fun b () Int 0)
+      (define-fun c () Int 0)
+    )
+
+    (
+      (define-fun a () Int (- 2))
+      (define-fun b () Int 0)
+      (define-fun c () Int (- 2))
+    )
+  ```
+
+- with the `--produce-models` option, you turn the model generation on. You can display
+  a model after a `(check-sat)` using the statement `(get-model)`. For instance the input file `INPUT.smt2`:
+
+  ```
+    (set-logic ALL)
+    (declare-fun a () Int)
+    (declare-fun b () Int)
+    (declare-fun c () Int)
+
+    (assert (= a (+ b c)))
+    (check-sat)
+    (get-model)
+
+    (assert (distinct a b))
+    (check-sat)
+
+  ```
+  and the command `$ alt-ergo --produce-models INPUT.smt2` produces the output model
+  ```
+    (
+      (define-fun a () Int 0)
+      (define-fun b () Int 0)
+      (define-fun c () Int 0)
+    )
+  ```
+
+- with the statement `(set-option :produce-models true)`, you turn the model generation on and
+  you can use the statement `(get-model)` as explained above. For instance the input file `INPUT.smt2`:
+  ```
+   (set-logic ALL)
+   (set-option :produce-models true)
+   (declare-fun a () Int)
+   (declare-fun b () Int)
+   (declare-fun c () Int)
+
+   (assert (= a (+ b c)))
+   (check-sat)
+   (get-model)
+
+  ```
+  and the command `$ alt-ergo --sat-solver tableaux INPUT.smt2` produces the output model
+  ```
+    (
+      (define-fun a () Int 0)
+      (define-fun b () Int 0)
+      (define-fun c () Int 0)
+    )
+  ```
+  Note: you need to select the SAT solver Tableaux as the model generation is not
+  supported yet by the other SAT solvers.
+
+You can fine-tune the moment Alt-Ergo computes its models by using the option `--interpretation=VALUE` where VALUE can be:
+  * "none", to disable model generation (by default);
+  * "first", to output the first model found;
+  * "every", to compute a model before each decision;
+  * "last", to output the last model computed before returning 'unknown'.
+
+  Note: this option only works with the option `--sat-solver tableaux`.
+
 
 ### Output
 The results of an Alt-ergo's execution have the following form :
--- a/docs/sphinx_docs/Usage/index.md
+++ b/docs/sphinx_docs/Usage/index.md
@@ -29,7 +29,7 @@ the context. The model format is a SMT-LIB compatible format, even if you use th
 
 #### Activation
 
-Model generation is disabled by default. There are two recommanded ways to enable it:
+Model generation is disabled by default. There are two recommended ways to enable it:
 - with the native language and the `--dump-models` option, Alt-Ergo tries to produce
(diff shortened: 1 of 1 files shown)


docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,101 @@ Alt-Ergo supports file extensions:
 See the [Input section] for more information about the format of the input files
 
 ### Generating models
-Since 2.5.0, Alt-Ergo also generates models in the case it concludes on the satisfiability of
-the formula.
-There is two ways to activate model generation:
+Alt-Ergo generates best-effort models in the case it cannot conclude the unsatisfiability of
```
bclement-ocp:
```suggestion
Alt-Ergo can generate best-effort models in the case it cannot conclude the unsatisfiability of
```

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,101 @@ Alt-Ergo supports file extensions:
...
+the context or the validity of a goal. By default, the model format is the SMT format.
+
+By default the model generation is disable. There is three ways to activate it:
```
bclement-ocp:
```suggestion
Model generation is disabled by default. There are three ways to enable it:
```

bclement-ocp: We should list the three ways succinctly, then expand on more details for each of them. We also should provide clear guidance on which way is recommended -- I think that's `--dump-models` in the native language, and `--produce-models` for smtlib2.

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,101 @@ Alt-Ergo supports file extensions:
...
+
+- with the `--dump-models` option, Alt-Ergo will produce a model after each `goal` in the native language
```
bclement-ocp: `--dump-models` is mostly useful with the native language because SMT-LIB users are more likely to want `--produce-models` to enable the standard `(check-sat)` + `(get-model)` workflow, so the example should be in the native language maybe?

Also, `goal` is confusing in the context of model generation because the formula is negated, we should advertise `check_sat` rather than `goal` for model generation.

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,101 @@ Alt-Ergo supports file extensions:
...
+    )
+
+    (
+      (define-fun a () Int (- 2))
+      (define-fun b () Int 0)
+      (define-fun c () Int (- 2))
+    )
+  ```
+
+- with the `--produce-models` option, you turn the model generation on. You can display
```
bclement-ocp: I think `--produce-models` and `(set-option :produce-models true)` should be bundled together, because they do the same thing: enable support for `(get-model)`, so that's probably the way to explain it (turn it around from "Model generation can be enabled. Once enabled, you can use `(get-model)`." into "We support `(get-model)`. To use it enable model generation with `--produce-models` or `(set-option :produce-models true)`.")

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,101 @@ Alt-Ergo supports file extensions:
...
+    )
+  ```
+  Note: you need to select the SAT solver Tableaux as the model generation is not
+  supported yet by the other SAT solvers.
+
+You can fine-tune the moment Alt-Ergo computes its models by using the option `--interpretation=VALUE` where VALUE can be:
+  * "none", to disable model generation (by default);
+  * "first", to output the first model found;
+  * "every", to compute a model before each decision;
+  * "last", to output the last model computed before returning 'unknown'.
```
bclement-ocp: I'm not sure we want to document the `--interpretation` option, except maybe advertise `--interpretation every` for debugging? If we talk about it, we also should be clear that `--interpretation last` is what `--dump-models` and `--produce-models` do.

Can't comment on `--interpretation first` because I am not sure what it does...

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,101 @@ Alt-Ergo supports file extensions:
...
+  ```
+  and the command `$ alt-ergo --sat-solver tableaux INPUT.smt2` produces the output model
+  ```
+    (
+      (define-fun a () Int 0)
+      (define-fun b () Int 0)
+      (define-fun c () Int 0)
+    )
+  ```
+  Note: you need to select the SAT solver Tableaux as the model generation is not
```
bclement-ocp:
```suggestion
  *Note*: you need to select the SAT solver Tableaux as the model generation is not
```

Also we should make it clear that `--sat-solver Tableaux` is only necessary with `(set-option :produce-models true)`, it is not needed for `--produce-models` or `--dump-models` (it is implicit currently but we should make it explicit).

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,101 @@ Alt-Ergo supports file extensions:
 See the [Input section] for more information about the format of the input files
 
 ### Generating models
-Since 2.5.0, Alt-Ergo also generates models in the case it concludes on the satisfiability of
-the formula.
-There is two ways to activate model generation:
+Alt-Ergo generates best-effort models in the case it cannot conclude the unsatisfiability of
+the context or the validity of a goal. By default, the model format is the SMT format.
```
bclement-ocp: We should make it clearer that models are printed in a SMT-LIB compatible format, even if the input problem is in the native language.

Also, we should probably remove the "By default" part -- if we keep it, we should document how to use a different format, and also check that the different forma still works, because I don't think it does...

bclement-ocp (commented):
Some wording/organization suggestions.

Do we want to mention `--model-type constraints`?

docs/sphinx_docs/Usage/index.md
Halbaroth: In my opinion, the SMT-LIB format would be the only one to output models. I will ask to remove the other formats for the next release or we can do it now but if there is no documentation about it, we can keep the code in the 2.5.0.

docs/sphinx_docs/Usage/index.md
Halbaroth: I agree. Actually I hesitated to remove it.

Halbaroth:
> Some wording/organization suggestions.
> 
> Do we want to mention `--model-type constraints`?

We didn't test this feature so we shouldn't mention it.

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,106 @@ Alt-Ergo supports file extensions:
...
-
-NB: the `--model` option is equivalent to `--interpretation every --sat-solver tableaux`.
-
-The default model format is the SMT format.
+Alt-Ergo can generates best-effort models in the case it cannot conclude the unsatisfiability of
+the context. The model format is a SMT-LIB compatible format, even if you use the native input language.
+
+#### Activation
+
+Model generation is disabled by default. There are two recommanded ways to enable it:
```
bclement-ocp:
```suggestion
Model generation is disabled by default. There are two recommended ways to enable it:
```

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,106 @@ Alt-Ergo supports file extensions:
...
+  a model after each `check_sat` that returns `I don't known` or
+  a counter-example after each `goal` it cannot prove `valid`. Note that both
+  `goal` and `check_sat` statements are independent in the native language.
+
+- with the SMT-LIB language and the `--produce-models` option, Alt-Ergo tries to
+  produce a model after each `(check-sat)` that returns `unknown`. Models are output
+  on demand using the statement `(get-model)`.
+
+  Alternatively, you can enable model generation using the statement
+  `(set-option :produce-models true)`.
```
bclement-ocp: Here we need to mention that `--sat-solver Tableaux` and `--frontend dolmen` must be used

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,106 @@ Alt-Ergo supports file extensions:
...
+      (define-fun a () Int 2)
+      (define-fun b () Int 0)
+      (define-fun c () Int 0)
+    )
+  ```
+  *Note*: In this example the model for the statement `check_sat c2` is not a
+  model for the statement `check_sat c1` since `check_sat` are independent in
+  the native language. The same goes for `goals`.
+
+  Using the SMT-LIB language in the input file `INPUT.ae`:
```
bclement-ocp:
```suggestion
  Using the SMT-LIB language in the input file `INPUT.smt2`:
```

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,106 @@ Alt-Ergo supports file extensions:
...
+    (check-sat)
+
+  ```
+  and the command `$ alt-ergo --produce-models INPUT.smt2` produces the output model
+  ```
+    (
+      (define-fun a () Int 0)
+      (define-fun b () Int 0)
+      (define-fun c () Int 0)
+    )
```
bclement-ocp: This is slightly confusing because when I run the command given by the doc I get:

```smt-lib
unknown
(
  (define-fun a () Int 0)
  (define-fun b () Int 0)
  (define-fun c () Int 0)
)
unknown
```

We should include the `unknown` results so that it's less confusing. Also did you intend to add a second `(get-model)` after the second `(check-sat)`?

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,106 @@ Alt-Ergo supports file extensions:
...
+
+    (assert (= a (+ b c)))
+    (check-sat)
+    (get-model)
+
+    (assert (distinct a b))
+    (check-sat)
+
+  ```
+  and the command `$ alt-ergo --produce-models INPUT.smt2` produces the output model
```
bclement-ocp:
```suggestion
  and the command `alt-ergo --produce-models INPUT.smt2` produces the output model
```

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,106 @@ Alt-Ergo supports file extensions:
...
+  Using the native language in the input file `INPUT.ae`:
+
+  ```
+    logic a, b, c : int
+    axiom A : a <> c
+
+    check_sat c1: a = b + c
+    check_sat c2: a <> b
+  ```
+  and the command `$ alt-ergo --dump-models INPUT.ae`, Alt-Ergo produces the
```
bclement-ocp:
```suggestion
  and the command `alt-ergo --dump-models INPUT.ae`, Alt-Ergo produces the
```

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,106 @@ Alt-Ergo supports file extensions:
...
+   (declare-fun a () Int)
+   (declare-fun b () Int)
+   (declare-fun c () Int)
+
+   (assert (= a (+ b c)))
+   (check-sat)
+   (get-model)
+
+  ```
+  and the command `$ alt-ergo --sat-solver tableaux INPUT.smt2` produces the output model
```
bclement-ocp:
```suggestion
  and the command `alt-ergo --sat-solver tableaux --frontend dolmen INPUT.smt2` produces the output model
```

Needs `--frontend dolmen` here

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,106 @@ Alt-Ergo supports file extensions:
...
+  ```
+    (
+      (define-fun a () Int 0)
+      (define-fun b () Int 0)
+      (define-fun c () Int 0)
+    )
+  ```
+  *Note*: you need to select the SAT solver Tableaux as the model generation is not
+  supported yet by the other SAT solvers. The options `--dump-models` and `--produce-models`
+  select the right SAT solver for you.
```
bclement-ocp: Also needs `--frontend dolmen` (which `--produce-models` uses, but not `--dump-models`).

bclement-ocp (changes requested):

docs/sphinx_docs/Usage/index.md
Halbaroth: Ok. I didn't add a `(get-model)` after the second `(check-sat)` as an illustration of the fact models are displayed on demand in this mode.

bclement-ocp: Ah, OK, I thought it was intended as an illustration of the fact that the `a = b + c` constraint was still satisfied in the 2nd model

Halbaroth: I will add an explanation.

docs/sphinx_docs/Usage/index.md
```diff
@@ -24,22 +24,116 @@ Alt-Ergo supports file extensions:
...
+  a counter-example after each `goal` it cannot prove `valid`. Note that both
+  `goal` and `check_sat` statements are independent in the native language.
+
+- with the SMT-LIB language and the `--produce-models` option, Alt-Ergo tries to
+  produce a model after each `(check-sat)` that returns `unknown`. Models are output
+  on demand using the statement `(get-model)`.
+
+  Alternatively, you can enable model generation using the statement
+  `(set-option :produce-models true)` and the options `--sat-solver tableaux`
+  and `--frontend dolmen`.
```
bclement-ocp:
```suggestion
  `(set-option :produce-models true)`. This currently requires using the options `--sat-solver tableaux`
  and `--frontend dolmen`.
```

docs/sphinx_docs/Usage/index.md
bclement-ocp: This is still relevant, the text should mention `--frontend dolmen`.

bclement-ocp (changes requested):

bclement-ocp (approved):

PR #760, opened 2023-07-27
Update dependencies

Halbaroth:
This PR updates the dependencies and the relative documentation for the release 2.5.0.
It will be backported on the branch 2.5.x.

### DIFF
--- a/alt-ergo-lib.opam
+++ b/alt-ergo-lib.opam
@@ -17,9 +17,9 @@ depends: [
   "ocaml" {>= "4.08.0"}
   "dune" {>= "3.0"}
   "dune-build-info"
-  "dolmen" {>= "0.8.1"}
-  "dolmen_type" {>= "0.8.1"}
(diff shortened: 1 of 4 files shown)


bclement-ocp (approved):
Oh right, good catch. We must remove the pin of dolmen to its master branch in `alt-ergo-lib.opam.template` as well.

Halbaroth:
> Oh right, good catch. We must remove the pin of dolmen to its master branch in `alt-ergo-lib.opam.template` as well.

Should we remove the pin on next? It quite useful to use the dev version of Dolmen on dev branch. If you agree, I will revert the last commit I just pushed and put it only on 2.5.x.

bclement-ocp:
I don't think we should have the pin on `next`, it's better that we test against released versions of our dependencies when possible. If we need to depend on unreleased Dolmen features again we will add the pin back.

PR #769, opened 2023-07-28
Add blocks for the notes

Halbaroth:
This PR uses the block of Sphinx to emphasize notes in the documentation.
It also includes minor fixes.

### DIFF
--- a/docs/sphinx_docs/Usage/index.md
+++ b/docs/sphinx_docs/Usage/index.md
@@ -56,7 +56,7 @@ Model generation is disabled by default. There are two recommended ways to enabl
 
 #### Examples
 
-  Using the native language in the input file `INPUT.ae`:
+  - Using the native language in the input file `INPUT.ae`:
 
   ```
     logic a, b, c : int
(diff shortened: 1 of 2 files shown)


docs/sphinx_docs/Usage/index.md
```diff
@@ -85,11 +85,16 @@ Model generation is disabled by default. There are two recommended ways to enabl
...
-  *Note*: In this example the model for the statement `check_sat c2` is not a
+
+  ```{admonition} Note
+
+  In this example the model for the statement `check_sat c2` is not a
   model for the statement `check_sat c1` since `check_sat` are independent in
   the native language. The same goes for `goals`.
 
-  Using the SMT-LIB language in the input file `INPUT.smt2`:
+  ```
```
bclement-ocp, line 95: I don't think this one should be an admonition, it is more a remark that doesn't deserve specific emphasis imo; it would be fine to keep it in the main content.

docs/sphinx_docs/Usage/index.md
```diff
@@ -116,10 +121,15 @@ Model generation is disabled by default. There are two recommended ways to enabl
 
     unknown
   ```
-  *Note*: There is no model printed after the second `(check-sat)` since we
+
+  ```{admonition} Note
+
+  There is no model printed after the second `(check-sat)` since we
   don't demand it with the statement `(get-model)`.
+  ```
```
bclement-ocp, line 129: Ditto, I think having this one as an admonition is more distracting than anything

bclement-ocp (approved):
Not 100% sure all of these deserve the extra emphasis but we can also revisit later if needed

PR #771, opened 2023-07-28
Ensure that definitions have the same ordering with dolmen

bclement-ocp:
The dolmen frontend is adding declarations to the end of the command queue instead of to the front, causing the final command list to be all shuffled. This is the cause of at least some of the regressions with the dolmen frontend, since the solver now processes things in a different order.

This patch ensures that the definitions are added with the correct ordering, so that the command queue should be the same with both frontends, barring additional bugs. This was tested with `-d commands` on a few benchmarks locally, where it produced the exact same output with both frontends.

The patch also refactors slightly the [handle_smtt] method in the solving loop: as part of the debugging for this, I realized that we can never end up with `Hyp` constructors with local/global hypotheses names there, because one difference between the Dolmen and legacy frontend is that the legacy frontend extracts the hypotheses from the goal during typechecking (so prior to cnf-conversion) while the Dolmen frontend extracts the hypotheses from the goal during cnf-conversion (i.e. when we encounter a `Solve`).

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -451,21 +451,17 @@ let main () =
               DStd.Loc.fmt loc
         in
         let stmt = { Typer_Pipe.id; contents; loc ; attrs } in
-        let rev_cnf =
-          D_cnf.make (State.get State.logic_file st).loc l stmt
+        let cnf, is_thm =
+          match D_cnf.make (State.get State.logic_file st).loc l stmt with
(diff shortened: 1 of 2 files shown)


bclement-ocp:
Probably worth having this for the first release of Dolmen in 2.5.0 @Halbaroth 

We will need to rerun benches but we probably will need to do that anyways due to the perf regressions

bclement-ocp:
(This fixes the two dolmen regressions on AE-format-fpa)

Halbaroth (approved):

PR #780, opened 2023-08-16
Use --enable-theories fpa to active the FPA builtin theory

Halbaroth:
This PR disables the fpa theories to prevent some regressions. We can active them using the option `--enable-theories fpa` or `--enable-theories fpa,nra,ria`. The theories `nra` and `ria` are enabled by default if you enable `fpa` and you got an error if you try to disable `nra` or `ria` after enabling `fpa`. For instance : 
```
alt-ergo --enable-theories fpa --disable-theories nra ...
```
produces an error now.

The old option `--use-fpa --prelude fpa-theory-2019-10-08-19h00.ae` still works and the prelude is loaded once.  

I also add an new category for non-regression tests in order to enable the fpa preludes.

We have to backport it on the branch 2.5.x.

~~I decide to restore the option instead of creating another as the combination `--use-fpa --prelude --prelude fpa-theory-2019-10-08-19h00.ae` still works (with a warning message).~~


bclement-ocp:
As I explained several times, we should not call the option `--use-fpa`, because we want old command lines to keep working as before.

If we reuse the same name for the new behavior, people that had command lines with `--use-fpa --prelude fpa-theory-XXX.ae` (in particular, all why3 users) will get the floating point prelude loaded *twice*. Since we are temporarily disabling the fpa prelude by default due to performance regressions, it would mean that all existing uses of the floating point prelude would be automatically affected by the performance regressions. Please use a different option name.

Halbaroth:
uhm maybe we should prevent to reload it a second time? Otherwise, we should create a new option to deprecate it in the next release and remove it in 2.7.x.

bclement-ocp:
I don't think we have an easy way to robustly avoid the duplication. Basically we need to either:

 - Detect that the user loads the shipped `fpa-theory` prelude, but detecting that based on the file name/file path rather than the file contents would lead to confusing behavior and seems brittle,
 - Or deduplicate the theory axioms, which we don't really have the infrastructure for

I think that using a new option is preferable. Maybe we can just make that option `--enable-theory fpa-full` or something? Or even make `--enable-theory fpa` automatically pull in the ria and nra deps, since the fpa theory is not very useful without them afaik.

Halbaroth:
We can just do that:
```ocaml
let set_preludes p =
  preludes := List.filter (fun prelude ->
    match prelude with
    | "fpa-theory-2017-01-04-16h00.ae"
    | "fpa-theory-2019-06-14-11h00.ae"
    | "fpa-theory-2019-10-08-19h00.ae" -> false
    | _ -> true
  ) p
```
There is no chance someone names is prelude as one of the fpa-theory and modify the warning when someone uses `--prelude fpa-theory` to explain this option as no effect at all.

We can check why3 doesn't rename the prelude. I found the commands in Why3:
```
command = "%e --use-fpa --prelude=fpa-theory-2019-06-14-11h00.ae  --timelimit %.t %f"
command_steps = "%e --use-fpa --prelude=fpa-theory-2019-06-14-11h00.ae --steps-bound=%S %f"
```

We can also use a checksum to detect the fpa prelude but I think it is overkill.

Halbaroth:
> Or even make `--enable-theory fpa` automatically pull in the ria and nra deps, since the fpa theory is not very useful without them afaik.

This option is preferable since we don't plan to deprecate `--enable-theory fpa`.

bclement-ocp:
The three preludes are different and define different symbols / axioms. Proofs may pass with `fpa-theory-2017-01-04-16h00.ae` that do not work with `fpa-theory-2019-10-08-19h00.ae`, and so that would break.

We can somewhat safely do this, but only with `fpa-theory-2019-10-08-19h00.ae`, and people using the other preludes will get the performance regression (and possibly weird things will happen because of some axioms being slightly different in both cases, but probably not).

If we do this, we should also make sure to only remove the preludes if the `--use-fpa` option is given, otherwise the user will end up with no prelude at all.

It just seems needlessly complex compared to adding a new option that does what we want (or simply telling people to use `--enable-theories fpa,ria,nra` which would also be a forward-compatible option once we make it the default, despite the verbosity).

Halbaroth:
I prefer the flag `--enable-theory fpa`.

docs/sphinx_docs/Input_file_formats/Native/05_theories.md
```diff
@@ -44,8 +44,9 @@ precisely, Alt-Ergo implements the second and third layers from the paper "[A
 Three-tier Strategy for Reasoning about Floating-Point Numbers in
 SMT](https://inria.hal.science/hal-01522770)" by Conchon et al.
 
-*Note*: Support for floating-point arithmetic is enabled by default in
-Alt-Ergo since version 2.5.0.  Previous versions required the use of command
+*Note*: Prelude for floating-point arithmetic is a built-in theory in
+Alt-Ergo since version 2.5.0. Use `--enable-theories fpa` to enable it.
+Previous versions required the use of command
```
bclement-ocp: "Prelude for floating-point arithmetic is a built-in theory" doesn't really make sense as a sentence.

What about something like: "Support for floating-point arithmetic is available as a built-in theory since version 2.5.0. It be enabled using flag `--enable-theory fpa`, which will become the default in a future release. Previous versions used the external prelude mechanism and required command line flags `--use-fpa` and `--prelude fpa-theory-2019-10-08-19h00.ae`."

src/bin/common/parse_command.ml
```diff
@@ -1325,9 +1325,11 @@ let parse_theory_opt =
...
 
   let use_fpa =
-    let doc = "Floating-point builtins are always enabled and this option has
-    no effect anymore. It will be removed in a future version." in
-    let deprecated = "this option is always enabled" in
+    let doc =
+      "Enable Floating-point builtins. This option requires to load an fpa \
+       prelude with `--prelude`. It will be removed in a future version."
+    in
+    let deprecated = "Please use the new option `--enable-theories fpa" in
```
bclement-ocp: I think this change is no longer relevant?

tests/dune
```diff
@@ -1,3 +0,0 @@
-(cram
```
bclement-ocp: Please don't remove this file.

src/bin/common/parse_command.ml
bclement-ocp: On the other hand, the warning message when a "fpa-theory" prelude is loaded should be changed to mention `--enable-theory fpa`, I think right now it still says "this is no longer needed".

Halbaroth: You are right. I thought we still need `use_fpa` but the fpa primitives are now builtins.

tests/dune
Halbaroth: Oh I don't now what's happened! Good catch ;)

Halbaroth: In fact I'm used to remove all the dune files in `tests/` when `gentest` has produced invalid rules. I didn't notice there is a versioning dune file in this folder now.

docs/sphinx_docs/Input_file_formats/Native/05_theories.md
```diff
@@ -44,10 +44,11 @@ precisely, Alt-Ergo implements the second and third layers from the paper "[A
 Three-tier Strategy for Reasoning about Floating-Point Numbers in
 SMT](https://inria.hal.science/hal-01522770)" by Conchon et al.
 
-*Note*: Support for floating-point arithmetic is enabled by default in
-Alt-Ergo since version 2.5.0.  Previous versions required the use of command
-line flags `--use-fpa` and `--prelude fpa-theory-2019-10-08-19h00.ae` to enable
-it.
+*Note*: Support for floating-point arithmetic is available as a built-in theory
+since version 2.5.0. It be enabled using flag `--enable-theory fpa`, which will
```
bclement-ocp:
```suggestion
since version 2.5.0. It can be enabled using flag `--enable-theory fpa`, which will
```

(or "It is")

bclement-ocp (commented):
I think the warning message for `fpa-theory` preludes still needs to be updated, unless I missed the change?

tests/dune
bclement-ocp: I was going to say that it's not needed since #729 but actually we didn't merge that one because of a CI issue! I will take a look

bclement-ocp (approved):
LGTM but please apply the modification for "It be enabled" before merging.

PR #782, opened 2023-08-18
Simplify output channels

Halbaroth:
This PR simplifies the output CLI by removing the `warning`, `debug` and `unsat core` channels.

### DIFF
--- a/src/bin/common/parse_command.ml
+++ b/src/bin/common/parse_command.ml
@@ -1438,7 +1438,7 @@ let parse_theory_opt =
 
 let parse_fmt_opt =
   let docs = s_fmt in
-  let docv = "CHANNEL" in
+  let docv = "OUTPUT" in
 
   let regular_output =
     let doc =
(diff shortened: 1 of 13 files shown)


bclement-ocp (approved):
`std` and `err` are standard OCaml names for these so I'd keep `get_fmt_err` and `get_fmt_std` but otherwise LGTM.

This will be for 2.6.0 so we will have had one release with deprecated `--std-formatter` and `--err-formatter` so I'm fine with removing them -- it'd cause less churn for users to keep them as aliases, but I don't think these options are used much anyways.

Halbaroth:
> `std` and `err` are standard OCaml names for these so I'd keep `get_fmt_err` and `get_fmt_std` but otherwise LGTM.

I think we shouldn't keep `get_fmt_err` and `get_fmt_std`. Indeed, the formatter can be something else. For instance, the user can retrieve the `diagnostic output` into a file. In my opinion `get_fmt_err` and `get_fmt_std` mean the standard output and standard error only.

bclement-ocp:
Right, the regular and diagnostic channels can be redirected independently of the standard and error outputs (that can also be redirected) and so it could get confusing. Let's keep the names as is

Issue #778, opened 2023-08-10
Wrong model in the non-regression tests


Halbaroth:
Alt-Ergo outputs a wrong model on this SMTLIB input:
```
(set-logic QF_UF)
(set-option :produce-models true) ; enable model generation
(declare-const p Bool)
(declare-const q Bool)
(assert (=> (not p) (not q)))
(check-sat)
(get-model)
```
The output:
```
unknown
(
)
```
and the command : `dune exec -- alt-ergo --frontend dolmen --produce-models`

bclement-ocp:
This is a dupe of #777, no? We probably should have a single issue for model unsoundness (although in this case the model is not unsound, it is incomplete).

Should we postpone these to 2.6.0? Although these are unfortunate, we don't claim anything on the soundness (or absence thereof) of models in 2.5.0, and the release is already way past due.

Halbaroth:
I don't think so. I guess AE failed to produce a complete model because functions returning boolean values are converted into predicates.

bclement-ocp:
I am not saying the cause is the same in both cases, but that I think it would make more sense to consolidate all the incorrect models into a single issue rather than having one issue for each possible cause. Since the outcomes are related, it would be easier to have a single place to go to for all the info/insight/etc. we have on this rather than have it be scattered.

Halbaroth:
After investigating more, I think I found the origin of the bug. The problem comes from case-splits. Consider this very simple problem:
``` 
(set-logic QF_UF)
(set-option :produce-models true) ; enable model generation
(declare-const p Bool)
(declare-const q Bool)
(assert (or p q))
(check-sat)
(get-model)
``` 
Alt-Ergo produces the partial model `{ p = true }` but we expect something as `{ p = true; q = whatever }`. While reasoning on `p`, AE found that `true` satisfies this formula and it didn't explore at all the other branch, which means it didn't add the `q` in the UF and AE ignores that the term `q` exists during the model generation phase.

If we permute p an q in the above problem, we got the partial model `{ q = true }` and replacing the `or` gate by an `and` gate produces a complete model.

Halbaroth:
I wrote a fix for this issue that is not completely satisfactory. The idea is really simple:
- The frontend has to remember which terms have been declared by the user. Notice that a term cannot be both declared and defined in the SMT standard or the native language. I create a separate hash table to collect declared terms in `d_cnf`.
- I converted this table into a list and feeds `output_concrete_model` with it.
[...]

PR #789, opened 2023-08-24
Use post-solve SAT environment for model generation

### LINKED ISSUE #777: Wrong model on elementary integer problem
Alt-Ergo produces a wrong model for this problem:
```
(set-logic ALL)
(declare-const i Int)
(define-fun C ((i Int)) Bool (> i 0))
(assert (C i))
(check-sat)
(get-model)
```
outputs
```
unknown
(
  (define-fun i () Int 0)
)
```
But the slightly different input: 
```
(set-logic ALL)
(declare-const i Int)
(assert (> i 0))
(check-sat)
(get-model)

```
produces a correct model:
```
unknown
(
  (define-fun i () Int 2)
)
```

I used the command `dune exec -- alt-ergo --produce-model --frontend dolmen`.

Halbaroth:
This PR fixes both issue #777 and partially #778.

The fix is only one character! The accumulator returns by `process_decl` in the module `Frontend` should contain the latest 
`pseudo-consistent` environment of the SAT solver. By pseudo-consistent, I mean that AE didn't manage to prove the context is unsatisfiable. Thus, we have to return the new environment even if the SAT returns `unknown`.

Nominating for 2.5.0. 

Note: the CI failed because the PR #784 is not merged yet.

### DIFF
--- a/examples/dune
+++ b/examples/dune
@@ -1,0 +1,16 @@
+(executable
+ (name Lib_usage)
+ (libraries AltErgoLib)
+ (modules Lib_usage)
+)
+
+(rule
+ (alias runtest)
+ (action
+  (ignore-stdout
+   (ignore-stderr
+    (run ./Lib_usage.exe)
+   )
+  )
+ )
+)
--- a/examples/lib_usage.ml
+++ b/examples/lib_usage.ml
@@ -65,7 +65,7 @@ open AltErgoLib
 
 module PA = Parsed_interface
 
-let x = PA.mk_var_type Loc.dummy "'a"
+let _x = PA.mk_var_type Loc.dummy "'a"
 
 let one = PA.mk_int_const Loc.dummy "1"
 let two = PA.mk_int_const Loc.dummy "2"
@@ -80,7 +80,7 @@ let goal_3 = PA.mk_goal Loc.dummy "toy_3" (PA.mk_not Loc.dummy eq1)
 
 let parsed = [goal_1; goal_2; goal_3]
 
-let typed, env = Typechecker.type_file parsed
+let typed, _env = Typechecker.type_file parsed
 
 let pbs = Typechecker.split_goals_and_cnf typed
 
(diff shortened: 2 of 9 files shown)


bclement-ocp:
> The fix is only one character! The accumulator returns by `process_decl` in the module `Frontend` should contain the latest
> `pseudo-consistent` environment of the SAT solver. By pseudo-consistent, I mean that AE didn't manage to prove the context is unsatisfiable. Thus, we have to return the new environment even if the SAT returns `unknown`.

A `Query` is not expected to change the environment. With this change, it would be unsound to reuse that environment for another query, such as if we ever call `solve` with a list that contains multiple queries.

Although my understanding is that this can currently not happen in the main solver (we only ever have a single query at the end of a command sequence, and always restart the search from an empty environment), it is a bomb ready to blow up on us at any time. In particular, `process_decl` is exposed to library users and this looks like a major breaking change, so I would be quite uncomfortable shipping this.

The suggestion in the comments sounds like a better fix:

> it'd be better to accumulate in `consistent` a 3-case adt and not a simple bool

This ADT should be isomorphic to `Sat of sat_env | I_dont_know of sat_env | Unsat`.

bclement-ocp:
Also can you please change the name of the PR to something more descriptive, such as "Use post-solve environment for model generation" or the like? The current name ("Fixing 777") requires multiple (mental) indirection steps to figure out what the PR actually is about.

Halbaroth:
> The suggestion in the comments sounds like a better fix:
> 
> > it'd be better to accumulate in `consistent` a 3-case adt and not a simple bool
> 
> This ADT should be isomorphic to `Sat of sat_env | I_dont_know of sat_env | Unsat`.

This is done in OptimAE. This PR was just a fix for the release 2.5.0.

Halbaroth:
I create a new ADT to save the post-status of the environment after each command. A better design consists in moving the printing feature of the type `status` in the binary and use the ADT status to accumulate the post-status of the environment. But this part of code changes a lot in OptimAE. I will write another PR after OptimAE to simplify this part of code.

bclement-ocp (commented):
Changes to the frontend look good but aren't the corresponding changes to the `solving_loop` missing?

bclement-ocp:
> This is done in OptimAE. This PR was just a fix for the release 2.5.0.

That makes sense and should have been explained in the PR. I am still uneasy about doing this due to `parse_decl` silently breaking for library users, so having a fix that changes the type (if it's a temporary hotfix, using `sat_env option` is fine) is better.

bclement-ocp (approved):
LGTM. I believe `examples/lib_usage.ml` and `src/bin/js/worker_js.ml` also need to be updated.

Halbaroth:
I fixed both Lib_usage and Worker_js. I also add a test to compile and execute `lib_usage` while running `make runtest`.

PR #790, opened 2023-08-24
Check ground models with Dolmen

Halbaroth:
Please do not merge this PR on `next` before #789.


Halbaroth:
As I explained during the last dev meeting, we cannot check models of problems involving custom builtin functions like `ae.round` because there is no mechanism in Dolmen's binary to register custom definitions: https://github.com/Gbury/dolmen/issues/203

But I believe we can do it by adding definitions in the input problems. I think this isn't very important and I'll finish this PR without
supporting this annoying case. In these few cases, the model will be check by `get-value` anyway ;)

PR #776, opened 2023-08-09
[WIP] OptimAE

Halbaroth:
This PR is mostly a rebase of the branch `optims`: https://github.com/OCamlPro/alt-ergo/compare/next...optims.
All the commits between e392d5b071f0414b78ea552693dc4b8bec9ed46a and c4b336c0cba5383275808fe215dde00cca1feb2f have been merged by the PR #530 

- I prevent `ocp-indent` to reindent the mly files. It seems there is a unfixed bug and the parser was a mess after running ocp-indent on it. See https://github.com/OCamlPro/ocp-indent/issues/218

Notice that I also ignored these commits: 
- c4b336c0cba5383275808fe215dde00cca1feb2f
- 3f01bf60eb0178964c83d9ee2f55bf4622c82782 (old code)
- 3c52d274bfb7247feb2f13a8d780160791dda75a (old code)
- fca77c9f6fc883794be0ffadef6c82442a1f282e (old code)
- 57d4af580b2edef18ab1554b071d3a11b7fd37a9 (old code)
- 78811d80b05e3f33d34de8c012dca5b1a77ca51d (already applied)
- 90bb216dca1146c9f3b3f9f140959a83f6c1e4e3 (already applied)
- 6405133eac1780a011809c73a5f51a4687a2c2bb (already applied)
- e8e6c2d3656888739aa04417956b24787cbccaa0 (old code)
- cb2e74aeb291bef28c4a5e9b55c0cf049d98ee6c (poetry)
- f7c655d2350625610d85384aaa1a51bfa4d9faea (old code)
- e87fb9d84270fe4edd50c4bc01a49daa9ee3b984 (already applied)
- 202c7684c4396cf1713b0ca722e1707fbc76a426 (already applied)
- b534dc957daa877ae2dabfcc68ceda80982b7eac (already applied)
- 6b344b6f4956c41998cf73c017209a1d5e279270 (already applied)
- ce50c902266fcfd3d2cdcdfdfe6c0f713e578319 (already applied)
- 71e7feeaa171c15e1d55d00baf74012b934c423a (already applied)
- 0a1e3307fb81d3d401be7ef980445b44d4506131 (already applied)
- 2523faef0ca35b52c73fd7c6f3070083da0bffb3 (already removed)
- 86876467ca9f71099a66ba534c39445c467a932c (already applied)
- e541e5ccf510bd444c3a82e9d1a6222f0825cbf1 (already applied)
[...]

PR #792, opened 2023-09-04
Fix model generation for records

Halbaroth:
Before this PR, it seems the model generation for records didn't work on `next`. The reason is simple. If we want to produce a value for a declared record, we need to accumulate all the accesses to this record. For instance, the following input file:
```
(set-logic ALL)
(declare-datatype Pair
((pair (first Int) (second Int))))
(declare-const a Pair)
(assert (= (first a) 5))
(check-sat)
(get-model)
```
didn't produce a correct model. It assigned `0` for any field of type `Int`.

Note: for functional arrays, we can use the `ModelMap.t` structure to accumulate the `store` instructions as functional arrays can be seen as a map. We cannot do the same thing for records since each destructor of the record can return a different type.

### DIFF
--- a/src/lib/frontend/models.ml
+++ b/src/lib/frontend/models.ml
@@ -261,29 +261,33 @@ module SmtlibCounterExample = struct
     MS.add record_name destrs records
 
   let mk_records_constr records record_name
-      { Ty.name = _n; record_constr = cstr; lbs = lbs; _} =
+      { Ty.record_constr = cstr; lbs; _ } =
     let find_destrs destr destrs =
       try let rep = MS.find destr destrs in
         Some rep
       with Not_found -> None
     in
 
-    let print_destr fmt (destrs,lbs) =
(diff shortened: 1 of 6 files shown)


bclement-ocp:
Hmm I believe this should be dealt with by purification? In fact if you run your example with `-d uf` you get:

```
a --> {first = 5; second = 0}
```

in the representative maps, so this is probably either a printing issue or the specialized code for records in model generation. I have seen that code do dubious things in the past (I didn't look into it too much because my understanding is that records in models are not officially supported).

bclement-ocp:
Ah, I did not see this was a PR and not an issue, sorry. I think the fix works, but it would be better to exploit the information we have from the semantic values instead (in contrast, functional arrays don't have semantic values).

bclement-ocp:
(Discussed offline; #776 changes this code quite a bit so we are going for the simplest possible fix in #793 )

PR #793, opened 2023-09-05
Restore record printing functionality

bclement-ocp:
This got lost somewhere along the way as the conditions for model generation changed.

Note that it is a bit weird that the accumulation of record accesses happens in `models.ml` while the accumulation of array accesses happens in `uf.ml`.  We should pick a single one for both (probably in `models` but even better would be to expose the appropriate APIs so that each theory can build its own model values), but for now this patch should wrangle the `if`s properly and restore the existing functionality.

This is an alternate version of #792 that uses the existing functionality in `models.ml` to collect record field information.

### DIFF
--- a/src/lib/frontend/models.ml
+++ b/src/lib/frontend/models.ml
@@ -416,7 +416,13 @@ module SmtlibCounterExample = struct
            else
              reps_aux representants
          in
-         print_fun_def fmt f xs_ty_named ty rep;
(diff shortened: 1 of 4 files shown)


bclement-ocp:
Test failures seems to be due to GitHub issues (errors downloading some stuff from GH); re-running

Halbaroth (approved):

PR #795, opened 2023-09-06
[WIP] Removing exceptions from the Sat solver API

Stevendeo:
This PR removes the Sat, Unsat and I_dont_know exceptions raised by the library. It now returns an ADT which is either Done, Sat, Unsat or I_dont_know.
Alt-ergo has the exact same behavior than before, but it is not a bit cleaner. 

This PR also removes a `raise Exit` lost in the middle of the library.

### DIFF
--- a/src/lib/frontend/frontend.ml
+++ b/src/lib/frontend/frontend.ml
@@ -383,16 +383,17 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
             print_status (Unsat (d, dep)) (Steps.get_steps ());
             (env, `Unsat, dep)
           | `Sat t ->
-            (* This case should mainly occur when a query has a non-unsat result,
-               so we want to print the status in this case. *)
+            (* This case should mainly occur when a query has a non-unsat
+               result, so we want to print the status in this case. *)
             print_status (Sat (d,t)) (Steps.get_steps ());
             res
           | `Unknown t ->
             (* In this case, it's not clear whether we want to print the status.
-               Instead, it'd be better to accumulate in `consistent` a 3-case adt
-               and not a simple bool. *)
+               Instead, it'd be better to accumulate in `consistent` a 3-case
+               adt and not a simple bool. *)
             print_status (Unknown (d, t)) (Steps.get_steps ());
-            (*if get_model () then SAT.print_model ~header:true (get_fmt_mdl ()) t;*)
+            (* if get_model () then
+                 SAT.print_model ~header:true (get_fmt_mdl ()) t; *)
             res
         end
       | ThAssume th_elt ->
--- a/src/lib/frontend/frontend.ml
+++ b/src/lib/frontend/frontend.ml
@@ -120,7 +120,7 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
       let env =
         List.fold_left
           (fun env f ->
-             match
+             process_sat_res @@
                SAT.assume env
                  {E.ff=f;
                   origin_name = "";
@@ -135,11 +135,6 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
                   from_terms = [];
                   theory_elim = true;
                  } Ex.empty
-             with
-             | Done e -> e
-             | Sat t -> raise (SAT t)
-             | Unsat e -> raise (UNSAT e)
-             | I_dont_know t -> raise (I_DONT_KNOW t)
           ) (SAT.empty ()) pb
       in
       ignore (SAT.unsat
--- a/src/lib/frontend/frontend.ml
+++ b/src/lib/frontend/frontend.ml
@@ -87,6 +87,18 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
     | Timeout of Commands.sat_tdecl option
     | Preprocess
 
+  exception SAT of SAT.t
+  exception UNSAT of Explanation.t
+  exception I_DONT_KNOW of SAT.t
+
+  (** Returns the environment if the API call was successful
+      (i.e., returned 'Done'). Otherwise, raises an exceptio*)
+  let process_sat_res : SAT.res -> SAT.t = function
+    | Done t -> t
+    | Sat t -> raise (SAT t)
+    | Unsat e -> raise (UNSAT e)
+    | I_dont_know t -> raise (I_DONT_KNOW t)
+
   let output_used_context g_name dep =
     if not (Options.get_js_mode ()) then begin
       let f = Options.get_used_context_file () in
@@ -108,20 +120,26 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
       let env =
         List.fold_left
           (fun env f ->
-             SAT.assume env
-               {E.ff=f;
-                origin_name = "";
-                gdist = -1;
-                hdist = -1;
-                trigger_depth = max_int;
-                nb_reductions = 0;
-                age=0;
-                lem=None;
-                mf=false;
-                gf=false;
-                from_terms = [];
-                theory_elim = true;
-               } Ex.empty
+             match
+               SAT.assume env
+                 {E.ff=f;
+                  origin_name = "";
+                  gdist = -1;
+                  hdist = -1;
+                  trigger_depth = max_int;
+                  nb_reductions = 0;
+                  age=0;
+                  lem=None;
+                  mf=false;
+                  gf=false;
+                  from_terms = [];
+                  theory_elim = true;
+                 } Ex.empty
+             with
+             | Done e -> e
+             | Sat t -> raise (SAT t)
+             | Unsat e -> raise (UNSAT e)
+             | I_dont_know t -> raise (I_DONT_KNOW t)
           ) (SAT.empty ()) pb
       in
       ignore (SAT.unsat
@@ -141,8 +159,11 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
                 });
       Errors.run_error Errors.Failed_check_unsat_core
     with
-    | SAT.Unsat _  -> ()
-    | (SAT.Sat _ | SAT.I_dont_know _) as e -> raise e
+    | UNSAT _  -> ()
+    | SAT _ | I_DONT_KNOW _ as e ->
+      (* Useless line, but keeping it to ease the reading. These should be the
+         only exceptions raised. *)
+      raise e
 
 
   let unused_context name context =
@@ -177,6 +198,7 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
           begin
             match consistent with
             | `Sat _ | `Unknown _ ->
+              process_sat_res @@
               SAT.assume env
                 {E.ff=f;
                  origin_name = n;
@@ -199,7 +221,9 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
         if unused_context name used_context then acc
         else
           let dep = mk_root_dep name f d.st_loc in
-          SAT.pred_def env f name dep d.st_loc, consistent, dep
+          process_sat_res @@ SAT.pred_def env f name dep d.st_loc,
+          consistent,
+          dep
 
       | RwtDef _ -> assert false
 
@@ -207,20 +231,28 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
         let dep =
           match consistent with
           | `Sat _ | `Unknown _ ->
-            let dep' = SAT.unsat env
-                {E.ff=f;
-                 origin_name = n;
-                 hdist = -1;
-                 gdist = 0;
-                 trigger_depth = max_int;
-                 nb_reductions = 0;
-                 age=0;
-                 lem=None;
-                 mf=(sort != Check);
-                 gf=true;
-                 from_terms = [];
-                 theory_elim = true;
-                } in
+            let dep' =
+              match
+                SAT.unsat env
+                  {E.ff=f;
+                   origin_name = n;
+                   hdist = -1;
+                   gdist = 0;
+                   trigger_depth = max_int;
+                   nb_reductions = 0;
+                   age=0;
+                   lem=None;
+                   mf=(sort != Check);
+                   gf=true;
+                   from_terms = [];
+                   theory_elim = true;
+                  }
+              with
+              | Unsat d -> d
+              | Done t | I_dont_know t ->
+                raise (I_DONT_KNOW t)
+              | Sat t -> raise (SAT t)
+            in
             Ex.union dep' dep
           | `Unsat -> dep
         in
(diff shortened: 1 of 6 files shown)


src/lib/reasoners/satml_frontend.ml
```diff
@@ -92,12 +92,28 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
   let empty_with_inst add_inst =
     { (empty ()) with add_inst = add_inst }
 
-  exception Sat of t
+  (* Leaving this exception in case we actually raise Sat one day.
+     Remember to update `safe_call` by catching this exception if you do. *)
```
bclement-ocp: `safe_call` already catches this exception?

src/lib/reasoners/satml_frontend.ml
```diff
@@ -597,7 +616,7 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
             let acc =
               List.fold_left (fun (mz,lz) f ->
                   match atoms_from_sat_branch f with
-                  | None -> raise Exit
+                  | None -> assert false
```
bclement-ocp: This code is within a `try .. with Exit -> None`, can you explain this change?

src/lib/reasoners/sat_solver_sig.mli
```diff
@@ -53,19 +55,19 @@ module type S = sig
...
-  val assume_th_elt : t -> Expr.th_elt -> Explanation.t -> t
+  val assume_th_elt : t -> Expr.th_elt -> Explanation.t -> res
 
   (** [pred_def env f] assume a new predicate definition [f] in [env]. *)
-  val pred_def : t -> Expr.t -> string -> Explanation.t -> Loc.t -> t
+  val pred_def : t -> Expr.t -> string -> Explanation.t -> Loc.t -> res
 
   (** [unsat env f size] checks the unsatisfiability of [f] in
       [env]. Raises I_dont_know when the proof tree's height reaches
       [size]. Raises Sat if [f] is satisfiable in [env] *)
```
bclement-ocp: Needs to be updated

src/lib/frontend/frontend.ml
```diff
@@ -108,20 +120,26 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
...
+                  lem=None;
+                  mf=false;
+                  gf=false;
+                  from_terms = [];
+                  theory_elim = true;
+                 } Ex.empty
+             with
+             | Done e -> e
+             | Sat t -> raise (SAT t)
+             | Unsat e -> raise (UNSAT e)
```
bclement-ocp: This should use `process_sat_res`?

bclement-ocp (commented):
I am a bit unsure about this change. While the goal to have a cleaner API interface is good, this means that we currently:

 - Raise an exception in the SAT solver,
 - Convert the exception to `res` in the SAT solver,
 - Immediately convert the `res` back into an exception in the frontend

This feels like artificial complexity to me, especially considering that this is all internal to the library (the interface for library users is `process_decl`, which already captures and deals with the underlying exceptions).

src/lib/reasoners/sat_solver_sig.mli
```diff
@@ -31,9 +31,11 @@
...
   type t
 
-  exception Sat of t
-  exception Unsat of Explanation.t
-  exception I_dont_know of t
+  type res =
+    | Done of t
+    | Sat of t
+    | Unsat of Explanation.t
+    | I_dont_know of t
```
bclement-ocp: I'd suggest using

```ocaml
type outcome = Sat of t | Unsat of Explanation.t | I_dont_know of t

type 'a res = ('a, outcome) Result.t
```

and use `t res` as return type for `assume` / `assume_th_elt` / `pred_def`, and `outcome` for `unsat`.

src/lib/reasoners/satml_frontend.ml
Stevendeo: Oops, I removed the Sat exception, wrote this comment, put it back and forgot about it.

src/lib/reasoners/satml_frontend.ml
Stevendeo: A mistake, I'll just revert the commit

src/lib/reasoners/sat_solver_sig.mli
Stevendeo: Your version adds an extra pointer in the `Error` case. Besides, the `Error` constructor of `Result.t` would be badly named. It is not an error, it is a standard response.

bclement-ocp:
> I am a bit unsure about this change. While the goal to have a cleaner API interface is good, this means that we currently:
> 
>     * Raise an exception in the SAT solver,
> 
>     * Convert the exception to `res` in the SAT solver,
> 
>     * Immediately convert the `res` back into an exception in the frontend
> 
> 
> This feels like artificial complexity to me, especially considering that this is all internal to the library (the interface for library users is `process_decl`, which already captures and deals with the underlying exceptions).

I think it would make more sense to change the interface of the `Frontend` module to have an abstract type `t` isomorphic to `Sat of SAT.t | I_dont_know of SAT.t | Unsat of Explanation.t` with an interface similar to the `Sat_solver_sig`. If the state is `Sat` or `I_dont_know`, adding things with `assume` etc. are passed over to the corresponding `SAT` function; if the state is already `Unsat`, silently accept everything. Then calling `unsat` (or a new function `query`) would return `Sat | I_dont_know | Unsat` as a public type.

This would get rid of process_decl entirely (although we could re-implement process_decl on top of that new API to avoid changing the current solving loop), and bring the public-facing API more in-line with what is expected of other solvers, which was the goal I understood at yesterday's meeting.

Edited to add: Actually, it would make sense to do this in the SAT solvers themselves rather than in the Frontend. The main point is that `assume` etc. should return the same type that they take as argument.

src/lib/reasoners/sat_solver_sig.mli
bclement-ocp: I don't think the extra pointer here matters, I was trying to optimise for readability —sure  `Error` is badly named, but there is a fundamental difference here between `Done` and `Sat / Unsat / I_dont_know` (as exemplified by the fact that `unsat` can only return `Sat / Unsat / I_dont_know`, never `Done`).

But actually, as mentioned in another comment, that `Sat of t | Unsat of Explanation.t | I_dont_know of t` should be enough, bcause it makes sense for `assume` / `assume_th_elt` / `pred_def` to return `I_dont_know`.

Stevendeo:
> I am a bit unsure about this change. While the goal to have a cleaner API interface is good, this means that we currently:
> 
>     * Raise an exception in the SAT solver,
> 
>     * Convert the exception to `res` in the SAT solver,
> 
>     * Immediately convert the `res` back into an exception in the frontend
> 
> 
> This feels like artificial complexity to me, especially considering that this is all internal to the library (the interface for library users is `process_decl`, which already captures and deals with the underlying exceptions).

IMO, the way the frontend handles the response of the sat solver should be independent from the sat solver functionning.  The next step of this work would be indeed to remove the process_decl function and replacing it with specific functions, possibly without  using exceptions at all. 
I can add to this PR the simplification of frontend, or put it in another PR.

src/lib/reasoners/sat_solver_sig.mli
Stevendeo: Yes we could indeed replace `Done` by `I_dont_know` as they seem to play the same role (and I kind of did not like the `Done` name)

bclement-ocp:
> IMO, the way the frontend handles the response of the sat solver should be independent from the sat solver functionning. The next step of this work would be indeed to remove the process_decl function and replacing it with specific functions, possibly without using exceptions at all. I can add to this PR the simplification of frontend, or put it in another PR.

I agree; however, we also said yesterday that it made sense for the SAT solver to use exceptions.

I think the main question is: do we want to keep the `Frontend` module, or get rid of it?

 - If we keep the `Frontend` module, then it should be the interface for users (and I agree, it should expose specific functions instead of `process_decl` as a single entry point) — and so it should be the one that catches exceptions and provide a nice API. If we choose to do this, we can keep the exceptions in the `Sat_solver_sig`, as long as we don't use them in the `Frontend` that the users will interact with.

 - If we remove (or intend to — e.g. we could keep `process_decl` for the purpose of the internal solving loop for now) the `Frontend` module, and decide to make `Sat_solver_sig` the "official" interface, I think this PR should go a bit further than what it currently does and have `Sat / Unsat / I_dont_know` be the "public" type exposed by the solver, as mentioned in another message, so that the signatures are e.g. `val assume : t -> Expr.gformula -> Explanation.t -> t` (but `t` could be `Unsat`, in which case we do nothing).

I originally replied under the assumption we were going for the first option, but I think the second option may actually be closer to what @Halbaroth had in mind (and thinking about it more, it is probably the best option long-term, because it avoid one additional layer of wrapping).

Stevendeo:
I'm not sure to see the difference between the two versions. The `Frontend` does some preprocessing that is common to `Fun_sat` and `Satml`. 
Also, it feels weird to be able to send `Unsat` to the Frontend, this not only will require the sat solver to check its input, but it is just not well typed.
I can go for the following, which is the most precise :

- `assume` returns `[`Unknown of t | `Unsat of Explanation.t]`
- `unsat` returns [`Unknown of t | `Sat of t | `Unsat of Explanation.t]
- all the other methods return an environment
- `process_decl` is split
- `commands.ml` is removed

Halbaroth:
> > IMO, the way the frontend handles the response of the sat solver should be independent from the sat solver functionning. The next step of this work would be indeed to remove the process_decl function and replacing it with specific functions, possibly without using exceptions at all. I can add to this PR the simplification of frontend, or put it in another PR.
> 
> I agree; however, we also said yesterday that it made sense for the SAT solver to use exceptions.
> 
> I think the main question is: do we want to keep the `Frontend` module, or get rid of it?
> 
>     * If we keep the `Frontend` module, then it should be the interface for users (and I agree, it should expose specific functions instead of `process_decl` as a single entry point) — and so it should be the one that catches exceptions and provide a nice API. If we choose to do this, we can keep the exceptions in the `Sat_solver_sig`, as long as we don't use them in the `Frontend` that the users will interact with.
> 
>     * If we remove (or intend to — e.g. we could keep `process_decl` for the purpose of the internal solving loop for now) the `Frontend` module, and decide to make `Sat_solver_sig` the "official" interface, I think this PR should go a bit further than what it currently does and have `Sat / Unsat / I_dont_know` be the "public" type exposed by the solver, as mentioned in another message, so that the signatures are e.g. `val assume : t -> Expr.gformula -> Explanation.t -> t` (but `t` could be `Unsat`, in which case we do nothing).
> 
> 
> I originally replied under the assumption we were going for the first option, but I think the second option may actually be closer to what @Halbaroth had in mind (and thinking about it more, it is probably the best option long-term, because it avoid one additional layer of wrapping).

I had the second option in mind. I agree the PR should remove the exceptions in the frontend but `Util.Timeout`. Removing `Util.Timeout` requires more works.

bclement-ocp:
The preprocessing performed by the frontend is very light and could be instead provided as a (shared) helper module for both solvers.

>  Also, it feels weird to be able to send Unsat to the Frontend, this not only will require the sat solver to check its input, but it is just not well typed.

More precise types are not always desirable. As a user, if I have 5 formulas to add to my context, most of the time I don't really care if the context becomes inconsistent after I added 2 of them or 4 of them, I just care about the conjunction of all of them. If we expose the API you suggest, I have to do something like:

```ocaml
let extract = function
  | Unknown t -> t
  | Unsat expl -> raise (Unsat expl)
in
let t = extract @@ assume t fmla1 in
let t = extract @@ assume t fmla2 in
let t = extract @@ assume t fmla3 in
let t = extract @@ assume t fmla4 in
let t = extract @@ assume t fmla5 in
match unsat t with
| Sat _ | Unknown _ -> (* Do something appropriate *) ()
| Unsat expl -> raise (Unsat expl)
```

But if we hide from the user the fact that we can actually detect inconsistencies early (this will cause an "early" Unsat exception, before the user ever calls the `unsat` function), then I can write:

```ocaml
(* `assume` returns [ `I_dont_know of env | `Unsat of Explanation.t ] as an abstract type `t` *)
let t = assume t fmla1 in
let t = assume t fmla2 in
let t = assume t fmla3 in (* Maybe we detect an inconsistency between fmla1 and fmla3 here *)
let t = assume t fmla4 in
let t = assume t fmla5 in
unsat t (* Returns [ `Sat of env | `I_dont_know of env | `Unsat of Explanation.t ] *)
```

which is much more pleasant. And we can expose a function:

```ocaml
val is_inconsistent : t -> option Explanation.t
```

to be able to check whether an inconsistency was detected.

Stevendeo:
> The preprocessing performed by the frontend is very light and could be instead provided as a (shared) helper module for both solvers.
> 
> > Also, it feels weird to be able to send Unsat to the Frontend, this not only will require the sat solver to check its input, but it is just not well typed.
> 
> More precise types are not always desirable. As a user, if I have 5 formulas to add to my context, most of the time I don't really care if the context becomes inconsistent after I added 2 of them or 4 of them, I just care about the conjunction of all of them. If we expose the API you suggest, I have to do something like:
> 
> ```ocaml
> let extract = function
>   | Unknown t -> t
>   | Unsat expl -> raise (Unsat expl)
> in
> let t = extract @@ assume t fmla1 in
> let t = extract @@ assume t fmla2 in
> let t = extract @@ assume t fmla3 in
> let t = extract @@ assume t fmla4 in
> let t = extract @@ assume t fmla5 in
> match unsat t with
> | Sat _ | Unknown _ -> (* Do something appropriate *) ()
> | Unsat expl -> raise (Unsat expl)
> ```
> 
> But if we hide from the user the fact that we can actually detect inconsistencies early (this will cause an "early" Unsat exception, before the user ever calls the `unsat` function), then I can write:
> 
> ```ocaml
> (* `assume` returns [ `I_dont_know of env | `Unsat of Explanation.t ] as an abstract type `t` *)
> let t = assume t fmla1 in
> let t = assume t fmla2 in
> let t = assume t fmla3 in (* Maybe we detect an inconsistency between fmla1 and fmla3 here *)
> let t = assume t fmla4 in
> let t = assume t fmla5 in
> unsat t (* Returns [ `Sat of env | `I_dont_know of env | `Unsat of Explanation.t ] *)
> ```
> 
> which is much more pleasant. And we can expose a function:
> 
> ```ocaml
> val is_inconsistent : t -> option Explanation.t
> ```
> 
> to be able to check whether an inconsistency was detected.

I understand your point. Note that you can already create a syntactic sugar for your example : 

```ocaml
let extract_and x f = match x with
  | Unknown t -> f t
  | Unsat expl -> raise (Unsat expl)

let (let*) = extract_and
 
let* t = assume t fmla1 in
let* t = assume t fmla2 in
let* t = assume t fmla3 in
let* t = assume t fmla4 in
let* t = assume t fmla5 in
unsat t

bclement-ocp:
> I understand your point. Note that you can already create a syntactic sugar for your example :
> 
> ```ocaml
> let extract_and x f = match x with
>   | Unknown t -> f t
>   | Unsat expl -> raise (Unsat expl)
> 
> let (let*) = extract_and
>  
> let* t = assume t fmla1 in
> let* t = assume t fmla2 in
> let* t = assume t fmla3 in
> let* t = assume t fmla4 in
> let* t = assume t fmla5 in
> unsat t
> ```

True, but then I am also either forced to use exceptions (as I did in my example and you do here, but is what we were trying to avoid), or to live in a monad.

Stevendeo:
Exceptions have now disapeared from both modules; the only remaining exceptions are used for debugging. I almost  "purified" the whole API, but there still are few options dependencies left.

bclement-ocp:
I had in mind that these changes would be mostly in the SAT solver modules rather than the frontend, sorry if I was not clear. But I suppose doing this in the frontend could work as well — at first glance this looks mostly OK, although I have not checked the new logic in detail.

I'll let @Halbaroth make a proper review since he mentioned in https://github.com/OCamlPro/alt-ergo/pull/789#issuecomment-1704930209 that this part of the frontend changes a lot in OptimAE and will know better if this risk causing issues.

Halbaroth:
I will review it tomorrow :)

Stevendeo:
As this PR will not be compatible with the optimae changes, I created another branch from scratch, based on optimae, that reworks the frontend to make it more user-friendly. Should I close this PR and open another one, or recycle this one?

PR #796, opened 2023-09-06
Fix the cram tests

Halbaroth:
The opam CI failed because `alt-ergo` binary is not available for the some cram tests located in `src/lib/`. You can see it with:
```sh
dune clean && dune build -p alt-ergo-lib @runtest
```
All the cram tests in `src/lib/` depend on the binary and should be a part of the `alt-ergo` package or depend on it.

Alas, we have to backport this fix for the release 2.5.0 :(

- I also set the minimum version for `ppx_blob` to the version `0.7.2`. The lower versions are not compatible with dune and the opam CI failed or recent OCaml compilers.

- I prevent dune to build examples.

### DIFF
--- a/src/plugins/AB-Why3/preludes/dune
+++ b/src/plugins/AB-Why3/preludes/dune
@@ -8,7 +8,7 @@
 )
 
 (cram
- (package alt-ergo)
+ (package alt-ergo-plugin-ab-why3)
  (alias runtest-ci)
  (deps
    %{bin:alt-ergo}
--- a/src/plugins/AB-Why3/preludes/dune
+++ b/src/plugins/AB-Why3/preludes/dune
@@ -8,5 +8,11 @@
 )
 
 (cram
+ (package alt-ergo)
  (alias runtest-ci)
- (deps b-set-theory-prelude-2018-09-28.ae b-set-theory-prelude-2020-02-28.ae))
+ (deps
+   %{bin:alt-ergo}
+   b-set-theory-prelude-2018-09-28.ae
+   b-set-theory-prelude-2020-02-28.ae
+ )
+)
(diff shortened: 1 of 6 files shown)


src/plugins/AB-Why3/preludes/dune
```diff
@@ -8,5 +8,11 @@
 )
 
 (cram
+ (package alt-ergo)
```
bclement-ocp: This should be part of package `alt-ergo-plugin-ab-why3, not `alt-ergo`

Halbaroth: Oh I didn't know `alt-ergo-plugin-ab-why3` depends on the binary also :)

bclement-ocp: Yeah, plugins have to depend on `alt-ergo` because they install stuff to the `alt-ergo` sites.

bclement-ocp:
LGTM; I think we need

```
(data_only_dirs examples)
```

in a toplevel `dune` file, otherwise dune builds the examples which we don't want

src/plugins/AB-Why3/preludes/dune
Halbaroth: Yep, I saw it. It's sad we cannot share dune sites between several packages.

Halbaroth:
Good catch!

src/plugins/AB-Why3/preludes/dune
bclement-ocp: What do you mean? The site mechanism is what we use to share data across several packages, it's just that we want the site to be called "alt-ergo", not "alt-ergo-lib" (which should be merged back into alt-ergo anyways)

bclement-ocp (approved):

PR #802, opened 2023-09-11
Use fmt 0.9.0 to prevent conflict with the result library

Halbaroth:
This PR prevent conflict with old version of the library `result`. After this PR, Alt-Ergo depends on this library through `Fmt >= 0.9.0`.

We have to backport this fix on the branch `v2.5.x`.

### DIFF
--- a/alt-ergo.opam.template
+++ b/alt-ergo.opam.template
@@ -6,3 +6,7 @@ license: [
   "LicenseRef-OCamlpro-Non-Commercial"
   "Apache-2.0"
 ]
+
+conflicts: [
+  "result" {< "1.5"}
+]
--- a/alt-ergo.opam.template
+++ b/alt-ergo.opam.template
@@ -6,7 +6,3 @@ license: [
   "LicenseRef-OCamlpro-Non-Commercial"
   "Apache-2.0"
 ]
-
-conflicts: [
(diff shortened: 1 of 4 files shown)


bclement-ocp:
`result` is only needed for compatibility with OCaml < 4.03. If `fmt` is the issue shouldn't we rather require `fmt >= 0.8.6` (0.8.6 dropped the dependency on `result`) or `fmt >= 0.9.0` (0.9.0 has some API incompatibilities, might as well stay safe from our own future mistakes)?

alt-ergo.opam.template
```diff
@@ -6,3 +6,7 @@ license: [
   "LicenseRef-OCamlpro-Non-Commercial"
   "Apache-2.0"
 ]
+
+conflicts: [
+  "result" {< "1.5"}
+]
```
bclement-ocp: Dependencies should be specified in the `dune-project` (here with a `(conflicts)` field in the package declaration, see https://dune.readthedocs.io/en/stable/dune-files.html#stanza-package ), not in the template file.

Halbaroth: Thanks! I didn't find the right stanza.

bclement-ocp (approved):

PR #803, opened 2023-09-11
Test AE on 4.08.1 and 4.14.1

Halbaroth:
As I explained on Zulip, we don't test AE with ocaml 4.08.1 but still we publish AE on opam with this minimal version. This PR modifies a bit the CI in such way we test AE on both 4.08.1 and 4.14.1.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 ## unreleased
-### deprecated
-* printing underscore instead of fresh value in model
+### Deprecated
+* printing underscore instead of fresh value in model (PR #805)
+
+### Build
+* use OCaml 4.08.1 as the minimal supported version (PR #803)
 
 ## v2.5.0
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
 permissions: {}
 
 env:
-  OCAML_DEFAULT_VERSION: 4.08.0
+  OCAML_DEFAULT_VERSION: 4.08.1
   # Add OPAMYES=true to the environment, this is usefill to replace `-y` option
   # in any opam call
   OPAMYES: true
@@ -26,7 +26,7 @@ jobs:
           - macos-latest
           - ubuntu-latest
         ocaml-compiler:
-          - 4.08.0
+          - 4.08.1
           - 4.14.1
 
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
 permissions: {}
 
 env:
-  OCAML_DEFAULT_VERSION: 4.10.1
+  OCAML_DEFAULT_VERSION: 4.08.0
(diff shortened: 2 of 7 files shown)


bclement-ocp (approved):
LGTM. Calling OCaml 5.0 a beta release is a bit of a stretch, we could (and probably should) add it to the set.

That said, I'm not sure we need to run the tests on all these versions since it takes a bit of time, compiling on all of them and running the tests on a single one is probably enough (and slightly better for the environment).

bclement-ocp:
> LGTM. Calling OCaml 5.0 a beta release is a bit of a stretch, we could (and probably should) add it to the set.
> 
> That said, I'm not sure we need to run the tests on all these versions since it takes a bit of time, compiling on all of them and running the tests on a single one is probably enough (and slightly better for the environment).

Although, why 4.08.0 and not 4.08.1?

Gbury:
Small note: the `5.0` version of ocaml is not a beta: while there may be some breakages for projects that directly use parts of the runtime (through C bindings and the like), as far as I know, quite a lot of care was taken to not break pure ocaml programs (though some subtle parts such as Gc controls and asynchronous exceptions can exhibit strange behaviors).

Halbaroth:
Ok, I suggest to test the compilation of AE on 4.08.1 and 5.0.0 only and tests AE on 4.08.1 only.

bclement-ocp:
4.14.1 needs to stay, I think (being an LTS and all). I'd also say it makes more sense to run the tests with a more recent version (4.14 or 5.0), but those are also the version we use for the benches (and the versions we are most likely to run tests on locally) — so having some variety there and using 4.08 for the tests on GH makes sense.

Halbaroth:
I'm not sure to understand your point. You just said we should run tests on only versions and now we should run them on several ones?

bclement-ocp:
I was saying:

 - In general, if we run the tests with only one version, it makes more sense to pick a recent version (either 4.14 or 5.0, not necessarily both) to run the tests (because that's what we expect most users to have), rather than the older 4.08
 - 4.14 and 5.0 are used for the benches (on Marvin, not the CI tests) and on dev machines already
 - Therefore, using the old 4.08 makes more sense than I initially thought

But really I am just overthinking an inconsequential decision here. Running the tests is also faster than I had in mind (I had 15-30min in mind, but that is for the whole CI -- tests are only about 2min i.e. a fraction of that) so you can probably ignore my previous message and just run the tests with all versions, sorry for the confusion.

CHANGES.md
```diff
@@ -1,4 +1,6 @@
 ## unreleased
+### Build
+* Use OCaml 4.08.1 has the minimal supported version
```
bclement-ocp:
```suggestion
* Use OCaml 4.08.1 as the minimal supported version
```

Having a link to the PR would be nice :)

bclement-ocp (changes requested):
OCaml 5.0 is still missing

bclement-ocp (approved):

PR #809, opened 2023-09-12
Use --enable-assertions in tests

bclement-ocp:
The `--enable-assertions` flag is used to make Alt-Ergo perform internal verification of invariants at various points in the solve. This verification is not enabled by default, but it makes sense to set a high bar for ourselves and check the invariants on the in-repo tests.


Halbaroth:
Good catch! I thought it was enabled using the dev profile.

Halbaroth (approved):

Issue #807, opened 2023-09-11
Do not expose internal symbols


bclement-ocp:
We should go through and replace the use of "special" identifiers (e.g. `@mod`, `@/`, `@*`, etc.) with a different namespace (`Builtin of Hstring.t` works, or maybe we should just reuse `Op`).

```console
$ cat quoted2.smt2 
(set-logic ALL)
(declare-fun |@mod| (Int Int) Int)
(assert (exists ((x Int)) (distinct (mod 1 x) (|@mod| 1 x))))
(check-sat)
$ alt-ergo quoted2.smt2 --frontend dolmen
; File "quoted2.smt2", line 4, characters 1-12: Valid (0.4211) (2 steps) (goal g_1)

unsat
$ z3 quoted2.smt2
sat
```

Note that:

 - This is not a big issue with `@`-symbols because symbols starting with `@` are reserved for solver use in the SMT-LIB format
 - This *may* be an issue with the "fresh" `!kXXX" symbols that we generate that are *not* reserved for solver use. I initially thought we coud use `.kXXX` (and also `.?__` prefix instead of `!?__` prefix for skolem variables) but there are comments in the code staying "do not remove the underscores thay are important for the ordering" so… we'll have to carefully test this.

Actually having different namespaces for internal / user symbols (instead of relying on prefixes and the such) would be ideal, although that will also have an impact on order.

Stevendeo:
Should we close this issue?

bclement-ocp:
Is this handled correctly by Dolmen now? If so yes.

Gbury:
This should have been fixed by Gbury/dolmen#193 (but that requires a non-yet released version of dolmen)

bclement-ocp:
That's fine — we have #894 which will have to be taken care of before the next release.

PR #810, opened 2023-09-12
Ensure that internal AC symbols do not end up in models

bclement-ocp:
The AC(X) algorithm uses special symbols as an indirection (I believe to help enforce a terminating rewrite order) for AC symbols appearing within other AC symbols. These internal symbols start with `@` which is otherwise not a valid first character for a symbol name (but see #807). They should not appear in models.

### DIFF
--- a/src/lib/reasoners/shostak.ml
+++ b/src/lib/reasoners/shostak.ml
@@ -327,13 +327,24 @@ struct
       false
     | _ -> assert false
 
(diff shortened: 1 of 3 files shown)


bclement-ocp:
Changed the implementation to use `is_solvable_theory_symbol` and filter these earlier.

bclement-ocp:
This is obsoleted by #829 closing

PR #811, opened 2023-09-12
Use fresh names for all the abstract values

Halbaroth:
We should print fresh names for all the abstract values in order to prevent accidental captures.


bclement-ocp (commented):
It would be cleaner (and safer!) to call `Hstring.fresh_string` directly in `pp_abstract_value_of_type` at this point

bclement-ocp (approved):

PR #814, opened 2023-09-13
Result conflict and ppxlib

Halbaroth:
We have to use ppxlib >= 0.30.0 and prevent conflict with result < 1.5.

### DIFF
--- a/alt-ergo-lib.opam
+++ b/alt-ergo-lib.opam
@@ -26,8 +26,12 @@ depends: [
   "fmt" {>= "0.9.0"}
   "stdlib-shims"
   "ppx_blob" {>= "0.7.2"}
+  "ppxlib" {>= "0.30.0"}
   "camlzip" {>= "1.07"}
   "odoc" {with-doc}
+]
+conflicts: [
+  "result" {< "1.5"}
 ]
 build: [
   ["dune" "subst"] {dev}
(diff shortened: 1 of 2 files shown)


bclement-ocp:
We shouldn't have a dependency on `result` at all, we don't use it. I thought we solved this by adding a version restriction on `fmt` instead; do we have other dependencies that bring in `result`?

We also don't use `ppxlib`, so we shouldn't have it as a dependency either. What is broken?

Halbaroth:
`ppxlib` is used by recent `ppx_blob`. We have to use recent `ppx_blob` as older versions are not compatible with `Dune > 3`. If we don't use a recent version of `ppxlib`, it doesn't recognize user-defined binding operators: 
https://v2.ocaml.org/manual/bindingops.html 
I guess the AST of OCaml is not correct into too old `ppxlib` libraries.

I will check for the result dependency.

EDIT:  if you prefer, I can add `ppx_lib < 0.30.0` as a conflict clause instead of a dependency. It makes more sense.

bclement-ocp:
> `ppxlib` is used by recent `ppx_blob`. We have to use recent `ppx_blob` as older versions are not compatible with `Dune > 3`. If we don't use a recent version of `ppxlib`, it doesn't recognize user-defined binding operators:

Ah, so `ppx_blob` is the one that probably doesn't have proper dependency constraints, OK. Adding a conflicts clause on old ppxlib (with a comment that it's for `ppx_blob`) makes more sense then, good idea!

bclement-ocp:
I think `result` is `logs` through `ocplib-simplex`, OK for that then :'(

bclement-ocp (approved):
Let's merge this as is then, since #815 was merged already it will make things easier for us. We can clean up the `ppxlib` dependency into a `conflicts` once we are done with 2.5.0

Issue #693, opened 2023-06-26
Make sure that the documentation on the website is that of the last release


bclement-ocp:
Not of unreleased code ;)

Halbaroth:
The current workflow is as follow:
There is a branch `gh-pages` and a GitHub action in `documentation.yml` that build and push the documentation from `next` to the branch `gh-pages` as soon as we push on `next`. 

The branch `main` is not up to date. We should push the branch 2.4.x on it and modify the GitHub action in order to use this branch as reference for the documentation.

bclement-ocp:
(Taking over this one)

PR #818, opened 2023-09-13
Deploy the documentation of the main branch

Halbaroth:
This PR should solve #693.

As soon as the opam CI accepts our packages, I will push `v2.5.x` on the branch `main`.
We have to backport this patch on `v2.5.x` otherwise the documentation won't be deployed while pushing on `main`.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -19,7 +19,7 @@ env:
 jobs:
   # For any push and PR, build the documentation from the ocaml comments
   # If this build fails, the documentation workflow stops
-  # If it succed, an artifact is made with the generated documentation
+  # If it succeed, an artifact is made with the generated documentation
   # (html format only). This artifact is used in the deploying job
   ocaml_docs:
     name: OCaml documentation
@@ -95,7 +95,7 @@ jobs:
 
   # On PR, or push on next/main, build the sphinx general documentation
   # If this build fails, the documentation workflow stops
-  # If it succed, an artifact is made with the generated documentation
+  # If it succeed, an artifact is made with the generated documentation
   # This artifact is used in the deploying job
   sphinx_docs:
     name: Sphinx documentation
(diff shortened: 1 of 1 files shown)


Halbaroth:
Alt-Ergo 2.5.0 is on opam!

bclement-ocp:
Are we sure we want to use `main` for this? Since 2.5.0 is already released I think it would be better to disable the automatic GH page generation for now and manually generate the doc for 2.5.0 (using the same commands as in the workflow) so that we have a bit more time to think about what we want to do. In particular, it would be great to keep the doc for old versions available.

Halbaroth:
Why do you want to keep old documentations here? Users can run `odig` to retrieve old documentations.

bclement-ocp:
It is a good practice to provide access to the documentation for all released versions (or recent-ish released versions) for users that do not use the latest release and do not have the sources readily available and want to access the documentation corresponding to their version of Alt-Ergo. Moreover, the Sphinx documentation that is built by the GH pages is not available through `odig`.

We used to have old versions (at least one) of the documentation, but the link to the doc for version 2.4.0 seems to be broken: https://ocamlpro.github.io/alt-ergo/API/index.html

(As an aside: OPAM is able to provide docs, see e.g. https://ocaml.org/p/cmdliner/latest/doc/index.html , but it looks like ours does not get built, we should investigate this sometime as well)

.github/workflows/documentation.yml
```diff
@@ -19,7 +19,7 @@ env:
 jobs:
   # For any push and PR, build the documentation from the ocaml comments
   # If this build fails, the documentation workflow stops
-  # If it succed, an artifact is made with the generated documentation
+  # If it succeed, an artifact is made with the generated documentation
```
bclement-ocp:
```suggestion
  # If it succeeds, an artifact is made with the generated documentation
```

.github/workflows/documentation.yml
```diff
@@ -95,7 +95,7 @@ jobs:
 
   # On PR, or push on next/main, build the sphinx general documentation
   # If this build fails, the documentation workflow stops
-  # If it succed, an artifact is made with the generated documentation
+  # If it succeed, an artifact is made with the generated documentation
```
bclement-ocp:
```suggestion
  # If it succeeds, an artifact is made with the generated documentation
```

bclement-ocp (approved):
After offline discussion: let's merge this and put it in 2.5.1 for now, we will figure out the issues with multiple versions later.

.github/workflows/documentation.yml
Halbaroth: This is an incremental spelling correction! :D

PR #820, opened 2023-09-14
Update the changes log for v2.5.1

Halbaroth:


CHANGES.md
```diff
@@ -5,6 +5,11 @@
 ### Build
 * use OCaml 4.08.1 as the minimal supported version (PR #803)
 
+## v2.5.1
+
+### Bug fixes
+* fix a critical unsoundness bug with bvnot primitive (#819)
```
bclement-ocp: Nit: these are usually called "soundness bug" rather than "unsoundness bug" because it is the soundness that is buggy (the unsoundness unfortunately works very well :sweat_smile: )

bclement-ocp (approved):

CHANGES.md
Halbaroth: You are right ;) I will change it.

PR #829, opened 2023-09-21
Model generation support for SatML

Halbaroth:
This PR includes two things:
- A refactoring of the way AE prints models.
- A support of the model generation for SatML.
I cannot separate them as the support for model generation in SatML has been done after the refactoring in the OptimAE branch.

**IMPORTANT**: This PR has to been merged before OptimAE.

~~I also disable some tests as the printers change a bit but the models (with FunSAT) are still valid. I am preparing another PR rebase on this one that fix printers and test the support of model generation with SatML.~~


bclement-ocp:
Thanks for splitting the PR up — I will take a look at this one in the afternoon.

(We could split up further by merging the refactoring first as well — but this is at a size and scope that is manageable on its own now I think)

src/lib/frontend/frontend.ml
```diff
@@ -246,27 +273,33 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
       (* This case should mainly occur when a query has a non-unsat result,
          so we want to print the status in this case. *)
       print_status (Sat (d,t)) (Steps.get_steps ());
-      (*if get_model () then SAT.print_model ~header:true (get_fmt_mdl ()) t;*)
-      env, `Sat t, dep
+      print_model env (Some SAT.NoTimeout);
```
Halbaroth, line 273: We should prevent from printing model here as we did before this PR.

src/lib/frontend/frontend.ml
```diff
@@ -246,27 +273,33 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
...
+      env , consistent, dep
```
Halbaroth: This line reverts what we did in #789. Still the model printed for the test `777.smt2` is correct. We should investigate.

src/lib/frontend/models.ml
```diff
@@ -477,55 +484,63 @@ let rec pp_value ppk ppf = function
...
-    Format.fprintf fmt "; This model is a best-effort. It includes symbols
-        for which model generation is known to be incomplete. @.";
-
-  Format.fprintf fmt "@[<v 2>(";
-  if Options.get_model_type_constraints () then
-    begin
-      Why3CounterExample.output_constraints fmt props;
-      Format.fprintf fmt "@ ; values"
+let output_concrete_model fmt m =
+  if Options.get_interpretation () then begin
```
Halbaroth: We should remove this test. Deciding printing models shouldn't be taken in the pretty printer but in some caller of this function.

src/lib/frontend/models.mli
```diff
@@ -30,16 +30,17 @@
...
+  functions : ModelMap.t;
+  arrays : ModelMap.t;
+}
+
 (** Print the given counterexample on the given formatter with the
     corresponding format set with Options.get_output_format.
     - functions: the functions of the model;
     - constants: the variables of the model;
     - arrays: (experimental) the arrays of the model.
 *)
```
Halbaroth: We have to up to date this comment.

src/lib/frontend/models.mli
```diff
@@ -30,16 +30,17 @@
...
-val output_concrete_model :
-  Format.formatter ->
-  Expr.Set.t ->
-  functions:ModelMap.t ->
-  constants:ModelMap.t ->
-  arrays:ModelMap.t ->
-  unit
+val output_concrete_model : Format.formatter -> t -> unit
```
Halbaroth: As we use `Fmt`, we can replace this signature by `t Fmt.t`.

src/lib/util/printer.ml
```diff
@@ -369,10 +369,12 @@ let print_status_unknown ?(validity_mode=true) loc
     ("I don't know","unknown","fg_cyan") loc
     time steps goal
 
+(* TODO: The timeout answer doesn't exist in the SMT-LIB standard.
+   We should change this function. *)
```
Halbaroth: I leave this comment here to fix it later. We should always print SMT-LIB format answer with SMT-LIB files. For the legacy output format, I have no strong opinion but having more possible answers could be confusing for the final user.

src/lib/reasoners/fun_sat.ml
```diff
@@ -1902,10 +1845,13 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
 
+  (** returns the latest model stored in the env if any *)
+  let get_model env = !(env.last_saved_model)
+
   let reinit_ctx () =
     (* all_models_sat_env := None; *)
-    latest_saved_env := None;
-    terminated_normally := false;
+    (* latest_saved_env := None;
+       terminated_normally := false; *)
```
Halbaroth, line 1862: Probably wrong, I will fix it in a commit on this PR.

src/lib/reasoners/fun_sat.ml
```diff
@@ -1810,10 +1750,12 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
     with
     | IUnsat (d, classes) ->
-      terminated_normally := true;
       Debug.bottom classes;
       raise (Unsat d)
-    | Util.Timeout when switch_to_model_gen env -> do_switch_to_model_gen env
+    | Util.Timeout ->
+      (* don't attempt to compute a model if timeout before
+         calling unsat function *)
+      raise (I_dont_know {env; timeout = Assume})
```
Halbaroth, line 1766: This modification makes sense in the native input language. We don't expect a counterexample of a wrong context. But in the SMT-LIB, we should check what's happened.

src/lib/reasoners/fun_sat.ml
```diff
@@ -196,9 +192,15 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
     { env with
       unit_facts_cache = ref refs.unit_facts}, guard
 
+  type timeout_reason =
+    | NoTimeout
+    | Assume
+    | ProofSearch
+    | ModelGen
+
```
Halbaroth: We should add a documentation for this ADT in this PR.

src/lib/reasoners/fun_sat.ml
```diff
@@ -1326,12 +1271,14 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
   let greedy_instantiation env =
     match Options.get_instantiation_heuristic () with
     | INormal ->
+      (* TODO: check if this test still produces a wrong model. *)
```
Halbaroth, line 1282: I added this comment as a remainder.

src/bin/common/solving_loop.ml
```diff
@@ -489,7 +489,17 @@ let main () =
...
+            begin
+              match SAT.get_model partial_model with
+              | Some (lazy model) ->
+                Models.output_concrete_model
+                  (Options.Output.get_fmt_regular ()) model;
+                st
+              | _ ->
+                (* TODO: is it reachable? *)
+                st
+            end
```
bclement-ocp, line 494: Why do we do this dance? It looks cleaner to call `SAT.get_model` at the time we set the `partial_model_key` (this is OK since the model is lazy).

src/lib/frontend/frontend.ml
bclement-ocp: Yes, let's undo this change.

src/lib/frontend/frontend.ml
```diff
@@ -246,27 +273,33 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
...
+      print_model env (Some timeout);
+      (* TODO: Is it an appropriate behaviour? *)
+      (*       if timeout != NoTimeout then raise Util.Timeout; *)
+      env, consistent, dep
+
     | Util.Timeout as e ->
       (* In this case, we obviously want to print the status,
          since we exit right after  *)
       print_status (Timeout (Some d)) (Steps.get_steps ());
+      print_model env None;
```
bclement-ocp, line 299: Here we probably don't want to print the model either (or if we do it should be a separate option).

src/lib/frontend/models.ml
bclement-ocp: Yes. Created an issue.

src/lib/frontend/models.mli
bclement-ocp: Sure. Do you want to do it in this PR, or create an issue to do it later?

src/lib/frontend/frontend.ml
bclement-ocp: The last generated model is now saved imperatively on the solver using a reference, so this means that when we get the model from the old environment we still find the model of the new environment. Not very fond of this design, but we can revisit it later.

src/lib/frontend/frontend.ml
```diff
@@ -246,27 +273,33 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
...
-      env, `Unknown t, dep
+      env , `Unsat, dep
+    | SAT.I_dont_know {env; timeout} ->
+      (* TODO: always print Unknown for why3 ? *)
+      let status =
+        if timeout != NoTimeout then (Timeout (Some d))
+        else (Unknown (d, env))
+      in
+      print_status status (Steps.get_steps ());
+      print_model env (Some timeout);
```
bclement-ocp: Same as above, probably shouldn't print here.

src/lib/reasoners/fun_sat.ml
bclement-ocp: I think this is the same independent of input language. If the comment is correct, for a SMT-LIB problem, if we timeout here it means that we timed out before processing the `(check-sat)` command, which means that we may not have processed all the assertions / definitions. We shouldn't try to produce a model in this case.

src/lib/reasoners/fun_sat.ml
bclement-ocp: Can you clarify what is probably wrong? These two references no longer exist.

src/lib/reasoners/satml_frontend.ml
```diff
@@ -972,13 +982,72 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+    let ti = Options.get_timelimit_interpretation () in
+    if not i || (* not asked to gen a model *)
+       !(env.model_gen_phase) ||  (* we timeouted in model-gen-phase *)
+       Stdlib.(=) ti 0. (* no time allocated for extra model search *)
+    then
+      raise (I_dont_know {env; timeout = ProofSearch})
+    else
+      begin
+        (* Beware: models generated on timeout of ProofSearch phase may
+           be incohrent wrt. the ground part of the pb (ie. if delta
```
bclement-ocp: Probably "may not be coherent" or "may be incoherent"?

src/lib/reasoners/satml.ml
```diff
@@ -703,17 +705,13 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
...
-      with Ex.Inconsistent (expl, _) ->
-        C_theory expl
-    else C_none
+    try
+      let tenv, _terms = Th.do_case_split env.tenv origin in
+      (* TODO: terms not added to matching !!! *)
+      env.tenv <- tenv;
+      C_none
+    with Ex.Inconsistent (expl, _) ->
+      C_theory expl
```
bclement-ocp, line 714: Shouldn't we keep the `if Options.get_case_split_policy () == origin` check here?

src/lib/reasoners/theory.mli
```diff
@@ -60,14 +62,7 @@ module type S = sig
...
   val get_assumed : t -> Expr.Set.t
 
-  val output_concrete_model :
-    Format.formatter ->
-    prop_model:Expr.Set.t ->
-    t ->
-    unit
-
   val reinit_cpt : unit -> unit
-  (** reinitializes the counter to zero *)
```
bclement-ocp, line 70: Should keep the comment?

src/lib/util/printer.ml
bclement-ocp: It is useful to inform users that an unknown result is due to a timeout being reached and that re-trying with a higher timeout may succeed. Z3 uses a non-standard "timeout" response in that case. The SMT-LIB way of getting this information is to implement `(get-info :reason-unknown)`, but this is out of scope for this PR. I think both "timeout" and "unknown" are acceptable here.

src/lib/frontend/models.mli
bclement-ocp: Sure. Want to do it?

bclement-ocp (changes requested):
There are a couple minor cleanups to do, and some other things we may want to change later (e.g. the use of a reference for the last env), but I think we should merge with only the minor cleanups for now so that we can move on.

src/lib/reasoners/satml.ml
bclement-ocp: (Sorry, forgot to remove this comment — the check is now in `Theory.ml`)

src/lib/frontend/frontend.ml
Halbaroth: I would be better to have a fully immutable environment as the API of the SAT solvers hints the environment is immutable. I agree we should fix that in another PR. Probably after the Steven's PR on the frontend.

src/lib/util/printer.ml
Halbaroth: Let's keep `unknown` here and we will fix this in another PR. I will open an issue and put a link here.

src/lib/reasoners/fun_sat.ml
Halbaroth: You are right. I had a doubt.

src/lib/frontend/models.mli
Halbaroth: I did it in this PR.

src/lib/frontend/frontend.ml
Halbaroth: On further consideration, the line is correct. A part of the model refactoring consists in pulling the model printing from the `Models.ml` to the `Frontend.ml`. This code is now correct and will be changed after Steven's PR.

src/lib/frontend/frontend.ml
Halbaroth: It works because we check if we want to print models in the function `output_concrete_model`. We will change this while merging Steven's PR on the frontend.

src/lib/reasoners/theory.mli
Halbaroth: I kept it even if it is pointless in my opinion. We should rename the function as we don't care the function resets an internal counter if we don't document the purpose of this counter.

src/lib/reasoners/satml_frontend.ml
Halbaroth: I chose the second one.

Halbaroth:
> There are a couple minor cleanups to do, and some other things we may want to change later (e.g. the use of a reference for the last env), but I think we should merge with only the minor cleanups for now so that we can move on.

Seems good to mean as we have another PR to fix and test the feature #832

src/lib/util/printer.ml
Halbaroth: Issue: #834

src/bin/common/solving_loop.ml
Halbaroth: It's done.

Halbaroth:
I think we can merge this PR. The PR #832 fixes almost everything but we have to merge #836 and #835 in order to pass the CI on it.

bclement-ocp:
Is #836 really needed? PROXY variables shouldn't end up in models, unless I am mistaken.

If #832 is ready and working we should merge #832 into this PR, then merge both into next at once.

Halbaroth:
> Is #836 really needed? PROXY variables shouldn't end up in models, unless I am mistaken.

They do :/ I will check why they leak.

> If #832 is ready and working we should merge #832 into this PR, then merge both into next at once.

Ok

src/bin/common/solving_loop.ml
Halbaroth: Besides, the `partial_model_key` is now an option on a closure of the model. But we will need again to access to the last consistent environment in order to implement `get-info` in `Solving_loop` (Steven is working on it). A better design consists in modifying the signature of `get_model` in the module `Sat_sig`.  The function should return a `Models.t option` instead of `Models.t Lazy.t option`. If you call `get_model`, you want it and the value has to be force in `get_model` directly and keep the last consistent environment in `partial_model_key`. We should rename this key by the way as I will use it for different thing. Maybe `consistent_env_key`?

Halbaroth:
The CI will fail because `FunSAT` failed to produce a complete model for the first `check-sat` in `bool1.smt2`. `SatML` manages to produce a complete model in this case but the expected output has to be the some for both solvers. I will fix the issue with `FunSAT` in another PR.

Halbaroth:
I forgot to add the support for `String.starts_with` in the `compat` module! This feature is not present in old stdlib. I'll do it right now.

You did it!

src/bin/common/solving_loop.ml
```diff
@@ -223,7 +223,7 @@ let main () =
     State.create_key ~pipe:"" "solving_state"
   in
 
-  let partial_model_key: SAT.t option State.key =
+  let partial_model_key: Models.t Lazy.t option State.key =
```
bclement-ocp: I believe we decided on undoing this change for compatibility with #837

src/lib/frontend/frontend.ml
```diff
@@ -246,27 +270,33 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
       (* This case should mainly occur when a query has a non-unsat result,
          so we want to print the status in this case. *)
       print_status (Sat (d,t)) (Steps.get_steps ());
-      (*if get_model () then SAT.print_model ~header:true (get_fmt_mdl ()) t;*)
-      env, `Sat t, dep
+      print_model env (Some SAT.NoTimeout);
+      env , consistent, dep
```
bclement-ocp: Shouldn't we still return `` `Sat t`` here?

src/lib/frontend/frontend.ml
```diff
@@ -246,27 +270,33 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
...
-      env, `Unsat, dep
-    | SAT.I_dont_know t ->
-      (* In this case, it's not clear whether we want to print the status.
-         Instead, it'd be better to accumulate in `consistent` a 3-case adt
-         and not a simple bool. *)
-      print_status (Unknown (d, t)) (Steps.get_steps ());
-      (*if get_model () then SAT.print_model ~header:true (get_fmt_mdl ()) t;*)
-      env, `Unknown t, dep
+      env , `Unsat, dep
+    | SAT.I_dont_know {env; timeout} ->
```
bclement-ocp: This `env` shadows whereas previously there was no shadowing, I think we need

```suggestion
    | SAT.I_dont_know {env = t; timeout} ->
```

and corresponding changes below.

src/bin/common/solving_loop.ml
Halbaroth: As explained offline, this modification will cause a conflict with Steven's PR. I reverted it.

tests/dune.inc
bclement-ocp, line 1: Shouldn't this have been excluded from the diff by #830 ?

src/lib/structures/symbols.ml
```diff
@@ -122,6 +122,12 @@ let is_ac x = match x with
   | Name(_, Ac, _) -> true
   | _           -> false
 
+let is_internal sy =
+  match sy with
+  | Name (hs, _, _) ->
+    Compat.String.starts_with ~prefix:"." (Hstring.view hs)
```
bclement-ocp: Also `"@"`

bclement-ocp (changes requested):
A couple very minor things to fix and we should be good to go :tada:

src/bin/common/solving_loop.ml
bclement-ocp: This was solved as I was reviewing

src/lib/frontend/frontend.ml
Halbaroth: Good catch!

src/lib/frontend/frontend.ml
Halbaroth: I agree this shadowing is misleading but we have to return in this branch the SAT environment fron the caught exception:
```ocaml
    | SAT.I_dont_know {env = t; timeout} ->
      (* TODO: always print Unknown for why3 ? *)
      let status =
        if timeout != NoTimeout then (Timeout (Some d))
        else (Unknown (d, t))
      in
      print_status status (Steps.get_steps ());
      print_model t (Some timeout);
      (* TODO: Is it an appropriate behaviour? *)
      (*       if timeout != NoTimeout then raise Util.Timeout; *)
      env, `Unknown t, dep

```

bclement-ocp: Yeah, the first element of the returned triple ``env, `Unknown t, dep`` is `env` which is shadowed I believe? From what I remember the first element needs to be `env` here for correctness (otherwise the env has additional assertions that are not part of the semantic) but I may be wrong

Halbaroth: Oh you are right! The previous code was wrong because of the shadowing. Now it's fixed. We should add some tests for push and pop, otherwise we cannot detect this kind of mistake easily.

tests/dune.inc
Halbaroth: I will investigate.

Halbaroth:
I removed tests for other solvers. I will modify `gentest` script later to manage tests with several expected outputs but I have more urgent stuff to do now and this PR should be merge as soon as possible.

tests/dune.inc
Halbaroth: I will solve this in another PR.

bclement-ocp (approved):
LGTM :tada: 

Now only check-all-sat and OptimAE itself remain?

Halbaroth:
Yes! I'm working on `CheckAllSat`.

PR #830, opened 2023-09-21
The file dune.inc is linguist-generated

Halbaroth:
This PR hides the file `tests/dune.inc` in statistics and diff as it is generated by `gentest.ml` and modifying only one test produces a huge diff on this file.


bclement-ocp (approved):
Oh, didn't know this was possible, good idea! Maybe add a comment on the `.gitattribute` to say this is for hiding the file on GitHub diffs? Otherwise it looks a bit obscure

PR #831, opened 2023-09-21
Merging v2.5.x on main

Halbaroth:


bclement-ocp:
Looks like there are conflicts here :( Should we revert the few commits for 2.3.3 that seems to be on `main` but not `next`?

Halbaroth:
We shouldn't remove the commit e0b89ab53a39fdd56c80ed075dc043dcf767f6b7 on main as the tag of the release `2.3.3` is on this commit. It seems a release has been done by pushing a branch to `master` (renamed `main` later) and the branch have never been pushed on `next`.

bclement-ocp:
Sorry, I meant revert the commits (as in `git revert`), not remove them.

Halbaroth:
Merging `v2.5.x` on `main` isn't a good idea as `main` is supposed to be a prefix of `next` and `v2.5.x` diverged from `next`.

PR #835, opened 2023-09-25
Use a separated counter for abstract value

Halbaroth:
Use a separated counter in `Models` to produce fresh names for abstract values in models. 

This fix is temporary as we plan to refactor the model generation in another PR.
A better design consists in producing fresh names while computing the model. I will do it during the refactoring.


src/lib/frontend/models.ml
```diff
@@ -238,7 +242,7 @@ module SmtlibCounterExample = struct
 
   let pp_abstract_value_of_type ppf ty =
     if not @@ Options.get_interpretation_use_underscore () then
-      Fmt.pf ppf "(as @@%s %a)" (Hstring.fresh_string ()) Ty.pp_smtlib ty
+      Fmt.pf ppf "(as @@a%i %a)" !fresh_counter Ty.pp_smtlib ty
```
bclement-ocp: This needs to increment the counter, otherwise this will always be `a0`.

bclement-ocp (changes requested):

src/lib/frontend/models.ml
Halbaroth: oupsi! We should have a test with two abstract values :p

bclement-ocp (approved):

PR #836, opened 2023-09-25
Reserve proxy names of SatML

Halbaroth:
The SAT solver SatML produces proxy terms while building the CNF form of an expression. These terms have top symbols `Sy.Name` whose the string is prefixed by `PROXY_`.
I replace the prefix by `.PROXY_` as names starting by a dot are reserved in the SMT-LIB standard.


bclement-ocp:
We probably need to run benchmarks to ensure we don't introduce regressions b/c of ordering issues before making this change (see also #807).

Halbaroth:
I will investigate and run tests ;)

bclement-ocp (approved):
Tests show no regressions LGTM

PR #837, opened 2023-09-25
Support of get-info SMT2 command

Stevendeo:
Adds support for the `(get-info)` command, based on https://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf#page=58

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -50,6 +50,9 @@ let empty_solver_ctx = {
   global = [];
 }
 
+let unsupported_opt () =
+  Printer.print_std "unsupported"
+
 let main () =
   let () = Dolmen_loop.Code.init [] in
 
@@ -399,8 +402,52 @@ let main () =
       | ":produce-unsat-assumptions"
       | ":print-success"
       | ":random-seed"), _
-      -> Printer.print_wrn "unsupported option %s" name
-    | _ -> Printer.print_wrn "unsupported option %s" name
+      ->
+      unsupported_opt ();
+      Printer.print_wrn "unsupported option '%s'" name
+    | _ ->
+      unsupported_opt ();
+      Printer.print_err "unknown option '%s'" name
+  in
+
+  let handle_get_info (st : State.t) (name: string) =
+    let print_std =
+      fun (type a) (pp :a Fmt.t) (a : a) ->
+        Printer.print_std "(%s %a)" name pp a
+    in
+    let pp_reason_unknown st =
+      let err () =
+        let msg = "Invalid (get-info :reason-unknown)" in
+        Printer.print_smtlib_err "%s" msg
+      in
+      match State.get partial_model_key st with
+      | None -> err ()
+      | Some sat ->
+        match SAT.get_unknown_reason sat with
+        | None -> err ()
+        | Some s ->
+          print_std
+            Format.pp_print_string
+            (Frontend.unknown_reason_to_string s)
+    in
+    match name with
+    | ":authors" ->
+      print_std (fun fmt -> Fmt.pf fmt "%S") "Alt-Ergo developers"
+    | ":error-behavior" ->
+      print_std Fmt.string "immediate-exit"
+    | ":name" ->
+      print_std (fun fmt -> Fmt.pf fmt "%S") "Alt-Ergo"
+    | ":reason-unknown" ->
+      pp_reason_unknown st
+    | ":version" ->
+      print_std Fmt.string Version._version
+    | ":all-statistics"
+    | ":assertion-stack-levels" ->
+      unsupported_opt ();
+      Printer.print_wrn "unsupported option '%s'" name
+    | _ ->
+      unsupported_opt ();
+      Printer.print_err "unknown option '%s'" name
   in
 
   let handle_stmt :
@@ -503,6 +550,10 @@ let main () =
               "Model generation disabled (try --produce-models)";
             st
           end
+
+      | {contents = `Get_info kind; _ } ->
+        handle_get_info st kind;
+        st
 
       | _ ->
         (* TODO:
--- a/src/lib/reasoners/fun_sat.ml
+++ b/src/lib/reasoners/fun_sat.ml
@@ -176,6 +176,7 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
     add_inst: E.t -> bool;
     unit_facts_cache : (E.gformula * Ex.t) ME.t ref;
     last_saved_model : Models.t Lazy.t option ref;
+    unknown_reason : Sat_solver_sig.unknown_reason option;
   }
 
   let reset_refs () =
@@ -192,25 +193,13 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
     { env with
       unit_facts_cache = ref refs.unit_facts}, guard
 
-  (* Specify the pass in which Alt-Ergo runs out of time. This information
-     is used to decide if the last produced model is relevant. *)
-  type timeout_reason =
-    | NoTimeout
-    | Assume
-    (* Timeout while assuming a ground formula. *)
-
-    | ProofSearch
-    (* Timeout while doing instantiation or backtracking phases
-       during the proof search. *)
-
-    | ModelGen
-    (* Timeout while generating a new model. *)
-
   exception Sat of t
   exception Unsat of Explanation.t
-  exception I_dont_know of { env : t; timeout : timeout_reason }
+  exception I_dont_know of t
   exception IUnsat of Ex.t * SE.t list
 
+  let i_dont_know env ur =
+    raise (I_dont_know {env with unknown_reason = Some ur})
 
   (*BISECT-IGNORE-BEGIN*)
   module Debug = struct
@@ -1144,14 +1133,14 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
         raise (IUnsat (expl, classes))
     end
 
-  let update_model_and_return_unknown env compute_model ~timeout =
+  let update_model_and_return_unknown env compute_model ~unknown_reason =
     try
       let env = may_update_last_saved_model env compute_model in
       Options.Time.unset_timeout ();
-      raise (I_dont_know {env; timeout })
+      i_dont_know env unknown_reason
     with Util.Timeout when !(env.model_gen_phase) ->
       (* In this case, timeout reason becomes 'ModelGen' *)
-      raise (I_dont_know {env; timeout = ModelGen })
+      i_dont_know env (Timeout ModelGen)
 
   let model_gen_on_timeout env =
     let i = Options.get_interpretation () in
@@ -1160,7 +1149,7 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
        !(env.model_gen_phase) ||  (* we timeouted in model-gen-phase *)
        Stdlib.(=) ti 0. (* no time allocated for extra model search *)
     then
-      raise (I_dont_know {env; timeout = ProofSearch})
+      i_dont_know env (Timeout ProofSearch)
     else
       begin
         (* Beware: models generated on timeout of ProofSearch phase may
@@ -1170,7 +1159,7 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
         Options.Time.unset_timeout ();
         Options.Time.set_timeout ti;
         update_model_and_return_unknown
-          env i ~timeout:ProofSearch (* may becomes ModelGen *)
+          env i ~unknown_reason:(Timeout ProofSearch) (* may becomes ModelGen *)
       end
 
   let reduce_hypotheses tcp_cache tmp_cache env acc (hyp, gf, dep) =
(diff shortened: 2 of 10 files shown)


src/lib/reasoners/sat_solver_sig.ml
```diff
@@ -75,6 +79,10 @@ module type S = sig
 
   (** [get_model t] produces the current model. *)
   val get_model : t -> unit
+
+  (** [get_unknown_reason t] returns the reason Alt-Ergo returned
+      [I_dont_know] if it did. If it did not, returns None. *)
+  val get_unknown_reason : t -> unknown_reason option
```
Halbaroth: Maybe `raised` instead of `returned`?

src/lib/reasoners/satml_frontend.ml
```diff
@@ -60,6 +60,7 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
     skolems : E.gformula ME.t; (* key <-> f *)
     add_inst : E.t -> bool;
     guards : guards;
+    ur : Sat_solver_sig.unknown_reason option;
```
Halbaroth: I am not very fond of very short names but in this case the type makes clear what is the purpose of the field.

src/lib/reasoners/fun_sat.ml
```diff
@@ -1330,8 +1333,15 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
          this returns a wrong model. *)
-      return_answer env (Options.get_last_interpretation ())
-        (fun e -> raise (I_dont_know e))
+      return_answer
+        env
+        (Options.get_last_interpretation ())
+        (fun e ->
+           i_dont_know
+             "I won't conclude 'SAT' because I am not sure, hence \
+              I conclude 'Unknown'."
```
Halbaroth: I am not sure "I am not sure" is appropriate here. We can answer "incomplete"?

src/bin/common/solving_loop.ml
```diff
@@ -407,8 +410,50 @@ let main () =
...
+  let handle_get_info (st : State.t) (kind: string) =
+    let print_std =
+      fun (type a) (pp :(Format.formatter -> a -> unit)) (a : a) ->
+        Printer.print_std "(%s \"%a\")" kind pp a
+    in
+    match kind with
+    | ":authors" ->
+      print_std Format.pp_print_string "Alt-Ergo developers"
+    | ":error-behavior" ->
+      print_std Format.pp_print_string "immediate-exit"
```
bclement-ocp: This must be an S-expr atom, not a string (the expected output is `(:error-behavior immediate-exit)`, not `(:error-behavior "immediate-exit")`).

src/bin/common/solving_loop.ml
```diff
@@ -407,8 +410,50 @@ let main () =
...
+    | ":version" ->
+      print_std Format.pp_print_string Version._version
+    | ":all-statistics"
+    | ":assertion-stack-levels" ->
+      unsupported_opt kind;
+      print_std Format.pp_print_string "-- unsupported --"
+    | _ ->
+      Printer.print_wrn "unknown option %s" kind;
+      print_std Format.pp_print_string "-- unknown option --"
+
```
bclement-ocp: This should just output `unsupported` (see section 4.2.8).

src/bin/common/solving_loop.ml
```diff
@@ -407,8 +410,50 @@ let main () =
...
+    | _ -> unsupported_opt name
+  in
+
+  let pp_reason_unknown fmt st =
+    match State.get partial_model_key st with
+    | None ->
+      (* This case should happen only when there is no check-sat before
+         a get-info statement. *)
+      Printer.print_err "Invalid (get-info :reason-unknown)";
+      Format.fprintf fmt ""
```
bclement-ocp: The SMT-LIB way of indicating errors is to return an sexpr of the form `(error "Blah")` (this is required to not confuse automated tools; response to all commands must go to the same output).

bclement-ocp (changes requested):
Except when an error occurs (in which case the error message should be descriptive), SMT-LIB results should generally be short atoms rather than strings, to facilitate parsing by automated tools. We can use the standard `incomplete` response most of the time, or be more precise when appropriate e.g. `(unsupported :feature array-models)` if we really want to.

As an aside: `%S` automatically escapes quotes and is often better for printing strings than `\"%s\"` or `\"%a\"`.

src/bin/common/solving_loop.ml
Halbaroth: Actually, a string is a valid s-expression in the SMT-LIB, see p. 25 in the standard.  But we have to use double-quoted string actually. So we can simply output `"immediate-exit"`.

Halbaroth: On further consideration, we have to print as we said.

bclement-ocp: A string is a valid s-expression, but if we want to be compatible with the standard, we must reply with either `immediate-exit` or `continued-execution` as atoms to a `(get-info :error-behavior)` query (see section 4.1.8).

src/lib/reasoners/satml_frontend.ml
Stevendeo: I agree with you on this one, I'm changing it (and adding some documentation)

src/bin/common/solving_loop.ml
```diff
@@ -50,6 +50,9 @@ let empty_solver_ctx = {
   global = [];
 }
 
+let unsupported_opt s =
+  Printer.print_wrn "unsupported option %S" s
+
```
Halbaroth: We should print a SMT-LIB error here. Please use `Printers.print_smtlib_err`.

src/bin/common/solving_loop.ml
```diff
@@ -407,8 +410,51 @@ let main () =
...
+    | Some sat ->
+      match SAT.get_unknown_reason sat with
+      | None -> err ()
+      | Some s -> Format.pp_print_string fmt s
+  in
+
+  let handle_get_info (st : State.t) (kind: string) =
+    let print_std =
+      fun (type a) (pp :(Format.formatter -> a -> unit)) (a : a) ->
+        Printer.print_std "(%s %a)" kind pp a
```
Halbaroth: You can use `a Fmt.t` as a shorter for `Format.formatter -> a -> unit`.

src/bin/common/solving_loop.ml
```diff
@@ -407,8 +410,51 @@ let main () =
...
+      print_std Format.pp_print_string Version._version
+    | ":all-statistics"
+    | ":assertion-stack-levels" ->
+      unsupported_opt kind;
+      print_std Format.pp_print_string "unsupported"
+    | _ ->
+      Printer.print_wrn "unknown option %S" kind;
+      print_std Format.pp_print_string "unsupported"
+
   in
```
Halbaroth: We plan to use `Fmt` and `Logs` everywhere. I prefer `Fmt.pf` rather than `Format.fprintf` and `Fmt.string` instead of `Format.pp_print_string`. I'm ok if you don't apply this modification.

Halbaroth (approved):

src/bin/common/solving_loop.ml
```diff
@@ -50,6 +50,9 @@ let empty_solver_ctx = {
   global = [];
 }
 
+let unsupported_opt s =
+  Printer.print_smtlib_err "unsupported option %S" s
+
```
bclement-ocp: This should be `unsupported`, not `(error)` (see e.g. 4.1.7)

src/bin/common/solving_loop.ml
```diff
@@ -407,8 +410,50 @@ let main () =
...
+    | ":name" ->
+      print_std (fun fmt -> Fmt.pf fmt "%S") "alt-ergo"
+    | ":reason-unknown" ->
+      print_std pp_reason_unknown st
+    | ":version" ->
+      print_std Fmt.string Version._version
+    | ":all-statistics"
+    | ":assertion-stack-levels" ->
+      unsupported_opt kind;
+      print_std Fmt.string "unsupported"
```
bclement-ocp: This will now reply twice to the standard output, there needs to be a single reply. The reply needs to be `unsupported` (as a single atom -- not `(:all-statistics unsupported)`). Calling `unsupported_opt` here and below should be enough.

src/lib/reasoners/fun_sat.ml
```diff
@@ -1387,7 +1394,9 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
     with | Util.Not_implemented s ->
       Printer.print_err "Feature %s is not implemented. \
                          I can't conclude." s;
-      raise (I_dont_know env)
+      i_dont_know
+        "I stopped while trying to produce a model."
```
bclement-ocp: This is now printed as an SMT-LIB atom and can't be a free-form string, should be `incomplete` here.

src/lib/reasoners/sat_solver_sig.mli
```diff
@@ -28,6 +28,9 @@
 (*                                                                        *)
 (**************************************************************************)
 
+(** A reason for which the solver returned unknown. *)
+type unknown_reason = string
```
bclement-ocp: Nit: It would be cleaner for this to be an ADT, but we can keep a string for now (although I think the string is always `"incomplete"` currently)

bclement-ocp (changes requested):
There are still some formats that are not properly printed I belive. Maybe add tests to check the output format of `get-info`?

src/lib/reasoners/sat_solver_sig.mli
Stevendeo: > Nit: It would be cleaner for this to be an ADT, but we can keep a string for now (although I think the string is always `"incomplete"` currently)

I added this comment in sat_solver_sig.ml

tests/everything/testfile-smt-instr-get-info.dolmen.expected
```diff
@@ -0,0 +1,11 @@
+(error "Invalid (get-info :reason-unknown)")
+
+unsat
+(:all-statistics unsupported)
+(:assertion-stack-levels unsupported)
```
bclement-ocp: These lines need to be:

```suggestion
unsupported
unsupported
```

tests/everything/testfile-smt-instr-get-info.dolmen.expected
```diff
@@ -0,0 +1,11 @@
...
+(:authors "Alt-Ergo developers")
+(:error-behavior immediate-exit)
+(:name "Alt-Ergo")
+(error "Invalid (get-info :reason-unknown)")
+(:version dev)
+(error "unknown option ':foo'")
```
bclement-ocp: This line also needs to be:

```suggestion
unsupported
```

src/lib/reasoners/sat_solver_sig.mli
Halbaroth: After merging #829, the ADT will be something like
```ocaml 
type unknown_reason = 
  | Incomplete 
  | Memout 
  | Timeout of timeout_reason
``` 
We can add the ADT right now but we will have to modify it just after merging the other PR.

src/bin/common/solving_loop.ml
Halbaroth: Sorry, it's my mistake.

bclement-ocp (changes requested):
Thank you for adding tests! I believe the expected outputs are not correct, can you fix them?

src/lib/reasoners/sat_solver_sig.mli
Halbaroth: We merged #829 so we can create the ADT now.

Stevendeo: This ADT is strange, it makes the possibility of building a value `Timeout NoTimeout`. I'm changing this a bit.

Stevendeo:
Okay I fixed the output, but I added warning/error messages to give more insights on the unsupported/invalid requests

src/lib/reasoners/sat_solver_sig.ml
```diff
@@ -31,18 +31,25 @@
...
+  | ProofSearch
+  | ModelGen
+
+type unknown_reason =
+  | Incomplete
+  | Memout
+  | Timeout of timeout_reason
+
+(* TODO: use an ADT and add Timeout when library starts acting like a
+   library. *)
```
bclement-ocp: I think this comment can go away?

bclement-ocp (approved):
Thanks! You can also use the "Re-request a review" button once the changes are made.

I think this looks good now, I think the TODO comment should be removed before merging since you added the ADT.

It's not very important, but I am surprised about the blank lines in the output (although the output machinery is very convoluted so maybe I shouldn't be surprised). Do you understand where they are coming from?

src/lib/reasoners/sat_solver_sig.ml
Halbaroth: @Stevendeo I can merge as soon as the comment has been removed ;)

Stevendeo:
> It's not very important, but I am surprised about the blank lines in the output (although the output machinery is very convoluted so maybe I shouldn't be surprised). Do you understand where they are coming from?

I did not investigate it but I thought about it as well... Maybe one day we'll find it

bclement-ocp (commented):
:+1:

Issue #799, opened 2023-09-08
Allow configuration of output channel for `--dump-models`


bclement-ocp:
Alt-Ergo uses the standard input to print models. This is in accordance to the SMT-LIB standard.

However, while this makes sense for the models that are printed as a result of a `(get-model)` command from the user, it makes less sense for the models printed by the `--dump-models` option. We should consider printing *those* to stderr by default, and bring back an option to let the user control this output channel independently.

(Seel also #393)

Halbaroth:
I guess this issue is solved?

bclement-ocp:
Yes fixed in #838

PR #838, opened 2023-09-25
Adding an option to dump models on a specific output channel

Stevendeo:
Adresses issue #799

### DIFF
--- a/src/bin/common/parse_command.ml
+++ b/src/bin/common/parse_command.ml
@@ -913,7 +913,8 @@ let parse_output_opt =
 
     let dump_models_on =
       let doc =
-        "Select a channel to output the models."
+        "Select a channel to output the models dumped by the option \
+         --dump-model."
       in
       let docv = "VAL" in
       let chan =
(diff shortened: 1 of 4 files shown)


src/bin/common/parse_command.ml
```diff
@@ -910,6 +911,23 @@ let parse_output_opt =
       Arg.(value & flag & info ["dump-models"] ~doc ~docs:s_models)
     in
 
+    let dump_models_on =
+      let doc =
+        "Select a channel to output the models."
```
bclement-ocp: This doc should make it clear that the option only applies to `--dump-model`

bclement-ocp (changes requested):
LGTM except for the wording on the option doc

src/bin/common/parse_command.ml
Stevendeo: Done!

bclement-ocp (approved):

Halbaroth (approved):

PR #839, opened 2023-09-26
BV Theory: Handling of zero extend

Stevendeo:
This MR adds to the alt-ergo bit vector theory processor the zero_extend primitive

https://smtlib.cs.uiowa.edu/logics-all.shtml

### DIFF
--- a/src/lib/reasoners/bitv.ml
+++ b/src/lib/reasoners/bitv.ml
@@ -139,7 +139,8 @@ module Shostak(X : ALIEN) = struct
 
   let is_mine_symb sy _ =
     match sy with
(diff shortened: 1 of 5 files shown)


bclement-ocp:
Why do we need to support `zero_extend` as a primitive? The existing support translating `zero_extend` to a `concat` with a zero BV is not enough?

Stevendeo:
The fact that by default (i.e. without dolmen), alt-ergo returned an error on zero_extend made me think there were no support of zero_extend. My bad!

bclement-ocp:
Ah indeed, BV support in the SMT-LIB format is only available through the Dolmen frontend. The SMT-LIB support of the legacy frontend (through `psmt2-frontend`) is no longer developed and does not get new features, including new built-in theories.

PR #841, opened 2023-09-27
[bitv] Add support for model generation

bclement-ocp:
This patch adds support for model generation in the bitvector theory.

To generate a value for a model, a simple brute-force backtracking search amongst all possible bitvector values is performed to find one that is different from the values that were asserted distinct. If none can be found (or if exactly one is found, in which case the value is forced) the `case_split` flag is set so that we report `unsat` instead of looping during model generation.

In the future, this should be augmented with proper support for case splits / bit-blasting in the relational theory. Note that even if we add support for case splits in the relational theory, having an enumeration in model generation is useful, because we may not always want to perform case splits on large bitvector types.


src/lib/reasoners/bitv.ml
```diff
@@ -1325,14 +1325,113 @@ module Shostak(X : ALIEN) = struct
...
+  let is_cte_abstract = List.for_all is_cte
+
+  let assign_value r distincts eq =
+    (* If we are trying to assign a value to a constant, or a term that has a
+       constant in its equivalence class, bail out *)
+    let biv = embed r in
+    if is_cte_abstract biv || List.exists
+         (fun (t, x) ->
+            let { E.f; ty; _ } = E.term_view t in
+            is_mine_symb f ty && X.leaves x == [] ) eq
```
Halbaroth: I agree that your code is correct but I'm not sure to understand why you don't use again `is_cte_abstract (embed x)` to check if `x` is a constant of the BV theory? 

Besides, `r` should be an element of its equivalent class, thus we don't need to check it twice.
I guess the above snippet does the same check:
```ocaml
List.exists (fun (_, x) -> 
  match embed x with 
  | Some biv -> is_cte_abstract biv 
  | None -> false
)
```
We can check if the values appear first in the list `eq`.

src/lib/structures/ty.ml
```diff
@@ -107,7 +107,9 @@ let print_generic body_of =
       | Treal -> fprintf fmt "real"
       | Tbool -> fprintf fmt "bool"
       | Tunit -> fprintf fmt "unit"
-      | Tbitv n -> fprintf fmt "bitv[%d]" n
+      | Tbitv n ->
+        if Options.get_output_smtlib () then fprintf fmt "(_ BitVec %d)" n
+        else fprintf fmt "bitv[%d]" n
```
Halbaroth, line 112: It's probably better to write a separated printer for the SMT-LIB format as you did with `pp_smtlib` but we can do it later in another PR.

src/lib/reasoners/bitv.ml
```diff
@@ -1325,14 +1325,113 @@ module Shostak(X : ALIEN) = struct
...
+  (* Pop the first bit, raises [Not_found] if there is no first bit *)
+  let pop_bit = function
+    | [] -> raise Not_found
+    | { bv = Cte b; sz } as st :: rst ->
+      Some b, if sz > 1 then { st with sz = sz - 1 } :: rst else rst
+    | { bv = Other _ | Ext _ ; sz } as st :: rst ->
+      None, if sz > 1 then { st with sz = sz - 1 } :: rst else rst
+
+  (* Fills an initial segment of [buf] from [0] to [sz] with a bitvector that is
+     different from all those in [abstracts].
```
Halbaroth:
```suggestion
  (* Fills the segment of [buf] from [pos] to [sz] with a bitvector that is
     different from all those in [abstracts].
```

src/lib/reasoners/bitv.ml
```diff
@@ -1325,14 +1325,113 @@ module Shostak(X : ALIEN) = struct
...
+        | None -> (t, nt, f, nf)) ([], 0, [], 0) abstracts
+    in
+    match t, f with
+    | _, [] -> Bytes.fill buf pos sz '0'
+    | [], _ -> Bytes.fill buf pos sz '1'
+    | _, _ ->
+      (* Prefer searching where there are less candidates. *)
+      let x, cx, y, cy =
+        if nt < nf then t, '1', f, '0' else f, '0', t, '1'
+      in
```
Halbaroth, line 1360: If `nt < nf`, there is more abstract values starting by `0` to avoid and we should start by calling `search buf (pos + 1) (sz - 1) f`?

src/lib/reasoners/bitv.ml
```diff
@@ -1325,14 +1325,113 @@ module Shostak(X : ALIEN) = struct
...
+        match b with
+        | Some true -> (bv :: t, 1 + nt, f, nf)
+        | Some false -> (t, nt, bv :: f, 1 + nf)
+        | None -> (t, nt, f, nf)) ([], 0, [], 0) abstracts
+    in
+    match t, f with
+    | _, [] -> Bytes.fill buf pos sz '0'
+    | [], _ -> Bytes.fill buf pos sz '1'
+    | _, _ ->
+      (* Prefer searching where there are less candidates. *)
```
Halbaroth:
```suggestion
      (* Prefer searching first where there are more constraints, that is more abstract values 
         to avoid. *)
```

src/lib/reasoners/bitv.ml
bclement-ocp: We must be distinct from the values provided. If `nt < nf`, there are more values starting by `0` we must be distinct from -- so we should try those starting with `1` first. In the limit, if `nt = 0`, there are no values at all that start with `1` — so we can pick any of them.

src/lib/reasoners/bitv.ml
bclement-ocp: We actually want to search where the are more candidates / less constraints.

```suggestion
      (* Prefer searching where there are more candidates, i.e. less constraints. *)
```

src/lib/reasoners/bitv.ml
bclement-ocp: Honestly, because I copied this code from `arith.ml` :sweat_smile:
I think we can indeed use `is_cte_abstract biv (X.extract x)` instead.

`r` is not actually guaranteed to be an element of its equivalence class, because the equivalence class only contains term representatives (i.e. values returned by `X.make` for some term) while `r` may be dynamically created, so we need to check `r` separately.

src/lib/structures/ty.ml
bclement-ocp: Agree to both. Opened #845

src/lib/reasoners/bitv.ml
Halbaroth: I agree, the comment was misleading ;)

src/lib/reasoners/bitv.ml
Halbaroth: Good point for the equivalence class. We should add a comment about this pitfall at some point in the union-find. I don't think putting a comment to explain this here is appropriate as the issue occurs in all theories.

Halbaroth (approved):
LGTM after applying the changes ;)

Issue #746, opened 2023-07-24
Smart constructors for the FP theory


Halbaroth:
The current `Expr` module does not expose smart constructors for FP theory as the `Z3.FloatingPoint` does. I noticed 
this lack while writing the interface for AE in the project https://github.com/wasp-platform/encoding. We should implement at least the constructors defined by the FP theory https://smtlib.cs.uiowa.edu/theories-FloatingPoint.shtml.

bclement-ocp:
We don't have support for the smtlib FloatingPoint theory as far as I'm aware (we only have support for floating-point rounding with underflow, but nothing related to NaNs, infinites, or overflow), so I expect more work than just providing smart constructors.

Halbaroth:
I know :/. We should work on it at some point.

PR #842, opened 2023-09-28
Fix error handling of dolmen

Stevendeo:
When dolmen fails, its default behavior is to call `exit` with a custom exit code. This MR overrides the `error` function of the dolmen loop so that alt-ergo keeps control over the exit code.

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -257,6 +257,9 @@ let main () =
     | Errors.Error e ->
       Printer.print_err "%a" Errors.report e;
       exit 1
+    | D_loop.DolmenError (i, descr) ->
+      Printer.print_err "Dolmen failed %s (code %i)" descr i;
+      exit 1
     | _ as exn -> Printexc.raise_with_backtrace exn bt
   in
   let finally ~handle_exn st e =
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -258,7 +258,7 @@ let main () =
       Printer.print_err "%a" Errors.report e;
       exit 1
     | D_loop.DolmenError (i, descr) ->
-      Printer.print_err "Dolmen failed %s (code %i)" descr i;
+      Printer.print_err "Failure %s (code %i)" descr i;
       exit 1
     | _ as exn -> Printexc.raise_with_backtrace exn bt
   in
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -257,9 +257,6 @@ let main () =
     | Errors.Error e ->
       Printer.print_err "%a" Errors.report e;
       exit 1
-    | D_loop.DolmenError (i, descr) ->
-      Printer.print_err "Failure %s (code %i)" descr i;
-      exit 1
     | _ as exn -> Printexc.raise_with_backtrace exn bt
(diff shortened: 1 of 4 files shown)


src/lib/frontend/d_loop.ml
```diff
@@ -28,10 +28,34 @@
 (*                                                                        *)
 (**************************************************************************)
 
+exception DolmenError of int * string
+
```
Halbaroth: Could you add a documentation to specify what is the `int` and `string` fields there or use a record instead?

Halbaroth (approved):

src/bin/common/solving_loop.ml
```diff
@@ -257,6 +257,9 @@ let main () =
     | Errors.Error e ->
       Printer.print_err "%a" Errors.report e;
       exit 1
+    | D_loop.DolmenError (i, descr) ->
+      Printer.print_err "Dolmen failed %s (code %i)" descr i;
```
bclement-ocp: We do not really want to mention Dolmen here because users don't know (or care) what Dolmen is.

bclement-ocp (commented):
Alt-Ergo's current infrastructure for error handling is centered around the `Errors` module. Would it be possible to add a new `DolmenError` constructor to `Errors.error` instead, and deal with printing in the same place as for other errors?

bclement-ocp:
BTW if the only goal is to keep control over the exit code, Dolmen provides facility to register our own exit codes (see usage in the Dolmen binary [here](https://github.com/Gbury/dolmen/blob/master/src/bin/options.ml#L21) )

Stevendeo:
> BTW if the only goal is to keep control over the exit code, Dolmen provides facility to register our own exit codes (see usage in the Dolmen binary [here](https://github.com/Gbury/dolmen/blob/master/src/bin/options.ml#L21) )

It is not only to have control over the exit code, but to have control on how alt-ergo exits. We don't want an external library to exit alt-ergo.

Halbaroth:
> Alt-Ergo's current infrastructure for error handling is centered around the `Errors` module. Would it be possible to add a new `DolmenError` constructor to `Errors.error` instead, and deal with printing in the same place as for other errors?

It depends if we consider the Dolmen frontend as a part of the library or a part of the binary. Currently, the situation is a mess as we have some parts of the Dolmen frontend in the binary (`solving_loop` manages statements as `get-info`,  `get-model`, `set-option`, ...) and some parts in the library (mainly in `d_cnf.ml`, `frontend.ml`). I haven't decided yet which design is the best in our case, so I have no strong opinion about the location of this error.

bclement-ocp:
Right now, the situation is that Dolmen is part of the library though :) I actually think the current situation makes sense given the infrastructure we have — `D_cnf` deals with parsing actual expressions while `Solving_loop` deals with state and solver management, which is also the separation of concerns within the legacy frontend.

Besides, Dolmen provides parsing and typechecking, which can be useful to users of the library (e.g. for parsing custom preludes or other SMT-LIB fragments), and as long as the preludes are external files, providing any meaningful FPA support in the library (e.g. #746 ) requires Dolmen (provided we keep the plan of deprecating and removing the legacy frontend).

Halbaroth:
If @bclement-ocp agrees, we can merge it :)

bclement-ocp (approved):

PR #843, opened 2023-09-28
Adding a test for push/pop

Stevendeo:
Adds a test for push & pop functionality. Also adds to gentest the possibility for alt-ergo to fail on something in the non regression tests.

### DIFF
--- a/tests/smtlib/testfile-push-pop1.err.dolmen.expected
+++ b/tests/smtlib/testfile-push-pop1.err.dolmen.expected
@@ -1,0 +1,4 @@
+
+unknown
+
+unknown
--- a/tools/gentest.ml
+++ b/tools/gentest.ml
@@ -178,9 +178,9 @@ end = struct
           pp_output tst
     in
     let accepted_ae_exit_code =
-      if tst.params.ae_should_succeed then
-        "0"
-      else "(or 1 123 124 125 142)"
+      if tst.params.ae_should_succeed
+      then "0"
+      else "1"
     in
     Format.fprintf fmt "\
 @[<v 1>\
--- a/tools/gentest.ml
+++ b/tools/gentest.ml
@@ -107,14 +107,24 @@ end = struct
 end
 
 module Test : sig
+
+  type params = {
+    exclude: string list;
+    filters: string list option;
+    compare_should_succeed: bool;
+    ae_should_succeed: bool;
+  }
+
   type t = private {
     cmd: Cmd.t;
     pb_file: string;
-    should_succeed: bool
+    params: params
   }
   (** Type of a test. *)
 
-  val make: cmd: Cmd.t -> pb_file: string -> should_succeed: bool -> t
+  val base_params : params
+
+  val make: cmd: Cmd.t -> pb_file: string -> params:params -> t
   (** Set up the test. *)
 
   val pp_expected_output: t printer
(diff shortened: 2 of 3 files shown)


tools/gentest.ml
```diff
@@ -152,6 +177,11 @@ end = struct
           pp_expected_output tst
           pp_output tst
     in
+    let accepted_ae_exit_code =
+      if tst.params.ae_should_succeed then
+        "0"
+      else "(or 1 123 124 125 142)"
```
bclement-ocp: Probably only `1` should be allowed here because it is the only one that actually indicates an error of the solver — `142` is a timeout (very different!), and `123`, `124` and `125` are for command-line parsing errors which shouldn't occur at all in tests.

tests/smtlib/testfile-push-pop1.err.dolmen.expected
```diff
@@ -0,0 +1,4 @@
+
+unknown
+
+unknown
```
bclement-ocp, line 4: Unrelated remark: we should have an `(error "something")` output here.

bclement-ocp (changes requested):
LGTM except for the error codes used.

bclement-ocp:
The CI is failing because after modifying `gentest.ml`, you need to re-run `make gentest` locally to make sure that the `dune.inc` file is up-to-date.

By the way, the test only tests that declarations go way when we pop the stack. I don't think we have tests that check that *assertions* go away (e.g. if we assert `x = 0` then pop and assert `x = 1` we should get `unknown`, not `unsat`) but we should (doesn't need to be in this PR).

Stevendeo:
> The CI is failing because after modifying `gentest.ml`, you need to re-run `make gentest` locally to make sure that the `dune.inc` file is up-to-date.
> 
> By the way, the test only tests that declarations go way when we pop the stack. I don't think we have tests that check that _assertions_ go away (e.g. if we assert `x = 0` then pop and assert `x = 1` we should get `unknown`, not `unsat`) but we should (doesn't need to be in this PR).

Not only, it was not rebased on the PR that was pending when I opened it. It is now rebased so the CI should be working. This PR only adds 1 test as well as modifying how error tests are handled, I thought it would be better to add other tests in another PR. Besides, now this PR introduces the smtlib test folder, I can add more tests at the right place.

tests/smtlib/testfile-push-pop1.err.dolmen.expected
Stevendeo: It should be done in a PR that centralizes how errors are handled. Should we open an issue about it?

bclement-ocp: Yeah, this is definitely unrelated to this PR, opening an issue for it is a good idea — I thought you had one on error behavior already but I can't find it?

bclement-ocp (approved):
Sounds good — I am fine with merging this as is since it adds `.err` and we can add more tests in another PR.

tests/smtlib/testfile-push-pop1.err.dolmen.expected
Stevendeo: It may have gotten lost in the depth of this repo's madness.

bclement-ocp: Actually it was a Zulip message — opened #851 feel free to assign yourself if you want to tackle it.

Issue #844, opened 2023-09-29
Supporting `get-assignment`


Halbaroth:
The current model generation support allows to print a boolean model aside the first-order one. The way this features is implemented isn't SMT-LIB compliant as the standard defines the statement `get-assignment`.

Issue #833, opened 2023-09-21
Bubble decision of printing models upwards


bclement-ocp:
We should remove this test. Deciding printing models shouldn't be taken in the pretty printer but in some caller of this function.

_Originally posted by @Halbaroth in https://github.com/OCamlPro/alt-ergo/pull/829#discussion_r1332941819_

Issue #382, opened 2020-12-18
incremental mode vs multi-goals: currently, work is done twice


OriginLabs-Iguernlala:
Once https://github.com/OCamlPro/alt-ergo/pull/371 is merged, Alt-Ergo will support SMT2 incremental mode (push/pop).

But the way things are done in
- psmt2_to_alt_ergo (translates `check-sat` and `check-sat-assuming` commands to `goal`)
- main_text (all context reassumed for successive goals in multi-goal mode)

is not optimal.

The main loop in main_text should be improved.

OriginLabs-Iguernlala:
When the handling in main_text is simplfied, a particular attention should be given for check and cut commands

PR #846, opened 2023-09-29
feat: check-all-sat support

Halbaroth:
This PR adds the support for the `check-all-sat` support. I think we can merge the PR without clean-up its implementation as the feature doesn't break the classical model generation. Please see #847 for a clean-up roadmap of this feature.

The PR includes two ways to use this new feature:
- using the new CLI option `--all-sat` (or `--all-models`), AE produces all the possible propositional models after each `check-sat`.
For instance using the input file:
```
(set-logic ALL)
(declare-const x Bool)
(declare-const y Bool)
(declare-const z Bool)
(assert (or x y))
(assert (or (not y) z))
(check-sat)
```
and the command `dune exec -- alt-ergo --all-sat --produce-models --sat-solver CDCL test.smt2`, we got the output:
<details>
  <summary>output of the command</summary>
<pre>                                
; File "test.smt2", line 7, characters 1-12: I don't know (0.4635) (7 steps) (goal g_1)

unknown
; Returned timeout reason = NoTimeout
(
  (define-fun x () Bool true)
  (define-fun y () Bool false)
  (define-fun z () Bool false)
)

; File "test.smt2", line 7, characters 1-12: I don't know (0.4636) (13 steps) (goal g_1)

unknown
; Returned timeout reason = NoTimeout
(
  (define-fun x () Bool false)
  (define-fun y () Bool true)
  (define-fun z () Bool true)
)

; File "test.smt2", line 7, characters 1-12: I don't know (0.4637) (17 steps) (goal g_1)

unknown
; Returned timeout reason = NoTimeout
(
  (define-fun x () Bool true)
  (define-fun y () Bool false)
  (define-fun z () Bool true)
)

; File "test.smt2", line 7, characters 1-12: I don't know (0.4638) (23 steps) (goal g_1)

unknown
; Returned timeout reason = NoTimeout
(
  (define-fun x () Bool true)
  (define-fun y () Bool true)
  (define-fun z () Bool true)
)

; File "test.smt2", line 7, characters 1-12: Valid (0.4638) (23 steps) (goal g_1)

unsat
</pre>
</details>

- A new keyword `check-all-sat` is available in the SMT-LIB input language only. For instance
```
(set-logic ALL)
(declare-const x Bool)
(declare-const y Bool)
(declare-const z Bool)
(assert (or x y))
(assert (or (not y) z))
(check-all-sat x y)
```
with the command `dune exec -- alt-ergo --produce-models --sat-solver CDCL --frontend legacy test.smt2`
outputs
<details>
  <summary>output of the commad</summary>
  <pre>
; File "test.smt2", line 7, characters 1-20: I don't know (0.0003) (7 steps) (goal g_1)

unknown
; Returned timeout reason = NoTimeout
(
  (define-fun x () Bool true)
  (define-fun y () Bool false)
  (define-fun z () Bool false)
)

; File "test.smt2", line 7, characters 1-20: I don't know (0.0005) (14 steps) (goal g_1)

unknown
; Returned timeout reason = NoTimeout
(
  (define-fun x () Bool false)
  (define-fun y () Bool true)
  (define-fun z () Bool true)
)

; File "test.smt2", line 7, characters 1-20: I don't know (0.0005) (16 steps) (goal g_1)

unknown
; Returned timeout reason = NoTimeout
(
  (define-fun x () Bool true)
  (define-fun y () Bool true)
  (define-fun z () Bool true)
)

; File "test.smt2", line 7, characters 1-20: Valid (0.0006) (16 steps) (goal g_1)

unsat
  </pre>
</details>
This feature is clearly more interesting as you can choose for which literals you want all the boolean models.

The PR includes also another option to print a boolean model while printing the first-order model. The feature is broken but I can restore it easily. I'm not sure if the option is really useful, at least while using the SMT-LIB input language, as there is a standard way to obtain this feature (in fact a more general feature), see #844. We should deprecate the option or reserve it for the native input language.

Some important remarks about the feature:
- The feature doesn't work properly with the `tableaux` solver. The reason is simple and I actually ran into this issue while testing outputted models with the Tableaux solver. I have a solution that I will explain in details in an appropriate issue.
- The solver outputs `unsat` at the end because it exhausted all the possible boolean models. This answer leaks the way AE works under the hood. I guess the best solution consists to print a comment telling we have exhausted all the boolean models. But fixing this issue will be easier after fixing #833.

### DIFF
--- a/src/bin/js/worker_js.ml
+++ b/src/bin/js/worker_js.ml
@@ -145,7 +145,107 @@ let main worker_id content =
         begin match kind with
           | Ty.Check
           | Ty.Cut -> { state with local = []; }
-          | Ty.Thm | Ty.Sat -> { state with global = []; local = []; }
+          | Ty.Thm | Ty.Sat | Ty.AllSat _ ->
+            { state with global = []; local = []; }
+        end
+      | Typed.TAxiom (_, s, _, _) when Ty.is_global_hyp s ->
+        let cnf = Cnf.make state.global td in
+        { state with global = cnf; }
+      | Typed.TAxiom (_, s, _, _) when Ty.is_local_hyp s ->
+        let cnf = Cnf.make state.local td in
+        { state with local = cnf; }
+      | _ ->
+        let cnf = Cnf.make state.ctx td in
+        { state with ctx = cnf; }
+    in
+
+    let (module I : Input.S) = Input.find (Options.get_frontend ()) in
+    let parsed () =
+      try
+        Options.Time.start ();
+        I.parse_file ~content ~format:None
+      with
+      | Parsing.Parse_error ->
+        Printer.print_err "%a" Errors.report
+          (Syntax_error ((Lexing.dummy_pos,Lexing.dummy_pos),""));
+        raise Exit
+      | Errors.Error e ->
+        begin match e with
+          | Errors.Run_error r -> begin
+              match r with
+              | Steps_limit _ ->
+                returned_status :=
+                  Worker_interface.LimitReached "Steps limit"
+              | _ -> returned_status := Worker_interface.Error "Run error"
+            end
+          | _ -> returned_status := Worker_interface.Error "Error"
+        end;
+        Printer.print_err "%a" Errors.report e;
+        raise Exit
+    in
+    let all_used_context = FE.init_all_used_context () in
+    let assertion_stack = Stack.create () in
+    let typing_loop state p =
+      try
+        let l, env = I.type_parsed state.env assertion_stack p in
+        List.fold_left (typed_loop all_used_context) { state with env; } l
+      with Errors.Error e ->
+        Printer.print_err "%a" Errors.report e;
+        raise Exit
+    in
+
+    let state = {
+      env = I.empty_env;
+      ctx = [];
+      local = [];
+      global = [];
+    } in
+
+    begin
+      try let _ : _ state =
+            Seq.fold_left typing_loop state (parsed ()) in ()
+      with Exit -> () end;
+
+    let get_status_and_print status n =
+      returned_status :=
+        begin match status with
+          | FE.Unsat _ -> Worker_interface.Unsat n
+          | FE.Inconsistent _ -> Worker_interface.Inconsistent n
+          | FE.Sat _ -> Worker_interface.Sat n
+          | FE.Unknown _ -> Worker_interface.Unknown n
+          | FE.Timeout _ -> Worker_interface.LimitReached "timeout"
+          | FE.Preprocess -> Worker_interface.Unknown n
+        end;
+      FE.print_status status n
+    in
+
+    let solve all_context (cnf, goal_name) =
+      let used_context = FE.choose_used_context all_context ~goal_name in
+      let consistent_dep_stack = Stack.create () in
+      SAT.reset_refs ();
+      let _,_,dep =
+        let env = SAT.empty_with_inst add_inst in
+        List.fold_left
+          (FE.process_decl
+             get_status_and_print used_context consistent_dep_stack)
+          (env, `Unknown env, Explanation.empty) cnf
+      in
+      if Options.get_unsat_core () then begin
+        unsat_core := Explanation.get_unsat_core dep;
+      end;
+    in
+
+    let typed_loop all_context state td =
+      match td.Typed.c with
+      | Typed.TGoal (_, kind, name, _) ->
+        let l = state.local @ state.global @ state.ctx in
+        let cnf = List.rev @@ Cnf.make l td in
+        let () = solve all_context (cnf, name) in
+        begin match kind with
+          | Ty.Check
+          | Ty.Cut -> { state with local = []; }
+          | Ty.Thm | Ty.Sat | Ty.AllSat _ ->
+            { state with global = []; local = []; }
         end
       | Typed.TAxiom (_, s, _, _) when Ty.is_global_hyp s ->
         let cnf = Cnf.make state.global td in
@@ -224,12 +324,10 @@ let main worker_id content =
             end
           | Some r ->
             let b,e = r.loc in
-            (r.name,b.Lexing.pos_lnum,e.Lexing.pos_lnum,
-             !nb,Worker_interface.Used)
+            (r.name,b.Lexing.pos_lnum,e.Lexing.pos_lnum,!nb,Worker_interface.Used)
             :: acc
         ) tbl []
     in
-
 
     (* returns a records with compatible worker_interface fields *)
     {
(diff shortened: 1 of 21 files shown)


src/bin/js/worker_js.ml
```diff
@@ -145,7 +145,107 @@ let main worker_id content =
         begin match kind with
           | Ty.Check
           | Ty.Cut -> { state with local = []; }
-          | Ty.Thm | Ty.Sat -> { state with global = []; local = []; }
+          | Ty.Thm | Ty.Sat | Ty.AllSat _ ->
```
Stevendeo, line 148: There seems to be twice the same code here. Why the duplication?

src/bin/text/main_text.ml
```diff
@@ -40,6 +40,7 @@ let parse_cmdline () =
   try Parse_command.parse_cmdline_arguments ()
   with Parse_command.Exit_parse_command i -> exit i
 
+
```
Stevendeo, line 43: Remove this line

src/lib/frontend/frontend.ml
```diff
@@ -280,25 +299,120 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
...
+         all-models listing *)
+      (*       print_model (SAT.get_model env) None; *)
       raise e
 
+  and process_unknown
+      print_status used_context consistent_dep_stack env t dep d
+      timeout_kind =
+    match d.st_decl with
+    | Assume _ | PredDef _ | RwtDef _ | Push _ | Pop _ | ThAssume _->
+      (* cannot raise Sat or Unknown in this case *)
```
Stevendeo, line 328: Can't alt-ergo Timeout during one of these steps?

src/lib/frontend/frontend.ml
```diff
@@ -280,25 +299,120 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
...
+        | AllSat l ->
+          (* 1.A SMT command check-all-sat *)
+          Some (List.fold_left
+                  (fun acc s ->
+                     (*transform string to propositional vars (E.t)*)
+                     let t = E.mk_term (Symbols.name s) [] Ty.Tbool in
+                     E.Set.add t acc
+                  )E.Set.empty l)
+
+        | Thm | Sat ->
```
Stevendeo, line 345: `| Thm | Sat when Options.get_all_models ()`

src/lib/frontend/frontend.ml
```diff
@@ -280,25 +299,120 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
...
+          Expr.Set.fold
+            (fun e acc -> E.mk_and e acc false) propositional E.vrai
+        in
+        if E.equal f E.vrai then
+          begin
+            (* this may happen if current propositional model is empty
+               (for instance, in case of timeout and no model computed
+               so far). We should stop to avoid infinite loop *)
+            if timeout_kind != NoTimeout then raise Util.Timeout;
+            (* TODO: is it the appropriate status? *)
```
Stevendeo, line 391: I don't know, is it?

src/lib/frontend/frontend.ml
```diff
@@ -280,25 +299,120 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
...
+             prevent process_decl to print itself results. *)
+          process_decl
+            print_status used_context consistent_dep_stack
+            (env, `Unknown t, dep) d
+
+      | _ ->
+        (* 2. default case + case where a simple interpretation is
+           requested *)
+        print_model t (Some timeout_kind);
+        (* TODO: is it the appropriate status? *)
```
Stevendeo, line 413: I don't know, is it?

src/lib/frontend/models.ml
```diff
@@ -490,7 +490,7 @@ let rec pp_value ppk ppf = function
 let pp_constant ppf (_sy, t) =
   Fmt.pf ppf "%a" SmtlibCounterExample.pp_abstract_value_of_type t
 
-let output_concrete_model fmt m =
+let output_concrete_model ~pp_prop_model:_ fmt m =
```
Stevendeo, line 493: Can't you remove this argument?

src/lib/structures/ty.mli
```diff
@@ -286,6 +286,9 @@ type goal_sort =
   (** The goal to be proved valid *)
   | Sat
   (** The goal to be proved satisfiable *)
+  | AllSat of string list
+  (** Rather generate all models involving the given list of
```
Stevendeo, line 290: Why 'Rather'?

bclement-ocp:
Before reviewing the code in detail, I want to better understand what we are trying to achieve with that feature, so I will only make high-level remarks here.

It was my understanding that this feature was required for OptimAE. After looking at the code, it doesn't appear to be the case, as OptimAE uses a different mechanism to iterate over the propositional models for optimization (in fact, in the OptimAE branch, the `all-models` feature has been implemented after the rest of the optimization work). Given that OptimAE does not need this feature, and that it is actually fairly small and self-contained (all changes but those in `frontend.ml` seem to be trivial plumbing), we can evaluate the feature on its own merit without rushing to merge the implementation as-is.

As an aside, this feature was initially implemented for an [unnamed intern project in 2011](https://frama-c.com/html/fc-discuss/2016-January/msg00016.html), dropped during the development of OptimAE (the removal was merged in #530) and then re-implemented later --- but if I had to wager, without that one-day hack in 2011, it would not have been implemented in OptimAE.

With that out of the way, I am not sure we want to merge this at all. I don't think there is much value in enumerating propositional models rather than actual models in general (in part because having a propositional model does not guarantee that a corresponding model of the input formula exists), and if we care about models, then model enumeration can be implemented outside the solver. 

The typical way to implement model enumeration uses incremental solving: first get a model using `(check-sat)`, get the value of the propositional term(s) you are interested in with `(get-assignment)` (see #844), assert that one of them should be false, then call `(check-sat)` again and repeat until you get unsat. Replace `(check-sat)` and `(get-assignment)` by the appropriate (incremental) API calls if not using SMT-LIB input.

I agree that providing this loop in a built-in way would be useful (in fact, the current implementation uses the exact loop I described above), but given that there is no rush for the feature I think it would be better to first clean up the interface to be able to do proper incremental solving (rather than giving a single list of commands and we're done, as we discussed) and to implement this feature on top of the incremental solving API once it exists (I think the relevant issue is #382 ).

Edit: as a side note, we should all this `check-allsat` rather than `check-all-sat` when using SMT-LIB to be compatible with MathSat5, see bottom of https://mathsat.fbk.eu/smt2examples.html

Halbaroth:
I agree to postpone this feature until we have a sufficient support for incremental mode.

Halbaroth:
I think we can close this PR. Keeping it up to date is painful and we do not plan to support this feature in the near future.

bclement-ocp:
agreed

PR #848, opened 2023-09-29
feat: get-assignment statement

Stevendeo:
This PR adds the `(get-assignment)` statement handling (issue #844)
- [x] Produce assertion option handler
- [x] Registering & accessing to named expressions in kernel
- [x] Evaluation of boolean named expression
- [x] SMT Output


bclement-ocp:
Please mark WIP PRs as draft so that others don't get notifications until it is ready (I just did it for this one). Thanks!

Stevendeo:
I did not see where to set the PR as a draft

bclement-ocp:
There is a drop-down on the "Create pull request" button:

image

For existing PRs there is a link below "Reviewers" (this is less effective WRT notifications because people are automatically subscribed once the PR becomes non-draft, but still indicates that the PR is still being worked on in the UI) :

image

To mark as non-draft once the PR is ready the button is at the bottom of the page:

image

src/bin/common/solving_loop.ml
```diff
@@ -297,6 +343,14 @@ let main () =
     State.create_key ~pipe:"" "optimize"
   in
 
+  let get_assignment: bool State.key =
```
bclement-ocp: This should be called `produce_assignments` rather than `get_assignment`, no?

src/bin/common/solving_loop.ml
```diff
@@ -95,6 +95,52 @@ let enable_maxsmt b =
...
+    | [] -> acc
+    | [] :: rest -> loop acc rest
+    | (term :: tl) :: rest ->
+      let terms_to_check = tl :: rest in
+      let terms_to_check =
+        match term.DStd.Expr.term_descr with
+        | DStd.Expr.Var _ | Cst _ -> terms_to_check
+        | App (t, _, tl) -> (t :: tl) :: terms_to_check
+        | Binder (b, t) ->
+          (t :: terms_of_dolmen_binder b) :: terms_to_check
```
bclement-ocp: Is this going to do the right thing wrt capture of bound variables (e.g. what happens if I do `(let ((x 2)) (+ (! x :named name)))`)?

src/bin/common/solving_loop.ml
```diff
@@ -778,6 +889,35 @@ let main () =
...
+                    ~get_value:(SAT.get_value partial_model)
+                    st
+                else
+                  recoverable_error
+                    "Produce assignments disabled; \
+                     add (set-option :produce-assignments true)";
+                st
+              | _ ->
+                recoverable_error
+                  "Model generation disactivated, cannot execute \
```
bclement-ocp: `disabled`, not `deactivated`. We may want to mention `(set-option :produce-model true)` here.

src/bin/common/solving_loop.ml
```diff
@@ -650,13 +714,58 @@ let main () =
...
+
+  let print_terms_assignments =
+    Fmt.list
+      ~sep:(fun fmt _ -> Fmt.pf fmt "@,")
+      (fun fmt (name, v) -> Fmt.pf fmt "(%s %s)" name v)
+  in
+
+  let handle_get_assignment ~get_value st =
+    let assignments =
+      Util.MS.fold
```
bclement-ocp, line 735: Nit: you can probably leverage `Fmt.iter_bindings` here.

src/bin/common/solving_loop.ml
bclement-ocp: Although do we actually need this case? It looks like we don't care about the result of `SAT.get_model` and I think it should work without?

bclement-ocp (changes requested):
We should figure out if we actually need model production for `get-assignment`. If we do, then `(set-option :produce-assignments true)` needs to activate model production, if we don't then we remove the check.

src/bin/common/solving_loop.ml
Stevendeo: I did not think about the semantics of named terms in bindings. Do you think we should not record the named formulas in bindings for now?

src/bin/common/solving_loop.ml
Stevendeo: The command should only be available in SAT mode, is that the good way to check it?

src/bin/common/solving_loop.ml
Stevendeo: I preferred to build the list and print it afterwards to make the code more readable

src/bin/common/solving_loop.ml
bclement-ocp: As the name indicates, the `partial_model_key` should only contain a value if we are in SAT mode, there is no need to also check `SAT.get_model` (even though AFAIK the call to `SAT.get_model` cannot fail currently).

src/bin/common/solving_loop.ml
bclement-ocp: If we don't have an easy way of checking capture avoidance then yes, that would be better (this is already what we do for quantifiers if I am not mistaken).

src/bin/common/solving_loop.ml
```diff
@@ -536,10 +587,18 @@ let main () =
               solver;
             st
       )
+    | ":produce-assignments",  Symbol { name = Simple b; _ } ->
+      begin
+        match bool_of_string_opt b with
+        | None ->
+          print_wrn_opt ~name:":verbosity" st_loc "integer" value;
```
```suggestion
          print_wrn_opt ~name:":verbosity" st_loc "boolean" value;
```

src/bin/common/solving_loop.ml
```diff
@@ -650,13 +709,58 @@ let main () =
...
+    in
+    match get_value ae_form with
+    | None -> "unknown" (* Not in the standard, but useful for recording when
+                           Alt-Ergo fails to guess the value of a term. *)
+    | Some v -> Fmt.to_to_string Expr.print v
+  in
+
+  let print_terms_assignments =
+    Fmt.list
+      ~sep:(fun fmt _ -> Fmt.pf fmt "@,")
```
bclement-ocp: Nit: Missed it earlier but `(fun ppf _ -> Fmt.pf ppf "@,")` is `Fmt.cut`.

bclement-ocp (commented):
We should have a test without `(set-option :produce-models true)`, otherwise looks good (modulo the `integer` / `boolean` typo).

bclement-ocp:
Thanks for the new tests. Can you add one without `(set-option :produce-models true)` (but with `(set-option :produce-assignments true)` as requested previously? Thanks

src/bin/common/solving_loop.ml
```diff
@@ -886,22 +886,32 @@ let main () =
...
+              st
+          end
+          else
+            begin
+              (* TODO: add the location of the statement. *)
+              recoverable_error
+                "(get-assignment) requires model generation, \
+                 but it is disabled (try --produce-models)";
+              st
+            end
```
bclement-ocp: I thought we determined earlier that `produce-assignments` should work without `produce-models`, why are we disabling it instead?

Stevendeo: I'm not sure why you requested an extra test for get-assignment if not for testing this specific case. Did you mean you wanted a non-error test with no models calculated but a get-assignment response anyway? Because there already was a test without `(set-option :produce-models true)`.
The standard states that, when in `sat` mode, "the solver will have identified a model A of the current logic". I think that's what confused me (the said model can be abstract, which is specified later on the standard).

bclement-ocp: Yes, I wanted a successful test :) Indeed the `sat` does not require to have built the model explicitly.

> Because there already was a test without (set-option :produce-models true).

Sorry, I must have missed it, but I still don't see it — can you point me to that test? The tests I see are `bool1`, `bool2`, `bool3` that have both `produce-models` and `produce-assignments`; `get-assignment1` that has neither, `get-assignment2` that has `produce-models` but not `produce-assignments` and only the new `get-assignment3` that has `produce-assignment` but not `produce-models`.

bclement-ocp (approved):

src/bin/common/solving_loop.ml
Stevendeo: > We should have a test without (set-option :produce-models true), otherwise looks good (modulo the integer / boolean typo).
 
`tests/smtlib2/testfile-get-assignment1.err.dolmen.smt2` indeed has neither

bclement-ocp: Ah! Sorry I was not clear, I meant without `produce-models` but with `produce-assignments` as clarified in my later message :)

Halbaroth (approved):

PR #852, opened 2023-10-02
Adding (reset) and (exit) statements

Stevendeo:
Adding the two instructions and associated tests. This PR is mostly motivated by removing the `assert false` raised on the `reset` instruction in the psmt2 frontend.

### DIFF
--- a/src/lib/frontend/cnf.mli
+++ b/src/lib/frontend/cnf.mli
@@ -31,10 +31,10 @@
 (* used in the typechecker for the text-mode *)
 val make :
   Commands.sat_tdecl list ->
-  ('a Typed.tdecl, 'a) Typed.annoted ->
+  _ Typed.atdecl ->
   Commands.sat_tdecl list
 
 (* used in the GUI *)
 val make_list :
-  ('a Typed.tdecl, 'a) Typed.annoted list ->
+  _ Typed.atdecl list ->
   Commands.sat_tdecl list
--- a/src/lib/frontend/cnf.mli
+++ b/src/lib/frontend/cnf.mli
@@ -31,10 +31,10 @@
 (* used in the typechecker for the text-mode *)
 val make :
   Commands.sat_tdecl list ->
-  (int Typed.tdecl, 'a) Typed.annoted ->
+  ('a Typed.tdecl, 'a) Typed.annoted ->
   Commands.sat_tdecl list
 
 (* used in the GUI *)
 val make_list :
-  (int Typed.tdecl, 'a) Typed.annoted list ->
+  ('a Typed.tdecl, 'a) Typed.annoted list ->
   Commands.sat_tdecl list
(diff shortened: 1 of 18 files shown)


bclement-ocp:
The PSMT2 frontend is deprecated, and is intended to be replaced with the Dolmen frontend — we generally don't add features to that frontend.

That said, in the current architecture of the frontend, I don't think `Reset` and `Exit` should be commands, because they only have an effect on the context we are building. They should be handled directly in the solving loop (`typed_loop` and `handle_stmt`) and immediately either reset the context or exit the solver, so that we don't keep the rest of the context around (with the current implementation, every `(check-sat)` after the `(reset)` will create a new solver with the old assertions then throw it away).

Stevendeo:
> The PSMT2 frontend is deprecated, and is intended to be replaced with the Dolmen frontend — we generally don't add features to that frontend.

Yeah I knew that, but as long as it is the default frontend I feel like it should not return `assert false` on a standard command.

Anyway, I agree with you about that fact they should not be commands

bclement-ocp:
> Yeah I knew that, but as long as it is the default frontend I feel like it should not return `assert false` on a standard command.

True, but a regular error would be fine too :) I agree that in this case we *should* support `(reset)` and `(exit)`, I am just pointing out that if it is easier to only add a feature to the Dolmen frontend, we shouldn't bother with adding it to the legacy frontend if it adds complexity.

Halbaroth:
I agree with Basile. We shouldn't modify at all the legacy frontend but adding new commands as you did shouldn't break anything. @Stevendeo Is your PR ready for reviewing?

Halbaroth (approved):
LGTM

src/lib/frontend/cnf.mli
```diff
@@ -31,10 +31,10 @@
...
 val make :
   Commands.sat_tdecl list ->
-  (int Typed.tdecl, 'a) Typed.annoted ->
+  ('a Typed.tdecl, 'a) Typed.annoted ->
   Commands.sat_tdecl list
 
 (* used in the GUI *)
 val make_list :
-  (int Typed.tdecl, 'a) Typed.annoted list ->
+  ('a Typed.tdecl, 'a) Typed.annoted list ->
```
bclement-ocp: Should this be `('a Typed.tdecl, 'b) Typed.annoted`?

bclement-ocp (approved):
Looks cleaner now, thanks! Just a doubt about a type annotation change, a bit surprised by the new type.

src/lib/frontend/cnf.mli
Stevendeo: No it must be the same type. I changed it by the alias defined in typed : `'a Typed.atdecl = ('a Typed.tdecl, 'a) annoted`

bclement-ocp: Hmm OK. I don't understand why `(int Typed.tdecl, 'a) Typed.annotated` worked before then :)  Anyways, using `atdecl` sounds good yeah, let's merge.

Issue #851, opened 2023-10-02
Inconsistent error handling strategy


bclement-ocp:
In SMT-LIB format:

- [ ] We should consistently honor the error behavior reported by `(get-info :error-behavior)`
- [ ] When an error occurs, we should return an `(error)` response even if we then exit (see e.g. [this comment](https://github.com/OCamlPro/alt-ergo/pull/843#discussion_r1340365349) for a case where we don't)

PR #853, opened 2023-10-02
Handling error strategy

Stevendeo:
Adresses #851


tests/cram.t/run.t
```diff
@@ -50,21 +50,21 @@ combinations of produce-models et al.
 First, if model generation is not enabled, we should error out when a
 `(get-model)` statement is issued:
 
-  $ echo '(get-model)' | alt-ergo --frontend dolmen -i smtlib2 -o smtlib2
+  $ echo '(get-model)' | alt-ergo --frontend dolmen -i smtlib2 -o smtlib2 --continue-on-error
```
bclement-ocp: Why do we need `--continue-on-error` here? Is it because cram tests require a 0 exit code?

src/bin/common/parse_command.ml
```diff
@@ -37,13 +37,21 @@ exception Error of bool * string
 (* Exception used to exit with corresponding retcode *)
 exception Exit_parse_command of int
 
+let pretty_error msg =
+  match Options.get_output_format () with
+  | Smtlib2 -> Format.asprintf "; %s" msg
+  | _ -> msg
+
```
bclement-ocp: We shouldn't be using SMT-LIB comments for error messages that come from argument parsing, because argument parsing is where we set the `output_format` and so whichever option is used depend on the order in which arguments are parsed. Further, errors during argument parsing are passed to `cmdliner` and should be consistent with `cmdliner`'s own error messages.

src/bin/common/parse_command.ml
```diff
@@ -496,8 +523,9 @@ let halt_opt version_info where =
     | Some w -> handle_where w; `Ok true
     | None -> if version_info then (handle_version_info version_info; `Ok true)
       else `Ok false
-  with Failure f -> `Error (false, f)
-     | Error (b, m) -> `Error (b, m)
+  with
+  | Failure f -> `Error (false, pretty_error f) (* S: dead code? *)
```
bclement-ocp: Nit: Having comments with questions/remarks on the code is good, but we should avoid using obscure shorthands in such notes --- using a full name or an explicit handle is better (I personally use `TODO(bclement):`).

src/bin/common/solving_loop.ml
```diff
@@ -50,8 +50,50 @@ let empty_solver_ctx = {
   global = [];
 }
 
-let unsupported_opt () =
-  Printer.print_std "unsupported"
+let custom_error
+    (type a)
+    ?(default : a option)
+    err_code
+    (msg : ('a, Format.formatter, unit, unit, unit, a) format6)
```
bclement-ocp: I understand the reason for it, but I feel like this type annotation is overly complicated for little benefit.

What about something like:

```ocaml
let recoverable_error ?(code = 1) msg =
  Format.kasprintf (fun msg ->
     let () =
      if msg <> "" then
        match Options.get_output_format () with
        | Smtlib2 -> Printer.print_smtlib_err "%s" msg
        | _ -> Printer.print_err "%s" msg
    in
    if Options.get_exit_on_error () then exit code)

let fatal_error ?code msg =
  recoverable_error ?code msg;
  exit (Option.value ~default:1 code)
```

The use of `fatal_error` vs `recoverable_error` is clearer than `error` vs `error ~default:v` IMO.

src/bin/common/solving_loop.ml
```diff
@@ -182,14 +224,15 @@ let main () =
       with
       | Util.Timeout ->
         FE.print_status (FE.Timeout None) 0;
-        exit 142
+        exit_as_timeout ()
       | Parsing.Parse_error ->
-        Printer.print_err "%a" Errors.report
-          (Syntax_error ((Lexing.dummy_pos,Lexing.dummy_pos),""));
-        exit 1
+        (* TODO: display a better error message *)
```
bclement-ocp: This is error messages for the legacy parser — the way to have better error messages is to use the Dolmen frontend (which we want people to use anyways).

bclement-ocp (changes requested):
I think the behavior in the solver is correct, but we should have some tests for the new "continue on error" mode. I'd rather avoid adding a n-th '.continue-on-error.' modifier to test file names, maybe we can just add a couple cram tests?

tests/cram.t/run.t
Stevendeo: Yes, if we don't add this option the output is suffixed with `[1]` (the error code); for clarity it is better to just ignore the error.

src/bin/common/parse_command.ml
Stevendeo: This is a best effort prettyfier. Also, I actually did that because of a test that added a ';' on an error message (when a plugin fails to be loaded). I can update the test, but I was not sure of the specification of the output message for parsing commands. I agree with you though, it would be better to just not consider parsing command errors as SMTLIB2 errors. I just did not want to argue on why I removed the ';' on the test.

bclement-ocp: Yes, removing the `;` should be fine in that case -- it looks out of place in the test and was probably kept there as a historical accident. This is a PR for cleaning up error strategies & error messages; it is expected that some error messages may slightly change :)

src/bin/common/solving_loop.ml
Stevendeo: Good, that means this comment will disappear when we switch to dolmen ;) In the meantime, there is explicitely in the code the printing of a dummy value: we should either remove it or document it. If we decide for any reason not to remove the legacy parser, it is better to have some documentation about it. I will make it more explicit.

bclement-ocp: Yeah, being more explicit with the comment (or even just not using `Errors.report` here because it doesn't really make sense) is OK but it would be a waste to dig too deep into this (unless we ultimately decide not to switch to Dolmen, as you say).

Now that I think about it -- the plan is to make Dolmen the default frontend for the next release, which means we should make it the default on `next` right now so that we can iron out the issues. I will make a PR to do so.

src/bin/common/parse_command.ml
Stevendeo: I'll revert these changes then :)

src/bin/common/solving_loop.ml
```diff
@@ -50,8 +50,34 @@ let empty_solver_ctx = {
...
 
-let unsupported_opt () =
-  Printer.print_std "unsupported"
+let recoverable_error ?(code = 1) =
+  Format.kasprintf (fun msg ->
+      let () =
+        if msg <> "" then
+          match Options.get_output_format () with
+          | Smtlib2 -> Printer.print_smtlib_err "%s" msg
+          | _ -> Printer.print_err "%s" msg
```
bclement-ocp, line 59: Food for thought (not in this PR): maybe this should be in `Printer.print_err`?

bclement-ocp (approved):
I think we are good :tada:

bclement-ocp:
I see this is marked as requesting a review from @hra687261 -- did you want him to take a look before we merge?

src/bin/common/solving_loop.ml
```diff
@@ -186,14 +211,18 @@ let main () =
       with
       | Util.Timeout ->
         FE.print_status (FE.Timeout None) 0;
-        exit 142
+        exit_as_timeout ()
       | Parsing.Parse_error ->
-        Printer.print_err "%a" Errors.report
-          (Syntax_error ((Lexing.dummy_pos,Lexing.dummy_pos),""));
-        exit 1
+        (* TODO(Steven): display a dummy value is a bad idea.
```
hra687261:
```suggestion
        (* TODO(Steven): displaying a dummy value is a bad idea.
```

src/bin/common/solving_loop.ml
```diff
@@ -546,15 +587,13 @@ let main () =
...
-            (* TODO: add the location of the statement. *)
-            Printer.print_smtlib_err
-              "Model generation disabled (try --produce-models)";
-            st
-          end
+          (* TODO: add the location of the statement. *)
+          let () =
+            recoverable_error "Model generation disabled (try --produce-models)"
+          in
+          st
```
hra687261:
```suggestion
          recoverable_error "Model generation disabled (try --produce-models)";
          st
```

hra687261 (approved):
LGTM! I don't have much to add to @bclement-ocp review

src/bin/common/solving_loop.ml
Stevendeo: We are in an ITE! I could use a `begin end` though...

bclement-ocp:
@Stevendeo there seems to be a line that is too long for the `check_style.sh` script, can you fix that so we can merge? Thanks :)

PR #856, opened 2023-10-03
[bitv] Add support for (bvnot) in the solver

### LINKED ISSUE #800: BV: Handle negation natively
Currently the bit-vector solver has limited support for negation (it can simplify negation within a single reasoning step, but not across reasoning steps / substitutions). Negation is fairly easy to handle in the solver directly: each bv has exactly one negated bv, and if we ever find one (atomic) bv and its negation in the same equivalence class after solving, we are inconsistent.

bclement-ocp:
This patch adds support for the negation (bvnot) operator in the
bitvector solver.

There are two components to this support:

 - First, the union-find data structure used by the solver is replaced
   with an actual union-find data structure implemented using Tarjan's
   efficient algorithm, rather than using an implementation with sets
   and maps. This is not only more efficient but also simplifies the
   implementation of the second part below.

 - Second, the new union-find data structure is augmented with a link
   between a class representative and the representative of its negated
   class. When merging two classes, their negated classes are merged as
   well (and, in particular, if a class gets forced to all ones or all
   zeroes, its negated classes gets forced to the other bit value). If a
   class is ever merged with its negated class, the problem is
   unsolvable.

Note that this also fixes a bug in the existing solver: previously, it
was possible to make bogus substitutions by asserting an equality
between a constant and a bit-vector term that depends on the constant,
for instance, `(= x (concat ((_ extract 0 0) x) ((_ extract 1 1) x)))`.
This would create the mappings `x -> x[0] @ x[1]` and `x[0] @ x[1] -> x`
in the `repr` table of the union-find, which is obviously incorrect and
confuses Alt-Ergo (see the `coherence` test).

The code is ready for review, although I have not ran the benchmarks
on BV and QF_BV yet.


bclement-ocp:
This is +8-0 on smtlib QF_BV. BV benchmarks are still running, I will fix the style issues once it is done to avoid issues with the bot.

bclement-ocp:
This is +3-0 on BV.

src/lib/reasoners/bitv.ml
```diff
@@ -64,16 +64,117 @@ let pp_sort ppf = function
...
-  | A, A | B, B | C, C -> 0
-  | A, (B | C) | B, C -> -1
-  | C, (A | B) | B, A -> 1
-
-type tvar = { var : int ; sorte : sort_var }
-
-let pp_tvar ppf { var; sorte } =
-  Format.fprintf ppf "%a_%d" pp_sort sorte var
+(** The [tvar] and [defn] type are used to implement Tarjan's Union-Find
+    algorithm with path compression but not union-by-rank.
```
Halbaroth, line 68: Why you don't use ranks as in the usual implementation of Tarjan's union-find? Is it because we have to kind of root in this union-find?

Halbaroth (approved):

src/lib/reasoners/bitv.ml
bclement-ocp: There are two reasons, mostly:

 - I didn't want to add extra complexity to the implementation — there would be some subtleties to take care of as you note (mostly you need to keep `Cte` when one of the members is a `Cte`), but union-by-rank would work otherwise and yield better complexity.
 - The size of the problems here are relatively small (for an equality on `(_ BitVec n)` there is at most one variable per bit, if I am not mistaken) that I'm not sure it'd matter too much in practice.

That said, it wouldn't hurt to have union-by-rank here, I can make an issue (however I have since realized a performance bug in the current representation of bitvectors which is that `x @ y = 0b0000000000` will use 2 variables but `x @ y = 0b1010101010` will use 10 and that I want to fix first, as it will have to change the representation of the solver a bit).

Halbaroth: I agree, we don't need a too complicated solution here. Let's merge.

PR #857, opened 2023-10-03
Make Dolmen the default frontend

### LINKED ISSUE #859: Update documentation for the frontend
This issue is a remainder to update the documentation as the default frontend of Alt-Ergo will change in 
the next release.

bclement-ocp:
The current plan is for Dolmen to be the default frontend in the next release. Making the switch as early as possible in the development branch will help us use it and test it more.


src/lib/util/options.ml
```diff
@@ -259,7 +259,7 @@ let output_with_colors = ref false
 let output_with_headers = ref true
 let output_with_formatting = ref true
 let output_with_forced_flush = ref true
-let frontend = ref "legacy"
+let frontend = ref "dolmen"
 let input_format = ref Native
```
Stevendeo: Should the default input format be Smtlib2 as well?

bclement-ocp: Somewhat confusingly, the default input format is never actually used, so it doesn't matter. It's fairly easy to clean that up, I will do it in another PR.

bclement-ocp: #858

Halbaroth (approved):
I create a remainder issue #859 to update the documentation for the frontend option. I agree you shouldn't change the documentation right now as some options could change again later.

bclement-ocp:
Hm, do we mention the frontend in the documentation outside of the release notes? I suppose we do, actually — then I'd rather update it now, it's not a huge change and I'd rather keep things coherent as much as possible.

bclement-ocp:
@Halbaroth can you take a look at the doc changes? Thanks!

Halbaroth (approved):
LGTM

PR #858, opened 2023-10-04
Remove infer_input_format option

bclement-ocp:
The use of `infer_input_format` / `input_format` is confusing:

 - By default, `infer_input_format` is `true`
 - We only ever look at `input_format` if `infer_input_format` is `false`
 - We only set `infer_input_format` to `false` when we explicitly set `input_format` to a non-default value

Since we can never look at the default value of `input_format`, this patch removes `infer_input_format` and makes `input_format` an `option` to better align with the way it is used.


Halbaroth (approved):
Thanks :)

PR #861, opened 2023-10-04
feat: OptimAE

Halbaroth:
This PR adds the value optimization in models.


src/lib/util/lists.ml
```diff
@@ -84,3 +84,18 @@ let rec equal eq l1 l2 =
...
+      else left, right
+    | x :: l ->
+      let left, right =
+        match f x with
+        | Ok x   -> x :: left, right
+        | Error x -> left, x :: right
+      in
+      part left right l
+  in
+  part [] [] l
```
Halbaroth: We should move this in `compat` module? Actually most of the `lists` module should be in `compat` but this function is not exactly the `partition_map` of the stdlib.

src/bin/common/parse_command.ml
```diff
@@ -361,8 +361,8 @@ let mk_limit_opt age_bound fm_cross_limit timelimit_interpretation
     `Ok()
 
 let mk_output_opt
-    interpretation use_underscore unsat_core output_format model_type
-    () () () ()
+    interpretation use_underscore objectives_in_interpretation unsat_core
+    output_format model_type () () () ()
```
Halbaroth, line 371: In my opinion, printing objectives is a debug message. We should print it on the diagnostic channel using SMT-LIB comments.

src/lib/frontend/models.ml
```diff
@@ -429,6 +437,29 @@ module SmtlibCounterExample = struct
          | _ -> ()
       ) fprofs;
     !records
+
+  let output_objectives fmt objectives =
+    (* TODO: we can decide to print objectives to stderr if
+       Options.get_objectives_in_interpretation() is enabled *)
```
Halbaroth, line 443: This should be the default behaviour indeed.

src/lib/reasoners/ac.ml
```diff
@@ -300,10 +304,55 @@ module Make (X : Sig.X) = struct
...
+            with
+            | Found kt ->
+              Some (kt , true)
+            | Not_found ->
+              let fresh = Expr.fresh_name ty in
+              Queue.push (fresh, X.term_embed fresh) !q;
+              cache := Ty.Map.add ty !q !cache;
+              Some (fresh , false)
+          end
+*)
```
Halbaroth, line 357: It seems the OptimAE contains a prototype for model generation while using AC symbols. We can keep it as a comment for later.

src/lib/reasoners/arith.ml
```diff
@@ -841,7 +841,12 @@ module Shostak
       | [] ->
         (* We do this, because terms of some semantic values created
            by CS are not created and added to UF *)
-        assert (P.is_const (embed r) != None);
+        if (P.is_const (embed r) == None) then begin
+          Printer.print_dbg
+            ~module_name:"Arith" ~function_name:"choose_adequate_model"
+            "no adequate model found for %a" X.print r;
+          assert false
+        end;
```
Halbaroth, line 850: We may replace it by a `Fmt.failwith`?

src/lib/reasoners/shostak.ml
```diff
@@ -635,7 +635,7 @@ struct
         if Expr.const_term t ||
            List.exists (fun (t,_) -> Expr.const_term t) eq then None
         else Some (Expr.fresh_name ty, false) (* false <-> not a case-split *)
-      | _               -> assert false
+      | _               -> AC.assign_value r distincts eq
```
Halbaroth: The function `assign_value` of the AC module does nothing.

src/lib/reasoners/enum_rel.ml
```diff
@@ -244,33 +244,40 @@ let assume env uf la =
...
-      Debug.case_split r r';
-      [LR.mkv_eq r r', true, Th_util.CS(Th_util.Th_sum, n)]
-    else []
-  | None ->
-    Debug.no_case_split ();
-    []
+let case_split env uf ~for_model ~to_optimize =
+  if to_optimize != None then
+    (* no classical splits if we have something to optimize *)
+    Sig_rel.Split []
```
Halbaroth: It seems we cannot do both optimization and case split.

src/lib/reasoners/relation.ml
```diff
@@ -122,23 +122,40 @@ let query env uf a =
...
+  let seq3 = Rel3.case_split env.r3 uf ~for_model ~to_optimize in
+  let seq4 = Rel4.case_split env.r4 uf ~for_model ~to_optimize in
+  let seq5 = Rel5.case_split env.r5 uf ~for_model ~to_optimize in
+  let seq6 = Rel6.case_split env.r6 uf ~for_model ~to_optimize in
+  let seq7 = Rel7.case_split env.r7 uf ~for_model ~to_optimize in
+  let splits, unbounded =
+    Lists.partition_map ~keep_ordering:false (function
+        | Sig_rel.Split l -> Ok l
+        | Optimized_split res -> Error res
+      ) [ seq1; seq2; seq3; seq4; seq5; seq6; seq7 ]
```
Halbaroth: We should use the constructor `Left` and `Right` of the stdlib here and redefine them in the module `compat`.

src/lib/reasoners/relation.ml
```diff
@@ -122,23 +122,40 @@ let query env uf a =
...
+  in
+  let unbounded =
+    List.fast_sort (fun a b -> a .Th_util.order - b.Th_util.order) unbounded in
```
Halbaroth: I'm not sure to understand the purpose of this sorting as the evaluation of the next statement will raise an uncaught exception if the list of unbounded splits contains several elements.

src/lib/reasoners/relation.ml
```diff
@@ -122,23 +122,40 @@ let query env uf a =
...
-       | _ -> assert false
-    )l
-
+  let seq1 = Rel1.case_split env.r1 uf ~for_model ~to_optimize in
+  let seq2 = Rel2.case_split env.r2 uf ~for_model ~to_optimize in
+  let seq3 = Rel3.case_split env.r3 uf ~for_model ~to_optimize in
+  let seq4 = Rel4.case_split env.r4 uf ~for_model ~to_optimize in
+  let seq5 = Rel5.case_split env.r5 uf ~for_model ~to_optimize in
+  let seq6 = Rel6.case_split env.r6 uf ~for_model ~to_optimize in
+  let seq7 = Rel7.case_split env.r7 uf ~for_model ~to_optimize in
```
Halbaroth: If I understand well, if the option `to_optimize` is of the form `Some _`, we don't do any case splits but we try to optimize one value. It would make more sense to add a new function `optimize` in `rel_sig.mli`.

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -1353,6 +1353,9 @@ let fm (module Oracle : S with type p = P.t) uf are_eq rclass_of env eqs =
 let is_num r =
   let ty = X.type_info r in ty == Ty.Tint || ty == Ty.Treal
 
+let is_num_term r =
+  let ty = E.type_info r in ty == Ty.Tint || ty == Ty.Treal
+
```
Halbaroth: We should use `is_num` function.

src/lib/reasoners/theory.ml
```diff
@@ -369,6 +373,62 @@ module Main_Default : S = struct
...
+     is coded differently *)
+  let next_optimization env ~for_model =
+    try
+      Util.MI.iter (fun _ x ->
+          match x.Th_util.value with
+          | Value _ ->
+            ()
+          | Pinfinity | Minfinity ->
+            (* We should block case-split at infinite values.
+               Otherwise. Otherwise, we may have soundness issues. We
```
Halbaroth:
```suggestion
            (* We should block case-split at infinite values.
               Otherwise, we may have soundness issues. We
```

src/lib/reasoners/theory.ml
```diff
@@ -369,6 +373,62 @@ module Main_Default : S = struct
...
+               may think an objective is unbounded, but some late
+               constraints may make it bounded.
+
+               An alternative is to only allow further splits when we
+               know that no extra-assumpptions will be propagated to
+               the env. Hence the test 'if for_model' *)
```
Halbaroth:
```suggestion
               know that no extra assumptions will be propagated to
               the env. Hence the test 'if for_model' *)
```

src/lib/util/lists.ml
bclement-ocp: `Compat` should only contain backports of `Stdlib` functions, so… let's keep this version for now.

src/bin/common/parse_command.ml
bclement-ocp: I agree but I am not sure how this is relevant for the lines you commented on? Do you want to remove the `objectives_in_interpretation` option?

src/lib/reasoners/arith.ml
bclement-ocp: I think having the extra info is good, but don't really care either way so my preference goes to the code that exists.

src/lib/reasoners/shostak.ml
bclement-ocp: Yeah, it just returns `None`. I am not sure why we go from `assert false` to `None` here, but it would be better to return `None` here directly and have a comment about AC terms.

src/lib/reasoners/enum_rel.ml
bclement-ocp: We can — the `case_split` function is used for both "regular" case splits and for optimizing case splits, probably for factorization (but I'm not a big fan of the result, it ends up being quite hard to read and understand). See `next_optimization` in `theory.ml`: as long as there is an objective for which we don't have a value, we try to optimize it; once we found a value (candidate optimum) for all the objectives, we start doing "regular" case splits. For most theories optimization does not make sense, and so they hold off until all we have a value for all the objectives before starting to make splits.

src/lib/reasoners/relation.ml
bclement-ocp: Yeah, I agree. Using the same `case_split` function for regular case splits and for optimization leads to confusing code. The name needs to be clear that this is still about making a choice, so I'd like to keep `split` somewhere. Maybe `best_split` or `optimizing_split`.

src/lib/reasoners/relation.ml
bclement-ocp: If we remove `to_optimize` and add an `optimize` or `best_split` function instead we won't need `partition_map` anymore, which would solve the problem. That said, there is a point where we should just give up and use https://github.com/thierry-martinez/stdcompat

src/lib/reasoners/relation.ml
bclement-ocp: The objectives are optimized lexicographically (according to `order`), and so I agree that it makes sense to keep them in that order when possible to avoid potential mistakes down the line.

We should at least add an explanatory comment warning people about caveats if multiple objectives at once become supported, but sorting a one-element list here is not going to be a performance penalty and so I'd rather keep the code, as a sort of defensive programming against our future selves.

src/lib/reasoners/intervalCalculus.ml
bclement-ocp: `is_num` has type `X.r -> bool`, this function has type `E.t -> bool`. Or do you mean we should check `is_num r` instead of `is_num_term e`?

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2082,20 +2085,95 @@ let model_from_unbounded_domains =
...
+    | Sim.Core.Unsat _ -> assert false (* we know sim is SAT *)
+    | Sim.Core.Unbounded _ ->
+      Format.eprintf
+        "%a is unbounded. Let other methods assign a value for it@."
+        E.print e;
+      let value = if to_max then Th_util.Pinfinity else Th_util.Minfinity in
+      Sig_rel.Optimized_split { opt with value }
+
+    | Sim.Core.Max(mx,_sol) ->
+      let {Sim.Core.max_v; _} = Lazy.force mx in
```
bclement-ocp: I am surprised we ignore `is_le` here. In fact, Alt-Ergo loops when given an objective that has a strict bound (e.g. maximize `x` under the constraint `x < 0.5`).

bclement-ocp (commented):
I think I roughly understand how this works now. As discussed there is probably some refactoring we should do before merging; but probably we shouldn't rewrite everything either.

One thing I really don't like is the way that optimization is added as terms with explicit order and then need to be internalized awkwardly. And once we get *that* in it will haunt us for a long time... One option would be to only add support for optimization in the SMT-LIB format with the Dolmen frontend (once Gbury/dolmen#190 is merged), because then we can have a slightly saner way of adding optimization objectives as commands.

And we can think about a different syntax for the native format (maybe using `optimize` instead of `goal`?), or just preprocess the raw `minimize` and `maximize` terms into commands earlier when typing goals.

I will let things sink in a bit and do a second pass later this week, probably tomorrow.

src/lib/reasoners/relation.ml
Halbaroth: I suggest to do it in a separated PR.

src/lib/reasoners/intervalCalculus.ml
Halbaroth: Oups, I didn't notice it wasn't the same type.

bclement-ocp: The fact that the term is called `r` doesn't help, to be fair :sweat_smile:

src/lib/reasoners/relation.ml
bclement-ocp: I don't know — I think we can do it here. You have done a great job over the last couple months maintaining the branch as some of the fixes were also made on the main branch, and then splitting what was once a huge PR into three self-contained and reasonably sized PRs (satML models, all-sat, and optimization). And since the PR is now feature-sized, we can evaluate it on its own, and I think it is reasonable to clean it up before merging. I can take that up if you're busy elsewhere.

Halbaroth: Ok, I will do it now.

src/bin/common/parse_command.ml
Halbaroth: The line isn't relevant. I don't want to remove it. We should add an extra case for the `get-info` statement to print the objectives on the diagnostic channel and keep this CLI option in case we use the native input language.

src/lib/reasoners/relation.ml
Halbaroth: After refactoring this part of the code, I got only one `optimized split` at this place and there is no array to sort. I let a comment to explain we need to sort them here if we support multiple optimizations of splits some day.

src/parsers/psmt2_to_alt_ergo.ml
```diff
@@ -456,6 +477,7 @@ module Translate = struct
     | Cmd_CheckAllSat _ -> not_supported "check-all-sat"; acc
     | Cmd_Maximize _ -> not_supported "maximize"; acc
     | Cmd_Minimize _ -> not_supported "minimize"; acc
+    | Cmd_Exit -> acc
```
Stevendeo: These last `Cmd_Maximize`, `Minimize` and `Exit` are redundant

Halbaroth: Thanks. I fixed it but I didn't push the commit yet ;) It was also the case for the two lines above it.

Stevendeo: I am concerned the CI passed with this. Should we reactivate this warning?

bclement-ocp: The warning is enabled but we don't have warning as errors on. It is good for local development (when building locally, we see the warnings and can make an informed decision about whether to treat them as errors now or fix them later to try things), but we should definitely use a profile with `-warn-error +A` on the CI, yes.

src/lib/reasoners/intervalCalculus.ml
Halbaroth: I fixed by giving up optimization when I got such optimization constraints.

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2035,6 +2041,79 @@ let case_split env uf ~for_model =
...
+          Some {Th_util.opt_ord = order; opt_val = Th_util.Value t2}
+        in
+        let s =
+          LR.mkv_eq r1 r2, true, Th_util.CS (o, Th_util.Th_arith, Q.one)
+        in
+        Some { opt_split with value = Value s; }
+
+      | Sim.Core.Max (lazy Sim.Core.{ is_le = false; _ }, _) ->
+        (* There is no upper bound as we try to optimize a strict bound. *)
+        None
```
bclement-ocp: This is not true for integers, I don't know if ocplib-simplex can return `is_le = false` for an integer problem here but we need to check (and probably assert `ty == Treal` here).

src/lib/reasoners/shostak.ml
```diff
@@ -646,7 +646,10 @@ struct
         if Expr.const_term t ||
            List.exists (fun (t,_) -> Expr.const_term t) eq then None
         else Some (Expr.fresh_name ty, false) (* false <-> not a case-split *)
-      | _               -> assert false
+      | _               ->
+        (* There is no mode-generation support for the AC symbols yet.
```
bclement-ocp:
```suggestion
        (* There is no model-generation support for the AC symbols yet.
```

src/lib/reasoners/sig_rel.mli
```diff
@@ -59,11 +59,14 @@ module type RELATION = sig
...
-    t -> Uf.t ->
-    for_model:bool ->
-    (Shostak.Combine.r Xliteral.view * bool * Th_util.lit_origin) list
+    t -> Uf.t -> for_model:bool -> Th_util.case_split list
   (** case_split env returns a list of equalities *)
 
+  val optimizing_split :
+    t -> Uf.t -> Th_util.optimized_split -> Th_util.optimized_split option
+  (** [optimizing_split env uf opt_split] try to optimize the value
+      contains in [opt_split]. *)
```
bclement-ocp:
```suggestion
  (** [optimizing_split env uf opt_split] try to optimize the expression
      contained in [opt_split]. *)
```

src/lib/reasoners/theory.ml
```diff
@@ -369,31 +373,118 @@ module Main_Default : S = struct
...
+           else
+             match v.Th_util.value with
+             | Th_util.Unknown -> acc (* not optimized yet *)
+             | Value _ -> Util.MI.add ord {v with value = Unknown} acc
+             | Pinfinity | Minfinity -> assert false (* may happen? *)
+        ) objectives objectives
 
   let look_for_sat ?(bad_last=None) ch env l ~for_model =
-    let rec aux ch bad_last dl env li =
+    let rec aux ~optimize ch bad_last dl env li =
```
bclement-ocp: This additional boolean flags make an already convoluted function even more convoluted.  I would suggest removing the flag and defining (either local or mutually recursive with aux) functions `optimizing_split` and  `case_split`, and to do something like:

```ocaml
match next_optimization ~for_model env with
| Some opt_split -> optimizing_split ch dl env opt_split
| None -> case_split ch dl env
```

in `aux` and:

```ocaml
match CC.optimizing_split env.gamma_finite opt_split with
| Some x -> (* same as currently *)
| None -> if for_model then case_split ch dl env else { env with choices = List.rev dl }, ch
```

in `optimizing_split`.

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1035,6 +1038,70 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+               seen_infinity := true; acc
+             | Minfinity ->
+               seen_infinity := true; acc
+             | Value (_,_, Th_util.CS (Some {opt_val = Value v; _},_,_)) ->
+               (* hack-ish to get the value of type expr *)
+               (e, v, is_max) :: acc
+             | Value _ ->
+               assert false
+             | Unknown ->
+               seen_infinity := true; acc
```
bclement-ocp: This doesn't seem correct. `Unknown` means `Unknown`, not `infinity`.

src/lib/reasoners/intervalCalculus.ml
bclement-ocp: Also I don't think returning `None` is good here — `None` means "I don't know how to optimize this expression", but we *do* know how to optimize this expression! We should have an additional constructor `StrictBound` in `optimized_split_value` to keep track of this instead.

src/lib/reasoners/satml_frontend.ml
bclement-ocp: See other comment above — if we use an explicit `StrictBound` as suggested above, we can get rid of this. Note that `StrictBound` needs to contain the value of the bound to be updated in the "should continue beyond SAT" below.

bclement-ocp (changes requested):
I think we are almost there — there are some places the code can be made clearer, but most importantly there are still remaining soundness issues wrt strict bounds. I believe we need to explicitly track the strict bounds in `optimized_split_value`, see my inline comments.

To demonstrate, with the following input file:

```smt-lib
; strictobj.smt2
(set-logic ALL)
(declare-const x Real)
(assert (or (<= x 20.0) (< x 10.0)))
(maximize x)
(check-sat)
```

I get:

```console
$ alt-ergo --frontend legacy strictobj.smt2 --dump-models 

; [Warning] optimize commands only work if the file contains check-sat


; [Debug] x is unbounded. Let other methods assign a value for it


; File "strictobj.smt2", line 5, characters 1-12: I don't know (0.0000) (3 steps) (goal g_1)

unknown
; Returned unknown reason = Incomplete
(
  (define-fun x () Real 0)
)
```

which is clearly incorrect (the maximum for `x` is `20.0`).

src/bin/common/parse_command.ml
Halbaroth: I open a remainder issue #891

src/parsers/psmt2_to_alt_ergo.ml
Halbaroth: Solved by #879

src/lib/reasoners/satml_frontend.ml
Halbaroth: I used an explicit `StrictBound` in this very location. I don't know why I modified this line. I revert the modification.
(Sorry I didn't push it but I forgot to modify the line on my local repository :D)

Halbaroth:
I have to do some clean-up before merging the PR but I think I fixed most of the issues. In particular, AE gives the right answer even if your problem includes strict bounds.

I start another PR in which I clean up the code gradually. 

Still, we continue to optimize bounds even if we see an impossible optimization. For instance, if your input file is:
``` 
(set-logic ALL)
(declare-const x Real)
(declare-const y Real)
(assert (< x 0.5))
(assert (<= y 2.0))
(maximize x)
(maximize y)
(check-sat)
(get-model)
```

Alt-Ergo will produce a correct model and optimize the variable `y`. The reason is because the default is to do 
casesplits after each asserts. It means AE cannot be sure after seeing the optimizing_split `x < 0.5`, there is no other constraints on `x` that makes the optimizing problem solvable. Fixing this problem is complicated without modifying a lot AE and the problem isn't a real issue as AE produces a correct model and respect the priority for optimization queries.

src/lib/reasoners/th_util.mli
```diff
@@ -30,13 +30,26 @@
...
+
+type limit_kind =
+  | Plus
+  | Minus
+  | None
+
+type 'a optimized_split_value =
+  | Minfinity
+  | Value of 'a * limit_kind
+  (** [Value (v, e)] represents a real with a possible epsilon. *)
```
Halbaroth: Oups this comment is outdated. I will modify it after your review ;)

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2026,6 +2026,12 @@ let model_from_unbounded_domains =
     let l2 = model_from_simplex int_sim true  env uf in
     List.fold_left mk_cs (List.fold_left mk_cs [] l1) l2
 
+let mk_const_term c ty =
+  let c = Q.to_string c in
+  match ty with
+  | Ty.Tint -> E.int c
+  | Ty.Treal -> E.real c
+  | _ -> assert false
```
bclement-ocp: Nit: now that we use `Z.t` and `Q.t` to represent integers and reals; it would be more efficient to do `E.Reals.of_Q c` and `E.Ints.of_Z (Q.to_z c)`.

Well, it wouldn't actually be more efficient because right now both of these also do a round-trip through `string`, but at least if/when we optimize those we would get the benefit here.

src/lib/reasoners/th_util.ml
```diff
@@ -30,16 +30,42 @@
...
   | Th_sum
   | Th_adt
   | Th_arrays
   | Th_UF
+[@@deriving show]
+
+type limit_kind =
+  | Plus
+  | Minus
+  | None
```
bclement-ocp: `None` is a well-known constructor of the `'a option` type and it is confusing to use it for another type, please choose an other name. Also, it would be good to have a comment explaining what `limit_kind` is used for — from the name of the type and constructors alone, I have absolutely no idea.

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2040,6 +2046,100 @@ let case_split env uf ~for_model =
...
+
+      | Sim.Core.Max (lazy Sim.Core.{ max_v; is_le }, _sol) ->
+        let max_p = Q.add max_v.bvalue.v c in
+        let optim = if to_max then max_p else Q.mult Q.m_one max_p in
+        let () =
+          if is_le then
+            Printer.print_dbg "%a has a %s: %a@."
+              E.print e
+              (if to_max then "maximum" else "minimum")
+              Q.print optim
```
bclement-ocp: This should either never print or always print; I think printing is good — if `is_le` is false we should use `"upper bound"` and `"lower bound"` instead of `"maximum"` and `"minimum"`.

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2040,6 +2046,100 @@ let case_split env uf ~for_model =
...
+            Some (LR.mkv_eq r1 r2, true, Th_util.CS (Th_util.Th_arith, Q.one))
+          else
+            (* As some bounds are strict, the value [r2] doesn't satisfy the
+               constraints of the problem. We don't produce a case-split
+               in this case because we don't want to propagate the equality
+               [r1 = r2] in the CC(X) environment [gamma_finite]. But
+               we propagate the new value [value] for this objective. Indeed,
+               we want to be able to print it while using the statement
+               [get-objective]. *)
+            None
```
bclement-ocp, line 2139: We can probably propagate something like `r1 > r2 - 1` instead to nudge model generation towards selecting "big" values of the optimized term. Or are there potential issues with cancellation?

src/lib/reasoners/relation.ml
```diff
@@ -131,14 +133,31 @@ let case_split env uf ~for_model =
...
+  | Ty.Trecord _ ->
+    Rel2.optimizing_split env.r2 uf opt_split
+  | Ty.Tbitv _ ->
+    Rel3.optimizing_split env.r3 uf opt_split
+  | Ty.Tfarray _ ->
+    Rel4.optimizing_split env.r4 uf opt_split
+  | Ty.Tadt _ ->
+    Rel5.optimizing_split env.r5 uf opt_split
+  | _ ->
+    opt_split
```
bclement-ocp: Ah, so this is why the `optimizing_split` are now returning the `opt_split` instead of `None`.

Can you explain the rational for this change? From a separation of concerns point of view, the `Relation` module should not have to care about which theories can or cannot optimize expressions of a given type.

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1035,6 +1038,110 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+
+     In our running example, after getting the model [2], the below function
+     run again the SAT solver with the extra assert:
+       (assert (or (> x 2) (and (= x 2) (> y 3)))
+
+     If this run produces the answer [unsat], we know that [2] was the best
+     model value for [x]. Otherwise, we got a better value for [x]. *)
+
+  let op is_max =
+    match is_max with
```
bclement-ocp: Don't use `match` on booleans — use `if` instead.

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1199,12 +1327,7 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
           )dep true
       end;
       dep
-    | (Util.Timeout | I_dont_know _) as e -> raise e
-    | e ->
-      Printer.print_dbg
-        ~module_name:"Satml_frontend" ~function_name:"unsat"
-        "%s" (Printexc.to_string e);
-      assert false
+    | (Util.Timeout | I_dont_know _ | Assert_failure _ ) as e -> raise e
```
bclement-ocp: If there is no catchall handler, this is the default behavior. On the fence about removing that debug print with `assert false` (if there is another assertion here we could be in big trouble and we want to know), but I think it's OK.

src/lib/reasoners/th_util.ml
bclement-ocp: After reading more of the code, looks like this is about knowing if the bound is strict or noe. Maybe: `Below | Above | Exact` would be clearer constructor names?

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1035,6 +1038,110 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+        raise unknown_exn;
+
+      | (e, tv, is_max) :: l ->
+        let neg =
+          List.fold_left
+            (fun acc (e, tv, is_max) ->
+               let eq = E.mk_eq e tv ~iff: false in
+               let acc = E.Core.and_ acc eq in
+               E.Core.(or_ acc (not ((op is_max) e tv)))
+            ) (E.Core.not ((op is_max) e tv)) l
```
bclement-ocp: In the case where we bound an upper bound rather than a maximum this is incorrect. As I mentioned earlier, the formula we generate here needs to take into account the strictness.

You can see the issue with the following input:
```smt-lib
(set-logic ALL)
(declare-const p1 Bool)
(declare-const p2 Bool)

(declare-const x Real)
(declare-const y Real)
(assert (or p1 p2))
(assert (=> p1 (< x 2.0)))
(assert (=> p2 (<= x 2.0)))
(maximize x)
(check-sat)
```

The current version of the code finds `2.0` as a *strict* bound, then keeps trying with the constraint `x > 2.0` and does not find a solution — but in fact, the `2.0` bound is reachable (the output is `(define-fun x () Real 0)` but changing `(=> p1 (< x 2.0))` to `(=> p1 (<= x 2.0))` correctly outputs `(define-fun x () Real 2)`).

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1095,6 +1202,27 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+          assert (!(env.objectives) != None);
+          (* objectives is a ref, it's necessiraly updated as a
+             side-effect to best value *)
+          raise (I_dont_know env)
+      end
+    | Util.Timeout as e -> raise e
+
+    | IUnsat (env, _) as e ->
+      if !(env.objectives) == None then raise e;
+      (* TODO: put the correct objectives *)
```
bclement-ocp, line 1230: Does this mean that we will output the objectives of the current SAT branch rather than the global objectives? But the objectives can only increase so that should actually be correct. Do you understand what "the correct objectives" means here?

src/lib/reasoners/theory.ml
```diff
@@ -338,110 +340,250 @@ module Main_Default : S = struct
...
+             The objective function given to Ocplib-simplex looks like:
+               5 * U + 2 * y + 3 where U = x * x
+             and the procedure will optimize the problem in terms of U and y. *)
+        assert false
+
+      | Pinfinity | Minfinity | Value (_, (Plus | Minus)) ->
+        (* We stop optimizing the split [opt_split] in this case, but
+           we continue to produce a model if the flag [for_model] is up. *)
+        if for_model then
+          propagate_choices env sem_facts acc_choices []
```
bclement-ocp: This should just be `aux env sem_facts acc_choices []`.

src/lib/reasoners/theory.ml
```diff
@@ -338,110 +340,250 @@ module Main_Default : S = struct
...
               Expr.print_tagged_classes classes;
-          aux ch None dl env [neg_c, lit_orig, CNeg, dep]
+          let objectives = partial_objectives_reset env.objectives order in
+          let env = { env with objectives } in
+          propagate_choices env sem_facts acc_choices
+            [neg_c, lit_orig, CNeg, dep, order]
+
+    and aux env sem_facts acc_choices new_choices =
+      Options.tool_req 3 "TR-CCX-CS-Case-Split";
+      match new_choices, next_optimization ~for_model env with
```
bclement-ocp: No need to call `next_optimization` if `new_choices` is not `[]`.

src/lib/reasoners/theory.ml
```diff
@@ -338,110 +340,250 @@ module Main_Default : S = struct
...
-          Options.tool_req 3 "TR-CCX-CS-Case-Split";
-          let l, base_env = CC_X.case_split env.gamma_finite ~for_model in
-          let env = {env with gamma_finite = base_env} in
-          match l with
-          | [] ->
-            { env with choices = List.rev dl }, ch
+          match opt_split.case_split with
+          | Some cs ->
+            let new_choice = add_explanations_to_split cs in
+            aux env sem_facts acc_choices [new_choice]
```
bclement-ocp: `aux` will always dispatch to `propagate_choices` here.

src/lib/reasoners/theory.ml
```diff
@@ -338,110 +340,250 @@ module Main_Default : S = struct
...
+           else
+             match v.Th_util.value with
+             | Th_util.Unknown -> acc (* not optimized yet *)
+             | Value _ ->
+               Util.MI.add ord {v with value = Unknown} acc
+             | Pinfinity | Minfinity ->
+               assert false (* may happen? *)
+        ) objectives objectives
+
+  let look_for_sat ~bad_last ~for_model env sem_facts new_choices =
```
bclement-ocp: Nit: we could take the opportunity to remove `bad_last` which is unused AFAICT.

src/lib/reasoners/intervalCalculus.ml
bclement-ocp: I think there are only issues with cancellation if we propagate under the optimization, if we add that split but still block optimization I believe it is OK (we do need to reset the objectives if it gets contradicted). Not a big deal though, we can do it later, let's leave it at it is for now.

src/lib/reasoners/theory.ml
```diff
@@ -480,34 +622,75 @@ module Main_Default : S = struct
...
             Options.tool_req 3 "TR-CCX-CS-Case-Split-Erase-Choices";
-            (* we replay the conflict in look_for_sat, so we can
-               safely ignore the explanation which is not useful *)
+            (* We replay the conflict in look_for_sat, so we can
+               safely ignore the explanation which is not useful. *)
             let uf =  CC_X.get_union_find t.gamma in
             let filt_choices = filter_choices uf t.choices in
+            let filt_choices =
+              if Util.MI.is_empty t.objectives ||
+                 has_no_infinity t.objectives
```
bclement-ocp: I believe we also must exclude strict bounds, not just infinities, here — they are not propagated either.  Although I *think* this is not actually an issue because we block splits under infinities/strict bounds now, but it's better to be safe and sound rather than reckless.

src/lib/reasoners/theory.ml
```diff
@@ -764,6 +1006,7 @@ module Main_Default : S = struct
       do_case_split_aux env ~for_model:true in
     CC_X.extract_concrete_model
       ~prop_model:env.assumed_set
+      ~optimized_splits:env.objectives
```
bclement-ocp: The objectives need to be taken from the result of `do_case_split_aux` above. Otherwise, we do not know about the optimizations made during model generation (under infinities/strict bounds), which breaks the invariants of `extract_concrete_model`. See for instance:

```smt-lib
(set-logic ALL)

(declare-const x Real)
(declare-const y Real)
(assert (< x 2.0))
(assert (<= y 10.0))
(maximize x)
(maximize y)
(check-sat)
```

crashes with `--objectives-in-interpretation`.

src/lib/reasoners/uf.ml
```diff
@@ -1006,89 +1006,184 @@ let is_a_good_model_value (x, _) =
...
+  MED.fold
+    (fun t mk acc ->
+       if SetX.mem mk pinfty then
+         (* mk's optimum is +infinity *)
+         compute_concrete_model_of_val env t acc (Some "+oo")
+       else
+       if SetX.mem mk minfty then
+         (* mk's optimum is -infinity *)
+         compute_concrete_model_of_val env t acc (Some "-oo")
+       else
```
bclement-ocp, line 1139: We should to have a 3rd set with `pinfty` and `minfty` to collect the strict bounds and have something like `(+ n epsilon)` and `(- n epsilon)` for the strict bounds.

bclement-ocp (changes requested):
There is still a crash and a tiny soundness bug with strict bounds that have to be fixed, I also have a few questions I am not sure about; the rest are minor perf remarks.

src/lib/reasoners/theory.ml
Halbaroth: I thought to do it in a separate PR but I do agree to remove it right now.

src/lib/reasoners/uf.ml
Halbaroth: Actually I have a better design for this. There is no reason to use a different ADT for the objective value during the model generation and printing objectives. I have a PR in which I use the same ADT for both. In particular, I use my `limit_kind` type to distinguish between `n`, `n + epsilon` and `n - epsilon`.  I will fix the printer in this PR also ;)

src/lib/reasoners/intervalCalculus.ml
Halbaroth: I will create a new debugging flag for optimization but I prefer do it in the next PR :)

I will remove the test in this one.

src/lib/reasoners/relation.ml
Halbaroth: In the previous code, we run `RelX.optimizing_split` on each split and stop after getting a `Some _` value. I changed it because in most of the codebase, we decide which of theories will manipulate a formula according to the type of its top symbol or with the constructor of `X.r`. As the type `X.r` is opaque, I used the type of the symbol.

src/lib/reasoners/theory.ml
Halbaroth: I wasn't completely sure of the semantic of this part of `try_it` so I didn't modify the function even if I knew I should do something. I will exclude it ;)

src/lib/reasoners/th_util.ml
Halbaroth: I agree that we shouldn't use `None`. The reason I use `Plus` or `Minus` is because I think the optimized value as a limit. In Math, we use to write `a+` for the limit from above and `a-` for the limit from below. I will use `Below | Above | Exact` and add documentation to explain their semantic.

src/lib/reasoners/theory.ml
Halbaroth: Good catch!

src/lib/reasoners/relation.ml
bclement-ocp: We generally only dispatch on the type or the semantic value "kind" for Shostak/CC(X) level operations. Things related to relations are usually independent of the type (for instance, all relations see equalities / disequalities / predicates etc. involving values of all types). I think it makes sense to keep it that way — whereas for Shostak theories there *needs* to be a separation by typing (see e.g. the infamous completeness issue for untyped theories) there is nothing preventing us from having multiple relations for the same type (e.g. linear vs non-linear arith).

src/lib/reasoners/adt_rel.ml
```diff
@@ -705,12 +707,14 @@ let case_split env _ ~for_model =
...
         (* cs on negative version would be better in general *)
         let cs =  LR.mkv_builtin false (Sy.IsConstr hs) [r] in
-        [ cs, true, Th_util.CS(Th_util.Th_adt, two) ]
+        [ cs, true, Th_util.CS (Th_util.Th_adt, two) ]
       with Not_found ->
         Debug.no_case_split ();
         []
     end
 
+let optimizing_split _env _uf opt_split = Some opt_split
```
bclement-ocp: This should return `None`.

src/lib/reasoners/arrays_rel.ml
```diff
@@ -396,6 +396,8 @@ let case_split env _ ~for_model:_ =
     Debug.case_split_none ();
     []
 
+let optimizing_split _env _uf opt_split = Some opt_split
```
bclement-ocp: Ditto.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -72,6 +72,7 @@ let case_split _ _ ~for_model:_ = []
 let add env uf r t =
   let delayed, eqs = Rel_utils.Delayed.add env.delayed uf r t in
   { delayed }, eqs
+let optimizing_split _env _uf opt_split = Some opt_split
```
bclement-ocp: Ditto.

src/lib/reasoners/enum_rel.ml
```diff
@@ -323,12 +323,15 @@ let case_split env uf ~for_model =
...
-      [LR.mkv_eq r r', true, Th_util.CS(Th_util.Th_sum, n)]
-    else []
+      [LR.mkv_eq r r', true, Th_util.CS (Th_util.Th_sum, n)]
+    else
+      []
   | None ->
     Debug.no_case_split ();
     []
 
+let optimizing_split _env _uf opt_split = Some opt_split
```
bclement-ocp: Ditto.

src/lib/reasoners/ite_rel.ml
```diff
@@ -221,7 +221,9 @@ let assume env uf la =
       raise e
   else assume env uf la
 
-let case_split _ _ ~for_model:_ = []
+let case_split _env _uf ~for_model:_ = []
+
+let optimizing_split _env _uf opt_split = Some opt_split
```
bclement-ocp: Ditto.

src/lib/reasoners/records_rel.ml
```diff
@@ -34,7 +34,8 @@ let empty _ = ()
 let assume _ _ _ =
   (), { Sig_rel.assume = []; remove = []}
 let query _ _ _ = None
-let case_split _ _ ~for_model:_ = []
+let case_split _env _uf ~for_model:_ = []
+let optimizing_split _env _uf opt_split = Some opt_split
```
bclement-ocp: Ditto.

src/lib/reasoners/relation.ml
```diff
@@ -131,14 +133,26 @@ let case_split env uf ~for_model =
...
+
+let (let*) = Option.bind
 
+let optimizing_split env uf opt_split =
+  Options.exec_thread_yield ();
+  let* opt_split = Rel1.optimizing_split env.r1 uf opt_split in
+  let* opt_split = Rel2.optimizing_split env.r2 uf opt_split in
+  let* opt_split = Rel3.optimizing_split env.r3 uf opt_split in
+  let* opt_split = Rel4.optimizing_split env.r4 uf opt_split in
+  let* opt_split = Rel5.optimizing_split env.r5 uf opt_split in
```
bclement-ocp: I don't think this is correct. Here we are requesting that all relations optimize the expression. We only expect one relation to optimize an expression, even if we don't know which one.

So we should stop once one of the relations returns `Some opt_split` (ideally, we should call them all and `assert` that there is only one that returned an optimization to avoid potentially gnarly bugs in the future, as was done in the original code).

src/lib/reasoners/satml_frontend.ml
bclement-ocp: > Don't use `match` on booleans — use `if` instead.

This is still relevant.

src/lib/reasoners/satml_frontend.ml
bclement-ocp: > In the case where we bound an upper bound rather than a maximum this is incorrect. As I mentioned earlier, the formula we generate here needs to take into account the strictness.
> 
> You can see the issue with the following input:
> 
> ```
> (set-logic ALL)
> (declare-const p1 Bool)
> (declare-const p2 Bool)
> 
> (declare-const x Real)
> (declare-const y Real)
> (assert (or p1 p2))
> (assert (=> p1 (< x 2.0)))
> (assert (=> p2 (<= x 2.0)))
> (maximize x)
> (check-sat)
> ```
> 
> The current version of the code finds `2.0` as a _strict_ bound, then keeps trying with the constraint `x > 2.0` and does not find a solution — but in fact, the `2.0` bound is reachable (the output is `(define-fun x () Real 0)` but changing `(=> p1 (< x 2.0))` to `(=> p1 (<= x 2.0))` correctly outputs `(define-fun x () Real 2)`).

This is still relevant.

src/lib/reasoners/satml_frontend.ml
bclement-ocp: > Does this mean that we will output the objectives of the current SAT branch rather than the global objectives? But the objectives can only increase so that should actually be correct. Do you understand what "the correct objectives" means here?

Repeating my question, I don't think you originally wrote this `TODO`, and I don't expect that you necessarily know the answer but given that you have been working on the PR, do you understand what it refers to (and if there are possible correctness implications)?

src/lib/reasoners/satml_frontend.ml
bclement-ocp: > If there is no catchall handler, this is the default behavior. On the fence about removing that debug print with `assert false` (if there is another assertion here we could be in big trouble and we want to know), but I think it's OK.

This is still relevant.

src/lib/reasoners/th_util.mli
bclement-ocp: > Oups this comment is outdated. I will modify it after your review ;)

I guess you should modify the comment now?

src/lib/reasoners/th_util.mli
```diff
@@ -30,13 +30,28 @@
...
   | Th_arrays
   | Th_UF
+[@@deriving show]
+
+type limit_kind =
+  | Above
+  | Below
+  | Exact
+  (* Type used to discriminate between limits from above or below. The
+     [Exact] constructor means the value isn't a limit but an exact value. *)
```
bclement-ocp: I think you are using this type to describe that "this is a limit from above" (the value is above the limit) and "this is a limit from below" (the value is below the limit), whereas my initial mental model was "the value is bounded above" (the bound is above the value) and "the value is bounded below" (the bound is below the value) which are exact opposites.

Thinking about it a bit more i do think your interpretation is more intuitive (especially if the order of the arguments in `Value` are changed: `Value (Below, 2.0)`, `Value (Exact, 2.0)`, etc.) but I am fine with either interpretations as long as the one used is documented somewhere :)

src/lib/reasoners/theory.ml
bclement-ocp: > This should just be `aux env sem_facts acc_choices []`.

This is still relevant.

bclement-ocp (changes requested):
There is still an issue with the way relations are handled, I believe.

I also have repeated a number of comments from my previous review that were silently ignored. All review comments that are not nits (but especially ones that exhibit incorrect behavior) should be addressed (either by making the required changes, providing requested details, otherwise engaging with the comment, or indirectly through a global message explaining why some comments were ignored) before re-requesting a review.

src/lib/reasoners/satml_frontend.ml
Halbaroth: Sorry! I didn't see this batch of reviews. There were hidden by GitHub. I thought GitHub hid only resolved conversations...

src/lib/reasoners/satml_frontend.ml
Halbaroth: Same here :/

src/lib/reasoners/satml_frontend.ml
bclement-ocp: Ah, right, I understand — GitHub UI is confusing at times. Only resolved conversations are *collapsed*:

image

but sometimes there are *hidden* items, with this type of marking

image

which can include any sort of things (unresolved comments, commits, etc.). Particularly annoying on bad network connections, I should note.

For PRs that have had many comments, I also like to do a quick check in the "Review" tab — it should show comments that apply to lines of code that have not changed.

src/lib/reasoners/satml_frontend.ml
Halbaroth: Actually, I knew the code was suspicious but I didn't manage to trigger the bug. I thought replacing the new constraint `x > 2.0` by `x >= 2.0` could produce an infinity loop. I guess there is no infinity loop because the constraint `x >= 2.0` and `x <= 2.0` will produce the equation `x = 2.0` and the `optimizing_split` of `IntervalCalculus` produces an equation case-split in this case. I restore a previous patch which pass the test now ;)

src/lib/reasoners/th_util.mli
Halbaroth: I think the best ADT would be:
```ocaml
type limit_kind =
| Above
| Below 

type 'a optimized_split_value =
| Minfinity
| Pinfinity
| Value of 'a 
| Limit of limit_kind * 'a
```
I don't like `Exact` or `None`. I agree `None` is worse than `Exact` but `Exact` means nothing for a limit.

bclement-ocp: That is because you think of it as a limit whereas I think of it as a bound on the set of possible values, in which case an "exact bound" makes sense :) But that ADT is also fine, yes.

src/lib/reasoners/satml_frontend.ml
Halbaroth: First of all, the mutable field `env.objectives` is only read in the wrapper around `unsat_rec` to outputs `unknown` if we got `unsat` while exploring a second branch during optimization in `analyze_unknown_for_objectives`. So `env.objectives` is just a flag.

Now, the field `env.objectives` is up to date at the beginning of `analyze_unknown_for_objectives`. If the problem is `sat`, this function is called at least once and the current state of objective in `Theory` is put in `env.objectives`. So, in my opinion `env.objectives` contains the latest valid version of the objectives. During the last call of `analyze_unknown_for_objectives`, we could imagine the solver produces a wrong model as the extended problem is `unsat` but this invalid version isn't in `env.objectives`. I guess the comment is outdated.

bclement-ocp: Yeah, that matches my understanding as well. But I think that this is actually unrelated to `env.objectives` —  I possibly this is related to the fact that the env is not updated? As in, the `env.objectives` is correct (in the sense that it is the optimum) but the `env` that we return in `I_dont_know` does *not* satisfy the objectives anymore (since it is an unsat environment). And model generation still works because when we compute the model we also use a reference to a lazy term that uses the *last known sat env* instead of using the current env…

Halbaroth: Good remark. I will investigate on it later. I think we should merge now :)

src/lib/reasoners/satml_frontend.ml
bclement-ocp: > I didn't manage to trigger the bug

To be fair it was a bit tricky to manage to do so :) In general however it is good to mention it (either as a comment in the code or on the PR) when you know that some code is suspicious or you are unsure it is correct, it hints at the reviewer to do a more careful review of that part of the code and helps reduce (but not eliminate, sadly) bugs.

bclement-ocp:
I think we are almost ready to merge this :) Thanks for all your hard work on this one. I want to do one last "full" review before approving — I am not sure I will have the time today, but it will be on the top of my list on Monday.

Halbaroth:
Thanks for your multiple reviews :D

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1035,6 +1038,117 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+        raise unknown_exn;
+
+      | (e, tv, is_max, is_le) :: l ->
+        let neg =
+          List.fold_left
+            (fun acc (e, tv, is_max, is_le) ->
+               let eq = E.mk_eq e tv ~iff: false in
+               let acc = E.Core.and_ acc eq in
+               E.Core.(or_ acc (not ((op is_max is_le) e tv)))
+            ) (E.Core.not ((op is_max is_le) e tv)) l
```
bclement-ocp, line 1125: Assuming `is_max` is always `true`:

Inside the `fold_left`, if we encounter a second expression.
 - With `is_le = true`, we have found `e ≤ tv`. We need to look for *either* `e > tv`, or  (due to the lexicographic ordering) `e = tv` and `acc`, which is what we build.
 - With `is_le = false`, we have found `e < tv`. To contradict that, we just need to look for `e ≥ tv` — the values for the other terms cannot be optimized yet (note that this does mean that once we find a "strict" bound the rest is best-effort, which we know already). Currently we look for `e ≥ tv ∨ (e = tv ∧ acc)`, but the second branch is subsumed by the first (`e = tv` implies `e ≥ tv`).

So technically this is correct but we should just return `E.Core.(not ((op is_max is_le) e tv))` when `is_le` is `true` in the accumulator.

src/lib/reasoners/theory.ml
```diff
@@ -338,110 +340,244 @@ module Main_Default : S = struct
...
+      Util.MI.fold
+        (fun ord v acc ->
+           if ord < opt_ord then
+             (*don't change older optims that are still valid splits*)
+             acc
+           else
+             match v.Th_util.value with
+             | Th_util.Unknown -> acc (* not optimized yet *)
+             | Value _ | Limit _ ->
+               Util.MI.add ord {v with value = Unknown} acc
```
bclement-ocp, line 419: I don't think the `Limit` case can happen here? Should be with `Pinfinity` and `Minfinity` (we block optims under all of these -- see `next_optimization`).

bclement-ocp (approved):
OK, I think we are good now :) Just a couple comments on minor points I had missed earlier, and then we can merge.

Thinking about it more, I am still not happy with the use of limit terminology, and I think I am finally able to articulate why. A limit typically implies that there is an input quantity that tends towards a value — but here there is no such input quantity. In particular, the terms "limit from above" and "limit from below" are (to my knowledge) used to describe the direction an input quantity (variable) is coming from (e.g. the limit, but here we are using it to describe the possible values of an output quantity (arbitrary multi-variate expression).

For instance, the limit from below of `x → ite(x ≥ 0, 17, |x|)` in `0` is `0` and refers to the fact that `x` tends towards `0` from negative values — even though the value of the function is always positive, and the limit from above is 17. But if we minimize this function Alt-Ergo would find `Limit (Above, 0)`, which feels wrong given the above (no pun intended).

Anyways, I am OK with merging with the limit terminology, there is no need for more bikeshedding here, but I wanted to explain my discomfort with the terminology.

src/lib/reasoners/satml_frontend.ml
Halbaroth: As it's an optimization, I will apply it in #921 ;)

Halbaroth:
> OK, I think we are good now :) Just a couple comments on minor points I had missed earlier, and then we can merge.
> 
> Thinking about it more, I am still not happy with the use of limit terminology, and I think I am finally able to articulate why. A limit typically implies that there is an input quantity that tends towards a value — but here there is no such input quantity. In particular, the terms "limit from above" and "limit from below" are (to my knowledge) used to describe the direction an input quantity (variable) is coming from (e.g. the limit, but here we are using it to describe the possible values of an output quantity (arbitrary multi-variate expression).
> 
> For instance, the limit from below of `x → ite(x ≥ 0, 17, |x|)` in `0` is `0` and refers to the fact that `x` tends towards `0` from negative values — even though the value of the function is always positive, and the limit from above is 17. But if we minimize this function Alt-Ergo would find `Limit (Above, 0)`, which feels wrong given the above (no pun intended).
> 
> Anyways, I am OK with merging with the limit terminology, there is no need for more bikeshedding here, but I wanted to explain my discomfort with the terminology.

I agree that it's not a good terminology, I'll change it in #921.

Issue #377, opened 2020-12-08
[design] should remove all "sensitive" global refs in lib modules (sat solvers, ...)


OriginLabs-Iguernlala:
If some soft uses alt-ergo-lib and maintains two instances of the lib's state, this is not correct.

PR #863, opened 2023-10-05
feat: support (get-info :all-statistics) and refactoring Profiling module

Stevendeo:
This PR was initially open to add the get-statistics option. Some bugs and (IMHO) bad design choices in the Timer and Profiling module changed the purpose of this PR to not only add the get-statistics feature, but also to rework it.

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -392,6 +392,7 @@ let main () =
 
   let handle_option st_loc name (value : DStd.Term.t) =
     match name, value.term with
+    (* Smtlib2 regular options *)
     | ":regular-output-channel", Symbol { name = Simple name; _ } ->
       Options.Output.create_channel name
       |> Options.Output.set_regular
@@ -444,6 +445,13 @@ let main () =
       | ":random-seed"), _
       ->
       unsupported_opt name
+    (* Alt-Ergo custom options *)
+    | ":profiling", Symbol { name = Simple level; _ } ->
+      begin
+        match float_of_string_opt level with
+        | None -> print_wrn_opt ~name st_loc "nonnegative integer" value
+        | Some i -> Options.set_profiling true i
+      end
     | _ ->
       unsupported_opt name
   in
@@ -486,9 +494,9 @@ let main () =
       print_std Fmt.string Version._version
     | ":all-statistics" ->
       if Options.get_profiling () then
-        Printer.print_std "%t" Profiling.print_get_statistics
+        Printer.print_std "%t" Profiling.print_statistics
       else
-        Printer.print_smtlib_err "Profiling disactivated (try --profiling)"
+        warning "Profiling disactivated (try --profiling)"
     | ":assertion-stack-levels" ->
       unsupported_opt name
     | _ ->
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -120,7 +120,6 @@ let main () =
       if Options.get_profiling() then
         Profiling.print true
           (Steps.get_steps ())
-          (Signals_profiling.get_timers ())
           (Options.Output.get_fmt_diagnostic ());
       (* If the status of the SAT environment is inconsistent,
          we have to drop the partial model in order to prevent
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -493,9 +493,8 @@ let main () =
     | ":version" ->
       print_std Fmt.string Version._version
     | ":all-statistics" ->
-      if Options.get_profiling () then
-        Printer.print_std "%t" Profiling.print_statistics
-      else
+      Printer.print_std "%t" Profiling.print_statistics;
+      if not (Options.get_profiling ()) then
         warning "Profiling disactivated (try --profiling)"
     | ":assertion-stack-levels" ->
       unsupported_opt name
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -484,7 +484,11 @@ let main () =
       pp_reason_unknown st
     | ":version" ->
       print_std Fmt.string Version._version
-    | ":all-statistics"
+    | ":all-statistics" ->
+      if Options.get_profiling () then
+        Printer.print_std "%t" Profiling.print_get_statistics
+      else
+        Printer.print_smtlib_err "Profiling disactivated (try --profiling)"
     | ":assertion-stack-levels" ->
       unsupported_opt name
     | _ ->
--- a/src/lib/structures/profiling.ml
+++ b/src/lib/structures/profiling.ml
@@ -709,14 +709,17 @@ let register_int_stat (key : string) (getter : unit -> int) =
   register_stat key (fun () -> string_of_int (getter ()))
 
 let register_float_stat (key : string) (getter : unit -> float) =
-  register_stat key (fun () -> string_of_float (getter ()))
+  register_stat
+    key
+    (fun () ->
+       Format.sprintf "%.9f" (getter ()))
 
 let print_get_statistics fmt =
-  Format.fprintf fmt "(@[<v 0>";
+  Format.fprintf fmt "(@[<v 0>@,";
   Hashtbl.iter
     (fun key getter -> Format.fprintf fmt ":%s %s@," key (getter ()))
     statistics_table;
-  Format.fprintf fmt ")@]"
+  Format.fprintf fmt "@])@,"
 
 (* Now, registering the statistics we want in get-statistics. *)
 let () =
--- a/src/lib/structures/profiling.ml
+++ b/src/lib/structures/profiling.ml
@@ -127,33 +127,6 @@ let set_sigprof () =
     (Unix.setitimer Unix.ITIMER_PROF
        { Unix.it_value = tm; Unix.it_interval = 0. })
 
-let init () =
-  let state = Lazy.force state in
-  state.decisions <- 0;
-  state.assumes <- 0;
-  state.queries <- 0;
-  state.instantiation_rounds <- 0;
-  state.instances <- 0;
-  state.decision_lvl <- 0;
-  state.instantiation_lvl <- 0;
-  state.assumes_current_lvl <- 0;
-  state.th_conflicts <- 0;
-  state.b_conflicts <- 0;
-  state.bcp_th_conflicts <- 0;
-  state.bcp_b_conflicts <- 0;
-  state.bcp_mix_conflicts <- 0;
-  state.t_red <- 0;
-  state.b_red <- 0;
-  state.t_elim <- 0;
-  state.b_elim <- 0;
-  state.instances_map <- MS.empty;
-  state.instances_map_printed <- false;
-  let timers = get_timers () in
-  Timers.set_timer_start (Timers.start timers);
-  Timers.set_timer_pause (Timers.pause timers);
-  Timers.reset timers;
-  set_sigprof ()
-
 (* update functions of the internal state *)
 
 let assume nb =
@@ -590,8 +563,8 @@ let print_timers header steps fmt =
     )columns;
   Format.fprintf fmt "|@."
 
-
 let print_instances_generation forced _steps fmt =
+  let state = Lazy.force state in
   let (@@) a b = if a <> 0 then a else b in
   if not forced && state.instances_map_printed then
     Format.fprintf fmt "[Instances profiling] No change since last print@."
@@ -771,19 +744,45 @@ let register_float_stat (key : string) (getter : unit -> float) =
        Format.sprintf "%.9f" (getter ()))
 
 let print_statistics fmt =
-  Fmt.pf fmt "(@[<v 0>@,";
+  Fmt.pf fmt "(@[<v 0>";
   Hashtbl.iter
-    (fun key getter -> Fmt.pf fmt ":%s %s@," key (getter ()))
+    (fun key getter -> Fmt.pf fmt "@,:%s %s" key (getter ()))
     statistics_table;
   Fmt.pf fmt "@])@,"
 
-(* Now, registering the statistics we want in (get-info :all-statistics). *)
-let () =
+let init () =
+  let state = Lazy.force state in
+  state.decisions <- 0;
+  state.assumes <- 0;
+  state.queries <- 0;
+  state.instantiation_rounds <- 0;
+  state.instances <- 0;
+  state.decision_lvl <- 0;
+  state.instantiation_lvl <- 0;
+  state.assumes_current_lvl <- 0;
+  state.th_conflicts <- 0;
+  state.b_conflicts <- 0;
+  state.bcp_th_conflicts <- 0;
+  state.bcp_b_conflicts <- 0;
+  state.bcp_mix_conflicts <- 0;
+  state.t_red <- 0;
+  state.b_red <- 0;
+  state.t_elim <- 0;
+  state.b_elim <- 0;
+  state.instances_map <- MS.empty;
+  state.instances_map_printed <- false;
+  let timers = get_timers () in
+  Timers.set_timer_start (Timers.start timers);
+  Timers.set_timer_pause (Timers.pause timers);
+  Timers.reset timers;
+  set_sigprof ();
+
+  (* Now, registering the statistics we want in (get-info :all-statistics). *)
   register_int_stat "steps" (fun () -> Steps.get_steps ());
   List.iter
     (fun (m : Timers.ty_module) ->
        let name = Format.sprintf "timer-%s" (Timers.string_of_ty_module m) in
        register_float_stat
          name
-         (fun () -> Timers.get_sum (Lazy.force state.timers) m))
+         (fun () -> Timers.get_sum state.timers m))
     Timers.all_modules
--- a/src/lib/structures/profiling.ml
+++ b/src/lib/structures/profiling.ml
@@ -45,210 +45,28 @@ type inst_info = {
 }
 
 type t = {
-  decisions : int ref;
-  assumes : int ref;
-  assumes_current_lvl : int ref;
-  queries : int ref;
-  instantiation_rounds : int ref;
-  instances : int ref;
-  decision_lvl : int ref;
-  instantiation_lvl : int ref;
-
+  mutable decisions : int;
+  mutable assumes : int;
+  mutable assumes_current_lvl : int;
+  mutable queries : int;
+  mutable instantiation_rounds : int;
+  mutable instances : int;
+  mutable decision_lvl : int;
+  mutable instantiation_lvl : int;
   (* 4 kinds of conflicts *)
-  th_conflicts : int ref;
-  b_conflicts : int ref;
-  bcp_th_conflicts : int ref;
-  bcp_b_conflicts : int ref;
-  bcp_mix_conflicts : int ref;
-
+  mutable th_conflicts : int;
+  mutable b_conflicts : int;
+  mutable bcp_th_conflicts : int;
+  mutable bcp_b_conflicts : int;
+  mutable bcp_mix_conflicts : int;
   (* 4 kinds of red/elim *)
-  t_red : int ref;
-  b_red : int ref;
-  t_elim : int ref;
-  b_elim : int ref;
-  (* first int: counter ok kept instances,
-     second int: counter of removed instances*)
-  instances_map : inst_info MS.t ref;
-  instances_map_printed : bool ref
+  mutable t_red : int;
+  mutable b_red : int;
+  mutable t_elim : int;
+  mutable b_elim : int;
+  mutable instances_map : inst_info MS.t;
+  mutable instances_map_printed : bool;
 }
-
-let state = {
-  decisions = ref 0;
-  assumes = ref 0;
-  assumes_current_lvl = ref 0;
-  queries = ref 0;
-  instantiation_rounds = ref 0;
-  instances = ref 0;
-  decision_lvl = ref 0;
-  instantiation_lvl = ref 0;
-
-  th_conflicts = ref 0;
-  b_conflicts = ref 0;
-  bcp_th_conflicts = ref 0;
-  bcp_b_conflicts = ref 0;
-  bcp_mix_conflicts = ref 0;
-
-  t_red  = ref 0;
-  b_red  = ref 0;
-  t_elim = ref 0;
-  b_elim = ref 0;
-
-  instances_map = ref MS.empty;
-  instances_map_printed = ref false
-}
-
-let set_sigprof () =
-  let tm =
-    let v = Options.get_profiling_period () in
-    if (Stdlib.compare v 0.) > 0 then v else -. v
-  in
-  ignore
-    (Unix.setitimer Unix.ITIMER_PROF
-       { Unix.it_value = tm; Unix.it_interval = 0. })
-
-let init () =
-  state.decisions := 0;
-  state.assumes := 0;
-  state.queries := 0;
-  state.instantiation_rounds := 0;
-  state.instances := 0;
-  state.decision_lvl := 0;
-  state.instantiation_lvl := 0;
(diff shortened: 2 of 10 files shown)


src/lib/structures/profiling.mli
```diff
@@ -63,5 +61,10 @@ val register_produced_terms :
   Expr.Set.t -> (* produced that are new *)
   unit
 
-val print : bool -> int -> Timers.t -> Format.formatter -> unit
+val print : bool -> int -> Format.formatter -> unit
 val switch : Format.formatter -> unit
+
+val get_timers : unit -> Timers.t
+
+(** Prints the statistics for the SMT instruction get-statistics *)
```
Halbaroth:
```suggestion
(** Prints the statistics for the SMT statement (get-info :all-statistics). *)
```

src/lib/structures/profiling.mli
```diff
@@ -63,5 +61,10 @@ val register_produced_terms :
...
+val print_get_statistics : Format.formatter -> unit
```
Halbaroth:
```suggestion
val print_statistics : Format.formatter -> unit
```

src/bin/common/solving_loop.ml
```diff
@@ -446,7 +445,11 @@ let main () =
       pp_reason_unknown st
     | ":version" ->
       print_std Fmt.string Version._version
-    | ":all-statistics"
+    | ":all-statistics" ->
+      if Options.get_profiling () then
+        Printer.print_std "%t" Profiling.print_get_statistics
```
Halbaroth:
```suggestion
        Printer.print_std "%t" Profiling.print_statistics
```

src/bin/common/solving_loop.ml
```diff
@@ -446,7 +445,11 @@ let main () =
...
+      else
+        Printer.print_smtlib_err "Profiling disactivated (try --profiling)"
```
Halbaroth: We could add an option to activate profiling: `(set-option :produce-statistics true)`.

bclement-ocp (commented):
I will take a look at the rest of the changes later, but this patch moves the timers from a global constant in the binary (`bin/common/signals_profiling.ml`) to a global constant in the library (`lib/structures/profiling.ml`), which is the opposite of the direction we should be going for (see also #377). Please revert that part of the change.

src/lib/util/timers.ml
```diff
@@ -149,6 +149,96 @@ let string_of_ty_function f = match f with
...
+        | F_subst         -> 10
+        | F_union         -> 11
+        | F_unsat         -> 12
+        | F_add_predicate -> 13
+        | F_add_terms     -> 14
+        | F_are_equal     -> 15
+        | F_none          -> 16
+        | F_new_facts     -> 17
+        | F_apply_subst   -> 18
+        | F_instantiate   -> 19
```
Halbaroth: We may use the function `Hashtbl.hash`here? I know using this function involves some risks but for a profiling module, the risk is limited.

src/lib/structures/profiling.ml
```diff
@@ -694,33 +663,71 @@ let timers_table =
...
+  Hashtbl.add statistics_table key getter
+
+let register_int_stat (key : string) (getter : unit -> int) =
+  register_stat key (fun () -> string_of_int (getter ()))
+
+let register_float_stat (key : string) (getter : unit -> float) =
+  register_stat
+    key
+    (fun () ->
+       Format.sprintf "%.9f" (getter ()))
```
Halbaroth, line 741:
```suggestion
       Fmt.str "%.9f" (getter ()))
```

src/lib/structures/profiling.ml
```diff
@@ -694,33 +663,71 @@ let timers_table =
...
+
+let print_get_statistics fmt =
+  Format.fprintf fmt "(@[<v 0>@,";
```
Halbaroth:
```suggestion
  Fmt.pf fmt "(@[<v 0>@,";
```

src/lib/structures/profiling.ml
```diff
@@ -694,33 +663,71 @@ let timers_table =
...
+  Hashtbl.iter
+    (fun key getter -> Format.fprintf fmt ":%s %s@," key (getter ()))
+    statistics_table;
+  Format.fprintf fmt "@])@,"
```
Halbaroth:
```suggestion
  Fmt.pf fmt "@])@,"
```

src/lib/structures/profiling.ml
```diff
@@ -694,33 +663,71 @@ let timers_table =
...
+let register_float_stat (key : string) (getter : unit -> float) =
+  register_stat
+    key
+    (fun () ->
+       Format.sprintf "%.9f" (getter ()))
+
+let print_get_statistics fmt =
+  Format.fprintf fmt "(@[<v 0>@,";
+  Hashtbl.iter
+    (fun key getter -> Format.fprintf fmt ":%s %s@," key (getter ()))
```
Halbaroth:
```suggestion
    (fun key getter -> Fmt.pf fmt ":%s %s@," key (getter ()))
```

src/lib/structures/profiling.ml
```diff
@@ -694,33 +663,71 @@ let timers_table =
...
+let register_int_stat (key : string) (getter : unit -> int) =
+  register_stat key (fun () -> string_of_int (getter ()))
+
+let register_float_stat (key : string) (getter : unit -> float) =
+  register_stat
+    key
+    (fun () ->
+       Format.sprintf "%.9f" (getter ()))
+
+let print_get_statistics fmt =
```
Halbaroth:
```suggestion
let print_statistics fmt =
```

src/lib/structures/profiling.ml
```diff
@@ -694,33 +663,71 @@ let timers_table =
...
+  Format.fprintf fmt "(@[<v 0>@,";
+  Hashtbl.iter
+    (fun key getter -> Format.fprintf fmt ":%s %s@," key (getter ()))
+    statistics_table;
+  Format.fprintf fmt "@])@,"
+
+(* Now, registering the statistics we want in get-statistics. *)
```
Halbaroth: What is get-statistics here?

src/bin/common/solving_loop.ml
bclement-ocp: If we add a `set-option`, we should use a name as close to the CLI option as possible to avoid confusion, such as `(set-option :profiling true)`.

Also, `(get-info :all-statistics)` should not return an error. We should simply return an empty s-expression here (possibly with a warning message on stderr).

src/lib/structures/profiling.ml
```diff
@@ -196,44 +226,47 @@ let new_instance_of axiom inst loc kept =
...
+  state.decisions <- state.decisions + 1;
+  state.decision_lvl <- state.decision_lvl + 1;
+  if state.decision_lvl <> d then begin
     Printer.print_err
       "state.decision_lvl = %d et d = %d"
-      !(state.decision_lvl) d;
+      state.decision_lvl d;
+    (* TODO(Steven): if nobody knows how to use the previous print to debug
+       anything, I suggest we either remove `d` or get rid the invariant:
+       #(calls to Profiling.decision) = d. *)
```
bclement-ocp: This is a safeguard against wrong code: the profiling module maintains a "shadow" trail of decision levels which is also in the SAT module. They must be kept in sync by the SAT module calling `reset_dlevel` at the appropriate time. If they get out of sync, the profiling information will be garbage; having this invariant helps ensure this does not happen.

src/lib/structures/profiling.ml
```diff
@@ -277,150 +295,163 @@ let float_resize f i = string_resize (Format.sprintf "%f" f) i
...
+
+(* Profiling columns:
+   - Name
+   - Description
+   - Length of the cell
+   - Is the column linked to a statistic (Some true), a timer (Some false) or
+     not (None)?
+   - Data formatter
+   Todo(Steven): use Pierre's solution with Printbox and put these function where
+   they belong (the trash).
```
bclement-ocp: Unless you wrote this code, let's not be that harsh, please :)

src/lib/util/timers.ml
bclement-ocp: What are the risks of `Hashtbl.hash` you have in mind?

(To be fair I am not sure why we have an explicit datatype to describe profiled functions)

bclement-ocp (changes requested):

src/lib/util/timers.ml
Halbaroth: I mean the `hash` function of `Hashtbl` doesn't always produce a good hash function to get a `O(1)` lookup function in your hash table but in our case, it doesn't matter as the hash table is tiny.

Stevendeo:
> I will take a look at the rest of the changes later, but this patch moves the timers from a global constant in the binary (`bin/common/signals_profiling.ml`) to a global constant in the library (`lib/structures/profiling.ml`), which is the opposite of the direction we should be going for (see also #377). Please revert that part of the change.

There is a fundamental issue in the whole profiling system. You can't have profiling global variables in the library to simplify profiling through the analyzer and want some other in the binary. In the previous architecture, the timer was defined in the library, declared in the binary and sent back to the library for profiling. Maybe you have discussed about it, and I would be glad to know your conclusions on the subject ; IMO I think the best choice, for now, is to centralize the whole profiling in the library because we don't want an extra argument for each module to handle profiling from the outside (but again, tell me if it was not your choice).
You will note I lazyfied the timers so that it is only created when the profiling option is on, so that we keep the same behavior than before, but . What I can do in this PR is to lazify the whole profling value instead of only the timer. One of the possibilities to address #377 would be to have `sessions` : when using alt-ergo-lib, you initialize a session that is linked to a specific profiling value.

src/lib/util/timers.ml
Stevendeo: > We may use the function Hashtbl.hashhere? I know using this function involves some risks but for a profiling module, the risk is limited. 

This function already was defined, I just copied it.

src/bin/common/solving_loop.ml
```diff
@@ -485,7 +492,11 @@ let main () =
       pp_reason_unknown st
     | ":version" ->
       print_std Fmt.string Version._version
-    | ":all-statistics"
+    | ":all-statistics" ->
+      if Options.get_profiling () then
+        Printer.print_std "%t" Profiling.print_statistics
+      else
+        warning "Profiling disactivated (try --profiling)"
```
bclement-ocp, line 498: In addition to the warning, there needs to be a reply to the standard output (e.g. `()` or maybe `unsupported`) that can be processed by automated tools.

src/bin/common/solving_loop.ml
```diff
@@ -445,6 +445,13 @@ let main () =
       | ":random-seed"), _
       ->
       unsupported_opt name
+    (* Alt-Ergo custom options *)
+    | ":profiling", Symbol { name = Simple level; _ } ->
+      begin
+        match float_of_string_opt level with
+        | None -> print_wrn_opt ~name st_loc "nonnegative integer" value
```
bclement-ocp, line 452: There needs to be a reply to the standard output here. This should be a `recoverable_error`.

src/lib/structures/profiling.ml
Stevendeo: Sorry if you actually wrote this code (or to anyone who actually wrote it), for I know the constraints were different when it was written. I use the projects I work on as free expression (maybe I should not) and I would have been as harsh with my own code :) I won't get offended by remarks on my code actually, I use these kinds of remarks as a relief and not to insult the person who wrote it.
Anyway, I'll be more respectful in the future

src/lib/util/timers.ml
Stevendeo: If we want a shorter version, we can use `Obj.magic` which would here work like a charm...  :smiling_imp:

bclement-ocp (commented):
I had a hard time understanding your message, but I see what you mean: what use is it having a `Timers.t` object that we pass around if it ends up stored in global `timer_start` and `timer_pause` functions anyways?

This is a fair point that I didn't consider, but am still not entirely convinced, because the `Timers` module does two things currently: it provides a data structure to accumulate and print timers, *and* hooks to register arbitrary profiling functions (that may or may not use the provided data structure, although realistically always will). This looks like good design, and keeps the use of global state to a minimum.

Regarding having an extra argument to each module for profiling: if we clean up the library, we will probably end up with some sort of `Config` module that is threaded around and has the role of `Options` currently, and it makes sense to root the profiling functions there. But you are right, that is an issue we will have anyways.

By the way, what is the reasoning for moving from arrays to hash tables for storing the profiling functions? Based on the comments mentioning hash tables that are in the code, it looks like the inverse change has been made at some point in the past.

src/lib/structures/profiling.ml
bclement-ocp: I didn't write this code, and I know you didn't intend any harm — as you say, this code has served its purpose, even though there are probably better solutions. I just don't think it is necessary to keep disparaging comments, and it makes for an unpleasant codebase to navigate :)

src/bin/common/solving_loop.ml
bclement-ocp: Actually no, this is a `set-option`, not a `get-option`, so it's OK

src/lib/structures/profiling.ml
Stevendeo: Why not just register the decision level of the SAT instead of incrementing it? This invariant forces the Profiling module to be used in a very specific manner, which is undocumented.

Stevendeo:
> I had a hard time understanding your message, but I see what you mean: what use is it having a `Timers.t` object that we pass around if it ends up stored in global `timer_start` and `timer_pause` functions anyways?
> 
> This is a fair point that I didn't consider, but am still not entirely convinced, because the `Timers` module does two things currently: it provides a data structure to accumulate and print timers, _and_ hooks to register arbitrary profiling functions (that may or may not use the provided data structure, although realistically always will). This looks like good design, and keeps the use of global state to a minimum.

I am not sure of the relevance of these hooks. What I expect of a hook is to be applied when a timer is started. Here it is quite different : the hook _is_ the start of the timer. Worse: the hooks are not timer dependent, they are library global references...
Anyway, I am not sure what my changes have to do with your point. In my version, the Timers still have these mutable hooks.

> By the way, what is the reasoning for moving from arrays to hash tables for storing the profiling functions? Based on the comments mentioning hash tables that are in the code, it looks like the inverse change has been made at some point in the past.

Using arrays is error prone: by forgetting to update a hardcoded list, the profiling option is broken in next. Also, using hashtables allows to not to have a memory saved for unused timers. This second reason is a bad reason actually; now that I think of it, I understand why they used arrays in the first place. My version allocates memory during the analysis, which alters the time spent actually analyzing stuff. With arrays, everything is allocated before the timers starts.
Hmm. We can either go back to arrays or explicitely create all hashtables with the correct sizes.

src/lib/structures/profiling.ml
bclement-ocp: When we make a decision the decision level is always incremented (by definition). As I say, it just looks like a safeguard. It should be correct to call `reset_dlevel d` here instead of incrementing `state.decision_lvl`, I'm not sure having the decision level in the profiling matters much anyways.

bclement-ocp:
> Anyway, I am not sure what my changes have to do with your point. In my version, the Timers still have these mutable hooks.

But they are now set in the library rather than the binary. Actually no, they are not set at all — which I think means that no profiling is done at all? Or am I missing something?

> Using arrays is error prone: by forgetting to update a hardcoded list, the profiling option is broken in next.

Is this due to `all_modules` / `all_functions` not being updated? This should have been caught by the `assert` on the length, but looks like `nb_mtag` was not updated either. I think this can be solved by adding an appropriate comment to the type definitions. We could also steal from C's playbook and have `M_Last` and `F_Last` as special timers, with `nb_mtag = mtag M_Last`.

> My version allocates memory during the analysis, which alters the time spent actually analyzing stuff. With arrays, everything is allocated before the timers starts.

Yeah, the hash tables allocating during profiling was one thing I was concerned about, indeed.

> We can either go back to arrays or explicitely create all hashtables with the correct sizes.

If we initialize the hash tables to have constant sizes, using a hashtable instead of an array will just use more memory and be less efficient, so I'd prefer keeping the arrays. And having a comment (maybe at the top of the module?) that explicitly states "we make sure that functions in this module don't allocate while profiling" or something to that regard.

(Well, "perform minimal allocations" — `start` will always allocate due to the stack)

src/bin/common/solving_loop.ml
Stevendeo: I updated the print: when profiling is off, no field is registered and `( )` is printed.

Stevendeo:
> But they are now set in the library rather than the binary. Actually no, they are not set at all — which I think means that no profiling is done at all? Or am I missing something?

They are set in the library on a call to `Profiling.init` (which is made in the binary).
 
> Is this due to `all_modules` / `all_functions` not being updated? This should have been caught by the `assert` on the length, but looks like `nb_mtag` was not updated either. I think this can be solved by adding an appropriate comment to the type definitions. We could also steal from C's playbook and have `M_Last` and `F_Last` as special timers, with `nb_mtag = mtag M_Last`.

You are right, it was not caught because the counter was not updated either. I'll just go back to arrays and document this design.

bclement-ocp:
> They are set in the library on a call to `Profiling.init` (which is made in the binary).

Ah, I got got by the fact that this is hidden by default in GitHub's view. Since `Profiling.init` is in the library, and the `Timers` function is in the library, if we are going with this design we need to either document that the `Timers` function are meant for internal use and should not be called by users of the library, or that calling `Profiling.init` and `Timers.set_timer_{start,pause}` are mutually exclusive.

src/lib/structures/profiling.ml
```diff
@@ -45,58 +45,79 @@ type inst_info = {
...
+  instances_map = MS.empty;
+  instances_map_printed = false;
+  timers = Timers.empty ()
 }
 
+(** Calls the continuation only when the profiling is on.
+    This allows to only force the state when the profiling is on. *)
+let (let*) state f =
+  if Options.get_profiling () then
+    f (Lazy.force state)
```
bclement-ocp: I did not notice the use of `let*` earlier, apologies — I think that using `let*` here is quite confusing because it leads to `let* state = state in` whose behavior is unclear. `let*` is also traditionally used for a monadic `bind` operator, which this is not.

Why not instead call this something like `if_profiling` or `when_profiling`:

```ocaml
let if_profiling f = if Options.get_profiling () then f (Lazy.force state)
```

and then use it with:

```ocaml
if_profiling @@ fun state ->
...
```

which makes the intent more precise?

bclement-ocp: (As an aside: in the current state of the code, this leads to duplicate `Options.get_profiling ()` tests because most calls to the `Profiling` module are guarded with `if Options.get_profiling`, which is required in some cases because computation has to be performed prior to calling the corresponding profiling function)

Stevendeo: I usually go for custom `let`s to avoid the extra indentation. It really is a taste choice, I don't mind going for explicit continuations

bclement-ocp: You don't have to use extra indentation with `@@ fun _ ->`, and most styles I am aware of don't (in fact, ocp-indent will not let you use extra indentation in this repo if you use this construct).

Stevendeo: Oh my bad, I was convinced the `@@` to add indentation! Okay, I'll go for it then

src/lib/util/timers.ml
Stevendeo: Final choice: ppx_deriving.enum, as it defines the values we want.

src/lib/structures/profiling.ml
Stevendeo: That's why my comment mentions that, if we don't know if there is a deep meaning behind the error message and what it is, maybe we should just get rid of it. If we ever get in that case and get this error message, I'm not sure we know how to interpret it.

src/lib/structures/profiling.ml
Stevendeo: https://www.reddit.com/r/ProgrammerHumor/comments/1103mbp/c_code_has_higher_code_quality_when_it_contains/

src/lib/structures/profiling.ml
bclement-ocp: This error message means that the SAT solver is not calling the profiling module in the expected way, and that the profiling module may be accumulating incorrect statistics, because the state that the profiling module expects the solver to be in and the state the solver actually is in have diverged (they have different decision levels). I am trying to reply to your comment and explain what the error message means.

Stevendeo: > This error message means that the SAT solver is not calling the profiling module in the expected way,

What is this "expected" way? This function is not documented. I reread the function and I noticed my remark was not correct, the invariant I state is false. My understanding of the function is the following : the decision reference counts the number of decisions taken (sounds good), and the decision level is the "depth" of the decision, in the sense that everytime we decide something we go one step deeper. As we can backtrack decisions, there also is a way to reset the decision level to a given value. Now, I really don't think the profiling module should take care of keeping track of how Alt-Ergo decides its environment handles the depth of its reasoning. Alt-Ergo has an environment depth value, which is a good thing to display while profiling, but that should be it. Calculating its decision level in the profiling looks like a hack, and the fact the printed message is in french comforts me in this idea.
I'll remove my comment anyway, we can decide in another PR if this code is relevant or not.

Stevendeo:
> > They are set in the library on a call to `Profiling.init` (which is made in the binary).
> 
> Ah, I got got by the fact that this is hidden by default in GitHub's view. Since `Profiling.init` is in the library, and the `Timers` function is in the library, if we are going with this design we need to either document that the `Timers` function are meant for internal use and should not be called by users of the library, or that calling `Profiling.init` and `Timers.set_timer_{start,pause}` are mutually exclusive.

Same would go in the old functionning of timers actually, or did I miss something? Their functionning was meant for internal use, and used in the binary actually set them the only way it should have been set.
I think a better design would be to associate the timers and their hooks, and that the hooks would be triggered on calls to start.

src/lib/structures/profiling.ml
bclement-ocp: I agree this code is not very useful and would be fine removing `decision_lvl` altogether. I was only replying to your comment asking what it meant. Does it make sense to keep track of the decision level twice? Probably not, but that is what we currently do, and this code checks that the two places we keep track of the decision level agree. I was trying to be helpful, I am sorry if I came across any other way.

bclement-ocp:
> Same would go in the old functionning of timers actually, or did I miss something? Their functionning was meant for internal use, and used in the binary actually set them the only way it should have been set.

With the current design, the `set_timer_start` and `set_timer_pause` functions are exposed by the library and are the only way to set the timers. This is (to my knowledge) the only interface that allows users of the library (not using the binary) to setup the timers, and so is the de-facto public API for doing so. Now this PR is changing things so that users of the library should only call `Profiling.init`, but the old functions are still available, and so we should give the information somewhere that the old way of doing things will no longer work as expected (even if I am not sure anyone actually uses the library…). Alt-Ergo is poorly documented, but we should improve the status-quo by writing more documentation, not less!

> I think a better design would be to associate the timers and their hooks, and that the hooks would be triggered on calls to start.

I agree — I actually don't think that the hooks should be in the `Timers` module at all, but rather in `Profiling` since that's where the global state is at.

src/lib/structures/profiling.ml
bclement-ocp: Good thing we are not writing C then ;)

Stevendeo:
I will leave this PR as it is, and open an issue about documenting/rewriting profiling & timers; I feel like this PR about adding all-statistics went a little too far from it's initial purpose :)

bclement-ocp (approved):
I am OK with that in this case, but I will note that I find it disingenuous to put it this way when the title of the PR includes "refactoring Profiling module" and most of the lines affected by the PR are not about `(get-info :all-statistics)` at all.

PR #864, opened 2023-10-05
--dump-models-on implies --dump-models

Halbaroth:
If we want to dump models somewhere, we definitively want models!

I also refactor a bit the code and fix spelling.

### DIFF
--- a/src/bin/common/parse_command.ml
+++ b/src/bin/common/parse_command.ml
@@ -362,8 +362,17 @@ let mk_limit_opt age_bound fm_cross_limit timelimit_interpretation
 
 let mk_output_opt
     interpretation use_underscore unsat_core output_format model_type
-    () () () ()
+    produce_models dump_models dump_models_on () ()
   =
+  let _ =
+    match dump_models_on with
+    | None ->
+      Options.set_dump_models dump_models
+    | Some ch ->
+      let out = Options.Output.create_channel ch in
+      Options.Output.set_dump_models out;
+      Options.set_dump_models true
+  in
   set_infer_output_format (Option.is_none output_format);
   let output_format = match output_format with
     | None -> Native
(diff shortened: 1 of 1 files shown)


bclement-ocp (commented):
We may disagree on this one, but I'd prefer if we keep the option terms small, e.g. with something like

```ocaml
let mk_dump_models dump_models dump_models_on =
  dump_models || Option.is_some dump_models_on
in
let dump_models = Term.(const mk_dump_models $ dump_models $ dump_models_on) in
```

This has several advantages:
 - The huge terms such as `mk_output_opt` etc. where the arguments change as we add/remove options are more prone to errors, because the values we get depend on the order of the arguments in a function call. On the other hand, using small terms, calling the `Options.set_XXX` at the level of terms and passing `()` to the "final" function makes it easier to follow what an argument is doing (including using the editor's "find reference" functionality).
 - It helps keeping dependency chains narrow (e.g. here, we shouldn't have to modify the computation for `interpretation` at all: `interpretation` depends on `produce_models` and `dump_models`, so we should just be changing the way that `dump_models` is computed — otherwise, it is easy to forget a case or a dependency somewhere where making changes).
 - In the future, we want to be moving away from the `Options` module (having a single huge repository of references in a library is not good design, see #377). Having small, focused terms with low responsibility makes it easier to migrate to a design where we incrementally build configuration objects rather than `unit` values.

Stevendeo (approved):

Stevendeo:
https://github.com/OCamlPro/alt-ergo/pull/864#pullrequestreview-1661304845

Maybe can we merge this PR and open an issue about improving the command parsing in general which is (IMO) already quite a mess?

bclement-ocp:
The PR shouldn't block additional improvements to command line parsing (which, indeed, is a mess); the behavior change itself is trivial and can easily be replicated if command line parsing is refactored; I'd rather close it (besides, there are conflicts).

PR #865, opened 2023-10-05
Update the current authors

Halbaroth:


bclement-ocp:
I feel uneasy about removing Albin and Mattias from there, they deserve recognition. Maybe we can change the "Original authors" into "Previous authors" and credit them there? Or add a separate "Previous authors" section if we want to keep "Original authors"?

Halbaroth:
> I feel uneasy about removing Albin and Mattias from there, they deserve recognition. Maybe we can change the "Original authors" into "Previous authors" and credit them there? Or add a separate "Previous authors" section if we want to keep "Original authors"?

Actually, I forgot to copy them in the other section. I thought about it this morning while commuting to ocp... 
Yes, they deserves to keep their names as contributor! I will add a new section.

src/bin/common/parse_command.ml
```diff
@@ -1501,11 +1501,14 @@ let main =
...
       `Pre "CURRENT AUTHORS\n\
-           \   Sylvain Conchon\n\
-           \   Albin Coquereau\n\
            \   Guillaume Bury\n\
+           \   Basile Clément\n\
+           \   Steven de Olivera\n\
+           \   Hichem Rami Ait El Hara\n\
+           \   Pierre Villemot";
+      `Pre "PREVIOUS AUTHORS\n\
+           \   Albain Coquereau\n\
```
```suggestion
           \   Albin Coquereau\n\
```

bclement-ocp (changes requested):

bclement-ocp (approved):

Issue #798, opened 2023-09-08
Improve UF implementation


bclement-ocp:
The Union-Find module is implemented with a mapping from semantic values to their representative. This means that accessing the representative is an O(log n) operation, where `n` is the number of live semantic values (roughly, number of distinct terms).

Benchmarking a few problems reveals that Alt-Ergo can easily spend 5-10% of its time just looking for representatives, and on long-running problems where we discover many terms, this can get upwards of 20%. Using a more efficient implementation of the Union-Find structure could bring tangible performance benefits: for instance, a properly implemented persistent version of Tarjan's Union-Find data structure would bring down the time to access the representatives to (amortized) O(α(n)) i.e. almost constant time.

(Note that these numbers include both the term → semantic value and semantic value → representative mappings, and are thus slightly inflated — in reality, they are probably closer to 3-8% and 15%. An improved union-find representation would only affect the semantic value → representative mapping, not the term → semantic value mapping. However, the term → semantic value mapping is independent from the environment (it always maps `t` to `X.make t`, and may no longer be needed at all now that we have the global cache on `X.make`).

After trying such an implementation, there are a few caveats:

 - First, this cannot be done in UF only. To have an efficient persistent implementation, the best way is to use an imperative implementation with checkpointing. But checkpointing is not free, and doing checkpoints before each UF operation is slow. Ideally, we want the checkpoints to be in sync with the SAT checkpoints, which requires lifting the checkpoints to the `Ccx` and `Theory` modules.
 - Second, we need to be extra careful when merging classes due to AC(X) requirements. In fact, AC(X) *forces* the class representative to be the lowest element in the class for the appropriate order, which seems to be causing subtle issues in relation to congruence closure.

bclement-ocp:
I have spent some more time implementing a better version of this, but hit another blocker. The crux of it is that Alt-Ergo has many copies of the union-find data structure (up to 8 when using Tableaux-CDCL, although at most 5 are used at once), which is mostly not an issue (although it explains some weird behaviors I was seeing), except for one specific use.

This use is `gamma_finite` in `Theory`, which is used for case splits. The way Alt-Ergo uses case splits is as follows.

 - The theory maintains two environments: `gamma`, that only sees propagations and decisions from the SAT solver, and `gamma_finite`.
[...]

PR #868, opened 2023-10-10
The formulae `top` and `bot` are constant

Halbaroth:
This PR makes clear that the environments of both CC(X) and UF(X) are constant. The previous signature of the constructors for these environments was `unit -> t` and not `t` only because they use the `top` and `bot` semantic values. As the module Combine in `Shostak` cannot contain constants because of limitations with recursive module in OCaml, we have put these constants in a trivial closure.

I create a new module X which is an alias of `Shostak.Combine` and contains the constants `top` and `bot`.

Note: I don't remove all the aliases `module X = Shostak.Combine` in many modules of AE. We can do it gradually and modifying all the files will be painful while rebasing our current PRs.

Note: I cannot turn the environment of Theory into a constant because this environment uses mutable states to enqueue facts. I plan to fix that in another PR (probably by writing a module for functional queues but I have to check how the current environment behaves while backtracking).

### DIFF
--- a/src/lib/reasoners/ccx.ml
+++ b/src/lib/reasoners/ccx.ml
@@ -41,7 +41,7 @@ module type S = sig
   type t
   type r = Shostak.Combine.r
 
-  val empty : unit -> t
+  val empty : t
 
   val empty_facts : unit -> r Sig_rel.facts
 
@@ -107,9 +107,9 @@ module Main : S = struct
 
   type r = Shostak.Combine.r
 
-  let empty () = {
+  let empty = {
     use = Use.empty ;
-    uf = Uf.empty () ;
+    uf = Uf.empty;
     relation = Rel.empty [];
   }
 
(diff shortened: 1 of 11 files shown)


bclement-ocp:
This is somewhat conflicting with the direction I thought were trying to take with #798. Can you explain why it is important that these environments are constants?

I would also advise against writing functional queues for the `Theory` module. As far as I know, the queues in `Theory` are used locally and are empty between steps, we would never make use of the extra functionality provided by a functional queue compared to an imperative one.

Halbaroth:
I think the main change here is to make clear that `top` and `bot` are constant. I can change the title. The reason why I did that it's because I want to understand why the environments of these modules seem to be imperative. If there is no reason to be imperative, we should make it clear in the signature. You can modify the signature again when you will merge your PR with a new UF.

bclement-ocp:
I have given a longer thought about this, and I really don't like the introduction of `X` as a toplevel module (especially when it is a common alias in the rest of the code, and even further when it is mostly empty). Can we instead add the definitions:

```ocaml
let top = top ()
let bot = bot ()
```

directly in `Shostak.Combine` (but not `CX`), with the appropriate signature change?

Halbaroth:
I cannot do that. Indeed `Combine` and `CX` shares the same signature and `CX` is a recursive module, thus we can't add constant in it. An ugly solution consists in duplicating the signature module `Sig.X`, which is clearly a bad idea.

bclement-ocp:
You should be able to do `module Combine : sig include Sig.X val top : r val bot : r end` shadowing the `unit -> r` versions of `top` and `bot` I think.

Halbaroth:
> You should be able to do `module Combine : sig include Sig.X val top : r val bot : r end` shadowing the `unit -> r` versions of `top` and `bot` I think.

I removed my comment because I found this solution... :D

bclement-ocp (approved):
Ah, I was not sure if this was a GitHub bug that it didn't appear in the UI hence why I replied by email :) All good then!

PR #869, opened 2023-10-11
Support delayed computation of partially interpreted functions

### LINKED ISSUE #801: Improve behavior of `mod`, `int2bv`, `bv2nat`, etc. when called on non-constant terms
This is an alternative / improved approach to (some of) #736 

The crux of the issue is that if we have a function that is interpreted only on constants (i.e. a function for which we can compute, but not reason about, such as `mod` or `bv2nat` and `int2bv`), once we fall back to its "term" version when the argument is not a constant, we will never evaluate the result.

Simple example:

```console
$ cat int2bv1.smt2
(set-logic ALL)

(assert (distinct ((_ int2bv 2) 0) #b00))
(check-sat)
$ alt-ergo --frontend dolmen int2bv1.smt2 2>/dev/null

; File "int2bv1.smt2", line 4, characters 1-12: Valid (0.0064) (1 steps) (goal g_1)

unsat

$ cat int2bv2.smt2
(set-logic ALL)

(define-fun x () Int 0)
(assert (distinct ((_ int2bv 2) x) #b00))
(check-sat)
$ alt-ergo --frontend dolmen int2bv2.smt2

; File "int2bv2.smt2", line 5, characters 1-12: I don't know (0.0060) (4 steps) (goal g_1)

unknown
```

It is not clear to me at the time of writing what the proper fix for this is. Preprocessing unconstrained equalities into hard rewrites would make sense, but we are not really equipped to do it because there are term-indexed maps everywhere. Probably the better fix would be to have a way to indicate that certain functions should trigger a second call to `X.make` when (one of) their argument gets substituted. I am not sure if this should be a different type of semantic values (say, `Lazy` or `Partial`) or just a flag on `Term` semantic values, although the latter is probably more pragmatic.

bclement-ocp:
This patch lifts the existing implementation in `IntervalCalculus` for delayed computation of the `pow` function (by "delayed computation" I mean that we remember `pow` terms with uninterpreted arguments and evaluate them once their arguments become interpreted). This is a generic way of integrating functions that we know how to compute but not necessarily how to reason about.

The patch is concerned about lifting the implementation for `pow` to a generic one and adds `int2bv` and `bv2nat` as delayed functions (with corresponding tests), but does not (yet) use it for other partially interpreted functions from the arithmetic theory (e.g. `Modulo`, `Real_of_int`, and the like); this will be done separately.

The implementation is placed in a new `Rel_utils` module that is intended to provide scaffolding for generic behavior (such as delayed computation) that can be useful for all relations. An alternative would be to support delayed computation in a fully generic way in the `Relation` module directly; providing the feature in `Rel_utils` is (slightly) simpler and keeps things more decoupled; the implementation can easily be moved to `Relation` later if we want to.

Fixes #801


src/lib/reasoners/rel_utils.ml
```diff
@@ -0,0 +1,206 @@
...
+
+        Other cases are either not equalities (NCS) or should (I believe...) be
+        subsumed by `Subst` (`Other`). The original code in the arithmetic
+        theory that this is lifted from from was only considering `Subst`, and
+        was allowing possibly incorrect models.
+
+        Note that we need to handle `CS` here to produce correct models,
+        but also this can cause eager enumeration, so we only use it when
+        model generation has been requested. *)
+    if X.is_constant r2 then
```
Halbaroth: We don't need to test `X.is_constant r1` here as the equation `Xliteral.Eq (r1, r2)` is oriented and if one of its terms is constant, `r2` is constant?

src/lib/reasoners/rel_utils.ml
```diff
@@ -0,0 +1,206 @@
...
+      )
+    | _ -> env, []
+
+  let update { dispatch; used_by } uf r1 eqs =
+    match MXH.find r1 used_by with
+    | exception Not_found -> eqs
+    | sm ->
+      OMap.fold (fun sy se eqs ->
+          let fn =
+            (* The [fn] mus be present because we only add symbols to [used_by]
```
```suggestion
            (* The [fn] must be present because we only add symbols to [used_by]
```

Halbaroth (approved):
LGTM :)

src/lib/reasoners/rel_utils.ml
bclement-ocp: Yeah, `Subst` and `CS` equalities are oriented. I will update the comment to clarify that fact.

bclement-ocp: Turns out that `CS` equalities are oriented… except for those that come from the Arrays module, which are not even equalities but disequalities so we also need to handle `NCS` here. I have updated the code and stopped trying to be smart, both `r1` and `r2` are now updated in the `CS` and `NCS` case.

PR #871, opened 2023-10-12
Documentation of the Enum theory

Halbaroth:


src/lib/reasoners/enum.ml
```diff
@@ -32,7 +32,12 @@ module Sy = Symbols
 module E  = Expr
 module Hs = Hstring
 
-type 'a abstract = Cons of Hs.t * Ty.t |  Alien of 'a
+type 'a abstract =
+  | Cons of Hs.t * Ty.t
+  (* [Cons(hs, ty)] reprensents a constructor of an enum type of type [ty]. *)
```
bclement-ocp:
```suggestion
  (* [Cons(hs, ty)] represents a constructor of an enum type of type [ty]. *)
```

src/lib/reasoners/enum.ml
```diff
@@ -32,7 +32,12 @@ module Sy = Symbols
...
+
+  | Alien of 'a
+  (* [Alien r] reprensents a uninterpreted enum semantic values. *)
```
bclement-ocp:
```suggestion
  (* [Alien r] represents a uninterpreted enum semantic value. *)
```

src/lib/reasoners/enum_rel.ml
```diff
@@ -108,11 +125,18 @@ module Debug = struct
...
   | Ty.Tsum (_,l) ->
     Some (List.fold_left (fun st hs -> HSS.add hs st) HSS.empty l)
-  | _ -> None
+  | _ ->
+    (* This case can happen since we don't dispatch the literals
+       processed in [assume] by their types in the Relation module. *)
+    None
 
+(* Update the domains of the semantic values [sm1] and [sm2] according to
+   the disequality [sm1 <> sm2]. *)
```
bclement-ocp: This is not really useful as it repeats in English what can be assumed by the name of the function. It would be more useful to document *how* the domains are updated when encountering `sm1 <> sm2`. I assume singleton values from one side are removed from the other, but maybe something else also happens?

src/lib/reasoners/enum_rel.ml
```diff
@@ -156,6 +180,8 @@ let add_diseq hss sm1 sm2 dep env eqs =
 
   |  _ -> env, eqs
 
+(* Update the domains of the semantic values [sm1] and [sm2] according to
+   the equation [sm1 = sm2]. *)
```
bclement-ocp: Same remark as for `add_diseq` above — how are the domains updated? I assume we compute the intersection of the existing domains?

src/lib/reasoners/enum_rel.ml
```diff
@@ -244,13 +272,22 @@ let assume env uf la =
 
 let add env _ r _ = add_aux env r, []
 
+(* Do a case-split by choosing a value for a uninterpreted semantic value
+   whose the domain in [env] is of minimal size. *)
```
bclement-ocp:
```suggestion
   whose domain in [env] is of minimal size. *)
```

src/lib/reasoners/enum_rel.ml
```diff
@@ -244,13 +272,22 @@ let assume env uf la =
...
 let case_split env uf ~for_model =
   let acc = MX.fold
       (fun r (hss, _) acc ->
          let x, _ = Uf.find_r uf r in
          match Sh.embed x with
-         | Cons _ -> acc (* already bound to an Enum const *)
-         | _ -> (* cs even if sz below is equal to 1 *)
+         | Cons _ ->
+           (* The equivalence class of [r] already contains a value and
```
bclement-ocp:
```suggestion
           (* The equivalence class of [r] already contains a value so
```

src/lib/reasoners/enum_rel.ml
```diff
@@ -44,11 +46,26 @@ module HSS = Set.Make (struct type t=Hs.t let compare = Hs.compare end)
...
+     lies in the domain. *)
 
-let empty classes = { mx = MX.empty; classes = classes;
-                      size_splits = Numbers.Q.one }
+  classes : Expr.Set.t list;
+  (* State of the union-find represented by its equivalence classes.
+     The field is updated while assuming literals of the theory. *)
+
+  size_splits : Numbers.Q.t
+  (* Limit the number of case-splits do by this theory. The counter is
```
bclement-ocp:
```suggestion
  (* Limit the number of case-splits performed by this theory. The counter is
```

I am not sure I understand the description — does the `size_splits` limit the number of case-splits performed (as in, we will not perform more than `size_splits` splits), or does it count them?

src/lib/reasoners/enum_rel.ml
```diff
@@ -44,11 +46,26 @@ module HSS = Set.Make (struct type t=Hs.t let compare = Hs.compare end)
...
+  mx : (HSS.t * Ex.t) MX.t;
+  (* Map of uninterpreted enum semantic values to domains of their possible
+     values. The explanation justifies that the value of the semantic value
+     lies in the domain. *)
 
-let empty classes = { mx = MX.empty; classes = classes;
-                      size_splits = Numbers.Q.one }
+  classes : Expr.Set.t list;
+  (* State of the union-find represented by its equivalence classes.
+     The field is updated while assuming literals of the theory. *)
```
bclement-ocp: Is this a list of all the equivalence classes of the union-find, or only all the classes for terms of enum types (or something else)? What do we use it for?

src/lib/reasoners/enum_rel.ml
```diff
@@ -44,11 +46,26 @@ module HSS = Set.Make (struct type t=Hs.t let compare = Hs.compare end)
 
 module LR = Uf.LX
 
-type t = { mx : (HSS.t * Ex.t) MX.t; classes : Expr.Set.t list;
-           size_splits : Numbers.Q.t }
+type t = {
+  mx : (HSS.t * Ex.t) MX.t;
+  (* Map of uninterpreted enum semantic values to domains of their possible
+     values. The explanation justifies that the value of the semantic value
+     lies in the domain. *)
```
bclement-ocp: Are the keys in the `MX.t` arbitrary semantic values or (current) normal forms only?

src/lib/reasoners/enum_rel.ml
```diff
@@ -195,6 +222,7 @@ let count_splits env la =
   in
   {env with size_splits = nb}
 
+(* Add the uninterpreted semantic value [r] to the environment. *)
```
bclement-ocp: Same as above, it is clear from the name that we add `[r]` to the environment, but what does it mean? Apparently we initialize the corresponding `mx` entry with all the possible constructors for its type.

src/lib/reasoners/enum_rel.ml
```diff
@@ -244,13 +272,22 @@ let assume env uf la =
...
-         | _ -> (* cs even if sz below is equal to 1 *)
+         | Cons _ ->
+           (* The equivalence class of [r] already contains a value and
+              we don't need to make another case-split for this semantic
+              value. *)
+           acc
+         | _ ->
+           (* We have to do a case-split, even if the domain of [r] is
+              of cardinal 1 as we have to put this value in the
+              equivalence class of [r]. *)
```
bclement-ocp: I think that "let `Uf` know about the equality" would be more precise — in particular this matters for model generation.

src/lib/reasoners/enum_rel.ml
bclement-ocp: (Looks like it is arbitrary semantic values)

Halbaroth: I'm not sure to understand what you mean by `normal form` here?  There is no canonization for Enum theory.

src/lib/reasoners/enum_rel.ml
Halbaroth: It's actually the same design in most theories. We have to save the state of the all union-find after assuming each literal in order to return the appropriate state if we perform a query in this theory and we got an inconsistency.

Halbaroth: Actually, it's the case in all theories, which means we can bubble this feature in `Relation`  but I'm not sure it will be clearer.

src/lib/reasoners/enum_rel.ml
Halbaroth: It counts them and an option in `Options` gives an upper bound. I will fix it :) (sorry, Github resolved the conversation as I accepted your commit suggestion...)

src/lib/reasoners/enum_rel.ml
```diff
@@ -156,6 +189,14 @@ let add_diseq hss sm1 sm2 dep env eqs =
...
   |  _ -> env, eqs
 
+(* Update the domains of the semantic values [sm1] and [sm2] according to
+   the equation [sm1 = sm2]. More precisely, we replace their domains by
+   the intersection of these domains.
+
+   In any case, we produce an equality if the domain of these semantic values
+   becomes a singleton.
+
+   @raise Ex.Inconsistent if the domains of [sm1] and [sm2] are disjoint. *)
```
Halbaroth: I know this comment is a bit wrong as the function doesn't produce a new equality when one of the semantic values is a constructor. I don't know why? Maybe we don't need to process it here because it will be done somewhere else.

src/lib/reasoners/enum_rel.ml
```diff
@@ -44,11 +46,28 @@ module HSS = Set.Make (struct type t=Hs.t let compare = Hs.compare end)
...
-type t = { mx : (HSS.t * Ex.t) MX.t; classes : Expr.Set.t list;
-           size_splits : Numbers.Q.t }
-
-let empty classes = { mx = MX.empty; classes = classes;
-                      size_splits = Numbers.Q.one }
+type t = {
+  mx : (HSS.t * Ex.t) MX.t;
+  (* Map of uninterpreted enum semantic values to domains of their possible
+     values. The explanation justifies that any value assigns to the semantic
+     value has to lie in the domain. *)
```
Halbaroth, line 53: I changed a bit the comment. It didn't make sense to talk about `the value of the semantic value`.

src/lib/reasoners/enum_rel.ml
bclement-ocp: Ah right, this for caching is the result of `cl_extract` that is used only for a debug message. I don't think we can lift this to `Relation` (it is needed to raise `Ex.Inconsistent` and the relations do raise `Ex.Inconsistent`). We should make it clear in the doc that this is a cache of `Uf.cl_extract` (if the `--bottom-classes` flag is not passed to Alt-Ergo, `Uf.cl_extract` is always `[]`).

src/lib/reasoners/enum_rel.ml
bclement-ocp: I mean normal form modulo congruence (see `find_r` → `find_or_normal_form` in `Uf`), e.g. if we have `f(x)` and `f(x) = y` then the normal form of `f(x)` is `y`; in other word, class representatives (modulo AC(X) shenaningans).

src/lib/reasoners/enum_rel.ml
bclement-ocp: If we execute this function with `Alien _, Cons _` or `Cons _, Alien _` we are already processing the equality that we would be producing :)

src/lib/reasoners/enum_rel.ml
```diff
@@ -195,6 +237,8 @@ let count_splits env la =
   in
   {env with size_splits = nb}
 
+(* Add the uninterpreted semantic value [r] to the environment [env] with the
+   total domain. *)
```
bclement-ocp:
```suggestion
(* Add the uninterpreted semantic value [r] to the environment [env] with the
   all the constructors of its type as domain. *)
```

src/lib/reasoners/enum_rel.ml
Halbaroth: That's why I said `I'm not sure it will be clearer` as we would have to create a new intermediate exception :/ 
I agree that the best solution is to write a clear comment.

src/lib/reasoners/enum_rel.ml
Halbaroth: It's arbitrary semantic values as the only way to remove elements in the field `mx` is by backtracking in the SAT solver. We could remove the old normal form key and use `Uf.find` to search in the `mx` map. I don't know if it's very important. We could save a tiny amount of memory and I guess doing this will be a bit slower.

bclement-ocp: I am not asking to change the implementation, just trying to understand what it does so that the documentation is accurate :)

(Note that using class representatives as the key would also cause the same issue we observe in #474 / #823 )

src/lib/reasoners/enum_rel.ml
```diff
@@ -44,11 +46,30 @@ module HSS = Set.Make (struct type t=Hs.t let compare = Hs.compare end)
...
+  mx : (HSS.t * Ex.t) MX.t;
+  (* Map of uninterpreted enum semantic values to domains of their possible
+     values. The explanation justifies that any value assigns to the semantic
+     value has to lie in the domain. *)
+
+  classes : Expr.Set.t list;
+  (* State of the union-find represented by all its equivalence classes.
+     This state is kept for debugging purposes only. It is updated after
+     assuming literals of the theory and returned by queries in case of
+     inconsistency. *)
```
bclement-ocp: Please mention `Uf.cl_extract` explicitly here.

src/lib/reasoners/enum_rel.ml
Halbaroth: I am not saying to change the implementation in this PR ;)

src/lib/reasoners/enum_rel.ml
```diff
@@ -156,13 +191,20 @@ let add_diseq hss sm1 sm2 dep env eqs =
...
+   @raise Ex.Inconsistent if the domains of [sm1] and [sm2] are disjoint. *)
 let add_eq hss sm1 sm2 dep env eqs =
   match sm1, sm2 with
   | Alien r, Cons(h,_)
   | Cons (h,_), Alien r  ->
     let enum, ex = try MX.find r env.mx with Not_found -> hss, Ex.empty in
     let ex = Ex.union ex dep in
     if not (HSS.mem h enum) then raise (Ex.Inconsistent (ex, env.classes));
+    (* We don't need to produce a new equality as we already know it
+       by transitivity of equality. *)
```
bclement-ocp: I don't think transitivity is involved here, maybe commutativity? We are processing the equality `r = h`, which is exactly the equality we would be adding.

bclement-ocp (changes requested):

src/lib/reasoners/enum_rel.ml
```diff
@@ -195,6 +241,8 @@ let count_splits env la =
   in
   {env with size_splits = nb}
 
+(* Add the uninterpreted semantic value [r] to the environment [env] with the
```
bclement-ocp:
```suggestion
(* Add the uninterpreted semantic value [r] to the environment [env] with
```

src/lib/reasoners/enum_rel.ml
```diff
@@ -244,13 +292,22 @@ let assume env uf la =
 
 let add env _ r _ = add_aux env r, []
 
+(* Do a case-split by choosing a value for a uninterpreted semantic value
```
bclement-ocp:
```suggestion
(* Do a case-split by choosing a value for an uninterpreted semantic value
```

bclement-ocp (approved):

PR #872, opened 2023-10-12
Remove `Util.failwith`

Stevendeo:
The file : 
```
(set-logic ALL)

(declare-const x Float16)
```
produces an error in alt-ergo, but exits with code 2. This PR fixes this issue to make it fail properly with a code 1, plus removes the `Util.failwith` function to replace it with its `Fmt` equivalent.

### DIFF
--- a/src/lib/frontend/d_cnf.ml
+++ b/src/lib/frontend/d_cnf.ml
@@ -43,7 +43,7 @@ module Id = DStd.Id
 module B = DStd.Builtin
 
 let unsupported msg =
-  Format.kasprintf
+  Fmt.kstr
     (fun str -> Errors.(run_error (Unsupported_feature str)))
     msg
 
--- a/src/lib/frontend/d_cnf.ml
+++ b/src/lib/frontend/d_cnf.ml
@@ -42,6 +42,11 @@ module DT = DE.Ty
 module Id = DStd.Id
 module B = DStd.Builtin
 
+let unsupported msg =
+  Format.kasprintf
+    (fun str -> Errors.(run_error (Unsupported_feature str)))
+    msg
+
 module Shared = struct
   (* Shared constants to avoid allocations*)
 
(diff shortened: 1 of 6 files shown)


src/lib/frontend/d_cnf.ml
```diff
@@ -42,6 +42,11 @@ module DT = DE.Ty
 module Id = DStd.Id
 module B = DStd.Builtin
 
+let unsupported msg =
+  Format.kasprintf
+    (fun str -> Errors.(run_error (Unsupported_feature str)))
+    msg
```
Halbaroth: I guess we can use `Fmt.kstr` here :)? Besides we can move the function in `Utils`.

bclement-ocp (approved):
LGTM.

There is a trend in your PRs of saying they do one thing "and also, by the way, another thing". A common guideline is that if what the PR does cannot be described without using the word "and", it should not be a single PR but rather two (or more) separate PRs.

Here this is not the case because the description of the PR should be "Remove `Util.failwith`" (where it is sometimes replaced with `Fmt.failwith` and sometimes replaced by an appropriate specific exception), but please keep that in mind :)

hra687261 (approved):

Stevendeo:
> There is a trend in your PRs of saying they do one thing "and also, by the way, another thing". A common guideline is that if what the PR does cannot be described without using the word "and", it should not be a single PR but rather two (or more) separate PRs.

You're right, I have a hard time staying focused on a single thing; I will make efforts to focus on one single task on my next PRs :)

src/lib/frontend/d_cnf.ml
Stevendeo: `Errors` depends on `Util` unfortunately, but I can use `Fmt`

bclement-ocp:
> You're right, I have a hard time staying focused on a single thing; I will make efforts to focus on one single task on my next PRs :)

FWIW to counteract this, I often work on a branch where I do multiple things at once, then use `git checkout -p` to extract more focused commits that make it into PRs.

PR #878, opened 2023-10-12
Add `Expr` constructors for the `Reals` SMT-LIB theory

bclement-ocp:
This patch follows `Expr.Int` and adds constructors from the `Reals` SMT-LIB theory to the `Expr` module.


bclement-ocp:
If we have the `Ints`, we should have the `Reals` (even though we don't use it currently). However, I am not sure having `Reals_Ints` this way is actually useful — removing it from the PR. We will figure it out when we need it.

Halbaroth (approved):

PR #881, opened 2023-10-13
Do not use existential variables for integer division

### LINKED ISSUE #875: Do not use existential variables for integer division
Currently when we have `a / b` or `a % b` where `a`, `b` have type `int` we do:

 - If `b` is not constant, `a / b` becomes uninterpreted (with some reasoning support in `IntervalCalculus`), `a % b` becomes uninterpreted (with no reasoning support in `IntervalCalculus` -- see also #365)
 - If `b` is a constant and `a` is a constant, we compute `a / b` and `a % b` exactly (using euclidean division)
 - If `b` is a constant and `a` is *not* a constant, `a / b` becomes `a * (1/b) - (a % b) * (1 / b)` (as a polynomial with rational coefficients) and `a % b` becomes `a - k * b` where `k` is a fresh constant such that `0 <= a - k * b < b`, so `k` is actually always equal to `a / b`.

if we combine both we indeed get that `a / b` becomes (after a couple reasoning steps): `a * (1/b) - (a - k * b) * (1 / b) = a * (1 / b) - a * (1 / b) + (k * b) * (1 / b) = k`.

This is wasteful — we could just make `a / b` uninterpreted instead and define `a % b` in terms of `a / b` rather than the other way around, which would avoid the existential variables 

For a simple assertion such as `(assert (distinct (div x 4) (div (- x (mod x 4)) 4)))` there is the following line when printing the equivalence class at the end of reasoning (all the `!k` are created by the existential rule described above):
```
!k1 -> !k1, !k2, !k3, !k4, !k5, !k6, (/ x 4), (/ (- x (% x 4)) 4)
```

bclement-ocp:
This patch implements integer remainder in terms of integer division rather than the opposite. This avoids the introduction of fresh existential variables.

Fixes #875

### DIFF
--- a/src/lib/reasoners/arith.ml
+++ b/src/lib/reasoners/arith.ml
@@ -134,24 +134,7 @@ module Shostak
     | Some p -> p
     | _ -> P.create [Q.one, r] Q.zero (X.type_info r)
 
-  (* Compute [t1 % t2] as [t1 - |t2| * (t1 / |t2|)].
-
-     Requires: [t1], [t2] have type [Tint]
-     Requires: [p1] is the representative for [t1]
-     Requires: [p2] is the representative for [t2]
-     Requires: [p2] is a non-zero constant *)
-  let mk_modulo t1 p1 p2 ctx =
-    match P.is_const p2 with
-    | Some n2 ->
-      let a2 = Q.to_z n2 |> Z.abs in
-      assert (Z.sign a2 > 0);
-
-      (* NB: [t1 % t2] = [t1 % |t2|] *)
-      let div, ctx' = X.make (E.Ints.(t1 / ~$$a2)) in
-      P.sub p1 (P.mult_const (Q.from_z a2) (embed div)), ctx' @ ctx
-    | None -> assert false
-
-  (* Add range information for [t = t1 / t2].
+  (* Add range information for [t = t1 / t2], where `/` is euclidean division.
 
      Requires: [t], [t1], [t2] have type [Tint]
      Requires: [p2] is the term representative for [t2]
@@ -161,19 +144,31 @@ module Shostak
 
     match P.is_const p2 with
     | Some n2 ->
-      let a2 = Q.to_z n2 |> Z.abs in
-      assert (Z.sign a2 > 0);
+      let n2 = Q.to_z n2 in
+      assert (Z.sign n2 <> 0);
+      let a2 = Z.abs n2 in
 
-      if Q.sign n2 < 0 then
-        (* t1 / (-|t2|) = -(t1 / |t2|)*)
-        let r, ctx' = X.make E.Ints.(t1 / ~$$a2) in
-        P.mult_const Q.m_one (embed r), ctx' @ ctx
-      else
-        (* 0 <= t1 % t2 = t1 - t2 * (t1 / t2) < |t2| *)
-        let t1_mod_t2 = E.Ints.(t1 - ~$$a2 * t) in
-        let c1 = E.Ints.(~$0 <= t1_mod_t2) in
-        let c2 = E.Ints.(t1_mod_t2 < ~$$a2) in
-        P.create [Q.one, X.term_embed t] Q.zero Tint, c2 :: c1 :: ctx
+      (* 0 <= t1 % t2 = t1 - t2 * (t1 / t2) < |t2| *)
+      let t1_mod_t2 = E.Ints.(t1 - ~$$n2 * t) in
+      let c1 = E.Ints.(~$0 <= t1_mod_t2) in
+      let c2 = E.Ints.(t1_mod_t2 < ~$$a2) in
+      P.create [Q.one, X.term_embed t] Q.zero Tint, c2 :: c1 :: ctx
+    | None -> assert false
+
+
+  (* Compute the remainder of euclidean division [t1 % t2] as
+     [t1 - t2 * (t1 / t2)], where `a / b` is euclidean division.
+
+     Requires: [t1], [t2] have type [Tint]
+     Requires: [p1] is the representative for [t1]
+     Requires: [p2] is the representative for [t2]
+     Requires: [p2] is a non-zero constant *)
+  let mk_euc_modulo t1 t2 p1 p2 ctx =
+    match P.is_const p2 with
+    | Some n2 ->
+      assert (Q.sign n2 <> 0);
+      let div, ctx = mk_euc_division E.Ints.(t1 / t2) t1 p2 ctx in
+      P.sub p1 (P.mult_const n2 div), ctx
     | None -> assert false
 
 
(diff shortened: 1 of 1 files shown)


src/lib/reasoners/arith.ml
```diff
@@ -134,39 +134,47 @@ module Shostak
...
+    | Some n2 ->
+      let a2 = Q.to_z n2 |> Z.abs in
+      assert (Z.sign a2 > 0);
+
+      if Q.sign n2 < 0 then
+        (* t1 / (-|t2|) = -(t1 / |t2|)*)
+        let r, ctx' = X.make E.Ints.(t1 / ~$$a2) in
+        P.mult_const Q.m_one (embed r), ctx' @ ctx
+      else
+        (* 0 <= t1 % t2 = t1 - t2 * (t1 / t2) < |t2| *)
```
Halbaroth:
```suggestion
        (* 0 <= t1 % t2 = t1 - |t2| * (t1 / t2) < |t2| *)
```

src/lib/reasoners/arith.ml
```diff
@@ -134,39 +134,47 @@ module Shostak
...
-      let rp =
-        P.mult_const (Q.div Q.one coef_p2) (embed r) in
-      P.sub p rp, ctx' @ ctx
-    | _ -> assert false
+  (* Compute [t1 % t2] as [t1 - |t2| * (t1 / |t2|)].
+
+     Requires: [t1], [t2] have type [Tint]
+     Requires: [p1] is the representative for [t1]
+     Requires: [p2] is the representative for [t2]
+     Requires: [p2] is a non-zero constant *)
```
Halbaroth: The previous comment gave a precise definition of the modulo used by `Arith`. 

```ocaml
  (* t1 % t2 = md  <->
     c1. 0 <= md ;
     c2. md < t2 ;
     c3. exists k. t1 = t2 * k + t ;
     c4. t2 <> 0 (already checked) *)
```
It's meaningful to know that we choose an integer between `0`and `t2` as the representative element of the equivalence classes of `Z/t2Z`.

bclement-ocp:
The reasoning for this one is mostly explained in #875. Currently, we rewrite `a / n` (where `n` is a non-zero constant value) into `a * (1/n) - (a % n) * (1 / n)` and we rewrite `a % n` into `a - !k * n` where `!k` is a fresh constant for each occurrence of `a % n`, and we have the constraint `0 ≤ a - !k * n < |n|`.

To avoid causing the creation of fresh constants, the PR flips things around: now, we rewrite `a % n` into `a - (a / n) * n`, and we just add the constraint `0 ≤ a - (a / n) * n < |n|` directly whenever we encounter `a / n`. In addition, we rewrite `a / (-n)` into `-(a / n)` for normalization purposes — we would be able to prove it anyways, but we learn it a bit earlier this way.

src/lib/reasoners/arith.ml
```diff
@@ -134,39 +134,47 @@ module Shostak
...
-  let mk_euc_division p p2 t1 t2 ctx =
-    match P.to_list p2 with
-    | [], coef_p2 ->
-      let md = E.mk_term (Sy.Op Sy.Modulo) [t1;t2] Ty.Tint in
-      let r, ctx' = X.make md in
-      let rp =
-        P.mult_const (Q.div Q.one coef_p2) (embed r) in
-      P.sub p rp, ctx' @ ctx
-    | _ -> assert false
+  (* Compute [t1 % t2] as [t1 - |t2| * (t1 / |t2|)].
```
Halbaroth: I know the division symbol `/` here means the integer division but at the first sight, it was a bit confusing as the fact `t1` and `t2` are integer is mentioned later. We may precise that `/` is the integer division?

src/lib/reasoners/arith.ml
```diff
@@ -134,39 +134,47 @@ module Shostak
...
+     Requires: [p1] is the representative for [t1]
+     Requires: [p2] is the representative for [t2]
+     Requires: [p2] is a non-zero constant *)
+  let mk_modulo t1 p1 p2 ctx =
+    match P.is_const p2 with
+    | Some n2 ->
+      let a2 = Q.to_z n2 |> Z.abs in
+      assert (Z.sign a2 > 0);
+
+      (* NB: [t1 % t2] = [t1 % |t2|] *)
```
Halbaroth: Here the equality `=` denotes the usual equality relation between integers and this equation is consequence of our choice of representative term for the equivalence classes. I think we should clarify this point.

src/lib/reasoners/arith.ml
```diff
@@ -134,39 +134,47 @@ module Shostak
...
+
+    match P.is_const p2 with
+    | Some n2 ->
+      let a2 = Q.to_z n2 |> Z.abs in
+      assert (Z.sign a2 > 0);
+
+      if Q.sign n2 < 0 then
+        (* t1 / (-|t2|) = -(t1 / |t2|)*)
+        let r, ctx' = X.make E.Ints.(t1 / ~$$a2) in
+        P.mult_const Q.m_one (embed r), ctx' @ ctx
```
Halbaroth: If I understand well, we don't need to produce here an interval for the case `n2 < 0` as the recursive call `X.make` will produce the appropriate interval for `-n2`?

src/lib/reasoners/arith.ml
bclement-ocp: `t1 % t2` here is not an equivalence class, it is defined as the remainder of euclidean division, which is an integer between `0` and `|t2|` by definition (note that the condition `c2` in the pre-existing comment was incorrect because `t2` can be negative). Should be clear with the added definition of `%` as remainder of Euclidean division.

src/lib/reasoners/arith.ml
bclement-ocp: This follows from the definition of `t1 % t2` as the remainder of euclidean division. Anyways, I have removed this to avoid recomputing `X.make t1` and call `mk_euc_division` directly instead.

src/lib/reasoners/arith.ml
bclement-ocp: Yeah, here we just rewrite `t1 / (-|t2|)` into `-(t1 / |t2|)` and then we let the recursive call take care of things. Although I am now realizing this means we will compute `X.make t1` twice; I will inline the code.

src/lib/reasoners/arith.ml
bclement-ocp: I think I prefer the version that is always true, even if in this case `t2` is positive so `t2 = |t2|`, I will clarify.

src/lib/reasoners/arith.ml
bclement-ocp: Sure, added a note that `/` is Euclidean division here, and that `%` is the remainder of Euclidean division. Also renamed the function `mk_euc_modulo`.

src/lib/reasoners/arith.ml
bclement-ocp: Actually I have removed the distinction; we will create `t1 / n` and `t1 / (-n)` and rely on reasoning to prove they are identical, it is less error-prone.

src/lib/reasoners/arith.ml
Halbaroth: > This follows from the definition of `t1 % t2` as the remainder of euclidean division.

It's exactly what I said ;)

src/lib/reasoners/arith.ml
Halbaroth: I have no strong opinion about it. For me it was clearer that the formula agrees syntactically with the code below.

Halbaroth (approved):

PR #887, opened 2023-10-13
Using `dune build @check` to build the project

Halbaroth:
The current target `all` of the `Makefile` is a bit complicated as we want to build only the codebase without running tests when we type `make`. After looking for a better solution, I found this issue on the `dune` GitHub (https://github.com/ocaml/dune/issues/7203) explaining we have to use `dune build @check` and `dune build` builds everything by design.

I modify the `Makefile` accordingly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,8 @@ plugins:
 # Alias to build all the bytecode artefacts using dune
 # Hopefully more efficient than making "all" depend
 # on "lib" and "bin", since dune can parralelize more
-all: dune build @check
+all:
+	$(DUNE) build $(DUNE_FLAGS) @check
 
 # declare these targets as phony to avoid name clashes with existing directories,
 # particularly the "plugins" target
--- a/Makefile
+++ b/Makefile
@@ -90,13 +90,10 @@ AB-Why3:
 plugins:
 	$(DUNE) build $(DUNE_FLAGS) @$(PLUGINS_DIR)/all
 
-# Alias to build all targets using dune
+# Alias to build all the bytecode artefacts using dune
 # Hopefully more efficient than making "all" depend
-# on "lib" and "bin", since dune can
-# parralelize more
-all:
-	$(DUNE) build $(DUNE_FLAGS) @$(LIB_DIR)/all @$(BTEXT_DIR)/all \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -90,13 +90,10 @@ AB-Why3:
...
-# Alias to build all targets using dune
+# Alias to build all the bytecode artefacts using dune
 # Hopefully more efficient than making "all" depend
-# on "lib" and "bin", since dune can
-# parralelize more
-all:
-	$(DUNE) build $(DUNE_FLAGS) @$(LIB_DIR)/all @$(BTEXT_DIR)/all \
-		@$(PARSERS_DIR)/all @$(BJS_DIR)/all @$(PLUGINS_DIR)/all
+# on "lib" and "bin", since dune can parralelize more
+all: dune build @check
```
bclement-ocp: This tells `make` that `dune`, `build`, and `@check` are dependencies of the `all` target. Plus, this must use the `$(DUNE)` variable.

This should be:
```suggestion
all:
	$(DUNE) build $(DUNE_FLAGS) @$(LIB_DIR)/check @$(BTEXT_DIR)/check \
		@$(PARSERS_DIR)/check @$(BJS_DIR)/check @$(PLUGINS_DIR)/check
```

bclement-ocp: The following should also work:

```suggestion
all:
	$(DUNE) build $(DUNE_FLAGS) @check
```

bclement-ocp:
Interesting, I didn't know about `@check`. Probably this mean that we should get rid of `@runtest-quick` as well?

Makefile
Halbaroth: I agree for `$(DUNE)` and `$(DUNE_FLAGS)` but I don't understand why we should keep enumerate subdirectories? I did that in a old PR only to prevent building tests in the subdirectory `tests`.

If we add another directory at the toplevel, we have to update this list... I doesn't make sense in my opinion.

bclement-ocp: Yes I just replaced `all` in the subcommand. Using naked `@check` should work (I think it will also build the tools but doesn't matter).

bclement-ocp (approved):

Issue #889, opened 2023-10-13
The distinct expression clashes with the SMT-LIB specification


Halbaroth:
The current implementation of the `distinct` expression in the module `Expr` clashes with the 
SMT-LIB specification. 

In Alt-Ergo, the expression `mk_distinct [x_1; ...; x_n]` means:
```
x_1 <> x2 and x2 <> x3 and ... x_(n-1) <> x_n
```
but the expected implementation is:
```
forall 1 <= i < j <= n, x_i <> x_j
```

This disagreement induces a soundness bug. For instance with the input file:
``` 
(set-logic ALL)
(declare-const a Int)
(declare-const b Int)
(declare-const c Int)
(assert (not (distinct a b c)))
(assert (distinct a b))
(check-sat)
```
we got `unsat` with AE but `sat` with other solvers.

Halbaroth:
A hot fix have been merged on the branch `v2.5.x` for the release `v2.5.2`, see #890.

Halbaroth:
After some offline discussions, it seems that solving the issue is not a priority for the next release as we don't expect to see valuable performance improvements because of the current implementation. The difficulty comes from the fact that the negation of `distinct` expressions are in fact clauses. Let's imagine we have the literal `distinct x y z`. After discovering this literal is inconsistent in the current context, we have to propagate its negation `x = y \/ x = z \/ y = z` to the union-find environment. But there is no adequate operation to perform on the union-find for this propagation. It means we have to decide some literal of this clause in the SAT solver.

We have a solution to solve this. Basically, if the SAT solver discovers that a decision on `distinct x y z` is inconsistent, while deciding `not (distinct x y z)`, we add to the context the formula:
```
(distinct x y z) \/ x = y \/ x = z \/ y = z
```
This formula is a tautology. Thus, if we forgot to add it at all the backtracking locations in the code, our solver is still sound. Now, as we have `not (distinct x y z)` in the context, we got also `x = y \/ x = z \/ y = z` and the SAT solver can decide some literal of this clause. In CC(X), we ignore completely `not (distinct)` of at least three elements, as the expanded version in the SAT solver will propagate the appropriate equality in the current branch.

This solution is complex. For the release `v2.6`, we plan to clean up the codebase by removing completely the `distinct` literal in `Expr` dans `Xliteral`. We only keep a smart constructor for `distinct` which represents this formula by a disjunction.

bclement-ocp:
[...]

PR #890, opened 2023-10-13
Fix distinct unsoundness

Halbaroth:
This PR quickly fixes the clash between the SMT-LIB specification of `distinct` and the implementation of this expression in AE.

Basically, we expand the expression `distinct x_1 ... x_n` into a conjunction of binary disequalities.

Notice that the model outputted for `uf2.models.smt2` changes because I don't preserve the old order used in `distinct`.

Fix issue #889 for the `v2.5.x` branch.

### DIFF
--- a/src/lib/frontend/cnf.ml
+++ b/src/lib/frontend/cnf.ml
@@ -239,9 +239,29 @@ and make_form up_qv name_base ~toplevel f loc ~decl_kind : E.t =
           let res = make_term up_qv name_base t in
           if negated then E.neg res else res
 
-        | TAneq lt | TAdistinct lt ->
+        | TAneq lt ->
           let lt = List.map (make_term up_qv name_base) lt in
           E.mk_distinct ~iff:true lt
+
+        | TAdistinct lt ->
+          (* The current implementation of the distinct expression in
+             Expr clashes with the SMT-LIB specification when used with
+             at least 3 arguments. To prevent a soundness bug, we
+             translate the expected expression into a conjonction
+             of disequations of size 2. *)
+          let args = Array.of_list lt in
+          let acc = ref E.vrai in
+          for i = 0 to Array.length args - 1 do
+            for j = i + 1 to Array.length args - 1 do
+              acc :=
+                E.(mk_and
+                     (mk_distinct ~iff:true
+                        [make_term up_qv name_base args.(i);
+                         make_term up_qv name_base args.(j)]) !acc false)
+            done;
+          done;
+          !acc
+
         | TAle [t1;t2] ->
           E.mk_builtin ~is_pos:true Sy.LE
             [make_term up_qv name_base t1;
--- a/src/lib/frontend/cnf.ml
+++ b/src/lib/frontend/cnf.ml
@@ -244,23 +244,7 @@ and make_form up_qv name_base ~toplevel f loc ~decl_kind : E.t =
           E.mk_distinct ~iff:true lt
 
         | TAdistinct lt ->
-          (* The current implementation of the distinct expression in
-             Expr clashes with the SMT-LIB specification when used with
-             at least 3 arguments. To prevent a soundness bug, we
-             translate the expected expression into a conjonction
-             of disequations of size 2. *)
-          let args = Array.of_list lt in
-          let acc = ref E.vrai in
-          for i = 0 to Array.length args - 1 do
-            for j = i + 1 to Array.length args - 1 do
-              acc :=
-                E.(mk_and
-                     (mk_distinct ~iff:true
-                        [make_term up_qv name_base args.(i);
-                         make_term up_qv name_base args.(j)]) !acc false)
-            done;
-          done;
-          !acc
+          E.mk_distinct ~iff:true (List.map (make_term up_qv name_base) lt)
 
         | TAle [t1;t2] ->
           E.mk_builtin ~is_pos:true Sy.LE
--- a/src/lib/frontend/cnf.ml
+++ b/src/lib/frontend/cnf.ml
@@ -239,13 +239,9 @@ and make_form up_qv name_base ~toplevel f loc ~decl_kind : E.t =
           let res = make_term up_qv name_base t in
           if negated then E.neg res else res
 
-        | TAneq lt ->
+        | TAneq lt | TAdistinct lt ->
           let lt = List.map (make_term up_qv name_base) lt in
           E.mk_distinct ~iff:true lt
-
-        | TAdistinct lt ->
-          E.mk_distinct ~iff:true (List.map (make_term up_qv name_base) lt)
-
         | TAle [t1;t2] ->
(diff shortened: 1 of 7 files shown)


bclement-ocp:
> This fix doesn't change the smart constructor `mk_distinct`. We shouldn't change the API on the branch v2.5.x and I add a comment to advertise the user that this smart constructor doesn't behave as expected.

I disagree on this point. `mk_distinct` is supposed to build a `distinct` term; even if the API is undocumented, the intent (as expressed by the name `mk_distinct`) is clear to any reasonable user. `mk_distinct` not building a an "all-distinct" term *is* the bug that should be fixed.

(One other way to think about it: if we have a function `mk_plus` that actually created a multiplication, we would fix it in the function itself, we would not add workarounds)

Halbaroth:
Besides, I checked that the constructor `TANeq` of the legacy parser is never used with a list of at least three elements.

src/lib/structures/expr.ml
```diff
@@ -1210,10 +1210,20 @@ let mk_nary_eq l =
...
 
-let mk_distinct ~iff tl =
-  match tl with
-  | [a; b] -> neg (mk_eq ~iff a b)
-  | _ -> neg (mk_nary_eq tl)
+let mk_distinct ~iff args =
+  (* This fix makes sure that the smart constructor agrees with
+     the SMT-LIB specification when used with at least 3 arguments.
+     To prevent a soundness bug, we translate the expected expression into a
+     conjonction of binary disequations. *)
```
bclement-ocp: Note: this is not specifically related to the SMT-LIB specification but to the usual definition of `distinct`.

src/lib/structures/expr.ml
```diff
@@ -1210,10 +1210,20 @@ let mk_nary_eq l =
...
+  let args = Array.of_list args in
+  let acc = ref vrai in
+  for i = 0 to Array.length args - 1 do
+    for j = i + 1 to Array.length args - 1 do
+      acc :=
+        mk_and (neg (mk_eq ~iff args.(i) args.(j))) !acc false
+    done;
+  done;
+  !acc
```
bclement-ocp, line 1227: Nit: it doesn't really matter here as this is a hotfix but conversion to an array is not needed:

```ocaml
let distinct_from x acc ys =
  List.fold_left (fun acc y -> mk_and (neg (mk_eq ~iff x y)) acc) acc ys
in let rec distinct_aux acc = function
  | [] -> acc
  | x :: ys -> distinct_aux (distinct_from x acc ys) ys
in distinct_aux acc args
```

src/lib/frontend/cnf.ml
```diff
@@ -239,9 +239,13 @@ and make_form up_qv name_base ~toplevel f loc ~decl_kind : E.t =
...
           if negated then E.neg res else res
 
-        | TAneq lt | TAdistinct lt ->
+        | TAneq lt ->
           let lt = List.map (make_term up_qv name_base) lt in
           E.mk_distinct ~iff:true lt
+
+        | TAdistinct lt ->
+          E.mk_distinct ~iff:true (List.map (make_term up_qv name_base) lt)
+
```
bclement-ocp: Can we avoid these changes? They don't seem useful, and keeping the patch as tiny as possible makes it easier to cherry-pick and move around if needed.

src/lib/frontend/d_cnf.ml
```diff
@@ -1337,7 +1337,7 @@ let rec mk_expr
             end
 
           | B.Distinct, _ ->
-            E.mk_distinct ~iff:true (List.map (fun t -> aux_mk_expr t) args)
+            E.mk_distinct ~iff:true (List.map aux_mk_expr args)
```
bclement-ocp: Same as above, can we undo this? There shouldn't be this sort of unrelated cleanup in a hotfix.

src/lib/structures/expr.ml
bclement-ocp: Also the comment should point to #889 which will track the fix in the `next` branch

bclement-ocp (changes requested):
> even if the API is undocumented

It actually is: https://ocamlpro.github.io/alt-ergo/Input_file_formats/Native/02_types/02_01_builtin.html#comparisons

Please remove the cleanups that have no place in a hotfix and it should be good.

src/lib/structures/expr.ml
Halbaroth: Actually I didn't plan to apply this patch on next but I can.

bclement-ocp: Sorry, I maybe was not clear. #889 is not about this patch, #889 is about the issue with `distinct`. We don't need to apply the patch on `next`. We will fix the issue with `distinct` on `next` in a different way, which is fine, but we should still refer to it here.

src/lib/structures/expr.ml
Halbaroth: I know but I think the imperative version is the easier to read and a bit less error-prone. It's a matter of taste.

src/lib/structures/expr.ml
Halbaroth: Ok :)

src/lib/frontend/cnf.ml
Halbaroth: Sure. Besides, we should use a different constructor for TAneq? It's not documented and I'm not sure what is the expected semantic for this constructor.

src/lib/structures/expr.ml
bclement-ocp: It is not only a matter of taste (I also find the imperative version easier to read in this specific case); the imperative version is slower and uses more memory because of the call to `Array.of_list`. It is entirely negligible here (hence why it's a nit), just pointing that out.

src/lib/frontend/cnf.ml
bclement-ocp: `TAneq [x; y]` is the constructor used for `x <> y`. "neq" is a fairly common shorthand for "**n**ot **eq**ual".

src/lib/frontend/cnf.ml
```diff
@@ -242,6 +242,7 @@ and make_form up_qv name_base ~toplevel f loc ~decl_kind : E.t =
         | TAneq lt | TAdistinct lt ->
           let lt = List.map (make_term up_qv name_base) lt in
           E.mk_distinct ~iff:true lt
+
```
bclement-ocp: Can you undo this whitespace change? Thanks.

bclement-ocp (changes requested):

src/lib/structures/expr.ml
Halbaroth: Sure ;)

bclement-ocp (approved):

Stevendeo (commented):
LGTM!

EDIT: I still had the notification to review the PR after it was merged

Halbaroth:
:D

PR #893, opened 2023-10-16
Locking the Dolmen to a dev commit

Halbaroth:
This commit locks the Dolmen used by AE to a specific commit in order to use the unreleased feature for custom statements (see #892).

I also notice that the `with_cache` in `d_cnf` was silently shadow by the same function from the module `Expr` of `Dolmen`. The signature of this latter changed recently. I fix it by using the `with_cache` from Dolmen. Notice that the hash table of the cache is hidden in the closure of `with_cache` and `bv2nat` and `int2bv` don't share their cache as before.

### DIFF
--- a/alt-ergo-lib.opam
+++ b/alt-ergo-lib.opam
@@ -59,3 +59,18 @@ license: [
   "LicenseRef-OCamlpro-Non-Commercial"
   "Apache-2.0"
 ]
+
+pin-depends: [
+  [
+  "dolmen.dev"
+  "git+https://github.com/Gbury/dolmen.git#baaf1f92ccd473294679dd9025c3ae9a441a82fa"
+  ]
+  [
+  "dolmen_loop.dev"
+  "git+https://github.com/Gbury/dolmen.git#baaf1f92ccd473294679dd9025c3ae9a441a82fa"
+  ]
+  [
+  "dolmen_type.dev"
+  "git+https://github.com/Gbury/dolmen.git#baaf1f92ccd473294679dd9025c3ae9a441a82fa"
+  ]
+]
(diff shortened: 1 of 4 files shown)


bclement-ocp (approved):
I don't think the caches for `bv2nat` and `int2bv` were shared before?

bclement-ocp:
Well we do need a Dolmen version that is compatible with 4.08 though, Gbury/dolmen#195

bclement-ocp (commented):
Actually now that I think about it -- I think it would be better to do this in the PR that will solve #892 so that we can track (using `git log` / `git blame`) the actual reason why we bump the dependencies.

Halbaroth:
> I don't think the caches for `bv2nat` and `int2bv` were shared before?

I know. I said I checked it was still the case after this commit.

Halbaroth:
> Actually now that I think about it -- I think it would be better to do this in the PR that will solve #892 so that we can track (using `git log` / `git blame`) the actual reason why we bump the dependencies.

Actually this commit was the very first commit of my PR for MaxSMT. I prefer separate it and create an remainder issue on this PR to remind we have to revert the lock before releasing.

bclement-ocp:
> > I don't think the caches for `bv2nat` and `int2bv` were shared before?
> 
> I know. I said I checked it was still the case after this commit.

Ah, sorry — I parsed "bv2nat and int2bv don't share their cache as before." as stating that `bv2nat` and `int2bv` shared their cache before but no longer do.

Halbaroth:
> > > I don't think the caches for `bv2nat` and `int2bv` were shared before?
> > 
> > 
> > I know. I said I checked it was still the case after this commit.
> 
> Ah, sorry — I parsed "bv2nat and int2bv don't share their cache as before." as stating that `bv2nat` and `int2bv` shared their cache before but no longer do.

My sentence was very misleading ;)

Halbaroth:
I prefer waiting the fix for Dolmen as we will need to change the lock again.

PR #896, opened 2023-10-16
Add ml_z_extract_small JS stub

### LINKED ISSUE #895: Missing primitives
It seems we don't implement some primitives for JS version:
```
There are some missing primitives                          
Dummy implementations (raising 'Failure' exception) will be used if they are not available at runtime.
You can prevent the generation of dummy implementations with the commandline option '--disable genprim'
Missing primitives:
  ml_z_extract_small
There are some missing primitives        
Dummy implementations (raising 'Failure' exception) will be used if they are not available at runtime.
You can prevent the generation of dummy implementations with the commandline option '--disable genprim'
Missing primitives:
  ml_z_extract_small
```
I got this message with `js_of_ocaml 5.4.0`

bclement-ocp:
ml_z_extract_small was added as a fast path in ZArith 1.13. It is not (yet) supported by zarith_stubs_js. Since it is a fast path for ml_z_extract, we can simply call ml_z_extract in missing_primitives.js for the time being.

Fixes #895

### DIFF
--- a/src/lib/missing_primitives.js
+++ b/src/lib/missing_primitives.js
@@ -39,3 +39,11 @@ function unix_getpid() {
 function unix_kill() {
   return 0;
 }
+
+// Fast path in ZArith 1.13, not in zarith_stubs_js
+// https://github.com/janestreet/zarith_stubs_js/issues/10
+// Provides: ml_z_extract_small
+// Requires: ml_z_extract
+function ml_z_extract_small(z1, pos, len) {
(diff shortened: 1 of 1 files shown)


Halbaroth (approved):
Thansk :)

The CI failed but it seems the issue comes from opam.

bclement-ocp:
Yeah, CI issue seems unrelated, forcing the merge.

PR #905, opened 2023-10-18
Uniformization of internal identifiers generation

Stevendeo:
This PR centralizes how internal identifiers are created. This work is a first step before a simpler task, which is to replace the '!' by '.' in name identifiers.

### DIFF
--- a/src/lib/frontend/typechecker.ml
+++ b/src/lib/frontend/typechecker.ml
@@ -2152,7 +2152,7 @@ let axioms_of_rules loc name lf acc env =
   let acc =
     List.fold_left
       (fun acc f ->
-         let name = (Hstring.fresh_string ()) ^ "_" ^ name in
+         let name = (Symbols.InternalNameId.fresh ()) ^ "_" ^ name in
          let td = {c = TAxiom(loc,name,Util.Default, f); annot = new_id () } in
          (td, env)::acc
       ) acc lf
--- a/src/lib/frontend/typechecker.ml
+++ b/src/lib/frontend/typechecker.ml
@@ -2152,7 +2152,7 @@ let axioms_of_rules loc name lf acc env =
   let acc =
     List.fold_left
       (fun acc f ->
-         let name = (Symbols.InternalNameId.fresh ()) ^ "_" ^ name in
+         let name = (Symbols.InternalId.fresh ()) ^ "_" ^ name in
          let td = {c = TAxiom(loc,name,Util.Default, f); annot = new_id () } in
          (td, env)::acc
       ) acc lf
--- a/src/lib/frontend/typechecker.ml
+++ b/src/lib/frontend/typechecker.ml
@@ -2152,7 +2152,11 @@ let axioms_of_rules loc name lf acc env =
   let acc =
     List.fold_left
       (fun acc f ->
-         let name = (Symbols.InternalId.fresh ()) ^ "_" ^ name in
+         let name =
+           Fmt.str "%s_%s"
+             (Symbols.fresh_internal_string ())
+             name
+         in
          let td = {c = TAxiom(loc,name,Util.Default, f); annot = new_id () } in
          (td, env)::acc
       ) acc lf
--- a/src/lib/structures/expr.ml
+++ b/src/lib/structures/expr.ml
@@ -811,17 +811,6 @@ let neg t =
 let is_int t = t.ty == Ty.Tint
 let is_real t = t.ty == Ty.Treal
 
-let is_fresh t =
-  match t with
-  | { f = Sy.Name (hs, _, _); xs = []; _ } ->
-    Hstring.is_fresh_string (Hstring.view hs)
-  | _ -> false
-
-let is_fresh_skolem t =
-  match t with
-  | { f = Sy.Name (hs, _, _); _ } -> Hstring.is_fresh_skolem (Hstring.view hs)
-  | _ -> false
-
 let name_of_lemma f =
   match form_view f with
   | Lemma { name; _ } -> name
@@ -938,7 +927,14 @@ let vrai =
 let faux = neg (vrai)
 let void = mk_term (Sy.Void) [] Ty.Tunit
 
-let fresh_name ty = mk_term (Sy.name (Hstring.fresh_string())) [] ty
+let fresh_name ty = mk_term (Sy.fresh_name ()) [] ty
+
+let is_internal_name t =
+  match t with
+  | { f; xs = []; _ } -> Sy.is_internal_name f
+  | _ -> false
+
+let is_internal_skolem t = Sy.is_skolem t.f
 
 let positive_int i = mk_term (Sy.int i) [] Ty.Tint
 
@@ -1230,7 +1226,7 @@ let is_skolem_cst v =
 
 let get_skolem =
   let hsko = Hsko.create 17 in
-  let gen_sko ty = mk_term (Sy.fresh "@sko") [] ty in
+  let gen_sko ty = mk_term (Sy.fresh_skolem "@sko") [] ty in
   fun v ty ->
     try Hsko.find hsko v
     with Not_found ->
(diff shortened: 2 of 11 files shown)


src/lib/frontend/typechecker.ml
```diff
@@ -2152,7 +2152,7 @@ let axioms_of_rules loc name lf acc env =
   let acc =
     List.fold_left
       (fun acc f ->
-         let name = (Hstring.fresh_string ()) ^ "_" ^ name in
+         let name = (Symbols.InternalNameId.fresh ()) ^ "_" ^ name in
```
Halbaroth: I suggest to use `InternalId` instead of `InternalNameId`.

src/lib/structures/symbols.mli
```diff
@@ -128,9 +128,23 @@ val print_clean : Format.formatter -> t -> unit
 
 (*val dummy : t*)
 
-val fresh : ?is_var:bool -> string -> t
+module type Id = sig
+  val fresh : ?base:string -> unit -> string
+  val reset_fresh_cpt : unit -> unit
+  val is_id : string -> bool
+  val make_as_fresh : string -> string
```
Halbaroth: Is it necessary to expose the function `make_as_fresh`?

src/lib/structures/symbols.ml
```diff
@@ -367,18 +367,61 @@ let print_clean fmt s = Format.fprintf fmt "%s" (to_string_clean s)
...
+  let len_pre = String.length S.prefix
+
+  let is_id s =
+    let len_s = String.length s in
+    let rec aux i =
+      if i = len_pre then true
+      else
+        Char.equal (String.unsafe_get s i) (String.unsafe_get S.prefix i)
+        && aux (i + 1)
+    in len_s >= len_pre && aux 0
```
hra687261:
```suggestion
    in len_s > len_pre && aux 0
```
Assuming that names as empty strings don't occur

src/lib/structures/symbols.mli
```diff
@@ -128,9 +128,23 @@ val print_clean : Format.formatter -> t -> unit
...
+
+module InternalNameId : Id
+module SkolemId : Id
+
+val fresh_name : unit -> t
+val is_internal_name : t -> bool
+
+val fresh_skolem : ?is_var:bool -> string -> t
+val make_as_fresh_skolem : string -> t
+val is_skolem : t -> bool
```
hra687261: Suggestion, just for consistency:
`fresh_name` -> `fresh_interna_name`
`is_internal_name` -> `is_fresh_internal_name`
`is_skolem` -> `is_fresh_skolem`

hra687261 (approved):
LGTM modulo a couple of minor details
If my memory serves me right, these fresh names are used for comparisons in some cases, out of prudence, we should probably run some benchmarks, before merging.

src/lib/structures/symbols.ml
Stevendeo: `||` is a valid SMTLib2 identifier corresponding to the empty string.

src/lib/structures/symbols.mli
Stevendeo: Unfortunately, yes... https://github.com/OCamlPro/alt-ergo/pull/905/files#diff-89722ab84fe15db7d6cd1e3a8533c9d5e0660599698ae64313faae9cf6f23730L1720

Halbaroth: You only need `make_as_fresh_skolem` here ;)

src/lib/structures/symbols.ml
Halbaroth: Oh gosh... the SMT-LIB accepts too much things as identifiers...

src/lib/structures/symbols.mli
Stevendeo: Oh okay I get your point; if I remove the `make_as_fresh`, I can also remove the `reset_fresh_counter` and hide the whole module.

bclement-ocp: We may need to keep `reset_fresh_counter` for alt-ergo-fuzz, @hra687261 can you confirm?

src/lib/structures/symbols.ml
bclement-ocp: I think this is a reimplementation of `Compat.String.starts_with ~prefix:S.prefix`?

src/lib/structures/expr.ml
```diff
@@ -1230,7 +1226,7 @@ let is_skolem_cst v =
 
 let get_skolem =
   let hsko = Hsko.create 17 in
-  let gen_sko ty = mk_term (Sy.fresh "@sko") [] ty in
+  let gen_sko ty = mk_term (Sy.fresh_skolem "@sko") [] ty in
```
bclement-ocp, line 1229: I think that `Sy.fresh` used to add an extra `__` prefix (with the comment "(* garder le suffixe "__" car cela influence l'ordre *)") but it looks like `Sy.fresh_skolem` doesn't, so unless I missed something we used to get `!?__@sko` and we now get `!?@sko` here.

bclement-ocp (changes requested):
Agree with @hra687261 — this needs to be tested at least on ae-format and benchtop-tests. Marking as "change requested" so that we don't merge by mistake yet.

(Note that preserving the exact names of the symbols don't matter; *if* there is no change in the problems solved — but the comments hint that there may be some)

src/lib/structures/symbols.mli
Stevendeo: > We may need to keep `reset_fresh_counter` for alt-ergo-fuzz, @hra687261 can you confirm?

I meant hide the module by wrapping the reset_fresh_cpt in a specific function (which is what I did in the last version of the PR)

src/lib/structures/symbols.ml
Stevendeo: > I think this is a reimplementation of `Compat.String.starts_with ~prefix:S.prefix`?

Oh yes indeed

src/lib/structures/expr.ml
hra687261: As it is outdated I can't reply to the right comment, but `reset_id_builders` will do instead of `reset_fresh_counter`, thanks for noticing that @bclement-ocp !

Stevendeo: > I think that `Sy.fresh` used to add an extra `__` prefix (with the comment "(* garder le suffixe "__" car cela influence l'ordre *)") but it looks like `Sy.fresh_skolem` doesn't, so unless I missed something we used to get `!?__@sko` and we now get `!?@sko` here.

That makes me think this whole check system is unsound, because an identifier can start with `.?` if is has been defined quoted.

bclement-ocp: This is why we want to replace `!` with `.`! Identifiers starting with `.` or `@` are reserved for solver usage and cannot be used in scripts, dolmen ensures this (or will in the next release, see Gbury/dolmen#185 ).

bclement-ocp: Ah, sorry, I misread your remark — indeed, according to the SMT-LIB it can, but I think after Gbury/dolmen#185 identifiers starting with `.?` are rejected even in quoted form (which is incorrect from a strict reading of the standard).

Stevendeo: > Ah, sorry, I misread your remark — indeed, according to the SMT-LIB it can, but I think after [Gbury/dolmen#185](https://github.com/Gbury/dolmen/issues/185) identifiers starting with `.?` are rejected even in quoted form (which is incorrect from a strict reading of the standard).

You did not, I miswrote it and edited, you might have seen the old version :)

bclement-ocp (commented):
I believe these changes are correct and should not change the identifiers we generate, so there should be zero performance impact here, but please run benchmarks on `ae-format` and `benchtop-tests` to confirm.

bclement-ocp:
I am not sure why you re-requested a review from me — I think we are good and just waiting on benchmark results, or is there a specific part you want me to look at?

Stevendeo:
> I am not sure why you re-requested a review from me — I think we are good and just waiting on benchmark results, or is there a specific part you want me to look at?

Still me discovering github, sorry about that

Stevendeo:
No regressions (and even 1 improvement), we can merge!

Stevendeo:
@bclement-ocp Can we merge now?

bclement-ocp (approved):

bclement-ocp:
I think so — sorry I got confused somehow, thought I had approved already.

Issue #806, opened 2023-09-11
Proper support for quoted identifiers in models


bclement-ocp:
By the way we probably have a problem with quoted identifiers as well… This is unrelated to this PR however.

_Originally posted by @bclement-ocp in https://github.com/OCamlPro/alt-ergo/issues/804#issuecomment-1714038192_
            

We do.

```console
$ cat quoted.smt2
(set-logic ALL)
(declare-const |fail))| Int)
(assert (= |fail))| 0))
(check-sat)
(get-model)
$ alt-ergo --frontend dolmen --produce-models quoted.smt2
                                       
; File "quoted.smt2", line 4, characters 1-12: I don't know (0.4269) (1 steps) (goal g_1)

unknown
(
  (define-fun fail)) () Int 0)
)
```

bclement-ocp:
Assigning @Halbaroth who said he was interested in tackling this

Stevendeo:
There are few ways to tackle this issue. If Dolmen does not provide the information that the identifier was defined quoted (which would be IMO the best way to do to precisely match the user input), we can : 
- quote all the names in the output (easy, no cost, 100% sound)
- check all names when they are built or printed that they do not contain special characters (not as easy, negligible-but-existent cost, error prone)

PR #909, opened 2023-10-18
Quoted identifiers

Stevendeo:
Tackles #806


src/lib/structures/symbols.ml
```diff
@@ -294,8 +294,21 @@ let string_of_form f = match f with
...
   | F_Xor -> "xor"
 
+let name_to_string =
+  let no_need_to_quote =
+    String.for_all
+      (function
+        | 'a' .. 'z'
+        | 'A' .. 'Z'
+        | '0' .. '9'
+        | '-' | '_' -> true
```
bclement-ocp: - The list to use is: ~ ! @ $ % ^ & * _ - + = < > . ? / according to the SMT-LIB standard.
 - Names that start with a digit must always be quoted
 - SMT-LIB keywords must always be quoted

Halbaroth: We may add @Gbury to add a tag on `quoted` name. It makes sense that Dolmen outputs this information.

Stevendeo: > SMT-LIB keywords must always be quoted

It does not make sens for alt-ergo to maintain the list of reserved SMTLib2 keywords for the only purpose of printing identifiers, I agree with @Halbaroth

bclement-ocp: To be fair we can probably just ignore the SMT-LIB keywords case, but the others are important

bclement-ocp (approved):

src/lib/structures/symbols.ml
Gbury: > It does not make sens for alt-ergo to maintain the list of reserved SMTLib2 keywords for the only purpose of printing identifiers, I agree with @Halbaroth

Actually yes it does, at least in my opinion, and for a few reasons, that also explain why simply tagging quoted identifiers from the source is not a good solution:
- alt-ergo can create new identifiers that it will later want to print: for those, alt-ergo must decide how to print them anyway since there can be no info coming from the source (since these do not appear in the source)
- since the function that determines how to print an identifier must exist, then why not use it for all identifiers, and save ourselves the trouble ?
- what happens when/if we one day want to have cross-formatting, for instance read a problem in tptp, and output a model in smtlib format ? Any annotation from parsing in dolmen will be useless as the lexical requirements are different. Same problem if/when we want to add a new format for models. In that case, we need the list of reserved keywords for the language in which we want to print (which may be different from the input format).
- at one point we may expect to have users of the alt-ergo library (e.g. opam ?), in which case alt-ergo might have to deal with terms that do not necessarily come from the dolmen frontend, so again, dolmen annotations on quoted identifiers would be useless.

Basically, the summary of those points is that correctly printing terms is not that easy. At one point in the future, dolmen will have printers that will abstract away most of the complexity, and that alt-ergo may or may not use (but as shown here, it's not trivial to have a correct printer). In the meantime, doing what @bclement-ocp suggests in his comment above seems like the best solution.

Stevendeo: I don't really want to argue about this minor issue, I just regret there is no link between the actual user input and the printable expressions. Let's go for this last solution.

PR #912, opened 2023-10-19
Use efficient `String.starts_with` implementation

bclement-ocp:
The existing implementation for `Compat.String.starts_with` needlessly makes a copy of the prefix. This patch instead uses the implementation of `String.starts_with` from recent versions of the OCaml standard library, which avoids the copy.

### DIFF
--- a/src/lib/util/compat.ml
+++ b/src/lib/util/compat.ml
@@ -1,11 +1,22 @@
+(* Locally restore the default polymorphic operators of the Stdlib, to make it
+   easier to copy code from there without modifications. *)
+open Stdlib
+
 module String = struct
   open Stdlib.String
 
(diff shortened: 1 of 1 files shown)


Halbaroth:
In #866, I replace our implementation of `starts_with` by the implementation from the `stdcompat` library. I guess the implementation of `stdcompat` does the right thing. Still we can merge this PR even if we'll revert the change soon.

Stevendeo (approved):
LGTM

bclement-ocp:
Yes, the implementation of `starts_with` in `stdcompat` is the same as this one and is also copied from the standard library. #866 for now depends on #861 which is not finished, and so can't be merged ATM.

Issue #897, opened 2023-10-18
Adding an `--optimize` flag option


Halbaroth:
This flag should behave as follow:
- In presence of the flag, the custom statements `maximize`, `minimize` and `get-objectives` should be available as SMT-LIB statements and Alt-Ergo may perform extra works to fulfill these objectives. In the current implementation, the `FunSAT` solver doesn't support optimization and we don't plan to support it in future. Thus, in the presence of `--optimize`, the binary should select the `CDCL-Tableaux` solver and produce a warning in presence of `--sat-solver Tableaux` or `--sat-solver Tableaux-CDCL` options.
- In absence of the flag, the name `maximize` and `minimize` are not reserved and `get-objectives` should produce a SMT-LIB error.

PR #916, opened 2023-10-19
feat: Option --optimize

Stevendeo:
Adresses #897

### DIFF
--- a/src/bin/common/parse_command.ml
+++ b/src/bin/common/parse_command.ml
@@ -1011,33 +1011,31 @@ let parse_output_opt =
         | Util.CDCL_Tableaux ->
           set_sat_solver sat_solver;
           set_cdcl_tableaux_inst cdcl_tableaux_inst;
-          set_cdcl_tableaux_th cdcl_tableaux_th
+          set_cdcl_tableaux_th cdcl_tableaux_th;
+          Ok ()
         | Util.CDCL ->
           set_sat_solver sat_solver;
           set_cdcl_tableaux_inst false;
-          set_cdcl_tableaux_th false
+          set_cdcl_tableaux_th false;
+          Ok ()
         | s when optim ->
-          (* Optimization is so far only sound for CDCL & CDCL_Tableaux. We set
-             CDCL_Tableaux by default. *)
-          Fmt.pf
-            Format.err_formatter
-            ";Warning: --optimize is not compatible with %a: forcing it to %a@."
-            Util.pp_sat_solver s
-            Util.pp_sat_solver Util.CDCL_Tableaux;
-          set_sat_solver Util.CDCL_Tableaux;
-          set_cdcl_tableaux_inst cdcl_tableaux_inst;
-          set_cdcl_tableaux_th cdcl_tableaux_th
+          Fmt.error
+            "--optimize is not compatible with Sat solver %a@."
+            Util.pp_sat_solver s;
         | _ ->
           set_sat_solver sat_solver;
           set_cdcl_tableaux_inst false;
-          set_cdcl_tableaux_th false
-      end;
-      ()
+          set_cdcl_tableaux_th false;
+          Ok ()
+      end
     in
-    Term.(
-      const set_sat_options $ sat_solver $ cdcl_tableaux_inst
-      $ cdcl_tableaux_th $ optim
-    )
+    let term =
+      Term.(
+        const set_sat_options $ sat_solver $ cdcl_tableaux_inst
+        $ cdcl_tableaux_th $ optim
+      )
+    in
+    Term.term_result' term
   in
 
   let use_underscore =
--- a/src/bin/common/parse_command.ml
+++ b/src/bin/common/parse_command.ml
@@ -363,7 +363,7 @@ let mk_limit_opt age_bound fm_cross_limit timelimit_interpretation
     `Ok()
 
 let mk_output_opt
-    interpretation optim use_underscore unsat_core output_format model_type
+    interpretation use_underscore unsat_core output_format model_type
     () () () ()
   =
   set_infer_output_format (Option.is_none output_format);
@@ -376,7 +376,6 @@ let mk_output_opt
     | Some v -> v
   in
   set_interpretation interpretation;
-  set_optimize optim;
   set_interpretation_use_underscore use_underscore;
   set_unsat_core unsat_core;
   set_output_format output_format;
@@ -1097,7 +1096,7 @@ let parse_output_opt =
   in
 
   Term.(ret (const mk_output_opt $
-             interpretation $ optim $
+             interpretation $
              use_underscore $ unsat_core $
              output_format $ model_type $
              set_dump_models $ set_dump_models_on $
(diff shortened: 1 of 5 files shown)


src/bin/common/parse_command.ml
```diff
@@ -997,22 +997,46 @@ let parse_output_opt =
...
+          set_cdcl_tableaux_th false
+        | s when optim ->
+          (* Optimization is so far only sound for CDCL & CDCL_Tableaux. We set
+             CDCL_Tableaux by default. *)
+          Fmt.pf
+            Format.err_formatter
+            ";Warning: --optimize is not compatible with %a: forcing it to %a@."
+            Util.pp_sat_solver s
+            Util.pp_sat_solver Util.CDCL_Tableaux;
+          set_sat_solver Util.CDCL_Tableaux;
```
bclement-ocp: I don't think overriding an explicitly requested solver is a good thing to do — we should rather fail using `term_result` if conflicting options are given.

src/bin/common/solving_loop.ml
```diff
@@ -509,10 +530,57 @@ let main () =
...
+    | Dolmen.Std.Id.{name = Simple "maximize"; _}, [term] ->
+      handle_maximize_term term st
+    | Dolmen.Std.Id.{name = Simple "get-objectives"; _}, args ->
+      handle_get_objectives args st
+    | Dolmen.Std.Id.{name = Simple (("minimize" | "maximize") as ext); _}, _ ->
+      recoverable_error
+        "Statement %s only expects 1 argument (%i given)"
+        ext
+        (List.length args);
+      st
```
bclement-ocp, line 598: These should be `when O.get_optimize ()`

src/bin/common/solving_loop.ml
```diff
@@ -88,6 +89,26 @@ let unsupported_opt opt =
...
   warning "unsupported option %s" opt
 
+let set_sat_solver s =
+  match s with
+  | Util.CDCL_Tableaux | CDCL -> O.set_sat_solver s
+  | _ when O.get_optimize () ->
+    warning
+      "Sat-solver %a is incompatible with optimization: ignoring command."
+      Util.pp_sat_solver
+      s
```
bclement-ocp: This is not correct — the sat solver is stlil set on the Dolmen state even though it is not updated through options. Either the check needs to be in `handle_option` and `State.set sat_solver_key` must not be called, or we must call both `O.set_sat_solver` and `O.set_optimize false` here.

(Yes, the way we manage options is currently a mess, sorry)

src/bin/common/solving_loop.ml
bclement-ocp: In fact, instead of using `O.get_optimize ()`, it would be better to:

 - Add a new boolean key `optimize` to the Dolmen state
 - Initialize the boolean key with `O.get_optimize`
 - Use `when State.get st optimize_key` here
 - Update the `optimize_key` on `set-option :optimize` but *not* call `O.set_optimize`
 - Reset the `optimize_key` to `O.get_optimize` when `(reset)` is encountered

This allows the set-option to be local to the Dolmen state, and is correct wrt the semantics of `(reset)` mandated by the SMT-LIB standard (which should reinitialize the state of the solver to what it was at the beginning of parsing the file, so should undo the `set-option` even though it currently doesn't — but we can move towards that slowly).

bclement-ocp (changes requested):

src/bin/common/solving_loop.ml
bclement-ocp: For further clarification — the SAT solver that is actually used by the binary is the one that is set on the Dolmen state. The options are only updated because of code in the SAT solvers that triggers different behavior depending on `Options.get_cdcl_tableaux` and related. We want to move away from code that depends on global `Options` values in the library as much as possible, but for now we have no choice but to update those options.

Stevendeo: I understand. In an ideal world, options should only be used to initialize dolmen's state and never read again afterwards.

src/bin/common/solving_loop.ml
```diff
@@ -722,6 +809,7 @@ let main () =
         Parser.parse_logic ~preludes logic_file
       in
       let st = State.set Typer.additional_builtins D_cnf.builtins st in
+      let st = State.set optimize_key (O.get_optimize ()) st in
```
bclement-ocp: I think this should rather be in `mk_state` — in fact the previous line should be moved there as well.

src/bin/common/solving_loop.ml
```diff
@@ -410,6 +415,36 @@ let main () =
...
+        "Sat-solver %a is incompatible with optimization: ignoring command."
+        Util.pp_sat_solver
+        sat;
+      st
+    | Tableaux | Tableaux_CDCL | CDCL | CDCL_Tableaux ->
+      O.set_sat_solver sat;
+      State.set
+        sat_solver_key
+        (make_sat ())
+        st
```
bclement-ocp: This deserves a comment explaining that `make_sat` dispatches on `O.get_sat_solver`

bclement-ocp (changes requested):
I think this is mostly good, but we should call `DStd.Extensions.Smtlib2.(enable maxsmt)` when initializing Dolmen as well, and either call `DStd.Extensions.Smtlib2.(disable maxsmt)` when disabling optimizations or have the proper guards in place in `handle_custom_statement` to error out on minimize/maximize/etc. if the option is not provided (2nd one is probably better so that we can instruct the user to set the `optimize` flag)

src/bin/common/solving_loop.ml
```diff
@@ -410,6 +417,44 @@ let main () =
...
+      else
+        DStd.Extensions.Smtlib2.(disable maxsmt)
+    in
+    let sat, _ = State.get sat_solver_key st in
+    match sat with
+    | Util.Tableaux | Tableaux_CDCL when optim ->
+      warning
+        "Sat-solver %a is incompatible with optimization: ignoring command."
+        Util.pp_sat_solver
+        sat;
```
bclement-ocp, line 451: You are not ignoring the command here since you call `DStd.Extensions.Smtlib2.(enable maxsmt)` above.

src/bin/common/solving_loop.ml
```diff
@@ -410,6 +417,44 @@ let main () =
...
+      st
+    | Tableaux | Tableaux_CDCL | CDCL | CDCL_Tableaux ->
+      O.set_optimize optim;
```
bclement-ocp: `O.set_optimize` should not be called here for proper `(reset)` behavior.

bclement-ocp (changes requested):

src/bin/common/solving_loop.ml
Stevendeo: For an obscure reason, moving the previous line to the `mk_state` makes alt-ergo fails on tests where it needs built-ins.

bclement-ocp (approved):

Halbaroth:
@Stevendeo  Can I merge?

bclement-ocp:
Sorry, forgot to do it

PR #918, opened 2023-10-27
Changing the internal variables prefix

Stevendeo:
Follows #905


bclement-ocp:
This needs benchmarks on ae-format and benchtop-tests to be safe.

Halbaroth (approved):
No regression :)

PR #920, opened 2023-10-30
Fix linguist-generated

Halbaroth:
Fix the PR #830. The syntax I used for the `tag` linguist-generated was wrong. Should work now :)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,4 +2,4 @@
 configure eol=lf
 
 # Hide the file in statistics and pull request diffs.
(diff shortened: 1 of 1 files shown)


bclement-ocp (commented):
I think this is working with the current syntax? I have the message "Some generated files are not rendered by default" on the `dune.inc` file already.

image

Halbaroth:
It seems you are right... there is no way to prevent GitHub to count the lines modified in `dune.inc`. I thought the feature didn't work.

Issue #904, opened 2023-10-18
Supporting the float theory of the SMT-LIB


Halbaroth:
The current implementation of Alt-Ergo doesn't support the float theory https://smtlib.cs.uiowa.edu/theories-FloatingPoint.shtml but 
a theory about round functions from the `real` to subset of `real` corresponding to finite floats with fixed precision. 
In particular, our implementation doesn't support infinity values or `NaN` values.

bclement-ocp:
Postponing after discussion with the Why3 team, not a priority for now.

Issue #901, opened 2023-10-18
Case-splitting directly in the SAT solver


Halbaroth:

bclement-ocp:
Some progress report:

I have a better understanding of satML and a couple partially working prototypes for doing this, but did not run any large-scale benchmarks with them yet. There are still some things I am not entirely sure how to do properly with the combined CDCL+Tableaux solver, and I have realized that the optimization part is also affected by this, which I somehow overlooked initially. It should not be too big of a deal, but it is a bit annoying — and does make me want to pause this work until #921 is completed (even though I am not sure how much #921 touches this specific part of the optimization code).

(The following are mostly intended to be notes for myself when I get back to this)

I currently have two prototypes:
 - One that "just" allows satML atoms to be of type `X.r Xliteral.view` and is well integrated with the existing solver
 - One that separates case splits from satML atoms and is slightly less well integrated with the solver, but trying to replicate the current behavior of `try_it` / `look_for_sat` as much as possible

(I have not tried implementing something for the Tableaux solver but as I mentioned last week I don't anticipate too much difficulty there, as the Tableaux solver is simpler — but also it is less important because mostly unused)

The main issue with the first solution is that I don't really know how to do "garbage collection", which means that all the case splits that we have done in one branch of the sat will still be present in other branches. I think I can pop them when cancelling (e.g. splits made at level 7 will be collected when cancelling to a lower level), but they could still end up in learnt clauses, which is an issue, especially since I think that being able to incorporate splits into learnt clauses is an advantage of this first method over the second one. Might want to disable learning for clauses that involve case splits.

The main issue with the second solution is that the case splits really feel bolted on. In particular, there is no clause learning, and there are subtleties that I don't 100% understand yet — with this implementation, I still have infinite loops of making the same case split over and over again, which I think is due to throwing away the splits too early, but it seems to me like not doing that would require re-implementing a whole chunk of what is done for booleans and might not be worth it.

My current opinion is that the right way forward is: `Xliteral` in atoms (it is not as hard as I thought — modulo, again, CDCL+Tableaux and Tableaux+CDCL shenanigans that I don't have quite looked at in detail yet), but the same logic that I wanted to do for the separate case splits.

PR #921, opened 2023-10-31
Clean up OptimAE

### LINKED ISSUE #891: Add support for `get-objectives`

### LINKED ISSUE #892: Supporting the MaxSMT extension of Dolmen

Halbaroth:
This PR refactors the way we manage optimization constraints in
Alt-Ergo.

- Optimization constraints aren't tag on subformulas anymore. Instead
  we use the MaxSMT syntaxes.
- We don't support optimization in Alt-Ergo native format. As this
  feature haven't been published, we don't need to deprecated it.
- We support `maximize`, `minimize` and `get-objectives`. There is no
  support for `assert-soft`. I think it's out of the scope of this PR. 
  Besides, it seems that `get-objectives` can take arguments in our current implementation of `Solving_loop` but I didn't find 
  any examples using this syntax, thus I silently ignore these arguments.
- The SAT solver API exposes two new functions: `optimize` and
  `get_objectives` which, respectively, register objective function and
  return the current objective model.


src/lib/frontend/frontend.ml
```diff
@@ -429,6 +437,8 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
 
   let th_assume = wrap_f internal_th_assume
 
+  let optimize = handle_sat_exn internal_optimize
```
bclement-ocp: This needs to be `wrap_f`, not `handle_sat_exn` to avoid using an invalid satML environment (see #946 and related issues).

src/lib/frontend/models.ml
```diff
@@ -604,4 +575,3 @@ let output_concrete_model fmt m =
   (*     SmtlibCounterExample.output_arrays_counterexample fmt m.arrays; *)
 
   Printer.print_fmt fmt "@]@,)";
```
bclement-ocp:
```suggestion
  Printer.print_fmt fmt "@]@,)"
```

src/lib/reasoners/fun_sat.ml
```diff
@@ -1123,9 +1123,11 @@ module Make (Th : Theory.S) = struct
     else begin
       try
         (* also performs case-split and pushes pending atoms to CS *)
-        let model = Th.compute_concrete_model env.tbox in
-        env.last_saved_model := model;
-        env
+        match Th.compute_concrete_model env.tbox with
```
bclement-ocp: When can the theory fail to compute a model?

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2045,17 +2045,35 @@ let case_split env uf ~for_model =
...
+
+let optimizing_objective env uf Objective.Function.{ e; to_max; _ } =
   (* soundness: if there are expressions to optmize, this should be
      done without waiting for ~for_model flag to be true *)
-  let {Th_util.r = r; is_max = to_max; e; value; _ } = opt_split in
-  assert (match value with
-      | Unknown -> true
-      | _ -> false
-    );
+  let uf, _ = Uf.add uf e in
```
bclement-ocp: This does not seem right; we don't return the `uf` so the added repr is lost.

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2045,17 +2045,35 @@ let case_split env uf ~for_model =
       end
   | _ -> res
 
-let optimizing_split env uf opt_split =
+(* Helper function used in [optimizing_objective] to pick a value
+   for the polynomial [p] in its interval. We used this function in the case
```
bclement-ocp:
```suggestion
   for the polynomial [p] in its interval. We use this function in the case
```

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2045,17 +2045,35 @@ let case_split env uf ~for_model =
...
-  let {Th_util.r = r; is_max = to_max; e; value; _ } = opt_split in
-  assert (match value with
-      | Unknown -> true
-      | _ -> false
-    );
+  let uf, _ = Uf.add uf e in
   let repr, _ = Uf.find uf e in
   let ty = E.type_info e in
-  let r1 = r in (* instead of repr, which may be a constant *)
+  let r1, _ = X.make e in (* instead of repr, which may be a constant *)
```
bclement-ocp: Since `e` was added to the union-find this should be `Uf.make uf e` to avoid calling `X.make` twice.

src/lib/reasoners/sat_solver_sig.ml
```diff
@@ -97,6 +97,10 @@ module type S = sig
   (* [pred_def env f] assume a new predicate definition [f] in [env]. *)
   val pred_def : t -> Expr.t -> string -> Explanation.t -> Loc.t -> unit
 
+  (* TODO: update this documentation. *)
```
bclement-ocp: Still a TODO I guess.

src/lib/reasoners/sat_solver_sig.mli
```diff
@@ -79,6 +79,10 @@ module type S = sig
   (** [pred_def env f] assume a new predicate definition [f] in [env]. *)
   val pred_def : t -> Expr.t -> string -> Explanation.t -> Loc.t -> unit
 
+  (* TODO: update this documentation. *)
```
bclement-ocp: Ditto

src/lib/reasoners/satml.ml
```diff
@@ -1857,4 +1859,7 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
       end;
     enqueue env g.neg 0 None
 
+  let optimize env ~to_max obj =
+    (* TODO: Should we add the objective function to unit_tenv? *)
```
bclement-ocp: No, the objective function is only relevant when the environment is used for splitting, and the unit environment is not used for splitting.

src/lib/reasoners/theory.ml
```diff
@@ -52,16 +52,19 @@ module type S = sig
     (E.t * Explanation.t * int * int) list -> t ->
     t * Expr.Set.t * int
 
+  val optimize : t -> to_max:bool -> Expr.t -> t
+  (** [optimize env obj] *)
+  (* TODO: update this documentation. *)
```
bclement-ocp: Still a TODO

src/lib/util/numbers.mli
```diff
@@ -155,4 +155,11 @@ module Q : sig
   val root_excess  : t -> int -> t option
   val sqrt_default : t -> t option
   val sqrt_excess  : t -> t option
+
+  val (~$$) : Z.t -> t
```
bclement-ocp: Unrelated to the PR but really we should just get rid of the `Numbers` module and use `Z` and `Q` directly, this is a relic of the time we supported either Num or Zarith.

(Well there is the issue of `Q.to_Z` not asserting in Zarith I guess)

src/lib/structures/typed.mli
```diff
@@ -285,6 +285,11 @@ and 'a tdecl =
   | TExit of Loc.t
   (** Exits the solver. *)
 
+  | TOptimize of Loc.t * 'a atterm * bool
+  (** Optimization declaration.
+      [TOptimize (loc, obj, to_max] declares an objective function [obj]. The
```
bclement-ocp:
```suggestion
      [TOptimize (loc, obj, to_max)] declares an objective function [obj]. The
```

src/lib/structures/objective.ml
```diff
@@ -0,0 +1,118 @@
...
+(*                                                                        *)
+(**************************************************************************)
+
+module Function = struct
+  type t = {
+    e : Expr.t;
+    to_max : bool;
+  }
+
+  let equal { e = e1; _ } { e = e2; _ } = Expr.equal e1 e2
```
bclement-ocp: I think this should also compare `to_max`, otherwise if I do `(minimize x)(maximize x)` it has the same effect as `(maximize x)` whereas it should have the same effect as `(minimize x)` (because we must first minimize x *then* maximize it, which does nothing).

src/lib/structures/objective.ml
```diff
@@ -0,0 +1,118 @@
...
+      let compare { order = o1; _ } { order = o2; _ } = o1 - o2
+    end)
+
+  type t = Value.t M.t
+
+  let empty = M.empty
+  let is_empty = M.is_empty
+  let fold g = M.fold (fun { f; _ } acc -> g f acc)
+
+  let add o v mdl =
```
bclement-ocp, line 82: Nit: I was slightly confused by calling a function `o` (would have expected an order), why not `g`?

Halbaroth:
Uhm, I thought it was a good idea to hide completely the priority order in the objective model but in fact, users could feed AE with a file like this one:
```
(set-logic ALL)
(declare-const x Int)
(declare-const y Int)
(assert (<= x 10))
(assert (< y 5))
(maximize x)
(maximize y)
(maximize x)
(check-sat)
(get-model)
(get-objectives)
```
With the current implementation, the second `maximize` will be completely ignored and output:
```
unknown
(
  (define-fun x () Int 10)
  (define-fun y () Real 0.0)
)
(objectives 
  (x 10)
  (y (- 5.0 epsilon))
)
```

Another tricky example is if we use both `maximize` and `minimize` on the same expression. The second one is ignored. Should we print a warning?

src/lib/reasoners/theory.ml
```diff
@@ -996,22 +888,35 @@ module Main_Default : S = struct
...
   let get_assumed env = env.assumed_set
 
-  let reinit_cpt () =
-    Debug.reinit_cpt ()
+  let optimize env ~to_max f =
+    let obj = Objective.Function.mk ~to_max f in
+    let objectives =
+      Objective.Model.add obj Objective.Value.Unknown env.objectives
+    in
+    let env = reset_case_split_env env in
```
bclement-ocp, line 908: There is a comment in `reset_case_split_env` that states: "be sure it's always correct when this function is called" (wrt the `cs_pending_facts`), but now we can call `reset_case_split_env` from a public function. Are we sure it's always correct?

(I think it is, because from my understanding setting `cs_pending_facts` to `[]` is always correct if we reset `gamma_finite` to `gamma`)

src/lib/reasoners/theory.ml
```diff
@@ -996,22 +888,35 @@ module Main_Default : S = struct
   let get_case_split_env t = t.gamma_finite
 
   let compute_concrete_model env =
-    let { gamma_finite; assumed_set; objectives; _ }, _ =
-      do_case_split_aux env ~for_model:true in
-    CC_X.extract_concrete_model
-      ~prop_model:assumed_set
-      ~optimized_splits:objectives
-      gamma_finite
+    if Options.get_interpretation () then
```
bclement-ocp: I don't think this is the right place to have that test. Calling `compute_concrete_model` should always compute a model, but we shouldn't call `compute_concrete_model` if `Options.get_interpretation ()` is false (for similar reasons as #833 ).

src/lib/reasoners/theory.ml
```diff
@@ -830,14 +722,14 @@ module Main_Default : S = struct
 
       let gamma, _ = CC_X.assume_literals t.gamma [] facts in
       (* update to optimize with the new gamma *)
-      let objectives, reset_cs_env =
-        update_objectives t.objectives assumed gamma in
+      (* let objectives, reset_cs_env =
+         update_objectives t.objectives assumed gamma in *)
```
bclement-ocp: Why is this no longer needed? What happens if an assertion renders an existing objective invalid?

src/lib/reasoners/theory.ml
```diff
@@ -552,24 +493,31 @@ module Main_Default : S = struct
...
+          | [] ->
+            generate_choices env sem_facts acc_choices
         end
       | _ ->
-        propagate_choices env sem_facts acc_choices new_choices
+        propagate_choices env sem_facts acc_choices (objs, order) new_choices
     in
-    aux env sem_facts (List.rev env.choices) new_choices
+    aux env sem_facts (List.rev env.choices)
+      (Objective.Model.functions env.objectives, 0) new_choices
```
bclement-ocp: Is starting at order `0` here really correct? What if some objectives have been optimized already?

src/lib/reasoners/theory.ml
```diff
@@ -474,20 +414,20 @@ module Main_Default : S = struct
         assert false
 
       | Pinfinity | Minfinity | Limit _ ->
-        (* We stop optimizing the split [opt_split] in this case, but
+        (* We stop optimizing the objective function [obj] in this case, but
            we continue to produce a model if the flag [for_model] is up. *)
         if for_model then
-          aux env sem_facts acc_choices []
+          aux env sem_facts acc_choices ([], 0) []
```
bclement-ocp: I believe this is a behavior change — previously we would keep optimizing further objectives while doing model generation, but now we no longer does (this call to `aux` goes immediately to `generate_choices`). Is that intentional?

bclement-ocp: Also as far as I can tell, this means that the case splits in the middle of intervals and the such that we create for infinite/limit cases are never used.

src/lib/reasoners/theory.ml
bclement-ocp: Hm actually it looks like we are optimizing all objectives again at each case split, that looks wasteful (possibly incorrect, but I don't think so since we use equalities for the splits).

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2097,9 +2115,19 @@ let optimizing_split env uf opt_split =
...
-          if to_max then Th_util.Pinfinity else Th_util.Minfinity
+          if to_max then
+            Objective.Value.Pinfinity
+          else
+            Objective.Value.Minfinity
         in
-        Some { opt_split with value }
+        (* As the problem is unbounded, we need to produce a value for the
+           objective function. In this case, we pick a value in the domain
+           of the polynomial [p]. *)
```
bclement-ocp, line 2124: Why do we need to explicitly pick a value for the objective? If my understanding is correct, this is only intended to be used during model generation; are the standard model generation mechanism not enough?

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2113,32 +2141,33 @@ let optimizing_split env uf opt_split =
...
-               [r1 = r2] in the CC(X) environment [gamma_finite]. But
-               we propagate the new value [value] for this objective. Indeed,
-               we want to be able to print it while using the statement
-               [get-objective]. *)
-            None
+               constraints of the problem. In this case, we pick a value
+               in the domain of the polynomial [p]. But we propagate the new
+               value [value] for this objective. Indeed, we want to be able
+               to print it while using the statement [get-objective]. *)
+            middle_value env ~to_max ty p None
```
bclement-ocp: Same remark as above; are standard model generation mechanisms failing somehow?

bclement-ocp (commented):
> Uhm, I thought it was a good idea to hide completely the priority order in the objective model but in fact, users could feed AE with a file like this one:
> [..]
> Another tricky example is if we use both maximize and minimize on the same expression. The second one is ignored. Should we print a warning?

Yeah, I was going to make a remark on that. We shouldn't print a warning, we should print all the objectives requested by the user in the correct order!

(We might want to *also* print a warning when we perform optimization under an unbounded/strict objective, which should only occur during model generation if I understand correctly)

Not sure how that's tricky — wouldn't removing the test for an existing function in `Objective.Model.add` just do The Right Thing™?

tests/models/arith/arith10.optimize.smt2
```diff
@@ -1,8 +1,11 @@
 (set-logic ALL)
+(set-option :produce-models true)
```
Stevendeo, line 1: This should happen before `(set-logic ...)`

tests/models/arith/arith10.optimize.expected
```diff
@@ -2,5 +2,9 @@
 unknown
 (
   (define-fun x () Real 0.0)
-  (define-fun y () Real 10.0)
+  (define-fun y () Real 0.0)
 )
+(objectives
```
Stevendeo, line 7: As y cannot be optimized, I believe we should warn the user when he invokes (get-objectives)

tests/models/arith/arith4.optimize.expected
```diff
@@ -4,3 +4,6 @@ unknown
   (define-fun x () Int 1)
   (define-fun y () Int 1)
 )
+(objectives
```
Stevendeo, line 7: Why is there only a value for `x`?

src/lib/structures/objective.ml
Halbaroth: For the input file:
```
(set-logic ALL)
(declare-const x Int)
(assert (<= x 10))
(minimize x)
(maximize x)
(check-sat)
(get-model)
(get-objectives)
```
I got
```
unknown
(
  (define-fun x () Int 0)
)
(objectives 
  (x -oo)
  (x (interval (- oo) (+ oo)))
)
```
It's consistent but the user cannot understand there are not the same objective in the output of `get-objectives`.

src/lib/structures/objective.ml
Halbaroth: I called it `o` because it's an `objective function`. Maybe `obj`?

src/lib/structures/typed.mli
```diff
@@ -285,6 +285,11 @@ and 'a tdecl =
...
+      flag [to_max] determines if we try to maximize of minimize [obj]. *)
```
Stevendeo: Don't want to quibble over it, but `to_max` is used as `is_max` everywhere else the code

src/lib/structures/objective.ml
bclement-ocp: Ah right, this is handled by the order which will be different.

I think that output is fine, by the way — the user can understand that these are not the same objective because there are two entries, and the goal is that `get-objectives` return one entry per `(minimize)` or `(maximize)` call that was made. And the problem is ill-specified anyways, so. FWIW, Z3 gives a similar output on this problem.

(The intended use case is that there will be a tool that will parse this output and match the list in order with the `minimize` and `maximize` calls)

src/lib/structures/symbols.ml
```diff
@@ -155,10 +154,6 @@ let compare_operators op1 op2 =
         let r = Int.compare i1 i2 in
         if r = 0 then Int.compare j1 j2 else r
       | Int2BV n1, Int2BV n2 -> Int.compare n1 n2
-      | Optimize {order=o1; is_max=b1}, Optimize {order=o2; is_max=b2} ->
```
Stevendeo, line 158: Why is the compare removed?

src/lib/structures/objective.ml
```diff
@@ -0,0 +1,118 @@
...
+    | Unknown
+
+  let pp ppf v =
+    match v with
+    | Minfinity -> Fmt.pf ppf "-oo"
+    | Pinfinity -> Fmt.pf ppf "+oo"
+    | Value w -> Expr.print ppf w
+    | Limit (Above, w) -> Fmt.pf ppf "(+ %a epsilon)" Expr.print w
+    | Limit (Below, w) -> Fmt.pf ppf "(- %a epsilon)" Expr.print w
+    | Unknown -> Fmt.pf ppf "(interval (- oo) (+ oo))"
```
Stevendeo, line 64: Why the space between `-` (`+`) and `oo` ?

src/lib/structures/objective.ml
```diff
@@ -0,0 +1,118 @@
...
+  let pp ppf mdl =
+    if M.is_empty mdl then
+      Fmt.pf ppf "@[<v 2>(objectives @]@,)"
+    else
+      Fmt.pf ppf "@[<v 2>(objectives @,%a@]@,)"
+        (Fmt.iter_bindings ~sep:Fmt.cut M.iter pp_binding) mdl
+
+  let functions mdl =
+    M.bindings mdl
+    |> List.map (fun ({ f; _ }, _) -> f)
```
Stevendeo, line 102: There is a way to fold on the map to do this in one go, but I don't expect the map to be too big

src/lib/reasoners/intervalCalculus.ml
Halbaroth: Before this PR, we use a wrapped case-split as input for this function and we have access to the last computed value for the first-order model through this wrapped case-split. 

An alternative implementation would be to give this value as an argument of `optimizing_objective`. 

Notice that my `middle_value` gives a value in the last interval (for the `to_max = true` case) of the domain of the polynomial `p`.  You told me other solvers try to give a bigger value when we maximize strict bound.

src/lib/reasoners/intervalCalculus.ml
Halbaroth: See the comment below for the `Sim.Core.Max`.

src/lib/reasoners/theory.ml
Halbaroth: Ok, I will bubble the test.

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1375,7 +1381,11 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
   let assume_th_elt env th_elt dep =
     SAT.assume_th_elt env.satml th_elt dep
 
-  let get_model env = env.last_saved_model
+  let optimize env ~to_max obj = SAT.optimize env.satml ~to_max obj
+
+  let get_model env =
+    Option.bind env.last_saved_model @@
+    fun (lazy mdl) -> Some mdl
```
Stevendeo: Isn't that `Option.map Lazy.force env.last_saved_model` ?

src/lib/reasoners/intervals.mli
```diff
@@ -140,6 +140,10 @@ val add_explanation : t -> Explanation.t -> t
 
 val equal : t -> t -> bool
 
+val pick : to_max:bool -> t -> Numbers.Q.t option
+(** [pick ~to_max t] returns an elements of the set of intervals [t]. If
+    [to_max] is [true], we pick the largest element of [t], if it exists.
+    We look for the smallest element if [to_max] is [false]. *)
```
Stevendeo: In the case `t = ]-oo;+oo[`, we don't pick the largest element of `t`

src/lib/reasoners/intervalCalculus.ml
bclement-ocp: Ah, right this is for that! Makes sense. What I was thinking for that case (and what other solvers do I believe) is add an inequality constraint rather than an equality in case that specific value end up not being feasible. For instance, with a nonlinear objective `x * x` with `x in `1 .. 2` range, the range for `x * x` is `1 .. 4` (AFAIK Alt-Ergo know this) but if we pick `2` or `3` as a value there is no model whereas if we just constrain `< 3` or `> 2` it is.

(I think in this specific case we might go into an infinite loop in the current PR if we actually used this split in `look_for_sat`, not sure)

src/lib/reasoners/intervals.ml
```diff
@@ -1163,11 +1163,41 @@ let match_interval_lower {Sy.sort; is_open; kind; is_lower} i imatch =
...
+    (* By normalization, an integer interval of the form (q, p| has to
+       contain at least one integer and thus [q+1] is an element of this
+       interval. *)
+    Some Q.(q + ~$1)
+
+  | (Minfty, (Strict (q, _) | Large (q, _))) :: _ -> Some Q.(q - ~$1)
+  | ((Strict (q, _) | Large (q, _)), Pinfty) :: _ -> Some Q.(q + ~$1)
+  | ((Strict (q1, _) | Large (q1, _)), (Strict (q2, _) | Large (q2, _))) :: _ ->
+    begin
+      assert (not is_int);
```
Stevendeo, line 1191: This assert is guaranteed by normalization?

src/lib/structures/expr.ml
```diff
@@ -2682,9 +2683,6 @@ module Purification = struct
           | _ -> failwith "unexpected expression in purify_form"
         end
 
-      | Sy.Op Sy.Optimize _ ->
-        purify_literal e
```
Stevendeo, line 2674: Is this purification performed somewhere else?

src/lib/reasoners/fun_sat.ml
```diff
@@ -1849,7 +1851,9 @@ module Make (Th : Theory.S) = struct
     {env with tbox = Th.assume_th_elt env.tbox th_elt dep}
 
   (** returns the latest model stored in the env if any *)
-  let get_model env = !(env.last_saved_model)
+  let get_model env =
+    Option.bind !(env.last_saved_model) @@
```
Stevendeo: `Option.map` ?

src/lib/reasoners/intervalCalculus.ml
Halbaroth: Uhm, I don't think we need to keep the representative element here. The expression `e` here isn't a part of our problem but the expression we try to optimize modulo our constraints. Now, I remember why I add this line before 
```ocaml 
let repr, _ = Uf.find uf e in
```
 
The reason is because the function `Uf.find` fails if the expression have never been added to the union-find before. As the expression can be something completely new, this can happen. Consider for instance the input file:
```
(set-logic ALL)
(declare-const x Int)
(assert (<= x 10))
(maximize (+ x 5))
(check-sat)
(get-model)
(get-objectives)
```
The expression `(+ x 5)` is completely new and the union-find will raise `Not_found` exception here.

Now if the expression is already knew, we want to get both its representative in the union-find (in the code it's `repr`) and the semantic values returned by `X.make` for this expression (in the code it's `r1`).

Another solution would be add the expression `e` in the union-find when we call the function `optimize` of the SAT solver.

src/lib/reasoners/intervals.ml
Halbaroth: No, it's guaranteed by the pattern matching. All the cases have been matched by a branch before this one if the domain is an integer interval.

src/lib/reasoners/intervals.mli
Halbaroth: As I said, `if it exists` :) Otherwise, we pick an element in the last interval of the set of intervals. In the case `]-oo, +oo[`, we pick `0`.

src/lib/reasoners/intervalCalculus.ml
bclement-ocp: If we want to get the representative if there is one or the result of `X.make` otherwise then the right code is probably along the lines of:

```ocaml
let repr, r1 =
  match Uf.find uf e with
  | repr, _ -> repr, Uf.make uf e
  | exception Not_found -> let r, _ = X.make e in r, r
```

That said, I didn't think about it, but you are right — we should add `e` in the call to `optimize` (this is what was done before when `Optimize` was a symbol). Note that this needs to use `Theory.add_term` to make sure that all sub-terms are properly added (`Uf.add` only adds the term itself but not the subterms).

The real reason to do this is that if we do `(maximize (bv2nat x))` for instance, and the problem does not otherwise contain `(bv2nat x)`, not adding `bv2nat` to the theory would lose some reasoning power (because it would not be seen by the relations, and knowledge about it would not be properly propagated).

src/lib/structures/objective.ml
Halbaroth: No good reason. We can consider `oo` as a constant and apply an unary `+` or an unary `-` on it. Or we consider two constants `+oo`, `-oo`. My code is not consistent with that. This is the output of `arith3.optimize.smt2` by `z3`:
```
sat
(
  (define-fun x () Int
    0)
  (define-fun z () Int
    0)
  (define-fun y () Int
    0)
)
(objectives
 (x oo)
 (y  (interval (* (- 1) oo) oo))
 (z  (interval (* (- 1) oo) oo))
)
```
So we can follow the same principal and consider `oo` as a constant and apply usual arithmetical operator on it.

Notice that mathematically, it's a heresy. For instance, `(- oo oo)` doesn't make any sense.

bclement-ocp: FWIW, I believe that `(* (- 1) oo)` is just the way Z3 displays $-\infty$ but that it does not perform other operations on it.

src/lib/reasoners/fun_sat.ml
Halbaroth: I think it cannot happen.

src/lib/reasoners/intervalCalculus.ml
Halbaroth: So I suggest to keep the code as it's now. We could make better change later.

tests/models/arith/arith4.optimize.expected
Halbaroth: Because we don't try to maximize `y` :D

src/lib/structures/typed.mli
Halbaroth: Actually I used both. I will put `is_max` everywhere.

src/lib/structures/objective.ml
Halbaroth: Sure, I said we have to be consistent and we may use the Z3 convention.

src/lib/structures/objective.ml
Halbaroth: Actually we have to preserve the order of the objectives. If I traverse my map with a fold or a iter, I will get a list in reversed order and I have to iterate on it a second times to get the correct order.

If we use the reversed order in the map, both `add` and `fold` will need to reverse the order :/

src/lib/util/numbers.mli
Halbaroth: I removed this code and remove the `module Q = numbers.Q` in `intervals`. We'll remove completely the module numbers later.

src/lib/reasoners/fun_sat.ml
bclement-ocp: Yeah, looks like this returns `None` only when `get_interpretation ()` is disabled (cf other comment)

src/lib/util/numbers.mli
bclement-ocp: Note, this may be dangerous — I am not sure that some functions such as `Q.floor` or `Q.ceiling` have the same semantics in `Q` and `Numbers.Q`; I'd prefer we did this in a seprate PR.

src/lib/reasoners/intervalCalculus.ml
bclement-ocp: Sure I am fine with keeping that code as is for now, but the code in `look_for_sat` needs to be fixed to actually use the split during model generation!

src/lib/util/numbers.mli
Halbaroth: ... no

src/lib/reasoners/theory.ml
Halbaroth: I bubble the test in `may_update_last_saved_model`. The current condition to trigger the computation of models is quite complicated. I'm not sure keeping the possibility of generate models at different stages of the SAT solver is very useful.

src/lib/reasoners/theory.ml
Halbaroth: The purpose of `update_objectives` function was to ensure we add new optimization discovered while exploring a term. At this time, we could have optimization everywhere in formulae. In particular, it means the theory module could ignore an optimization constraint before the underlying formula was involved in a formula of the current trail of the SAT solver. I didn't check, but I guess we could even forget an optimization constraint by backtracking.

If we discover a new constraint, we have to throw away all the case splits. A flag in `update_objectives` was used to remember that and we call `reset_case_split_env` if the flag is up.

Now, we feed the SAT solver with all the optimization constraints at once.  I thought it was fine but in fact we perform case splits after each assume. Thus we have to reset casesplits after each `optimize` in `Theory`.

Halbaroth: Actually I did! I just forget to remove the dead code commented. :) Thanks for the remark. I will add a comment in `optimize`.

bclement-ocp: Ah, right, `update_objectives` is for adding new objectives — I mixed it up in my head with `partial_objectives_reset` which is used when an objective becomes invalid, thanks for the clarification!

> I thought it was fine but in fact we perform case splits after each assume.

Yeah, either after each assume or before matching or after matching depending on the split policy. After each assume is the most aggressive version tho, so I think if we deal with that properly we are good.

> I didn't check, but I guess we could even forget an optimization constraint by backtracking.
> Now, we feed the SAT solver with all the optimization constraints at once. I thought it was fine but in fact we perform case splits after each assume. Thus we have to reset casesplits after each optimize in Theory.

This remark makes me wonder: previously, we had optimizations in assertions, which are obviously stored in the assertion stack (and so would be normally affected by `push` and `pop`), but I am not sure that is still the case in this PR (I didn't see any mechanism that would take care of it, but also was not aware it could be an issue prior to your comment). Might be worth checking -- for instance the following should output `0`, not `10`.

```smt-lib
(set-logic ALL)
(declare-const x Int)
(assert (<= 0 x 10))
(push 1)
(maximize x)
(pop 1)
(minimize x)
(check-sat)
(get-model)
```

Halbaroth: Uhm I thought it was fine because I use the same semantic as the assert but I probably miss something in `solving_loop`. Thanks for the test.

bclement-ocp: Ah dang, I take it that means it does not work then :/ I think it is probably an issue in the `Satml_frontend` / `Satml` rather than `solving_loop`. The way `push` and `pop` work in the Satml solver is that we create guards, and assertions within a push/pop assertions are, well, guarded: if the guard is false, the assertion is just ignored.

So for instance the following:

```smt-lib
(push 1)
(assert f1)
(check-sat)
```

is treated as if we had:

```smt-lib
(declare-const guard Bool)
(assert (=> guard f1))
(assert guard) ; Generated for all assertions levels that were not popped on (check-sat)
(check-sat)
```

and the following:

```smt-lib
(push 1)
(assert f1)
(pop 1)
(check-sat)
```

is treated as if we had:

```smt-lib
(declare-const guard Bool)
(assert (=> guard f1))
(assert (not guard)) ; Generated on (pop 1)
(check-sat)
```

and so `Th.assume` (which is called during propagation) will never be called on `f1`. But I believe that with the current implementation (from my recollection of it, I did not check in detail) we will always call `optimize` without taking the guards into account.

In the previous implementation this "just worked" because optimization markers were expressions inside assertions that were not `assume`d when the guards are false, but now that optimization markers are declarations there is a little bit more work that needs to be duplicated. In fact there is probably a bit of overlap with #901, now that I think about it.

Given that optimization is still experimental (and that you've said you're sick of working on it for so long, which is understandable :) ), I think dropping support for optimization within push/pop in this PR is acceptable. We can then either deal with that as part of #901 (which I think is on track for inclusion in the next release, since postponing #904 freed up quite some time for me) or otherwise in a separate PR; your call.

Halbaroth: I think I'll take it after we solve #901. I'll print a warning if we try to optimize with `push` and `pop`. An error isn't necessary here as we still got a correct model but we don't respect the optimization constraint.

Halbaroth: I open an issue #933 and add a warning when we process files involving both optimization constraints and `pop` and `push`.

src/lib/reasoners/theory.ml
Halbaroth: Let's clarify a bit the previous behaviour and the current one.
- Before this PR, the function `look_for_sat` proceeds choices in this order when `for_model = false`:
  + First, it propagates new choices in the pending queue.
  + Then it try to optimize as much as it can objectives for which we didn't get a best value yet (in the current SAT branch of course). If it see an objective whose the previous value is a limit (+oo, -oo or a strict bound), it tries to optimize it again.
  + If we fail to optimize a bound, we stop in `optimizing_objective`.
  + Otherwise, it asks theories to generate new case splits and it goes back to the first point.

Now, if we're generating model (so `for_model = true`), the previous code ignore objectives whose the value is a limit and we keep optimize next objectives (with a lower priority).

This is not what we want. If we got a limit for some objective `o`, it doesn't make sense to try to optimize objectives with lower priority than `o`. But in model generation, we want to keep generating model, even if we can fulfill some objectives.

Let's pick an example:
``` smt
(set-logic ALL)
(declare-const x Int)
(declare-const y Int)
(maximize x)
(minimize y)
(assert (>= y (- 10)))
(assert (>= x 0))
(check-sat)
(get-model)
(assert (<= x 10))
(check-sat)
(get-model)
```
If you run AE `next` on it with the command
```sh
dune exec -- alt-ergo test.smt2 --optimize --objctives-in-interpretation --frontend legacy --dump-models
```
AE outputs:
```
unknown
Returned unknown reason = Incomplete
(
  (define-fun x () Int +oo)
  (define-fun y () Int (- 10))
)

unknown
Returned unknown reason = Incomplete
(
  (define-fun x () Int 10)
  (define-fun y () Int (- 10))
)
```
We don't expect that AE keep optimizing objectives after `x` before the first `check-sat`, but we keep doing it... 

Now, if we don't the same on this PR with the command
```sh
dune exec -- alt-ergo test2.smt2 --optimize --objectives-in-interpretation
```
AE outputs:
```
unknown
(
  (define-fun x () Int 1)
  (define-fun y () Int 0)
)

; File "test2.smt2", line 12, characters 1-12: I don't know (0.0009) (6 steps) (goal g_2)

unknown
(
  (define-fun x () Int 10)
  (define-fun y () Int (- 10))
)
```

The new code does the right thing... by chance! It's because during each case split phase, we do again all the optimizations, even if we got an infinity value before and we stop optimization if we got an infinity value in `optimizing_objective`. If the flag `for_model` is `true`, we stop also optimization but we keep generate new casesplits. 

I add a new function `next_unknown` in `Objective.Model` whose the semantic is closed to the old `next_optimization`:
```ocaml
  exception Found of Function.t * int

  let next_unknown mdl =
    try
      M.iter (fun { f; order } v ->
        match (v : Value.t) with
        | Value _ -> ()
        | Pinfinity | Minfinity | Limit _ -> raise Exit
        | _ -> raise (Found (f, order))
      ) mdl;
      None
    with
    | Found (f, order) -> Some (f, order)
    | Exit -> None
```

Notice this function isn't very efficient because we keep visiting `Value` node in the map again and again. The behaviour of `Objective.Model` is basically a queue with some printers. I guess I will rewrite it with a functional queue but let's keep map now, early optimization is risky.

Halbaroth: Ok, it's still incorrect for this test:
```smt
; This test checks if the optimization doesn't stop after trying to
; optimize a strict bound.
(set-logic ALL)
(declare-const x Real)
(declare-const y Real)
; If the SAT solver explores first the right branch of the following OR gate, it
; shouldn't conclude that x cannot be optimized.
(assert (or (<= x 4.0) (< x 1.5)))
(maximize x)
; The current implementation of SatML explores the right branch of OR gates
; first. We add the following line to be sure the optimization feature still
; works after changing the order of branches.
(assert (or (< y 1.5) (<= y 4.0)))
(maximize y)
(check-sat)
(get-model)
(get-objectives)
``` 
During the first assert we don't know that `y` will be bound and we get `infinity` here. Thus, we have to treat infinity value again after the second assert, otherwise we will print `+oo` for `y`. This was explained in a comment of the original OptimAE PR.

Gbury: > We don't expect that AE keeps optimize value after `x` after the first `check-sat`, but we keep doing it...

I disagree: in that case I fully expect alt-ergo to re-optimize the model because of the new `get-sat` and the new constraints added by `assert` since the last `check-sat` !

Halbaroth: Ok my example is ambiguous. I was talking of the first `check-sat` only. I totally agree we have to keep optimize if new constraints are involved.

src/lib/reasoners/theory.ml
Halbaroth: Actually, I'm not even sure we need to invoke `reset_case_split` here. If it's the first time we register the objective `f`,  we may have propagate case split for `f` because `f` was involved in some assertions. But the objective value of `f` is `Unknown` and we will optimize it anyway.
If we have optimize some subexpressions of `f` here, it means their objectives have a higher priority and there is no reason to throw away these choices.

bclement-ocp: Don't regular (non-optimizing) splits still need to be reset?

For instance if we assert ` 0 ≤ x ≤ 10`, there is currently no optimization under way so we pick `x = 0` as a regular split, then we do `(maximize x)`, we need to undo the `x = 0` choice.

src/lib/structures/expr.ml
Halbaroth: oups I forget this one. I'll check it now :)

src/lib/reasoners/satml_frontend.ml
```diff
@@ -60,12 +62,12 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
     mutable skolems : E.gformula ME.t; (* key <-> f *)
     add_inst : E.t -> bool;
     guards : guards;
+    mutable model_gen_phase : bool;
```
bclement-ocp: This has been removed in #982

src/lib/reasoners/satml_frontend.ml
```diff
@@ -985,11 +988,14 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
       if not (Options.get_first_interpretation ()) then compute
       else env.last_saved_model == None
     in
-    if compute then begin
+    if compute && Options.get_interpretation () then begin
```
bclement-ocp: I believe `compute` being true here already implies `Options.get_interpretation ()` and that this new test is not necessary.

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1337,7 +1345,10 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
   let assume_th_elt env th_elt dep =
     SAT.assume_th_elt env.satml th_elt dep
 
-  let get_model env = env.last_saved_model
+  let optimize env ~is_max obj = SAT.optimize env.satml ~is_max obj
+
+  let get_model env =
+    Option.map (fun (lazy mdl) -> mdl) env.last_saved_model
```
```suggestion
    Option.map Lazy.force env.last_saved_model
```

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2097,48 +2115,59 @@ let optimizing_split env uf opt_split =
...
+          else
+            Objective.Value.Minfinity
         in
-        Some { opt_split with value }
+        (* As the problem is unbounded, we need to produce a value for the
+           objective function. In this case, we pick a value in the domain
+           of the polynomial [p]. *)
+        let case_split =
+          LR.mkv_eq r1 (middle_value env ~is_max ty p None), true, Th_util.CS
+            (Th_util.Th_arith, Q.one)
```
bclement-ocp, line 2127: As far as I can tell, this value is never used ; the `case_split` field is only read when the value is `Value` in `look_for_sat`, never for `Pinfinity` or `Minfinity`. Let's not add dead code.

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2097,48 +2115,59 @@ let optimizing_split env uf opt_split =
...
-               [r1 = r2] in the CC(X) environment [gamma_finite]. But
-               we propagate the new value [value] for this objective. Indeed,
-               we want to be able to print it while using the statement
-               [get-objective]. *)
-            None
+               constraints of the problem. In this case, we pick a value
+               in the domain of the polynomial [p]. But we propagate the new
+               value [value] for this objective. Indeed, we want to be able
+               to print it while using the statement [get-objective]. *)
+            middle_value env ~is_max ty p None
```
bclement-ocp: As far as I can tell, this value is never used; the `case_split` field is only read when the value is `Value` in `look_for_sat`, never for `Limit`. Let's not add dead code.

tests/models/arith/arith7.optimize.expected
```diff
@@ -3,8 +3,12 @@ unknown
...
 )
-
+(objectives 
+  (x (- (/ 1 2) epsilon)))
 unknown
 (
   (define-fun x () Real (/ (- 3) 2))
 )
+(objectives 
+  (x (- (- 0.0 (/ 1 2)) epsilon))
```
bclement-ocp: This should be displayed as `(- (/ 1 2) epsilon)` following #981 — please rebase the PR.

src/bin/common/solving_loop.ml
```diff
@@ -677,35 +682,58 @@ let main () =
...
+        | Some Model ((module SAT), partial_model) ->
+          let objectives = SAT.get_objectives partial_model in
+          begin
+            match objectives with
+            | Some o ->
+              if not @@ Objective.Model.has_no_limit o then
+                warning "Some objectives cannot be fulfilled";
+              if State.get suspicious_objectives st then
+                warning "Optimization constraints in presence of push \
+                         and pop statements are not correctly processed.";
```
bclement-ocp: Nit: it'd be better to flat out reject this, but since it will be fixed ( #993 ) this is OK as well.

src/bin/common/solving_loop.ml
```diff
@@ -350,6 +350,10 @@ let main () =
     State.create_key ~pipe:"" "named_terms"
   in
 
+  let suspicious_objectives : bool State.key =
```
bclement-ocp: This key is not actually related to optimization since what it does is determine whether the problem has `push` or `pop` commands. Its name should reflect that — for instance, `has_push_pop` or `has_incremental_commands`.

src/bin/common/solving_loop.ml
```diff
@@ -955,10 +983,16 @@ let main () =
...
+      | {contents = `Other (custom, args); loc; _} ->
+        handle_custom_statement loc custom args st
 
-      | _ ->
+      | td ->
+        let st =
+          match td.contents with
+          | `Pop _ | `Push _ ->
+            State.set suspicious_objectives true st
+          | _ -> st
```
bclement-ocp: Nit: This is too strict as it prevents any combined use of `(push)` / `(pop)` and optimization, whereas as long as optimization primitives are at the toplevel it should be fine; we could have a cleaner implementation by using an integer key that tracks the current depth and check that we are at level 0 in `handle_optimize_stmt`.

src/lib/reasoners/fun_sat.ml
```diff
@@ -1119,12 +1119,12 @@ module Make (Th : Theory.S) = struct
       if not (Options.get_first_interpretation ()) then compute
       else !(env.last_saved_model) == None
     in
-    if not compute then env
+    if not compute || not @@ Options.get_interpretation () then env
```
bclement-ocp: I believe that `compute` can only be true if `Options.get_interpretation ()` is true, and so this check is meaningless.

src/lib/reasoners/fun_sat.ml
```diff
@@ -1849,7 +1849,8 @@ module Make (Th : Theory.S) = struct
     {env with tbox = Th.assume_th_elt env.tbox th_elt dep}
 
   (** returns the latest model stored in the env if any *)
-  let get_model env = !(env.last_saved_model)
+  let get_model env =
+    Option.map (fun (lazy mdl) -> mdl) !(env.last_saved_model)
```
bclement-ocp:
```suggestion
    Option.map Lazy.force !(env.last_saved_model)
```

src/lib/structures/objective.ml
bclement-ocp: Given that the current code already iterates on the map twice, folding to build a reversed list then calling `List.rev` is actually going to be slightly faster still, but it is probably not significant.

src/lib/structures/objective.ml
```diff
@@ -0,0 +1,134 @@
...
+        ) mdl;
+      None
+    with
+    | Found (f, order) -> Some (f, order)
+    | Exit -> None
+
+  let reset_until mdl o =
+    M.fold
+      (fun { f; order } v acc ->
+         if order < o then M.add { f; order } v acc
```
bclement-ocp: Nit: this could just take `mdl` as the initial value for `acc` and not do anything in this case.

src/lib/structures/objective.ml
```diff
@@ -0,0 +1,134 @@
...
+      ) mdl
+
+  exception Found of Function.t * int
+
+  let next_unknown ~for_model mdl =
+    try
+      M.iter (fun { f; order } v ->
+          match (v : Value.t) with
+          | Value _ -> ()
+          | Limit _ | Pinfinity | Minfinity when for_model -> raise Exit
```
bclement-ocp: This is a behavior change — the previous code was returning `()` in this case, which allows for further optimization during model generation. I believe the old behavior is better — if we optimize both `x` and `y` and `x` turns out to be unbounded, the user probably would still be happy to have a large value for `y` if possible.

(Recalling that the main use case for optimization is not necessarily to find the exact lexicographic optimum but rather have generally higher-quality models)

bclement-ocp (changes requested):
I think the behavior is mostly correct now — still some minor changes to do, notably I'd like to understand where the case splits for infinities/"limits" are used (or remove them if they are indeed unused).

src/lib/structures/objective.ml
Halbaroth: For me it's a very strange behavior but since it's the behavior of some SMT solvers, let's keep the old code.

src/lib/reasoners/intervalCalculus.ml
Halbaroth: Now, I propagate it :)

src/bin/common/solving_loop.ml
```diff
@@ -350,6 +350,10 @@ let main () =
     State.create_key ~pipe:"" "named_terms"
   in
 
+  let has_incremental : int State.key =
```
bclement-ocp: If it's an `int`, it should not be called `has_incremental`. I guess two of my comments in the previous round somewhat contradicted each other. Maybe `incremental_depth`?

src/bin/common/solving_loop.ml
```diff
@@ -677,35 +682,60 @@ let main () =
...
-    st
+  let handle_optimize_stmt ~is_max loc id (term : DStd.Expr.Term.t) st =
+    let contents = `Optimize (term, is_max) in
+    let stmt = { Typer_Pipe.id; contents; loc; attrs = []; implicit = false } in
+    let cnf =
+      D_cnf.make (State.get State.logic_file st).loc
+        (State.get solver_ctx_key st).ctx stmt
+    in
+    (* Using both optimization and incremental mode may be wrong if
+       some optimization constraints aren't at the toplevel. *)
```
bclement-ocp: Should have a link back to #993

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2045,17 +2045,35 @@ let case_split env uf ~for_model =
...
+
+let optimizing_objective env uf Objective.Function.{ e; is_max; _ } =
   (* soundness: if there are expressions to optmize, this should be
      done without waiting for ~for_model flag to be true *)
-  let {Th_util.r = r; is_max = to_max; e; value; _ } = opt_split in
-  assert (match value with
-      | Unknown -> true
-      | _ -> false
-    );
+  (*   let uf, _ = Uf.add uf e in *)
```
bclement-ocp:
```suggestion
```

src/lib/reasoners/theory.ml
```diff
@@ -474,20 +411,22 @@ module Main_Default : S = struct
...
       | Pinfinity | Minfinity | Limit _ ->
-        (* We stop optimizing the split [opt_split] in this case, but
+        (* We stop optimizing the objective function [obj] in this case, but
            we continue to produce a model if the flag [for_model] is up. *)
         if for_model then
-          aux env sem_facts acc_choices []
+          let new_choice =
+            add_explanations_to_split ~order opt_split.case_split
+          in
+          aux env sem_facts acc_choices [new_choice]
```
bclement-ocp:
```suggestion
          propagate_choices env sem_facts acc_choices [new_choice]
```

src/bin/common/solving_loop.ml
```diff
@@ -955,10 +985,18 @@ let main () =
...
+        handle_custom_statement loc custom args st
 
-      | _ ->
+      | td ->
+        let st =
+          match td.contents with
+          | `Pop _ ->
+            State.set has_incremental (State.get has_incremental st - 1) st
+          | `Push _ ->
+            State.set has_incremental (State.get has_incremental st + 1) st
```
bclement-ocp: This is not +1 and -1; Push and Pop have an integer argument which indicates how many levels are pushed/popped.

```suggestion
          | `Pop n ->
            State.set has_incremental (State.get has_incremental st - n) st
          | `Push n ->
            State.set has_incremental (State.get has_incremental st + n) st
```

bclement-ocp (approved):
I think we are good with a couple minor omissions noted inline; please fix these before merging!

PR #923, opened 2023-11-02
Update the lock file

Halbaroth:
It seems the lock file haven't been updated correctly in the PR #893. We have to lock `dolmen` to the dev version.


bclement-ocp (approved):
Maybe we can have a CI action that checks this? Not sure how easy that would be with opam (I mean check that the deps are the same in the lockfile as in the pinned version in `.opam` if they are both present...)

Halbaroth:
Actually, it should fail with the CI. I notice it while using `opam 2.2`. Maybe something change between `opam 2.1` and `opam 2.2`. I will check :)

Issue #917, opened 2023-10-24
Enable prelude by default


bclement-ocp:
Or solve the perf issues.

Halbaroth:
The performance issue is as follow:
The branch v2.5.x is slightly slower on problems of the `ae-format` dataset when we enable the theory `fpa`.  More precisely for the `unsat` output:
| Prover | Alt-Ergo v2.5.x with `--enable-theories fpa` | Alt-Ergo v2.5.x |
|----------|----------------------------------------------------------|---------------------|
| number  | 51561 | 51548 |
| mean time | 0.347s | 0.335s |
| deviation time | 0.792s | 0.801s |
| total time | 4h29m14s | 4h16m55s |

We also observe that Alt-Ergo proved +14 tests with `--enable-theories fpa` and +4 tests without this flag.

bclement-ocp:
The total time above is for all problems, not only `unsat` problems I believe?

Halbaroth:
~~It's a bug in my PR for benchpress. I'll fix it right now :)~~
It's a bit misleading but the current `total_time` outputs by benchpress is the total time for `unsat + sat` problems only. In our case, total time for `unsat` and `total_time` agree as we never output `sat`.

bclement-ocp:
EDIT: I made a dumb typo. The average don't look right, but the slowdown is always around 4%.

Ah, I seemed to recall different numbers. But there is still a discrepancy:

 - `4h29m14s` is `16154s` and `16154/51561 = 0.313` (not `0.303` as I initially wrote).
 - `4h16m55s` is `15415s` and `15415/51548 = 0.300`.

The total time figure shows around 4% slowdown (`0.313/0.300 ≈ 1.043`, similar calculation on the rather meaningless `15614/15415 ≈ 1.048`). The mean time figure shows around 3.6% slowdown (`0.347/0.335 ≈ 1.036`, similar calculation on the rather meaningless `0.347*51561/0.335*51548 ≈ 1.036`)~, which is significantly larger~.

Which one should we trust? Although in both cases the slowdowns are on the very high side of what I'd consider acceptable.

(Although in both cases we are kind of comparing apples to oranges — we should probably either compare the median rather than the mean, or the mean over the problems solved by both solvers, because we don't want to advantage a solver that solves very few problems quickly. Here we have so few problems solved by a single one that it probably doesn't matter much, however)

Issue #412, opened 2021-01-12
Invalid_argument("List.fold_left2") when trying to get a model


iguerNL:
command:
```
./alt-ergo tables-Map-VC_add_max_binding.ae --interpretation last --timelimit 1 --output smtlib2 --timelimit-interpretation 1
```

[bug.ae.zip](https://github.com/OCamlPro/alt-ergo/files/5803877/bug.ae.zip)

Output:

```
 ; Functions
  (define-fun key ((arg_0 (t1, '_c0) tuple2)) t1 k0)
  (define-fun key ((arg_0 (t1, '_c892) tuple2)) t1 !k54)
  (define-fun Tuple2_proj_1 ((arg_0 (t1, '_c0) tuple2)) t1 k0)
  (define-fun Tuple2_proj_1 ((arg_0 (t1, '_c892) tuple2)) t1 !k54)
  (define-fun field ((arg_0 (t3 '_c0 ))) (t2 '_c0 ) !k116)
  (define-fun singleton ((arg_0 (t1, '_c0) tuple2)) (seq (t1, '_c0) tuple2 ) (ite (= arg_0 {Tuple2_proj_1 = k0; Tuple2_proj_2 = v1}) !k134 !k131))
  (define-fun seq_model ((arg_0 (tree '_c0 ))) (seq (t1, '_c0) tuple2 ) !k43)
  (define-fun seq1 ((arg_0 '_c0  m)) (seq (t1, '_c0) tuple2 ) !k43)
  (define-fun repr ((arg_0 (t2 '_c0 ))) (tree '_c0 ) !k122)
  (define-fun Tuple2_proj_2 ((arg_0 (t1, '_c0) tuple2)) '_c0 (ite (= arg_0 {Tuple2_proj_1 = k0; Tuple2_proj_2 = v1}) v1 !k34))
  (define-fun fc ((arg_0 (t3 '_c0 ))) (Array t1 '_c0) !k120)
  (define-fun func ((arg_0 '_c0  m2)) (Array t1 (t1, '_c0) tuple2) !k118)
  (define-fun func ((arg_0 '_c892  m2)) (Array t1 (t1, '_c892) tuple2) !k196)
  (define-fun domn ((arg_0 '_c0  m2)) (Array t1 Bool) !k117)
  (define-fun domn ((arg_0 '_c876  m2)) (Array t1 Bool) !k191)
[...]

PR #925, opened 2023-11-07
Refactoring models representation

Halbaroth:
This PR changes the way Alt-Ergo stores models. 

Before the PR, Alt-Ergo didn't produce a printable model in memory but performed some basic computations in the model printers. 

Now, Alt-Ergo will produce in memory a printable model. The PR mostly consists in moving some codes from `Models.ml` to `ModelMap.ml` or `Uf.ml`.

- All the require computation is now done in the `output_concrete_model` of the union-find module (included the computation for records).
- All the printings stuff are in `ModelMap.ml`.
- We use `Expr.t` to store model values.
- The only reason I kept `Models.ml` is because we are the printer of `constraint model` here. This will be remove later.

Besides, as I notice, the symbols we define in models are all `Sy.Name`. Thus I create an new module to manage identifiers `Id.ml` and I store identifiers instead of symbols in models.


src/lib/reasoners/uf.ml
```diff
@@ -1020,7 +1026,9 @@ let model_repr_of_term t env mrepr unbounded =
     let mk = try ME.find t env.make with Not_found -> assert false in
     let rep,_ = try MapX.find mk env.repr with Not_found -> assert false in
     match unbounded with
-    | Some string_repr -> (rep, string_repr), mrepr
+    | Some string_repr ->
+      assert (is_a_good_model_value rep && is_const_term rep);
```
Stevendeo: Can you add a quick comment explaining the reason for this assertion?

src/lib/reasoners/uf.ml
```diff
@@ -1030,69 +1038,157 @@ let model_repr_of_term t env mrepr unbounded =
         try List.rev_map (fun s -> s, ME.find s env.make) cls
         with Not_found -> assert false
       in
-      let e = X.choose_adequate_model t rep cls in
-      e, ME.add t e mrepr
+      let rep, string_repr = X.choose_adequate_model t rep cls in
+      assert (is_a_good_model_value rep && is_const_term rep);
```
Stevendeo: Same here

src/lib/reasoners/uf.ml
```diff
@@ -1030,69 +1038,157 @@ let model_repr_of_term t env mrepr unbounded =
...
+  (X.is_solvable_theory_symbol f ty
+   && not (Shostak.Records.is_mine_symb f ty))
+  || is_optimization_op f
+  || Sy.is_internal f
 
-let compute_concrete_model_of_val
-    env t ((fprofs, cprofs, carrays, mrepr) as acc) unbounded =
+let compute_concrete_model_of_val env t ((mdl, mrepr) as acc) unbounded =
   let { E.f; xs; ty; _ } = E.term_view t in
   (* Keep record constructors because models.ml expects them to be there *)
```
Stevendeo: Isn't this comment linked to the `is_forbidden_symbol` ?

src/lib/reasoners/uf.ml
```diff
@@ -1030,69 +1038,157 @@ let model_repr_of_term t env mrepr unbounded =
...
-          ModelMap.add (f_ta,[X.type_info i], ty) ([e], rep) carrays,
-          mrepr
-
-        | _ -> assert false
-      end
+    begin
+      let arg_vals, arg_tys, mrepr =
+        List.fold_left
+          (fun (arg_vals, arg_tys, mrepr) arg ->
+             let rep_arg, mrepr = model_repr_of_term arg env mrepr unbounded in
```
Stevendeo: There was an assertion here, why is it removed?

src/lib/reasoners/uf.ml
```diff
@@ -1030,69 +1038,157 @@ let model_repr_of_term t env mrepr unbounded =
...
+             rep_arg :: arg_vals,
+             (Expr.type_info arg) :: arg_tys,
+             mrepr
+          )
+          ([], [], mrepr) (List.rev xs)
+      in
+      let ret_rep, mrepr = model_repr_of_term t env mrepr unbounded in
```
Stevendeo: Same here

src/lib/reasoners/uf.ml
```diff
@@ -1030,69 +1038,157 @@ let model_repr_of_term t env mrepr unbounded =
...
+      let hs = Hstring.make string_repr in
+      (rep, hs), ME.add t (rep, hs) mrepr
 
 let is_optimization_op = function
   | Sy.Op Sy.Optimize _ -> true
   | _ -> false
 
-let save_cache () =
-  LX.save_cache ()
+module Cache = struct
```
Stevendeo: Can you add a short documentation for this module?

src/lib/reasoners/uf.ml
```diff
@@ -1107,9 +1203,38 @@ module MED = Map.Make
...
-let terms env = ME.fold MED.add env.make MED.empty
+let is_suspicious_name hs =
+  match Hstring.view hs with
+  | "@/" | "@%" | "@*" -> true
+  | _ -> false
+
+(* The model generation is known to be imcomplete for FPA and Bitvector
+   theories. *)
+let is_suspicious_symbol = function
+  | Symbols.Op (Float | Fixed | Abs_int | Abs_real | Sqrt_real
```
Stevendeo: Symbols -> Sy

src/lib/reasoners/uf.ml
```diff
@@ -1107,9 +1203,38 @@ module MED = Map.Make
...
+let terms env =
+  ME.fold
+    (fun t r ((terms, suspicious) as acc) ->
+       let Expr.{ f; _ } = Expr.term_view t in
+       match f with
+       | Name (_, _, true) ->
+         (* We don't store names defined by the user. *)
+         acc
+       | _ ->
+         let suspicious = if is_suspicious_symbol f then true else suspicious in
```
Stevendeo:
```suggestion
         let suspicious = is_suspicious_symbol f || suspicious
```

src/lib/reasoners/uf.ml
```diff
@@ -1122,29 +1247,72 @@ let compute_concrete_model ?(optimized_splits=Util.MI.empty) env =
          | Unknown -> acc
       ) optimized_splits (SetX.empty, SetX.empty, SetX.empty)
   in
+
```
Stevendeo, line 1071:
```suggestion
```

src/lib/reasoners/uf.ml
```diff
@@ -1122,29 +1247,72 @@ let compute_concrete_model ?(optimized_splits=Util.MI.empty) env =
...
-       else
-       if is_bounded || SetX.mem mk bounded then
-         (* either the pb is bounded (or it isn't an optimization pb)
-            or we have an optimum for mk *)
-         compute_concrete_model_of_val env t acc None
-       else
-         acc
-    ) (terms env)
-    (ModelMap.empty, ModelMap.empty, ModelMap.empty, ME.empty)
+
```
Stevendeo:
```suggestion
```

src/lib/reasoners/uf.ml
```diff
@@ -1122,29 +1247,72 @@ let compute_concrete_model ?(optimized_splits=Util.MI.empty) env =
...
+        let arr_val =
+          Hashtbl.fold (fun i v arr_val ->
+              `Store (arr_val, i, v)
+            ) vals (`Abstract (abstract, [], ty))
+        in
+        let id =
+          let Expr.{ f; _ } = Expr.term_view t in
+          match f with
+          | Sy.Name (id, _, _) -> id
+          | _ -> assert false
```
Stevendeo: Quick explanation about this assertion

src/lib/structures/modelMap.ml
```diff
@@ -30,100 +30,179 @@
...
-
-let[@inline always] is_suspicious { suspicious; _ } = suspicious
-
-let fold f { values; _ } acc = P.fold f values acc
-
-let empty = { values = P.empty; suspicious = false }
-
-let is_empty { values; _ } = P.is_empty values
+let add ((_, arg_tys, _) as sig_) arg_vals ret_val { values; suspicious } =
+  assert (List.compare_lengths arg_tys arg_vals = 0);
```
Stevendeo: Add some documentation about this assertion

src/lib/util/util.mli
```diff
@@ -119,3 +119,5 @@ val print_list_pp:
   Format.formatter -> 'a list -> unit
 
 val internal_error : ('a, Format.formatter, unit, 'b) format4 -> 'a
+
+val quoted_string : string -> string
```
Stevendeo:
```suggestion
(** Returns the string in argument as a smtlib2 valid identifier (either itself or quoted). *)
val quoted_string : string -> string
```

src/lib/structures/modelMap.ml
```diff
@@ -30,100 +30,179 @@
...
+
+let pp_seq ppf seq =
+  match seq () with
+  | Seq.Nil -> ()
+  | Cons _ -> Fmt.pf ppf "@,%a" Fmt.(seq ~sep:cut pp_define_fun) seq
+
+let pp ppf {values; suspicious} =
+  if suspicious then begin
+    Fmt.pf ppf "; This model is a best-effort. It includes symbols
+        for which model generation is known to be incomplete. @."
```
Stevendeo:
```suggestion
    Fmt.pf ppf "; This model is a best-effort. It includes symbols \
                for which model generation is known to be incomplete.@."
```

Stevendeo (dismissed):

src/lib/reasoners/uf.ml
Halbaroth: I moved it in `model_repr_of_term` because we need to enforce this invariant twice.

src/lib/reasoners/uf.ml
Halbaroth: Same answer :D

src/lib/structures/modelMap.ml
Halbaroth: I replaced it by an `invalid_argument` as explained offline ;D

tests/models/uf/uf2.models.expected
```diff
@@ -1,7 +1,7 @@
 
 unknown
 (
-  (define-fun f ((arg_0 Int)) Int (ite (= arg_0 0) 2 0))
+  (define-fun f ((arg_0 Int)) Int (ite (= arg_0 (- 2)) 0 2))
```
bclement-ocp, line 4: Why did this model change?

tests/float/test_float2.models.expected
```diff
@@ -1,5 +1,7 @@
 
 unknown
+; This model is a best-effort. It includes symbols
+; for which model generation is known to be incomplete.
```
bclement-ocp: Note that after #886 model generation should only be incomplete when the FPA prelude is disabled — which I don't think is a configuration we want to support.

tests/bitv/testfile-bv2nat-models.dolmen.expected
```diff
@@ -1,5 +1,7 @@
 
 unknown
+; This model is a best-effort. It includes symbols
+; for which model generation is known to be incomplete.
```
bclement-ocp: After #886 model generation for bv2nat and int2bv should be complete.

src/lib/util/util.mli
```diff
@@ -119,3 +119,7 @@ val print_list_pp:
   Format.formatter -> 'a list -> unit
 
 val internal_error : ('a, Format.formatter, unit, 'b) format4 -> 'a
+
+(** Returns the string in argument as a smtlib2 valid identifier
+    (either itself or quoted). *)
+val quoted_string : string -> string
```
bclement-ocp: Note for the future: this should be replaced by the appropriate functions in Dolmen once Gbury/dolmen#198 is merged.

src/lib/structures/modelMap.mli
```diff
@@ -28,33 +28,52 @@
 (*                                                                        *)
 (**************************************************************************)
 
-(** Maps of values for alt-ergo's models.
-    Elements are sorted by symbols/types (P) and accumulated as sets
-    of expressions matching the P.key type (V).
-*)
+type sy = Id.t * Ty.t list * Ty.t
+(** Typed symbol of function. *)
```
bclement-ocp: I am not sure how to link the types in the tuple to the description "typed symbol of function". I believe this is: `(symbol, args_ty, ret_ty)` tuples where `symbol` is the function name, `args_ty` the type of arguments, `ret_ty` the return type?

src/lib/structures/modelMap.mli
```diff
@@ -28,33 +28,52 @@
...
+module Value : sig
+  type simple = [
+    | `Abstract of sy
+    (** An unique abstract value. *)
 
-module V : Set.S with type elt =
-                        (Expr.t * (Shostak.Combine.r * string)) list *
-                        (Shostak.Combine.r * string)
+    | `Constant of Hstring.t
+    (** A string representation of a semantic value. *)
```
bclement-ocp: Why do we use a string representation here? I thought the point of the PR was to build a structured representation of the model prior to printing, but if we just have strings everywhere I don't really understand the difference.

Besides, I am not sure I understand why we need these strings to be interned (i.e. why is this `Hstring.t` rather than simply `string`), can you clarify?

src/lib/structures/modelMap.mli
```diff
@@ -28,33 +28,52 @@
...
-    Elements are sorted by symbols/types (P) and accumulated as sets
-    of expressions matching the P.key type (V).
-*)
+type sy = Id.t * Ty.t list * Ty.t
+(** Typed symbol of function. *)
 
-module P : Map.S with type key =
-                        Symbols.t * Ty.t list * Ty.t
+module Value : sig
+  type simple = [
```
bclement-ocp: Why is this a polymorphic variant rather than a regular sum type?

src/lib/structures/modelMap.mli
```diff
@@ -28,33 +28,52 @@
...
-module V : Set.S with type elt =
-                        (Expr.t * (Shostak.Combine.r * string)) list *
-                        (Shostak.Combine.r * string)
+    | `Constant of Hstring.t
+    (** A string representation of a semantic value. *)
+  ]
 
-type key = P.key
-type elt = V.t
+  type record = Hstring.t * simple list
```
bclement-ocp: Records can contain other records, why are the fields of type `simple`? I guess this is the same question as above wrt using strings for constants inside `simple`.

src/lib/structures/modelMap.mli
```diff
@@ -28,33 +28,52 @@
...
 
-type t
+  type array = [
+    | `Abstract of sy
+    (** An unique abstract array value. *)
+
+    | `Store of array * Hstring.t * Hstring.t
```
bclement-ocp: Same as above; why are `key` and `value` strings rather than structured representations?

src/lib/structures/modelMap.mli
```diff
@@ -28,33 +28,52 @@
...
+
+  type t = [
+    | `Array of array
 
-val add : key -> V.elt -> t -> t
+    | `Record of record
+    (** A string representation of a record definition. *)
 
-val iter : (key -> elt -> unit) -> t -> unit
+    | simple
```
bclement-ocp: This type allows both `` `Abstract sy`` and `` `Array (`Abstract sy)`` and I am not sure I understand the difference between these two.

src/lib/structures/modelMap.mli
```diff
@@ -28,33 +28,52 @@
...
+(** Type of model. *)
 
-val empty : t
+val add : sy -> Value.t list -> Value.t -> t -> t
+(** [add sy args ret mdl] adds the binding [args -> ret] to the partial graph
+    associated with the symbol [sy]. *)
 
-val is_empty : t -> bool
+val empty : suspicious:bool -> t
+(** An empty model. *)
```
bclement-ocp: Should briefly explain the `suspicious` argument here.

src/lib/structures/modelMap.ml
```diff
@@ -30,100 +30,182 @@
...
-            List.iter2
-              (fun ty1 ty2 ->
-                 let d = Ty.compare ty1 ty2 in
-                 if d <> 0 then begin c := d; raise Exit end
-              ) b1 b2;
-            0
-          with
-          | Exit -> assert (!c <> 0); !c
-          | Invalid_argument _ -> assert false
+type sy = Id.t * Ty.t list * Ty.t [@@deriving ord]
```
bclement-ocp, line 34: The order on these values has changed, not sure if that was intentional (also unsure whether the order matters or not here, but since it did not use lexicographic ordering initially it may).

src/lib/structures/modelMap.ml
```diff
@@ -30,100 +30,182 @@
...
+      (struct
+        type t = Value.t list [@@deriving ord]
+      end)
+
+  type t = Value.t M.t
+
+  let empty = M.empty
+  let add = M.add
+
+  module Fiber = struct
```
bclement-ocp, line 53: While the term "fiber" may be common on some areas of mathematics, it is not in the SMT community; if you want to use that term, please include a short comment explaining what it means.

src/lib/structures/modelMap.ml
```diff
@@ -30,100 +30,182 @@
...
+    *)
+    let rec pp_args ctr ppf = function
+      | [] -> ()
+      | [arg] ->
+        Fmt.pf ppf "(= arg_%i %a)" ctr Value.pp arg
+      | arg :: args ->
+        Fmt.pf ppf "(and (= arg_%i %a) %a)"
+          ctr
+          Value.pp arg
+          (pp_args (ctr + 1)) args
```
bclement-ocp: Note that the `and` SMT-LIB constructor is n-ary, it would be more readable to generate `(and (= arg_0 x1) (= arg_1 x2) (= arg_2 x3))` rather than nesting `and`s.

src/lib/structures/modelMap.ml
```diff
@@ -30,100 +30,182 @@
...
+    let pp ppf fiber =
+      let rec aux ppf seq =
+        match seq () with
+        | Seq.Nil -> ()
+        | Cons (args, seq) when Stdcompat.Seq.is_empty seq ->
+          Fmt.pf ppf "%a" (pp_args 0) args
+        | Cons (args, seq) ->
+          Fmt.pf ppf "(or %a %a)"
+            (pp_args 0) args
+            aux seq
```
bclement-ocp: The `or` constructor is also n-ary.

src/lib/structures/modelMap.ml
```diff
@@ -30,100 +30,182 @@
...
+      in
+      aux ppf (to_seq fiber)
```
bclement-ocp: I am not sure the overhead of converting to a sequence is needed here; we could just inspect the cardinal and iterate the set directly.

src/lib/structures/modelMap.ml
```diff
@@ -30,100 +30,182 @@
...
-            List.iter2
-              (fun (_,(x,_)) (_,(y,_)) ->
-                 let d = X.hash_cmp x y in
-                 if d <> 0 then begin c := d; raise Exit end
-              ) l1 l2;
-            !c
-          with
-          | Exit -> !c
-          | Invalid_argument _ -> List.length l1 - List.length l2
+      type t = sy [@@deriving ord]
```
bclement-ocp: It feels icky to use `@@deriving ord` to provide the arguments to a functor, it feels very implicit. Defining `let compare = compare_sy` explicitly would be more explicit.

src/lib/structures/modelMap.ml
```diff
@@ -30,100 +30,182 @@
...
+
+  let graph = try P.find sy values with Not_found -> Graph.empty in
+  let values = P.add sy (Graph.add arg_vals ret_val graph) values in
+  { values; suspicious }
+
+let empty ~suspicious = { values = P.empty; suspicious }
+
+let pp_named_arg_ty ~unused ppf (arg_name, arg_ty) =
+  let pp_unused ppf unused = if unused then Fmt.pf ppf "_" else () in
+  Fmt.pf ppf "(%aarg_%i %a)" pp_unused unused arg_name Ty.pp_smtlib arg_ty
```
bclement-ocp, line 159: If we want to add `_` to the unused arguments, we should either do it properly or not do it. Currently we only add `_` to the arguments if they are *all* unused, which is a middle ground that I don't think makes much sense; I would prefer we did nothing here.

src/lib/structures/modelMap.ml
```diff
@@ -30,100 +30,182 @@
...
+  match seq () with
+  | Seq.Nil -> ()
+  | Cons _ -> Fmt.pf ppf "@,%a" Fmt.(seq ~sep:cut pp_define_fun) seq
+
+let pp ppf {values; suspicious} =
+  if suspicious then begin
+    Fmt.pf ppf "; This model is a best-effort. It includes symbols\n\
+                ; for which model generation is known to be incomplete.@."
+  end;
+  Fmt.pf ppf "@[<v 2>(%a@]@,)" pp_seq (P.to_seq values)
```
bclement-ocp: We can check `P.is_empty` and then `Fmt.iter P.iter` rather than converting to a sequence.

src/lib/frontend/models.mli
```diff
@@ -38,15 +38,11 @@ type objective_value =
 
 type t = {
   propositional : Expr.Set.t;
-  constants : ModelMap.t;
-  functions : ModelMap.t;
-  arrays : ModelMap.t;
+  model : ModelMap.t;
   objectives: (Expr.t * objective_value) Util.MI.t;
-  terms_values : (Shostak.Combine.r * string) Expr.Map.t
+  terms_values : (Shostak.Combine.r * Hstring.t) Expr.Map.t
```
bclement-ocp: Not sure why we need the values to be interned `Hstring.t`? Also, shouldn't we be using `Value.t` here?

src/lib/reasoners/uf.ml
```diff
@@ -999,100 +999,206 @@ let assign_next env =
...
+      abstract
+
+  let clear () =
+    Hashtbl.clear arrays_cache;
+    Hashtbl.clear records_cache;
+    Hashtbl.clear abstracts_cache
+end
+
+let is_forbidden_symbol f ty =
   (* Keep record constructors because models.ml expects them to be there *)
```
bclement-ocp: The corresponding code that expects the record constructors to be there in `models.ml` seems to have been removed, was it intentional?

src/lib/reasoners/uf.ml
```diff
@@ -999,100 +999,206 @@ let assign_next env =
...
 
 let is_optimization_op = function
   | Sy.Op Sy.Optimize _ -> true
   | _ -> false
 
-let save_cache () =
-  LX.save_cache ()
+(* Caches used during the computation of the model.
+   All these caches are flushed before generating a new model. *)
+module Cache = struct
```
bclement-ocp: Is there a reason why the caches are stored in global variables? It would be cleaner (and thread-safe) to have a type that bundles the hash tables and is passed to `compute_concrete_model_of_val`.

src/lib/reasoners/uf.ml
```diff
@@ -999,100 +999,206 @@ let assign_next env =
...
+          | None ->
+            (* We have to add an abstract array in case there is no
+               constraint on its values. *)
+            Hashtbl.add Cache.arrays_cache t (Hashtbl.create 17);
+            acc
+        end
+
+      | Sy.Op Sy.Set, _, _ ->
+        (* As arrays are immutable, the result of the set operator is
+           never a user-defined array and has to be ignored. *)
```
bclement-ocp: This is not true. The result of the set operator can be equal to a user-defined array, as in #929 (in fact this may or may not be the source of #929)

src/lib/reasoners/uf.ml
```diff
@@ -999,100 +999,206 @@ let assign_next env =
...
+        end
 
-    | _ ->
-      if tys == [] then
-        fprofs, ModelMap.add (f, tys, ty) (xs, rep) cprofs, carrays,
-        mrepr
-      else
-        ModelMap.add (f, tys, ty) (xs, rep) fprofs, cprofs, carrays,
-        mrepr
+      | Sy.Name (id, _, _), [], Ty.Trecord trecord ->
```
bclement-ocp: Are we sure nothing special needs to happen for functions returning records?

src/lib/reasoners/uf.ml
```diff
@@ -1107,9 +1213,38 @@ module MED = Map.Make
...
+
+(* The model generation is known to be imcomplete for FPA and Bitvector
+   theories. *)
+let is_suspicious_symbol = function
+  | Sy.Op (Float | Fixed | Abs_int | Abs_real | Sqrt_real
+          | Sqrt_real_default | Sqrt_real_excess
+          | Real_of_int | Int_floor | Int_ceil
+          | Max_int | Max_real | Min_int | Min_real
+          | Pow | Integer_log2 | Int2BV _ | BV2Nat
+          | BVand | BVor | Integer_round) -> true
```
bclement-ocp: After #886, `BV2Nat`, `Int2BV` and `Pow` should no longer be treated as suspicious. Further, `Float`, `Abs_int`, `Abs_real`, `Sqrt_real_default`, `Sqrt_real_excess`, `Real_of_int`, `Int_floor`, `Int_ceil`, `Max_int`, `Max_real`, `Min_int`, `Min_real`, `Integer_log2` and `Integer_round` are now only suspicious if the FPA prelude is not loaded.

This is because we now propagate their arguments during model generation, which should prevent us from generating models with invalid values for these functions.

bclement-ocp (changes requested):
This PR feels like it stops halfway through: from reading the PR description I thought that we would be building a structured representation of the model in memory prior to printing, but we actually build a fairly shallow map where most things are strings. I think we should go all the way and keep the structured representation throughout, rather than build (and intern!) intermediate strings that are only used for printing; although there are maybe blockers to that I don't immediately see — thoughts?

> Besides, as I notice, the symbols we define in models are all Sy.Name. Thus I create an new module to manage identifiers Id.ml and I store identifiers instead of symbols in models.

I am not entirely convinced this is a good idea — I am not opposed to it but we may want to include things that are not `Sy.Name` in the models at some point (namely, values for partially interpreted functions such as division by zero) where we may regret this change.

(Unrelated thought: at some point we should probably use the `Id` module from Dolmen to represent identifiers as well)

src/lib/reasoners/uf.ml
Halbaroth: I will check but as it checks there is no argument and a function without argument is the same as a constant in the SMT-LIB standard, it should be ok.

src/lib/reasoners/uf.ml
Halbaroth: No reason, I will create the cache in `compute_concrete_model` directly.

src/lib/reasoners/uf.ml
bclement-ocp: What I mean is that we are calling `store_record_sig` for constant records, but not for functions that return records, and I am wondering if we should do something similar for functions that return records.

See e.g. #412 where there is an uninterpreted function that returns a record and where Alt-Ergo gets confused. On a side note, I thought I had fixed the first half of the bug (treating all functions with a single record argument as constructors) but apparently I never made that PR, oops :innocent: . Would be worth checking what happens there with this implementation.

src/lib/structures/modelMap.ml
Halbaroth: I didn't know, I will change it :)

Halbaroth:
> This PR feels like it stops halfway through: from reading the PR description I thought that we would be building a structured representation of the model in memory prior to printing, but we actually build a fairly shallow map where most things are strings. I think we should go all the way and keep the structured representation throughout, rather than build (and intern!) intermediate strings that are only used for printing; although there are maybe blockers to that I don't immediately see — thoughts?

As I explained in the description of this PR, the purpose of it is to solve the issue #833. Actually, this PR comes from a first commit I did in a old PR I've never pushed because we had to merge OptimAE first. I keep it as simple as possible because I want to merge it before #921.

In a next PR, I plan to produce models with the following ADT:
```
  type simple =
    | Int of Z.t
    | Real of Q.t
    | Bool of bool
    | Constructor of string

  type array =
    | AbstractArray of Id.t * Ty.t * Ty.t
    | Store of array * t * t

  and t =
    | Abstract of sy
    | Constant of simple
    | Array of array
    | Record of string * t list
```
This should be sufficient to perform evaluation of expression on the model, and so to perform checking  models.

Actually, I'm not really happy of the next commits of my old PR. The situation is complicated because we have a conflict between two designs: 

- First, the `semantic value` type is abstract, which means we cannot using directly the semantic values to get the model values because we have no access to their recursive structures in the `uf` module.
The case of records is particularly eloquent here. Generating the model value in `shostak` would be so easy. You just need to write a recursive function. This is what I did in my old PR. It works very fine with all the current theories but `arrays`...

- Second, arrays has no semantic value. In other words, we cannot retrieve directly the model value from the semantic values by a simple recursive call in `shostak`. Currently, we explore all the union-find in `uf` and retrieve the mode value by catching all the `get` expressions and accumulating values in caches.

There is  another annoying issue. The current function `choose_adequate_model` needs the class of the value for which we want a model value. For recursive values, as records or ADT, we will need more as we will perform recursive call. So the new `choose_adequate_model` requires the state of the union-find.... which means the best place to perform this computation is `uf` as it is actually? I don't know, may be we should create semantic values for arrays?

I fixed the issue you noticed with records and add some tests.

src/lib/reasoners/uf.ml
Halbaroth: I fixed it and add some tests. This PR solves #412 .

src/lib/reasoners/uf.ml
Halbaroth: Uhm I'm not sure. The function `is_forbidden_symbol` is used to filter the symbols on which we will iterate during the model generation. I think the comment means we have to keep `Access` symbol, even if this symbol is interpreted by the theory of records, otherwise we cannot construct the model value for this theory.

Halbaroth:
> I am not entirely convinced this is a good idea — I am not opposed to it but we may want to include things that are not `Sy.Name` in the models at some point (namely, values for partially interpreted functions such as division by zero) where we may regret this change.

If we need some other symbols in the futur, we can create a new type for it in `ModelMap`. A lot of the symbols of `Symbols` are fully interpreted and shouldn't appear in the signature of model values.

tests/float/test_float2.models.expected
Halbaroth: For this one, it seems the test is running without enabling the `fpa` theory:
```
  (rule
   (target test_float2.models_tableaux.output)
   (deps (:input test_float2.models.smt2))
   (package alt-ergo)
   (action
    (chdir %{workspace_root}
     (with-stdout-to %{target}
      (ignore-stderr
       (with-accepted-exit-codes 0
        (run %{bin:alt-ergo}
             --timelimit=2
             --enable-assertions
             --output=smtlib2
             --frontend dolmen
             --sat-solver Tableaux
             %{input})))))))
```

src/lib/util/util.mli
Halbaroth: I will add a comment in the code.

src/lib/structures/modelMap.mli
Halbaroth: I will remove the `Hstring.t`. As I explained, this is not a definitive version of the model representation.

src/lib/structures/modelMap.mli
Halbaroth: Actually, I didn't want to use a different constructor for abstract of array and abstract of anything. I rolled back to a previous version and now it's just a classical ADT with two constructors.

src/lib/structures/modelMap.mli
Halbaroth: Fixed

src/lib/structures/modelMap.mli
Halbaroth: Fixed

src/lib/structures/modelMap.mli
Halbaroth: It was done on purpose. I rolled back to a previous version.

src/lib/reasoners/uf.ml
bclement-ocp: I wrote this comment. It refers to the fact that in we (used to) need to preserve the record constructors for `check_records` et al in `models.ml`, but those functions have been removed. This means that either some stuff will no longer be working, or the new code in `uf.ml` is meant to replace those functions and we don't need to keep the record constructors here anymore (probably the latter).

src/lib/structures/modelMap.ml
Halbaroth: I think it doesn't matter. This order is only use for the printing.

src/lib/structures/modelMap.mli
```diff
@@ -28,33 +28,37 @@
...
-    of expressions matching the P.key type (V).
-*)
+type sy = Id.t * Ty.t list * Ty.t
+(** Typed symbol of function. *)
 
-module P : Map.S with type key =
-                        Symbols.t * Ty.t list * Ty.t
+module Value : sig
+  type array =
+    | AbstractArray of Id.t * Ty.t * Ty.t
```
bclement-ocp: You changed the name from `Abstract` to `AbstractArray` but my question remains: what is the difference between `Abstract (n, tys, ty)` and `Array (AbstractArray (n, tys, ty))`? I feel like just having `Store of t * t * t` should be enough here and would avoid this possible confusion.

src/lib/structures/modelMap.ml
Halbaroth: I didn't known `Fmt.iter_bindings` and `Fmt.iter` ;). I replaced them.

src/lib/structures/modelMap.ml
Halbaroth: Our expressions to print the graph have a very particular form. If we use one argument in our expression, we use all of them.

bclement-ocp: Ah, that is correct indeed!

tests/float/test_float2.models.expected
bclement-ocp: My point is that we don't inspect the preludes for the "known to be incomplete" message and that I think we should never print it for the `float` function (rather than always print it), because we expect users to use them with the preludes (or to know what they are doing).

(BTW the fact that we reply `"unknown"` and not `"sat"` is also an indication that the model might be incomplete)

bclement-ocp (commented):
I don't understand why these changes are needed to fix #833. #833 is about when we decide to print the model or not. This PR doesn't change that, it changes the way we store the model prior to printing it. Besides, the offending call to `Options.get_interpretation` that was the source of #833 is no longer present in `output_concrete_model`, so to me it looks like #833 is already fixed.

Anyways, arrays cannot have (meaningful) semantic values because the theory of arrays cannot have a Shostak solver, so I think we are stuck with a few contorsions for arrays specifically.

That doesn't mean we have to do the same contorsions for records. In fact, for Shostak theories we can (and probably should) use semantic values, as you suggest here and we also discussed a couple months ago. I don't think that the type of semantic values being abstract is a big issue; we just need to figure out the proper API to expose from the `Shostak` module.

I think that the best way might be to expose:
 - `children : r -> r list` that returns the list of direct alien children of a semantic value
 - `pp_model : r Fmt.t -> r Fmt.t` that prints a semantic value, using the first argument for its children

This should be enough for the union-find part to do its job. In fact, since model generation is done through the case split mechanism (`case_split ~for_model:true` is first called before extracting the model), we may only need `pp_model`?

This way, we can check with `X.is_a_leaf` -- if it returns true, this is an uninterpreted term (or AC symbol) and we can either print the associated abstract constant or (if it's an array) the structure reconstructed from UF. And we could get rid of most of the strings laying around…

(As an aside, I am not sure I understand the issue with `choose_adequate_model` and recursion. At the point we reach `choose_adequate_model` all values should be (conceptually) ground and have been assigned by the previous calls to `assign_value`. In fact, I think that arrays (and records, in the current implementation) are the only ones that truly need the equivalence class in `choose_adequate_model` — in my understanding any other theory needing this equivalence class is a symptom of a bug or a design issue.)

Halbaroth:
So I close this PR?

bclement-ocp:
I think this PR does some things right (even fixes some bugs!), even though it doesn't seem needed to fix #833. In particular, having a structured representation of models is useful (notably because for some theories such as arrays the solver doesn't really store the information in the appropriate way in the first place we we noted), even if we may want to change the way that representation is built in the future.

That said, if it is the first in a sequence of PRs that you are no longer happy with, maybe we should first take some time to brainstorm the API we want to have (my gut feeling is that it should be centered around *canonical semantic values*, not symbols, associated with an auxiliary mapping for uninterpreted terms that don't have ground semantic values, i.e. arrays, functions) to make sure we are going towards the direction we want and one that fixes the design bugs we can encounter with the current implementation.

src/lib/structures/modelMap.mli
Halbaroth: I add a documentation to explain the meaning of its fields.

src/lib/reasoners/uf.ml
Halbaroth: This comment isn't relevant anymore as we don't need to fold on all the terms in the union-find to produce model values for record symbols.

src/lib/frontend/models.mli
Halbaroth: It's not relevant anymore as we use `Expr.t` to store model values now.

src/lib/reasoners/uf.ml
Halbaroth: I put caches in the closure of the function `extract_concrete_model`.

src/lib/reasoners/uf.ml
Halbaroth: My comment was misleading. In fact `Set` symbols occur here by accident because `X.is_my_symbol` isn't the appropriate function to filter terms here. I just removed the comment and I'll do another PR in which a create a better predicate for this use case.

src/lib/structures/modelMap.mli
Halbaroth: Not relevant anymore.

tests/float/test_float2.models.expected
Halbaroth: I don't understand your point. I don't know if I'm supposed to print something if the user tries to produce models in presence of FPA symbols without loading the FPA theory. I guess we should print something because users can forget flags.

Halbaroth:
I update this PR to use `Expr.t` instead of strings in order to store model values. I also add a new function `to_const_term` in `sig.mli`. This function returns a constant interpreted term from a constant semantic value. Please refer to the commit message of [caae55f](https://github.com/OCamlPro/alt-ergo/pull/925/commits/caae55f0c7a40c6610dca07386d1a7611975e4e5) for more details.

src/lib/frontend/models.mli
```diff
@@ -32,16 +32,12 @@
 
 type t = {
   propositional : Expr.Set.t;
-  constants : ModelMap.t;
-  functions : ModelMap.t;
-  arrays : ModelMap.t;
-  terms_values : (Shostak.Combine.r * string) Expr.Map.t
+  model : ModelMap.t;
+  terms_values : Expr.t Expr.Map.t
```
bclement-ocp: Nit: this should be `term_values`.

src/lib/frontend/models.ml
```diff
@@ -245,221 +231,19 @@ module Pp_smtlib_term = struct
 
 end
 
-module SmtlibCounterExample = struct
-  let fresh_counter = ref 0
-
-  let reset_counter () = fresh_counter := 0
+(* of module SmtlibCounterExample *)
```
bclement-ocp: Not sure what this means?

src/lib/reasoners/shostak.ml
```diff
@@ -369,7 +386,14 @@ struct
...
     | Adt t -> ADT.is_constant t
     | Ite t -> ITE.is_constant t
-    | Ac _ | Term _ -> false
+    | Term t ->
+      begin
+        let Expr.{ f; xs; _ } = Expr.term_view t in
+        match f, xs with
+        | Symbols.(True | False | Void), [] -> true
+        | _ -> false
+      end
```
bclement-ocp: There should be a comment here explaining why this is not `Expr.is_const_term` (because other constants are semantic values).

src/lib/reasoners/sig.mli
```diff
@@ -128,11 +128,7 @@ module type SHOSTAK = sig
      [Some (t, false)], then there must be no context in which
      [solve r (fst X.make t)] raises [Unsolvable]. You have been warned! *)
 
-  (* choose the value to print and how to print it for the given term.
-     The second term is its representative. The list is its equivalence class
-  *)
-  val choose_adequate_model : Expr.t -> r -> (Expr.t * r) list -> r * string
-
+  val to_const_term : r -> Expr.t option
```
bclement-ocp: Should be documented here.

src/lib/reasoners/uf.ml
```diff
@@ -999,95 +998,42 @@ let assign_next env =
...
     in
-    let cls =
-      try List.rev_map (fun s -> s, ME.find s env.make) cls
-      with Not_found -> assert false
+    let tt =
+      match cls with
+      | [] ->
+        (* We do this because terms of some semantic values created by CS
+           aren't created and added to the union-find. *)
+        X.to_const_term rep
```
bclement-ocp: Not sure what this comment means; terms of *most* semantic values created by Alt-Ergo are not added to `env.make`. I'd have expected `X.to_const_term rep` to always be the correct answer (except for arrays). When is that not the case?

(My understanding is that `cls` is guaranteed to be either `[]` or `[rep]` and any other value would break Shostak invariants; is that correct?)

src/lib/structures/expr.ml
```diff
@@ -3106,3 +3124,20 @@ module BV = struct
...
+(** Constructors from the smtlib theory of functional arrays with
+    extensionality logic.
+
+    https://smtlib.cs.uiowa.edu/theories-ArraysEx.shtml *)
+module ArraysEx = struct
+  let select a i =
+    let rty =
+      match type_info a with
+      | Tfarray (_, rty) -> rty
+      | _ -> invalid_arg "[Expr.ArraysEx.select]"
```
```suggestion
      | _ -> invalid_arg "Expr.ArraysEx.select"
```

src/lib/util/util.ml
```diff
@@ -213,3 +212,25 @@ let rec print_list_pp ~sep ~pp fmt = function
 
 let internal_error msg =
   Format.kasprintf (fun s -> raise (Internal_error s)) msg
+
+(* TODO: we have to update the way we manage quoted symbols after
+   the merge of https://github.com/Gbury/dolmen/pull/198 *)
```
bclement-ocp: This has been merged, I believe (#955 )

tests/models/arith/arith7.optimize.expected
```diff
@@ -1,5 +1,7 @@
 
 unknown
+; This model is a best-effort. It includes symbols
+; for which model generation is known to be incomplete.
```
bclement-ocp: Which ones? This problem only uses the builtin `<` and `<=` on reals, as far as I can tell.

src/lib/structures/expr.ml
```diff
@@ -1015,14 +1023,24 @@ let mk_ite cond th el =
     if ty == Ty.Tbool then mk_if cond th el
     else mk_term (Sy.Op Sy.Tite) [cond; th; el] ty
 
-let [@inline always] const_term e =
-  (* we use this function because depth is currently not correct to
+let rec is_const_term e =
```
bclement-ocp: I thought this was replacing the pre-existing `const_term` function. Given that it's not, it is very confusing to have both `const_term` and `is_const_term` performing checks but returning different values. Can we maybe call this `is_model_term` or `is_value_term` instead?

bclement-ocp (commented):
It looks like there are still some issues with the ways arrays and abstract types are printed.

See for instance:

```smt-lib
(set-logic ALL)
(declare-sort S 0)
(declare-const s S)
(declare-datatype Pair ((pair (first (Array Int S)) (second (Array Int S)))))
(declare-const x Pair)
(assert (= (select (first x) 0) s))
(assert (= (first x) (second x)))
(check-sat)
(get-model)
```

This prints:

```smt-lib
unknown
(
  (define-fun s () S (as @a0 S))
  (define-fun x () Pair (pair .k4 .k4))
  (define-fun .k4 () (Array Int S) (store (as @a1 (Array Int S)) 0 s))
)
```

but the `.k4` symbol should not appear in the model as it is an internal name (and only symbols defined by the user should appear in models). The expected output should be along the lines of:

```smt-lib
unknown
(
  (define-fun s () S (as @a0 S))
  (define-fun x () Pair
    (pair
      (store (as @a1 (Array Int S)) 0 s)
      (store (as @a1 (Array Int S)) 0 s)))
)
```

We probably need one last substitution step of fresh array and abstract constants.

src/lib/frontend/models.ml
Halbaroth: I guess it's a rebase artefact. It was used to delimit a module that doesn't exist anymore.

src/lib/structures/expr.ml
Halbaroth: Sure

src/lib/reasoners/shostak.ml
Halbaroth: In fact, we could use `Expr.is_const_term`. I didn't use it because when I wrote this code, the predicate `Expr.is_const_term` should accept variables as we use it in the smart constructor of let bindings. I decided to use two predicates and we don't need to accept variables with `Expr.is_const_term` anymore. Still, it makes sense to write a specific code here. I add a comment to explain the situation.

src/lib/reasoners/uf.ml
Halbaroth: I agree. I reproduced the old code here but we don't need to filter the class at all. This code seems to work:
```ocaml 
    let mk = try ME.find t env.make with Not_found -> assert false in
    let rep, _ = try MapX.find mk env.repr with Not_found -> assert false in
    match X.to_const_term rep with 
    | Some v -> v, ME.add t v mrepr
    | None -> assert false
```
If `X.to_const_term rep` fails to produce a model term here, it's a bug.

src/lib/reasoners/shostak.ml
Halbaroth: After simplifying the code `model_repr_of_term` in the union-find, we don't need to change `X.is_constant` anymore. But I think it would be better to have this invariant:
```ocaml
let mk, _ = X.make e in
if Expr.is_const_term e then assert (X.is_constant mk)
```
In other words, `X.make` sends model values to constant semantic values.

```ocaml 
  let is_constant r =
    match r.v with
    | Arith t -> ARITH.is_constant t
    | Records t -> RECORDS.is_constant t
    | Bitv t -> BITV.is_constant t
    | Arrays t -> ARRAYS.is_constant t
    | Enum t -> ENUM.is_constant t
    | Adt t -> ADT.is_constant t
    | Ite t -> ITE.is_constant t
    | Term t ->
      begin
        let Expr.{ f; xs; _ } = Expr.term_view t in
        match f, xs with
        | Symbols.(True | False | Void | Name _), [] -> true
        | _ -> false
      end
    | Ac _ -> false
```

tests/models/arith/arith7.optimize.expected
Halbaroth: I didn't fix this issue because I don't understand your point as I explained here: https://github.com/OCamlPro/alt-ergo/pull/925#discussion_r1418918197

tests/float/test_float2.models.expected
bclement-ocp: My point is that it is fine to not print anything since replying "unknown" is already saying that the model may be incomplete. At least we should remove this when #917 is taken care of.

tests/models/arith/arith7.optimize.expected
bclement-ocp: I replied there (sorry, it got lost int the notifications somewhere) but I don't think this is related; `arith7` doesn't look like it is using any suspicious symbols.

src/lib/reasoners/sig.mli
Halbaroth: I put the documentation in the second signature. I do a copy.

src/lib/reasoners/shostak.ml
```diff
@@ -646,8 +676,8 @@ struct
           end
 
       | Term t, ty      -> (* case disable_adts() handled here *)
-        if Expr.const_term t ||
-           List.exists (fun (t,_) -> Expr.const_term t) eq then None
+        if Expr.is_model_term t ||
+           List.exists (fun (t,_) -> Expr.is_model_term t) eq then None
```
bclement-ocp, line 680: This changes from the old `Expr.const_term` to the new `Expr.is_model_term` — is that intended?

src/lib/reasoners/uf.ml
```diff
@@ -1099,15 +1035,193 @@ module MED = Map.Make
...
+  in fun env t ((mdl, mrepr) as acc) ->
+    let { E.f; xs; ty; _ } = E.term_view t in
+    if X.is_solvable_theory_symbol f ty
+    (* TODO: BVand, BVor and BVxor are not symbols of the Shostak theory
+       of bitvectors as the solve function of this theory cannot support them.
+       Still these symbols will appear in the union-find.
+       We should use a different predicate to discriminate appropriate terms
+       here. *)
+    || Sy.equal f Sy.(Op BVand) || Sy.equal f Sy.(Op BVor)
+    || Sy.equal f Sy.(Op BVxor)
```
bclement-ocp: This can be removed after #1007

src/lib/reasoners/uf.ml
```diff
@@ -1099,15 +1035,193 @@ module MED = Map.Make
...
+            let Expr.{ f; _ } = Expr.term_view t in
+            match f with
+            | Sy.Name { hs; _ } -> hs
+            | _ ->
+              (* We only store array declarations as keys in the cache
+                 [array_selects]. *)
+              assert false
+          in
+          let mdl =
+            if not @@ Id.Namespace.Internal.is_id (Hstring.view id) then
```
bclement-ocp, line 1197: Would checking the `defined` field in the `Sy.Name` make more sense here?

src/lib/structures/errors.ml
```diff
@@ -252,6 +257,18 @@ let report_mode_error fmt = function
   | Forbidden_command s ->
     fprintf fmt "Command %s" s
 
+let report_model_error ppf = function
```
bclement-ocp, line 260: I believe raising these indicates an internal error, in which case we should say so in the error message (otherwise the user will not understand why we are trying to substitute these identifiers, since they didn't ask).

src/lib/structures/expr.mli
```diff
@@ -326,8 +335,13 @@ val print_th_elt : Format.formatter -> th_elt -> unit
...
 
-val const_term : t -> bool
-(** return true iff the given argument is a term without arguments *)
+val is_model_term : t -> bool
+(** [is_model_term e] checks if the expression [e] is a model terms.
+    A model term can be:
+    - A record definition involving only model terms.
+    - A constructor application involving only model terms,
+    - A literal of a basic type (integer, real, boolean, unit or bitvector),
+    - An application of function of arity [0]. *)
```
bclement-ocp: Is this *any* function of arity 0 or a specific subset (I guess internal & abstract IDs? Or maybe just abstract IDs?)

src/lib/reasoners/arrays.ml
```diff
@@ -60,28 +60,12 @@ module Shostak (X : ALIEN) = struct
   let abstract_selectors _ _ = assert false
   let solve _ _ = assert false
   let assign_value r _ eq =
-    if List.exists (fun (t,_) -> Expr.const_term t) eq then None
+    if List.exists (fun (t,_) -> Expr.is_model_term t) eq then None
     else
       match X.term_extract r with
       | Some _, true ->
         Some (Expr.fresh_name (X.type_info r), false)
```
bclement-ocp, line 67: Should this be `mk_abstract` now that that exists?

bclement-ocp (commented):
I think we are mostly there; a couple comments mostly on use of internal vs abstract terms & related conditions.

I feel like most identifiers used in model generation should be abstract identifiers (which we will substitute if we learn their actual value) since that's their purpose but there might be issues with this.

src/lib/structures/errors.ml
Halbaroth: It's not necessary an internal error as the user could apply `subst` on a model while using the library. I prefix the error message with `Model error: ` to clarify where the exception could be raised.

src/lib/structures/expr.mli
Halbaroth: I thought it could be only abstract and internal names, but it seems to be any name. For instance, 
```smt2 
(set-logic ALL)
(declare-sort s 0)
(declare-const x s)
(declare-fun f (Int) s)
(assert (distinct (f 0) x))
(check-sat)
(get-model)
```
Here, the class representative of `x` is `x` itself. As we use `Expr.is_model_term` in `X.to_model_term`, the name `x` will occur in `Expr.is_model_term`.

src/lib/reasoners/shostak.ml
Halbaroth: It is. The function `Expr.const_term` isn't exposed anymore and it's not appropriate for model generation.

src/lib/reasoners/arrays.ml
Halbaroth: We could be but it's not very useful. Currently, the only non-regression test calling this part of the code is `embedded-array.models.smt2`:
```smt2
(set-logic ALL)
(declare-sort S 0)
(declare-const s S)
(declare-datatype Pair ((pair (first (Array Int S)) (second (Array Int S)))))
(declare-const x Pair)
(assert (= (select (first x) 0) s))
(assert (= (first x) (second x)))
(check-sat)
(get-model)
```
Applying your suggestion, I tried to retrieve the abstract names returned by `assign_value` but it seems to be complicated. For instance, with the above input file, AE will produces a fresh variables for each field of the pair `x` while generating the model. Let's say we have a fresh name `.k` and the equality `first x = .k` and `second x = .k`. Then we produce our abstract array with `assign_value`, say `@a`. Now, CC(X) uses Shostak's solve function to determine the better class representative and the solve function uses `X.str_cmp` to decide if we have to perform `.k -> @a` or `@a -> .k`. As these names have the same depth, the tag is used to determine the largest one. In this situation `.k` wins the game and becomes the class representative of `first x`.

Of course, we could look in the class of `first x` to retrieve the abstract name `@a` but we don't need to do that most of the time. In general, the class representative is the adequate model value.

In my opinion, the current implementation is fine. The returned value by `assign_value` isn't the definitive value but an internal name used during the model generation.

src/lib/reasoners/uf.ml
Halbaroth: It doesn't. The `defined` field means the user defined the semantic of the symbols in the input file with `define-fun`. Here we're looking for array identifiers that haven´t been declared by the user. For instance in the input:
```smt2
(set-logic ALL)
(declare-sort S 0)
(declare-const s S)
(declare-datatype Pair ((pair (first (Array Int S)) (second (Array Int S)))))
(declare-const x Pair)
(assert (= (select (first x) 0) s))
(assert (= (first x) (second x)))
(check-sat)
(get-model)
```
We'll produce a identifier for embedded array in `x` but this array isn't declared by the user. Thus we don't want to display the definition of this array at the top level of the model.

bclement-ocp: Right — we'd need `declared`, not `defined`, but we don't have that, of course. Thanks for the explanation

src/lib/structures/expr.mli
bclement-ocp: Ah yes, we don't generate abstract identifiers for things that are already identifiers, there would be no point!

src/lib/reasoners/arrays.ml
bclement-ocp: `.k` is a `model_term` (I believe) hence we shouldn't reach this branch that calls `fresh_name` in your example, and so I am not sure I understand your point.

My suggestion to use `mk_abstract` here is that, since we have a concept of abstract identifiers to indicate variables related to model generation, it would make sense to use it for the variables we create during model generation (even if not all of them will end up not in the final model). The current implementation also works, of course :)

bclement-ocp (approved):
I'll let you choose whether to use internal identifers or abstract identifiers during model generation; I think on the other points we are ready to merge.

PR #931, opened 2023-11-13
Fix issue 926

### LINKED ISSUE #926: Models for reals must be in real format
Alt-Ergo can output things like: `(declare-fun x () Real 2)` but this is not correct it must be `(declare-fun x () Real 2.0)`.

Halbaroth:
This PR fixes the issue #926 about printing reals in models.

The current printer always displays positive reals as rational numbers of the
form `(/ x y)` where `x` and `y` are positive integers. Negative reals are
represented by the expression `(/ (- x) y)` where `x` and `y` are positive
integers. The printer simplifies the expression if the denominator is one.

Notice that by the current implementation of Zarith:
```ocaml
type t = {
    num: Z.t; (** Numerator. *)
    den: Z.t; (** Denominator, >= 0 *)
  }
(* Type of rationals.
   Invariants:
   - den is always >= 0;
   - num and den have no common factor;
   - if den=0, then num is -1, 0 or 1.
   - if num=0, then den is -1, 0 or 1.
 *)
```
Any rational number of the form `0/q` satisfies `q = -1` or `q = 0` or `q = 1`.
Thus the printer will always output `0.0` in this case but for the case `0/0`
which shouldn't occur.


src/lib/reasoners/arith.ml
```diff
@@ -813,23 +813,27 @@ module Shostak
...
-      | Some c ->
-        let sg = Q.sign c in
-        if sg = 0 then "0"
-        else if sg > 0 then pprint_positive_const c
-        else Format.sprintf "(- %s)" (pprint_positive_const (Q.abs c))
+  let pp_positive_real ppf q =
+    if Z.equal (Q.den q) Z.one then
+      Fmt.pf ppf "%a.0" Z.pp_print (Q.num q)
+    else
+      Fmt.pf ppf "(/ %a.0 %a.0)" Z.pp_print (Q.num q) Z.pp_print (Q.den q)
```
bclement-ocp: Regardless of how we deal with the integer case, fractions should be `(/ 1 4)`, not `(/ 1.0 4.0)`.

src/lib/reasoners/arith.ml
```diff
@@ -813,23 +813,27 @@ module Shostak
...
-      match P.is_const (embed r) with
-      | None -> assert false
-      | Some c ->
-        let sg = Q.sign c in
-        if sg = 0 then "0"
-        else if sg > 0 then pprint_positive_const c
-        else Format.sprintf "(- %s)" (pprint_positive_const (Q.abs c))
+  let pp_positive_real ppf q =
+    if Z.equal (Q.den q) Z.one then
+      Fmt.pf ppf "%a.0" Z.pp_print (Q.num q)
```
bclement-ocp: Interestingly, the [SMT-LIB theory of Reals](https://smtlib.cs.uiowa.edu/theories-Reals.shtml) (see the `:values` section) actually mandates that integers don't have the `.0`… which is what we were doing, but not what other solvers are doing!

Maybe we should say that we respect the SMT-LIB standard and that #926 is a Why3 bug, not an Alt-Ergo bug :)

The standard however also mandates that negative numbers be written `(- n)`, not `-n` (for both reals and integers) which we don't do.

bclement-ocp (commented):
> It seems that AE (or Zarith?) replaces 0/0 by 0/1. It makes sense as the division by zero has no semantic in the SMT-LIB standard and asserting x = 0/0 means that x can have any value.

Alt-Ergo only uses `Q.t` to represent rationals, `0/0` should not happen.

src/lib/reasoners/arith.ml
Halbaroth: Dolmen refuses such terms as input, thus I guess it will refuse it in model also.

src/lib/reasoners/arith.ml
Halbaroth: I don't understand your last sentence. The printer will print `(- n)` if the input is `-n` where `n` is a positive integer.

src/lib/reasoners/arith.ml
bclement-ocp: Dolmen has separate grammars for scripts and responses. It accepts terms such as `(/ 1 4)` in models (it would be surprising if it didn't since that is both what the spec says and the way Z3 and CVC5 display rationals).

src/lib/reasoners/arith.ml
Halbaroth: Or maybe, you mean we have to print `(/ (- p) q)` instead of `(- (/ p q)` for the number `-p/q`?

bclement-ocp: Ah, you are right — I misread the code. I does seem that Dolmen refuses integers as reals values in models however… but I am pretty sure the spec says it should accept them.

bclement-ocp: > Or maybe, you mean we have to print `(/ (- p) q)` instead of `(- (/ p q)` for the number `-p/q`?

As a matter of fact, it looks like that would be the spec-compliant way of writing it, yeah.

src/lib/reasoners/arith.ml
Halbaroth: Seriously... it's ... bang

src/lib/reasoners/arith.ml
bclement-ocp: > Ah, you are right — I misread the code. I does seem that Dolmen refuses integers as reals values in models however… but I am pretty sure the spec says it should accept them.

Ah, it should *only when the logic is specificaly set to RA*, which it does properly. Yeah, let's print with the `.0` then.

Halbaroth: Uhm, it's annoying but I guess we should follow the specification and fix the bug in Dolmen. We should ask Why3 fix it also.

Halbaroth: > Or maybe, you mean we have to print `(/ (- p) q)` instead of `(- (/ p q)` for the number `-p/q`?

I'm not sure this one is a mandatory.

bclement-ocp: I mixed things up, it is not a bug in Dolmen — integers as reals are only allowed for models of the "Reals" theory, not of the "Reals_Ints" theory. Because of course.

And I agree, `(- (/ p q))` should be fine (even though it is not in the grammar defined in the link above), but please make `p` and `q` integers.

Gbury: For what it's worth, models are currently quite underspecified in smtlib. Additionally, as far as I know, the only two solutions to verify models (i.e. `dolmen` and the old python script) are pretty lenient and accept basically anything that is: 1) of the correct type and 2) evaluates down to a value.
The issue with reals is mostly with point 1, because when both the `Ints` and `Reals` theory are present then `1` is an `Int` (independently of the context), and to get a `Real` you need to specify `1.0` (well, most of the time, there are some exceptions where the smtlib specifies that a `to_real` conversion` should be added).

Halbaroth: Let's summarise the situation. For the theory `Reals`, we have to print `5`, not `5.0` for a real without decimal part and for the theory `Reals_Ints`, we have to print `5.0`? But we cannot make this difference right? Thus, we should print `5.0` in any case?

src/lib/reasoners/arith.ml
Gbury: Technically, dolmen has the same grammar/syntax for both scripts and response, and the typechecking is mostly the same (the biggest difference would be that most of the linearity warnings are silenced during models, since it should not matter).

src/lib/reasoners/arith.ml
bclement-ocp: Yes, that's about right.

For fractions however there are no reasons to add the `.0` I think, @Gbury can you confirm? I mostly insist on making them integers because well it is a representation of a rational as an irreducible fraction.

Halbaroth: I agree for the rationals ;) 
I will print `(/ (- p) q)` instead of `(- (/ p q)`.

src/lib/reasoners/arith.ml
bclement-ocp: Yes, I meant that things happen differently in both cases. Hm, but it is still true in my tests that `(/ 1 4)` is accepted by Dolmen in models and actually it seems to also accept it in scripts on my machine?

src/lib/reasoners/arith.ml
bclement-ocp: I am fine with `(- (/ 1 4))` but not `(- (/ 1.0 4.0))` :)

Halbaroth: > The issue with reals is mostly with point 1, because when both the `Ints` and `Reals` theory are present then `1` is an `Int` (independently of the context), and to get a `Real` you need to specify `1.0` (well, most of the time, there are some exceptions where the smtlib specifies that a `to_real` conversion` should be added).

I'm not saying you're wrong but it sounds strange we have to write `1.0` to specify the type of the value as we have to write its type in the model. For instance,
``` 
(define-fun () Real 1)
```
isn't ambiguous.

bclement-ocp (approved):
LGTM, feel free to do `(/ (- p) q)` if you prefer.

src/lib/reasoners/arith.ml
bclement-ocp: Sadly the [SMT-LIB theory](https://smtlib.cs.uiowa.edu/theories-Reals_Ints.shtml) is clear that the NUMERAL constants (i.e. integers) are of sort `Int` and the DECIMAL constants (with a `.`) are of sort `Real`. I guess it made someone's implementation simpler.

Gbury: Actually, `5.0` is accepted as a real in both the `Reals` and `Reals_Ints` theories, so we could just always print `5.0`.

For the fractions, as @bclement-ocp said, there is no need to add the `.0` because the smtlib spec specifies that in that specific case (i.e. division), the typer/solver should automatically add some `to_real` casts/function calls on both arguments, so it's allowed to provide arguments as integers literals.

Gbury: For the exact placement of the negation in fractions, `dolmen` does not case, at least in the models. The placement matters mainly for the linearity criterion if I recall correctly (which, as I said in another comment, is not enabled for models).

Gbury: @Halbaroth you're right that in your example it wouldn't be ambiguous, but it (sightly) complexifies the implementation because that means that you need some context to decide the type of `1`, so the typechecking is not strictly bottom-up anymore, and that's probably annoying and hard to do when you do both parsing and typechecking at the same time. In any case, the smtlib spec is pretty clear on that point, ^^

bclement-ocp: > Actually, `5.0` is accepted as a real in both the `Reals` and `Reals_Ints` theories, so we could just always print `5.0`.

Not that it matters much but I have a hard time reading the spread out spec sometimes, do you have a reference for that?  https://smtlib.cs.uiowa.edu/theories-Reals.shtml states:

> The set of values for the sort Real consists of 
>  - an abstract value for each irrational algebraic number
>  - all numerals
>  - all terms of the form (- n) where n is a numeral other than 0
>  - all terms of the form (/ m n) or (/ (- m) n) where 
>    - m is a numeral other than 0,
>    - n is a numeral other than 0 and 1,
>    - as integers, m and n have no common factors besides 1.

which does not include decimals.

bclement-ocp: Ah, I guess you mean that `5.0` is a valid term of real type, not necessarily a value of real type.

Gbury: I think the problem here is mainly a problem of interpretation, so here is the state of things:
- the `value` part of the spec is there to specify the semantics of the theory; in particular it is meant to be quite restrictive, most notably to try and make it so that on these values semantic equality if the same as equality of representation.
- there is a distinction between values and literals: as you saw, for `Reals`, the `value` part uses integer literals, rationals (as normalized pairs of integer literals), and irrationals constants for the values. That being said, decimal literals are still typed as reals.
- as mentioned in other comments, `dolmen` does not really use the `value` specification as a constraint on models: models are not, strictly speaking, required to define only `value`s for symbols. Instead, the only two requirements that `dolmen` currently imposes on model definitions are : 1) the definition is correctly typed, and 2) the defining expression of the definition can be evaluated (i.e. reduced down to a value). So for example `(+ 1 2.0)` would be accepted as defining expr for a real.

Gbury: > Ah, I guess you mean that `5.0` is a valid term of real type, not necessarily a value of real type.

Indeed, and also that any term of type real, which can be evaluated down to a value can be used in a model definition (which might be the surprising/important point actually).

bclement-ocp: Right, thanks for the clarifications! I was confused by the fact that the [language spec](https://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2021-05-12.pdf) mandates (page 65) that the result of *`get-value`* be a value term ("returns for each tᵢ a value term vᵢ") but relaxes that requirement for `get-model` ("The term t is expected, although not required, to be a value when f is a constant").

Gbury: Right, but even then for a `get-value`, I'd expect it to return a **well-typed** term that is a value.

In a context with the `Reals_Ints` theory, that would mean `(define-fun a () Real 5.0)`, because `(define-fun a () Real 5)` would not be well-typed.

I think it's reasonable because, in my opinion, the main reason behind why the `value` specification of reals uses integers literals is to have three non-overlapping categories (integer literals, non-integer rationals, and irrationals; which would be harder to specify using decimal literals), and should not be understood as a syntactic restriction on the set of real values. Instead any term that trivially maps to one of these values, should be considered as a value term.

Lastly, and to further my point, the `value` specification does not really give any concrete way to express irrationals, and the current and planned way to solve that is to accept algebraic numbers using new primitives, so I don't think the `value` specification is really to be used with too strict an interpretation, otherwise it would make models impossible to express in some situations.

bclement-ocp: > Right, but even then for a get-value, I'd expect it to return a well-typed term that is a value.

I would agree with you except for the footnote that states "Recall that value terms are particular ground terms defined in a logic for each sort (see Subsection 5.5.1)." and explicitly points to the `"values"` specification of the logic!

To clarify, I agree with what you are saying and am perfectly happy with ignoring that part of the spec and using well-typed ground terms in models, I am just explaining why I was confused by the wording in the spec.

Gbury: That's a good point, I'd be interested in knowing whether any solver actually respects that (and therefore outputs suspiciously-typed terms), or if it's something that really everyone ignores.

PR #932, opened 2023-11-14
Bubble printing model

### LINKED ISSUE #833: Bubble decision of printing models upwards

Halbaroth:
This PR solves the issue #833. The most of the refactoring have been done in OptimAE but still AE prints model in the library. Now, it will make this decision in the binary only.

Notice that AE printed model after getting timeout. I didn't keep this behaviour but I can restore it if you prefer. In my opinion, if we get a model after a timeout, the model is probably wrong and shouldn't be outputted.

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -154,21 +154,6 @@ let main () =
     (module SatCont.Make(TH) : Sat_solver_sig.S)
   in
 
-  let print_model ppf (Model ((module SAT), env)) =
-    match SAT.get_model env with
-    | None ->
-      let ur = SAT.get_unknown_reason env in
-      Printer.print_fmt (Options.Output.get_fmt_diagnostic ())
-        "@[<v 0>It seems that no model has been computed so \
-         far. You may need to change your model generation strategy \
-         or to increase your timeouts. \
-         Returned unknown reason = %a@]"
-        Sat_solver_sig.pp_unknown_reason_opt ur;
-
-    | Some (lazy model) ->
-      Models.output_concrete_model ppf model
-  in
-
   let solve (module SAT : Sat_solver_sig.S) all_context (cnf, goal_name) =
     let module FE = Frontend.Make (SAT) in
     if Options.get_debug_commands () then
@@ -209,7 +194,7 @@ let main () =
             Printer.print_fmt (Options.Output.get_fmt_diagnostic ())
               "@[<v 0>Returned unknown reason = %a@]"
               Sat_solver_sig.pp_unknown_reason_opt ur;
-            print_model (Options.Output.get_fmt_models ()) mdl
+            FE.print_model (Options.Output.get_fmt_models ()) partial_model
           end;
           Some mdl
         end
@@ -862,8 +847,9 @@ let main () =
       | {contents = `Get_model; _ } ->
         if Options.get_interpretation () then
           let () = match State.get partial_model_key st with
-            | Some model ->
-              print_model (Options.Output.get_fmt_regular ()) model
+            | Some (Model ((module SAT), env)) ->
+              let module FE = Frontend.Make (SAT) in
+              FE.print_model (Options.Output.get_fmt_regular ()) env
             | None ->
               (* TODO: add the location of the statement. *)
               recoverable_error "No model produced."
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -154,15 +154,11 @@ let main () =
     (module SatCont.Make(TH) : Sat_solver_sig.S)
   in
 
-  let pp_unknown_reason_opt ppf = function
-    | None -> Fmt.pf ppf "Decided"
-    | Some ur -> Sat_solver_sig.pp_unknown_reason ppf ur
-  in
-
   let print_model ppf (Model ((module SAT), env)) =
     let ur = SAT.get_unknown_reason env in
     Printer.print_fmt (Options.Output.get_fmt_diagnostic ())
-      "@[<v 0>; Returned unknown reason = %a@]" pp_unknown_reason_opt ur;
+      "@[<v 0>; Returned unknown reason = %a@]"
+      Sat_solver_sig.pp_unknown_reason_opt ur;
     match SAT.get_model env with
     | None ->
       Printer.print_fmt (Options.Output.get_fmt_diagnostic ())
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -73,16 +73,7 @@ let recoverable_error ?(code = 1) =
 let fatal_error ?(code = 1) =
   Format.kasprintf (fun msg -> recoverable_error ~code "%s" msg; exit code)
 
-let print_timeout_reason tr =
-  match tr with
-  | Some tr ->
-    Printer.print_fmt (Options.Output.get_fmt_diagnostic ())
-      "@[<v 0>; Returned timeout reason = %a@]" Util.pp_timeout_reason tr;
-  | None -> ()
-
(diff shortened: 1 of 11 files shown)


Halbaroth:
Oups, I realize that I don't print timeout reason anymore as it's done during printing model ...

Halbaroth:
In fact, we don't need to print the timeout reason in the binary when we catch the exception `Util.Timeout`. Indeed, in this case it could mean only one thing: Alt-Ergo received the alarm signal and the reason is `Signal`. Thus I revert my last commit. If we raise another timeout in the SAT solver, we convert it to `Unknown env` in `Frontend` and the reason will be print in `solving_loop`.

src/lib/reasoners/sat_solver_sig.ml
```diff
@@ -35,12 +35,22 @@ type timeout_reason =
   | Assume
   | ProofSearch
   | ModelGen
+[@@deriving show]
 
 type unknown_reason =
   | Incomplete
   | Memout
   | Timeout of timeout_reason
```
Stevendeo, line 43: Why not `[@@deriving show]` ?

Stevendeo (approved):

src/lib/reasoners/sat_solver_sig.ml
Halbaroth: Not sure to understand your remark :D

src/bin/common/solving_loop.ml
```diff
@@ -153,6 +153,23 @@ let main () =
     O.get_sat_solver (),
     (module SatCont.Make(TH) : Sat_solver_sig.S)
   in
+
+  let print_model ppf (Model ((module SAT), env)) =
+    let ur = SAT.get_unknown_reason env in
+    Printer.print_fmt (Options.Output.get_fmt_diagnostic ())
+      "@[<v 0>; Returned unknown reason = %a@]"
```
bclement-ocp: There shouldn't be `;` comments on the diagnostic channel.

src/bin/common/solving_loop.ml
```diff
@@ -153,6 +153,23 @@ let main () =
...
+      Sat_solver_sig.pp_unknown_reason_opt ur;
+    match SAT.get_model env with
+    | None ->
+      Printer.print_fmt (Options.Output.get_fmt_diagnostic ())
+        "@[<v 0>It seems that no model has been computed so \
```
bclement-ocp: This box is not closed.

src/bin/common/solving_loop.ml
```diff
@@ -837,21 +857,14 @@ let main () =
...
-            (* TODO: add the location of the statement. *)
-            recoverable_error "No model produced."; st
+          let _ = match State.get partial_model_key st with
+            | Some model ->
+              print_model (Options.Output.get_fmt_regular ()) model
+            | None ->
+              (* TODO: add the location of the statement. *)
+              recoverable_error "No model produced."
+          in
+          st
```
bclement-ocp: Not convinced we should print the unknown reason when we have a model, it will just be noise. If the user wants to see the unkonwn reason, they can call `(get-info :unknown-reason)`, we shouldn't force it down their throat.

(For `--dump-models` I am not entirely convinced either but it kind of makes sense b/c the user has no control & it helps distinguish the intermediate models from the final models with `--interpretation every`, but for `(get-model)` there is a mechanism in place to display that information with `(get-info :unknown-reason)` already)

src/bin/common/solving_loop.ml
```diff
@@ -153,6 +153,23 @@ let main () =
...
+    let ur = SAT.get_unknown_reason env in
+    Printer.print_fmt (Options.Output.get_fmt_diagnostic ())
+      "@[<v 0>; Returned unknown reason = %a@]"
+      Sat_solver_sig.pp_unknown_reason_opt ur;
+    match SAT.get_model env with
+    | None ->
+      Printer.print_fmt (Options.Output.get_fmt_diagnostic ())
+        "@[<v 0>It seems that no model has been computed so \
+         far. You may need to change your model generation strategy \
+         or to increase your timeouts."
```
bclement-ocp: Why change the error message?

bclement-ocp (changes requested):
Not a very big fan of duplicating the `print_model` function, could we keep it in the frontend and call it from both binary & worker?

src/bin/common/solving_loop.ml
Halbaroth: I didn't change it but I print the timeout reason just before. Now, if we don't print always the `unknown reason` while outputting the model, it makes sense to keep print it if we fail to produce a model here.

src/bin/js/worker_js.ml
```diff
@@ -120,6 +120,22 @@ let main worker_id content =
...
+      | None ->
+        Printer.print_fmt (Options.Output.get_fmt_diagnostic ())
+          "@[<v 0>It seems that no model has been computed so \
+           far. You may need to change your model generation strategy \
+           or to increase your timeouts."
+
+      | Some (lazy model) ->
+        Models.output_concrete_model ppf model
+    in
+
```
bclement-ocp: All the remarks made in the previous round on the version of this function in `solving_loop.ml` apply. Please either make the same changes as in `solving_loop.ml` or (better, because it would avoid this type of issues) keep the implementation in the library and call it from both `solving_loop.ml` and `worker.ml`.

bclement-ocp (changes requested):

src/bin/js/worker_js.ml
Halbaroth: I know ;) actually it's a bit annoying to factorize this function because of the presence of the first-class module as argument. I've several options but none of them are really satisfying. If I move back the function `print_model` in the module `Frontend`, I will need to recreate the module `Frontend` in the `get-model` handler as this function requires the module `SAT`. 
```ocaml 
          let () = match State.get partial_model_key st with
            | Some (Model ((module SAT), env)) ->
              let module FE = Frontend.Make (SAT) in
              FE.print_model (Options.Output.get_fmt_regular ()) env
            | None ->
              (* TODO: add the location of the statement. *)
              recoverable_error "No model produced."
          in
          st
```
I don't know what's the overhead of recreate a module here.

I can also retrieve the model in the `Solving_loop` and feed the function `print_model` with its return but `print_model` is just a wrapper around the pretty print of models to display a warning if we don't produce a model as we should according to the option `produce-models` and the status of the last `check-sat`. Thus, we have to feed the function `print_model` with an `option`, which is a bit ugly.

Halbaroth: I factorize the code using the first method I described.

bclement-ocp: There is no references in the frontend so it would be fine to re-create a `Frontend.Make` here, and the cost would be virtually zero compared to actually computing the model.

Having a `print_model_opt` in `Frontend` (not `Frontend.Make`) would also work. `Option.map` is not ugly :)

bclement-ocp: Actually `print_model` in `Frontend` could take the first-class module, which would probably be the best solution.

bclement-ocp (approved):

src/bin/js/worker_js.ml
Halbaroth: I thought I could do that but `Frontend` is already a functor taking the `SAT` module as argument. Besides, if I want to do that, I will need to move the gadt you created in `Solving_loop` into `Frontend` and expose it here. I prefer using the function `Frontend.Make`. If the cost is negligible, it's acceptable.

Issue #899, opened 2023-10-18
Imperative API for FunSAT


Halbaroth:
The FunSAT solver exposes a purely functional API which isn't convenient while executing a SMT-LIB script. Exposing a new imperative API simplify a lot the `Frontend` module.

The frontend `satml_frontend` exposes the same functional API built on the top of an imperative one... We may only expose an imperative API for all the SAT solvers and the current functional one of FunSAT will be an internal API only.

PR #936, opened 2023-11-16
Cleaner handling of step limit

### LINKED ISSUE #934: Reporting steps limit exceeded is awkward
Currently, when Alt-Ergo reaches the steps limit given using `--steps-bound`, the reporting is weird, e.g.:
```
[Error]; Fatal Error: Steps limit reached: 10
```
There is no reason for considering the reaching of the steps limit as a "fatal error" or even an error. It should instead report a regular `unknown` or `I don't know` result, as it does when the time limit is reached

Moreover, it would be better to comply to the more or less SMT compliant answer, where one gets extra information using `(get-info :reason-unknown)`

### LINKED ISSUE #935: feature wish: improve the reporting of the number of steps taken
Currently, to get the number of steps taken by Alt-Ergo (using the Dolmen frontend) one has to parse the SMT comment of the form
```
; File \".*\", line [0-9]+, characters [0-9]+-[0-9]+:.*(\\([0-9]+.?[0-9]*\\) steps)
```
which indeed comes before the answer `unsat` or `unknown`. There should be a better way to get this info, e.g. by issuing a command of the form
```
(get-info :steps-taken)
```
after the `(check-sat)` query, and get back an answer of a standardized form, e.g.
```
resource::steps = ...
```

(moreover, this should be documented)

Thanks!

Stevendeo:
This PR aims at cleanly stop when the step limit is reached instead of returning a Fatal Error

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -626,6 +626,12 @@ let main () =
         | Some b ->
           set_optimize b st
       end
+    | ":steps-bound", Symbol { name = Simple level; _ } ->
+      begin
+        match int_of_string_opt level with
+        | None -> print_wrn_opt ~name st_loc "integer" value; st
+        | Some i -> set_steps_bound i st
+      end
     | _ ->
       unsupported_opt name; st
   in
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -312,20 +312,13 @@ let main () =
     State.create_key ~pipe:"" "sat_state"
   in
 
-  let steps_bound: int State.key =
-    State.create_key ~pipe:"" "steps-bound"
-  in
-
   let named_terms: DStd.Expr.term Util.MS.t State.key =
     State.create_key ~pipe:"" "named_terms"
   in
 
   let set_steps_bound i st =
-    try
-      Steps.set_steps_bound i;
-      State.set steps_bound i st
-    with
-      Invalid_argument _ -> (* Raised by Steps.set_steps_bound*)
+    try DO.Steps.set i st with
+      Invalid_argument _ -> (* Raised by Steps.set_steps_bound *)
       fatal_error
         "Error while setting steps bound to %i: current step = %i."
         i
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -461,7 +461,6 @@ let main () =
     State.empty
     |> State.set solver_ctx_key solver_ctx
     |> State.set partial_model_key partial_model
-    |> set_steps_bound (O.get_steps_bound ())
     |> State.set named_terms Util.MS.empty
     |> DO.init
     |> State.init ~debug ~report_style ~reports ~max_warn ~time_limit
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -312,8 +312,24 @@ let main () =
     State.create_key ~pipe:"" "sat_state"
   in
 
+  let steps_bound: int State.key =
+    State.create_key ~pipe:"" "steps-bound"
+  in
+
   let named_terms: DStd.Expr.term Util.MS.t State.key =
     State.create_key ~pipe:"" "named_terms"
+  in
+
+  let set_steps_bound i st =
+    try
+      Steps.set_steps_bound i;
+      State.set steps_bound i st
+    with
+      Invalid_argument _ -> (* Raised by Steps.set_steps_bound*)
+      fatal_error
+        "Error while setting steps bound to %i: current step = %i."
+        i
+        (Steps.get_steps ())
   in
 
   let debug_parsed_pipe st c =
@@ -452,6 +468,7 @@ let main () =
     State.empty
     |> State.set solver_ctx_key solver_ctx
     |> State.set partial_model_key partial_model
+    |> set_steps_bound (O.get_steps_bound ())
     |> State.set named_terms Util.MS.empty
     |> DO.init
     |> State.init ~debug ~report_style ~reports ~max_warn ~time_limit
@@ -477,6 +494,7 @@ let main () =
         sat;
       st
     | Tableaux | Tableaux_CDCL | CDCL | CDCL_Tableaux ->
+      O.set_sat_solver sat;
       (* `make_sat` returns the sat solver corresponding to the new sat_solver
          option. *)
       DO.SatSolver.set sat st
--- a/src/bin/js/worker_js.ml
+++ b/src/bin/js/worker_js.ml
@@ -179,13 +179,8 @@ let main worker_id content =
         raise Exit
       | Errors.Error e ->
         begin match e with
-          | Errors.Run_error r -> begin
-              match r with
-              | Steps_limit _ ->
-                returned_status :=
-                  Worker_interface.LimitReached "Steps limit"
-              | _ -> returned_status := Worker_interface.Error "Run error"
-            end
+          | Errors.Run_error _ ->
+            returned_status := Worker_interface.Error "Run error"
           | _ -> returned_status := Worker_interface.Error "Error"
         end;
         Printer.print_err "%a" Errors.report e;
--- a/src/lib/frontend/d_state_option.ml
+++ b/src/lib/frontend/d_state_option.ml
@@ -99,11 +99,11 @@ let msatsolver =
 
 module SatSolver = (val msatsolver)
 
-let mstate =
+let msteps =
   let on_update _ sat st = Steps.set_steps_bound sat; st in
   (create_opt ~on_update "steps_bound" O.get_steps_bound)
 
-module Steps = (val mstate)
+module Steps = (val msteps)
 
 (* Some options can be initialized to gain some performance. *)
 let options_requiring_initialization = [
--- a/src/lib/frontend/d_state_option.ml
+++ b/src/lib/frontend/d_state_option.ml
@@ -99,6 +99,12 @@ let msatsolver =
 
 module SatSolver = (val msatsolver)
 
+let mstate =
+  let on_update _ sat st = Steps.set_steps_bound sat; st in
+  (create_opt ~on_update "steps_bound" O.get_steps_bound)
+
+module Steps = (val mstate)
+
 (* Some options can be initialized to gain some performance. *)
 let options_requiring_initialization = [
   (module SatSolverModule : S);
--- a/src/lib/frontend/d_state_option.mli
+++ b/src/lib/frontend/d_state_option.mli
@@ -64,5 +64,10 @@ module SatSolver : S with type t = Util.sat_solver
     on SatSolver: when SatSolver is updated, this one also is. *)
 module SatSolverModule : Accessor with type t = (module Sat_solver_sig.S)
 
+(** Option for setting the max number of steps. Interfaces with Steps.
+    The [set] function may raise Invalid_arg from the Steps.set_steps_bound call
+    if the new option value is lower than the current number of steps. *)
+module Steps : S with type t = int
+
 (** Initializes the state with options that requires some preprocessing. *)
 val init : D_loop.Typer.state -> D_loop.Typer.state
--- a/src/lib/frontend/d_state_option.mli
+++ b/src/lib/frontend/d_state_option.mli
@@ -64,7 +64,8 @@ module SatSolver : S with type t = Util.sat_solver
     on SatSolver: when SatSolver is updated, this one also is. *)
 module SatSolverModule : Accessor with type t = (module Sat_solver_sig.S)
 
-(** Option for setting the max number of steps. Interfaces with Steps.
+(** Option for setting the max number of steps. Interfaces with the toplevel
+    Steps module.
     The [set] function may raise Invalid_arg from the Steps.set_steps_bound call
     if the new option value is lower than the current number of steps. *)
 module Steps : S with type t = int
--- a/src/lib/frontend/frontend.ml
+++ b/src/lib/frontend/frontend.ml
@@ -273,10 +273,11 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
                 from_terms = [];
                 theory_elim = true;
                } Ex.empty
-          ) (SAT.empty ()) pb
+          )
+          pb
       in
       ignore (SAT.unsat
-                env
+                satenv
                 {E.ff=E.vrai;
                  origin_name = "";
                  gdist = -1;
@@ -397,7 +398,7 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
 
   let check_step_limit f env =
     match SAT.get_unknown_reason env.sat_env with
-    | Some (Step_limit _) -> env
+    | Some (Step_limit _) -> ()
     | _ -> f env
 
   let handle_sat_exn f ?loc x env =
@@ -428,7 +429,7 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
 
   let th_assume = wrap_f internal_th_assume
 
-  let process_decl ~hook_on_status env d =
+  let process_decl ?(hook_on_status=(fun _ -> ignore)) env d =
     try
       match d.st_decl with
       | Push n -> check_step_limit (internal_push ~loc:d.st_loc n) env
@@ -443,14 +444,14 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
           (* If we have reached an unknown state, we can return it right
              away. *)
           match SAT.get_unknown_reason env.sat_env with
-          | Some (Step_limit _) -> raise (SAT.I_dont_know env.sat_env)
+          | Some (Step_limit _) -> raise SAT.I_dont_know
           | Some _ ->
             (* For now, only the step limit is an unknown step reachable
                here. We could raise SAT.I_dont_know as in the previous case,
                but we have choosen a defensive strategy. *)
             assert false
           | None ->
-            internal_query ~loc:d.st_loc env (n, f, sort);
+            internal_query ~loc:d.st_loc (n, f, sort) env;
             match env.res with
             | `Unsat ->
               hook_on_status (Unsat (d, env.expl)) (Steps.get_steps ())
--- a/src/lib/frontend/frontend.ml
+++ b/src/lib/frontend/frontend.ml
@@ -142,7 +142,7 @@ module type S = sig
     mutable expl : Explanation.t
   }
 
-  type 'a process = ?loc : Loc.t -> env -> 'a -> unit
+  type 'a process = ?loc : Loc.t -> 'a -> env -> unit
 
   val init_env : ?sat_env:sat_env -> used_context -> env
 
@@ -226,7 +226,7 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
     mutable expl : Explanation.t
   }
 
-  type 'a process = ?loc : Loc.t -> env -> 'a -> unit
+  type 'a process = ?loc : Loc.t -> 'a -> env -> unit
 
   let init_env ?(sat_env=SAT.empty ()) used_context =
     {
@@ -273,11 +273,10 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
                 from_terms = [];
                 theory_elim = true;
                } Ex.empty
-          )
-          pb
+          ) (SAT.empty ()) pb
       in
       ignore (SAT.unsat
-                satenv
+                env
                 {E.ff=E.vrai;
                  origin_name = "";
                  gdist = -1;
@@ -300,14 +299,14 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
     if Options.get_unsat_core () then Ex.singleton (Ex.RootDep {name;f;loc})
     else Ex.empty
 
-  let internal_push ?(loc = Loc.dummy) (env : env) (n : int) : unit =
+  let internal_push ?(loc = Loc.dummy) (n : int) (env : env) : unit =
     ignore loc;
     Util.loop ~f:(fun _ res () -> Stack.push res env.consistent_dep_stack)
       ~max:n ~elt:(env.res, env.expl) ~init:();
     Steps.apply_without_step_limit (fun () -> SAT.push env.sat_env n);
     SAT.push env.sat_env n
 
-  let internal_pop ?(loc = Loc.dummy) (env : env) (n : int) : unit =
+  let internal_pop ?(loc = Loc.dummy) (n : int) (env : env) : unit =
     ignore loc;
     let res, expl =
       Util.loop ~f:(fun _n () _env -> Stack.pop env.consistent_dep_stack)
@@ -319,8 +318,8 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
 
   let internal_assume
       ?(loc = Loc.dummy)
-      (env : env)
-      ((name, f, mf) : string * E.t * bool) =
+      ((name, f, mf) : string * E.t * bool)
+      (env : env) =
     let is_hyp = try (Char.equal '@' name.[0]) with _ -> false in
     if is_hyp || not (unused_context name env.used_context) then
       let expl =
@@ -350,13 +349,13 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
       | `Unsat ->
         env.expl <- expl
 
-  let internal_pred_def ?(loc = Loc.dummy) env (name, f) =
+  let internal_pred_def ?(loc = Loc.dummy) (name, f) env =
     if not (unused_context name env.used_context) then
       let expl = mk_root_dep name f loc in
       SAT.pred_def env.sat_env f name expl loc;
       env.expl <- expl
 
-  let internal_query ?(loc = Loc.dummy) env (n, f, sort) =
+  let internal_query ?(loc = Loc.dummy) (n, f, sort) env =
     ignore loc;
     let expl =
       match env.res with
@@ -386,8 +385,8 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
 
   let internal_th_assume
       ?(loc = Loc.dummy)
-      env
-      ({ Expr.ax_name; Expr.ax_form ; _ } as th_elt) =
+      ({ Expr.ax_name; Expr.ax_form ; _ } as th_elt)
+      env =
     if not (unused_context ax_name env.used_context) then
       match env.res with
       | `Sat | `Unknown ->
@@ -401,8 +400,8 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
     | Some (Step_limit _) -> env
     | _ -> f env
 
-  let handle_sat_exn f ?loc env x =
-    try f ?loc env x with
+  let handle_sat_exn f ?loc x env =
+    try f ?loc x env with
     | SAT.Sat -> env.res <- `Sat
     | SAT.Unsat expl ->
       env.res <- `Unsat;
@@ -414,8 +413,8 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
   (* Wraps the function f to check if the step limit is reached (in which case,
      don't do anything), and then calls the function & catches the
      exceptions. *)
-  let wrap_f f ?loc env x =
-    check_step_limit (fun env -> handle_sat_exn f ?loc env x) env
+  let wrap_f f ?loc x env =
+    check_step_limit (handle_sat_exn f ?loc x) env
 
   let push = wrap_f internal_push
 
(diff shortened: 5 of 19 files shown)


src/bin/js/worker_js.ml
```diff
@@ -171,9 +171,6 @@ let main worker_id content =
         begin match e with
           | Errors.Run_error r -> begin
               match r with
-              | Steps_limit _ ->
-                returned_status :=
-                  Worker_interface.LimitReached "Steps limit"
               | _ -> returned_status := Worker_interface.Error "Run error"
             end
```
hra687261: Suggestion:
```
          | Errors.Run_error _ -> 
            returned_status := Worker_interface.Error "Run error"
```

hra687261 (approved):
LGTM modulo a couple of superficial suggestions

src/lib/reasoners/sat_solver_sig.ml
```diff
@@ -39,6 +39,7 @@ type timeout_reason =
 type unknown_reason =
   | Incomplete
   | Memout
+  | Steps_limit of int
```
hra687261: `Step_limit`*

src/lib/util/util.mli
```diff
@@ -29,6 +29,7 @@
 (**************************************************************************)
 
 exception Timeout
+exception Step_limit_reached of int
```
bclement-ocp, line 32: Those two have a very similar role. We could factor that out into a single `exception Limit_reached of resources_limit` and `type resources_limit = Time | Steps of int` to ensure uniform handling.

(I am fine with keeping the PR as is and making this a long-term cleanup issue instead)

bclement-ocp (approved):

bclement-ocp (changes requested):
I approved but actually — please add a test with a very low step limit (e.g. 1) and `(get-info :reason-unknown)` to ensure that:

 - We properly return `unknown` when reaching the step lmit, and
 - We properly say that the reason is due to the step limit

src/lib/util/util.mli
Stevendeo: I'll do the change right now, I like this factorization

bclement-ocp: (Note that the type should be `resource_limit`, not `resources_limit`, because of English rules being different from French ones)

Stevendeo: Actually this change is a little trickier than I thought. Right now the Timeout exception is catched by the fronend while, in my version, the limit_step exception is catched by the solver. I will leave it as it until we do a whole rework of the exception handling in the solver.

src/bin/common/solving_loop.ml
```diff
@@ -351,10 +351,26 @@ let main () =
...
   let named_terms: DStd.Expr.term Util.MS.t State.key =
     State.create_key ~pipe:"" "named_terms"
   in
 
+  let set_steps_bound i st =
+    try
+      Steps.set_steps_bound i;
+      State.set steps_bound i st
+    with
+      Invalid_argument _ -> (* Raised by Steps.set_steps_bound*)
```
bclement-ocp: I this is raised by `Steps.set_steps_bound` the `try .. with ..` block should only include `Steps.set_steps_bound` here.

src/bin/js/worker_js.ml
```diff
@@ -181,9 +181,6 @@ let main worker_id content =
         begin match e with
           | Errors.Run_error r -> begin
               match r with
-              | Steps_limit _ ->
-                returned_status :=
-                  Worker_interface.LimitReached "Steps limit"
               | _ -> returned_status := Worker_interface.Error "Run error"
```
bclement-ocp: @hra687261's suggestion of removing the match here still applies.

src/lib/reasoners/fun_sat.ml
```diff
@@ -1621,16 +1621,24 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+                        { acc.guards with
+                          current_guard = new_guard;
+                          guards = guards;
+                        }}
+          with
+          | Util.Step_limit_reached _ ->
+            (* If we reached the step limit while pushing, we have to cancel the
+               push.
+               TODO: properly cancel the push. *)
+            env
```
bclement-ocp: Can you explain why the steps limit increases while pushing? We call: `E.fresh_name`, `save_guard_and_refs` (which only pushes things on a stack), `mk_gf` (which is just a dumb record constructor) and `ME.add`, none of which seem to increase the step count.

src/lib/reasoners/sat_solver_sig.ml
bclement-ocp: @hra687261's remark still applies.

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1270,13 +1270,20 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+          Steps.push_steps ();
+          {acc with guards =
+                      {acc.guards with
+                       current_guard = expr_guard;} }
+        with
+        | Util.Step_limit_reached _ ->
+          (* If we reached the step limit while pushing, we have to cancel the
+             push.
+             TODO: properly cancel the push. *)
+          env
```
bclement-ocp: Returning `env` here is not correct for satML. Because the underlying satml solver uses mutability, it is shared between `env` and `acc`; if we reach the steps limit here, `new_vars` (called through `create_guard`) can leave the satml solver in an inconsistent state with variables partially added. Plus, `env` will be out of sync with the content of the satml solver and weird things will follow.

I see two solutions here — either we make sure that `new_vars` returns a valid environment here, so that `acc` is correct in the exception handler and we can then pop a corresponding amount, or we temporarily disable the step limit while pushing. I'd prefer the second option (even though it can cause going slightly over the step limit) because the first one is opening up a can of worms where we *will* mess up and introduce subtle bugs in the future.

bclement-ocp (changes requested):
I don't think we can do this this way (see also #946 )

For satML in particular we would need to make sure that whenever the "step limit reached" exception is raised it cannot leave the solver in an invalid intermediate state, which we will very probably mess up. I have made comments on that fact wrt `push` but I now realize that it applies more generally — the sat solver was written with the fact that it could get interrupted by an exception at any time and then re-used. Trying to make it do so is probably a recipe for many bugs ranging from crashes to soundness holes.

Fixing this properly would be a huge amount of extremely error-prone work and frankly I don't think that it is worth it. I'd be in favor of simply putting the solver in a "imited" state once the steps limit (or time limit, for that matter) is exhausted, and make it reject all commands except for diagnostic ones (e.g. `get-info`). This would fix #934 without an equivalent implicit requirement (you really shouldn't keep using the solver once the steps limit has been exceeded because its output can be complete garbage at that point).

bclement-ocp:
> you really shouldn't keep using the solver once the steps limit has been exceeded because its output can be complete garbage at that point

Note that this really only applies to satML, not to the Tableaux solver, since the Tableaux solver is purely functional and won't get stuck in an intermediate invalid state. But satML is the default solver that 99% of users will use, so…

bclement-ocp:
> Note that this really only applies to satML, not to the Tableaux solver, since the Tableaux solver is purely functional and won't get stuck in an intermediate invalid state. But satML is the default solver that 99% of users will use, so…

Well and even then only "pure" Tableaux, not Tableaux-CDCL, which has an internal CDCL solver that could potentially end up in an invalid state as well!

src/lib/reasoners/sat_solver_sig.ml
Stevendeo: This is just a copy of the previous ADT, which uses plural for steps. I may not be as fluent as you are in English, but my guess is "step limit" is okay to mean "this step is the limit" and "steps limit" is also ok to mean "this number of steps is the limit".

src/bin/common/solving_loop.ml
Stevendeo: If set_steps_bound fails, we do not want `State.set` to be called. It was written initially when I put "recoverable_error" instead of "fatal_error" in the next line.

src/lib/reasoners/sat_solver_sig.ml
bclement-ocp: It doesn't really matter (and I wouldn't have said anything if @hra687261 did not make a remark), but I'm pretty sure in this situation "steps limit" is incorrect, or at least not idiomatic (see for instance [here](https://english.stackexchange.com/questions/229406/how-to-correctly-write-chains-of-nouns-with-plurals-like-messages-list-and-ap)). It is a "resource limit", a "character limit", and a "step limit". This is different from French where we would say "la limite des pas", and "la limite du pas" would be wrong (I guess we could say "the limit of the steps" but it would be weird).

src/lib/reasoners/fun_sat.ml
Stevendeo: I thought `fun_sat` and `satml_frontend` had the same behavior regarding the steps; I was wrong.

src/bin/common/solving_loop.ml
bclement-ocp: In that case, the best way to write the code in modern OCaml is to use `match .. exception`:

```ocaml
match Steps.set_steps_bound i with
| () -> State.set steps_bound i st
| exception Invalid_argument _ ->
  fatal_error "..."
```

This helps controlling the scope of exception handling (I make the remark here because for standard exceptions such as `Not_found`, `Failure`, or `Invalid_argument` it is very easy to include an unrelated function in the `try .. with ..` block that could also raise that same exception and cause puzzling bugs).

src/lib/reasoners/satml_frontend.ml
Stevendeo: > Returning env here is not correct for satML. 

That's the point of the `TODO`, we need to actually cancel the push instead of just returning the env. But disabling the step limit is a better idea, I'll go for it.

bclement-ocp: What I mean by "not correct" here is worse — it will cause the type of crashes that we see in #946 , see my toplevel comment.

src/lib/reasoners/sat_solver_sig.ml
Stevendeo: Just noted I defined the exception Step_limit_reached :sweat_smile:

Stevendeo:
Just in case, I disactivated the steps exception at the Frontend level to make sure we never break any invariant in the future.

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1270,13 +1270,19 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+          Stack.push expr_guard acc.guards.stack_guard;
+          Steps.push_steps ();
+          {acc with guards =
+                      {acc.guards with
+                       current_guard = expr_guard;} }
+        with
+        | Util.Step_limit_reached _ ->
+          (* This function should be called without step limit
+             (see Steps.apply_without_step_limit) *)
+          assert false
```
bclement-ocp, line 1273: Why not directly call `Steps.apply_without_step_limit` from inside `Satml_frontend.push`?

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1335,6 +1341,7 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
       end;
       dep
     | (Util.Timeout | I_dont_know _ ) as e -> raise e
+    | Util.Step_limit_reached n -> i_dont_know env (Step_limit n)
```
bclement-ocp: As I explained, this is probably incorrect (i.e. there is no guarantee that `env` is in a valid state), because there is no guarantee that the sat solver be in a consistent state when the `Step_limit` is reached.

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1350,6 +1357,10 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
            (* don't attempt to compute a model if timeout before
               calling unsat function *)
            i_dont_know env (Timeout Assume)
+         | Util.Step_limit_reached _ ->
+           (* When reaching the step limit on an assume, we do not want to
+              answer 'unknown' right away. *)
+           env
```
bclement-ocp: Same as above

bclement-ocp (changes requested):

src/lib/reasoners/satml_frontend.ml
Stevendeo: > Just in case, I disactivated the steps exception at the Frontend level to make sure we never break any invariant in the future.

For this reason

src/lib/reasoners/satml_frontend.ml
bclement-ocp: For instance we could raise `Steps_limit` in the middle of the `Satml.theory_propagate` loop and end up with the queues in a weird state that may or may not be valid.

Stevendeo: I am adding a wrapper at the Frontend level that blocks the calls to the SAT (sorry it came just after the review request, I forgot to add it)

Stevendeo: It does not solve the issue where the number of steps is updated after failing once, which cannot be solved by putting the solver in a "limited" state without making it permanent.
I'm starting to think the fatal error was a good choice while there is no deep work on the solver to make sure environments are consistent even when failing in the middle of the execution of an instruction.

src/lib/reasoners/satml_frontend.ml
Stevendeo: By blocking the instructions at the frontend level, we ensure the environment cannot be used in a wrong way. Besides, we may want to display information from the invalid environment when the step limit is reached, so I would rather keep it like that.

src/lib/reasoners/satml_frontend.ml
bclement-ocp: > It does not solve the issue where the number of steps is updated after failing once, which cannot be solved by putting the solver in a "limited" state without making it permanent.

If the solver is in a "step limit exceeded" state, that shouldn't change if the step limit is updated, so that should be fine, no?

src/lib/reasoners/satml_frontend.ml
bclement-ocp: But if `Step_limit_reached` was raised, the `env` is potentially invalid to use. How do you avoid it being reused in this case?

src/lib/frontend/frontend.ml
```diff
@@ -398,26 +401,37 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
...
 
+  (* Wraps the function f to check if the step limit is reached (in which case,
+     don't do anything), and then calls the function & catches the
+     exceptions. *)
+  let wrap_f f ?loc env x =
+    check_step_limit (fun env -> handle_sat_exn f ?loc env x) env
+
   let push = handle_sat_exn internal_push
 
   let pop = handle_sat_exn internal_pop
```
bclement-ocp: These also need `wrap_f`

Stevendeo: I did not put the wrapper on these because I did not want to silently fail (fail = do nothing) on `push` and `pop`.

bclement-ocp: Ah, I did not catch this, but the wrapper should *not* silently fail, it should return an error.

If the underlying state of the SAT solver is invalid, there is no guarantee that `push` and `pop` are correct either, so this could possibly crash as well.

Stevendeo: Not now `push` can no longer fail with the disabled steps check

Stevendeo: Oh I see your point sorry, we may push a wrong environment; but is it really an issue?

bclement-ocp: The issue is not that `push` can fail. The issue is that if you call another function, say `assume`, and *that* function reaches the step limit, the environment becomes invalid (i.e. arbitrary internal invariants may be broken), and we shouldn't be calling `push` (or `pop`, for that matters) on an invalid environment.

bclement-ocp: > Oh I see your point sorry, we may push a wrong environment; but is it really an issue?

If a step limit is reached, there may or may not be some invariants in the environment that are broken. Without looking at the code, I don't know which invariants are broken, or how calling `push` or `pop` interacts with those broken invariants. It might be that this is perfectly fine, it might not be, but we won't know without tracking all the invariants that can potentially get broken, and even then if we can only call `push` and `pop` it is not very interesting, so we should just prevent that.

Stevendeo: The sat environment should keep track of its own inconsistency wrt the steps; with this, we could actually save invalid environments, update the step number and make sure we don't ever work with invalid environments. What do you think?

Stevendeo: Nevermind, it's pretty much what is already done. When the unknown_reason is Some (Step_limit _) we should not be able to do anything

bclement-ocp: I am not sure I understand why updating the step number seems to be such a huge issue. Once we have an unknown reason with `Step_limit` updating the step number is not going to change that unknown reason. What am I missing here?

Stevendeo: Now we use the Limit step as a reason not to keep using the environment, it is not an issue anymore.

src/lib/reasoners/fun_sat.ml
```diff
@@ -1753,6 +1755,10 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
       (* don't attempt to compute a model if timeout before
          calling unsat function *)
       i_dont_know env (Timeout Assume)
+    | Util.Step_limit_reached _ ->
+      (* When reaching the step limit on an assume, we do not want to answer
+         'unknown' right away. *)
+      env
```
bclement-ocp: I believe that this is still leaking a potentially invalid environment into the wild without *protecting* it with an unknown reason. This should raise `i_dont_know` and it should be caught in `handle_sat_exn` in the frontend to set the state there to `` `Unknown`` and following operations will be protected by `wrap_f`.

Edit: accidentally a word ("invalid")

bclement-ocp: Sorry, this comment shouldn't be here — it should be in the corresponding position in `satml_frontend`

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1350,6 +1357,8 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
            (* don't attempt to compute a model if timeout before
               calling unsat function *)
            i_dont_know env (Timeout Assume)
+         | Util.Step_limit_reached _ ->
+           env
```
bclement-ocp: I believe that this is still leaking a potentially invalid environment into the wild without protecting it with an unknown reason. This should raise i_dont_know and it should be caught in handle_sat_exn in the frontend to set the state there to  `Unknown and following operations will be protected by wrap_f.

Stevendeo: If we raise `I_dont_know` we will answer `unknown` during an assertion; is that really what we want to do? I would rather update the environment with the correct `unknown_reason`

bclement-ocp: > If we raise I_dont_know we will answer unknown during an assertion

If we raise `I_dont_know` I think it should be caught by the frontend which will update the unknown reason as needed, and only display `unknown` once the solving loop is done processing, but worth checking.

Stevendeo: The unknown reason is on the sat solver side

bclement-ocp: Ah right — yeah returning the env with `Step_limit` unknown reason is what needs to happen then.

src/bin/common/solving_loop.ml
```diff
@@ -471,6 +480,7 @@ let main () =
     State.empty
     |> State.set solver_ctx_key solver_ctx
     |> State.set partial_model_key partial_model
+    |> set_steps_bound (O.get_steps_bound ())
```
bclement-ocp: Should this be taken care of by `DO.init`?

src/bin/common/solving_loop.ml
```diff
@@ -496,6 +506,7 @@ let main () =
         sat;
       st
     | Tableaux | Tableaux_CDCL | CDCL | CDCL_Tableaux ->
+      O.set_sat_solver sat;
```
bclement-ocp, line 489: Is it necessary to add this back?

src/lib/frontend/d_state_option.ml
```diff
@@ -99,6 +99,12 @@ let msatsolver =
 
 module SatSolver = (val msatsolver)
 
+let mstate =
```
bclement-ocp: Why `mstate`?

src/lib/frontend/d_state_option.mli
```diff
@@ -64,5 +64,10 @@ module SatSolver : S with type t = Util.sat_solver
     on SatSolver: when SatSolver is updated, this one also is. *)
 module SatSolverModule : Accessor with type t = (module Sat_solver_sig.S)
 
+(** Option for setting the max number of steps. Interfaces with Steps.
```
bclement-ocp: Should clarify "the toplevel Steps module" or some such — otherwise it is a bit confusing since it seems like this is interfacing with itself.

tests/smtlib/testfile-steps-bound.dolmen.smt2
```diff
@@ -0,0 +1,22 @@
+(set-logic ALL)
+
+(declare-const x Int)
+(declare-const y Int)
+(assert (= (* x x) 3))
+
+; Should answer 'unknown'
+(push 1)
+(set-option :steps-bound 2)
```
bclement-ocp, line 12: I am not sure it is worth being able to define the `set-option` after stuff has been defined/pushed/etc. This is an option that should be supported at the top-level (i.e. before any constants/asserts) but we really don't need to support it past that — is there an use case here?

tests/smtlib/testfile-steps-bound.dolmen.expected
```diff
@@ -0,0 +1,5 @@
+
+unknown
+(:reason-unknown Step limit: 3)
```
bclement-ocp, line 3: This is not a valid syntax. This should be `(:reason-unknown s)` where `s` is a valid s-expression, "Step limit: 3" is not a valid s-expression. It should be something similar to `(:reason-unknown (:step-limit 3))`.

bclement-ocp (changes requested):
There are still some issues with the smt interface, see inline comments.

I wonder if we should call the option `steps-bound`. I know that this is the name of the CLI flag, but we use "limit" for the time limit and the Fourier-Motzkin cross-limit (not sure what that limits) and "limit" is closer to what it actually do. The other place we use "bound" is for the "age bound" that controls the behavior of matching and is not a resource limit. If we change the name, it should also be changed for the CLI options, so best to keep "steps-bound" for now and do the rename everywhere in a separate PR.

src/bin/common/solving_loop.ml
Stevendeo: This, like the previous one, are rebase artifacts; thanks for pointing them out.

src/bin/common/solving_loop.ml
Stevendeo: We don't even need to initialize this variable

tests/smtlib/testfile-steps-bound.dolmen.smt2
Stevendeo: Not a use case, but I did find a bug while implementing this when setting an option after a push, so I think we should keep the test as it is.

tests/smtlib/testfile-steps-bound.dolmen.expected
Stevendeo: https://github.com/OCamlPro/alt-ergo/issues/966

tests/smtlib/testfile-steps-bound.dolmen.smt2
bclement-ocp: I don't think we should have tests for features that we don't want to support, unless if the tests says that it should be an error. Otherwise, that is implicitly saying that we support the feature.

Stevendeo: Let's say that I want to support this feature then, the whole test is a use case (let's try something with a given limit, then with a greater limit).

Stevendeo: Besides, we don't want alt-ergo to throw random `assert false` even if the user writes non-sense in its smt file

bclement-ocp: You are making a false dichotomy in your second message. The options are not "support this use case" or "raise an assertion failure"; there is a mechanism for reporting proper errors to the user that can be used. If we currently raise an assertion failure for this test case, that is a bug, what we should do is display an appropriate error instead.

I don't really understand the point of setting a limit then a greater limit in the current infrastructure, since the solver will have to re-do everything it did the second time, so it seems to me you'd just be wasting time. If the internal solver infrastructure was able to support this and we could say "ignore the previously set limit and keep solving", why not, I'd be happy to throw that in. But currently, we can't; if the limit is reached, we have to start from scratch, which induces additional complexity (need to reset the solver, replay the assertion queue, make sure everything is properly synced, etc.) of the kind you just advocated for removing yesterday.

In any case: this is currently not supported (as evidenced by the fact it is neither documented nor working), and it would require committing to additional complexity to support properly given the current infrastructure (again, because we can't just re-use the solver). This doesn't sound like a decision that should be taken by introducing a test case in a PR; it sounds like a change that an issue should be created for and evaluated separately. 

Given that it is also a thing I have never needed when using solvers, nor heard anyone requesting — to my knowledge, resource limits are typically set once at the beginning and that's it — I would be interested if you have practical use cases for changing limits during execution (besides "just because"), but that discussion should be happening in a separate issue if it has to happen, not holding up the merge of this PR.

bclement-ocp: > Let's say that I want to support this feature then, the whole test is a use case (let's try something with a given limit, then with a greater limit).

Also note that this test uses `(reset)`, which will work and is easy to support. I am just asking for the `set-option :steps-bound` to be moved to the start of the file / immediately after the reset.

Stevendeo: > You are making a false dichotomy in your second message.
 
Sorry if I was not clear. I meant to say two separate things: that *I* did want this feature (because I'm tired of arguing for that, it is a *test*) ; and separately, that I didn't want `assert false` when setting an option between a push and a pop (which was a real issue when implementing, that this PR fixed). 

It is not even like push and pop had any effect on the matter, push and pop only affect the assertions. IMO, the simple fact that it triggered an assertion at some point may imply that we lack tests that push alt-ergo to its edge.

So, if you *really* want this line gone, can we at least agree on creating another test file that presses alt-ergo on strange cases? I would be fine keeping the line for this exact same reason.

bclement-ocp: > It is not even like push and pop had any effect on the matter, push and pop only affect the assertions. IMO, the simple fact that it triggered an assertion at some point may imply that we lack tests that push alt-ergo to its edge.

This is definitely 100% true, and somewhere where we have to improve! Not arguing with that at all.

> So, if you really want this line gone, can we at least agree on creating another test file that presses alt-ergo on strange cases? I would be fine keeping the line for this exact same reason.

(Note, I don't want that line gone — I want that line moved before the `push`)

I would be happy with having tests that explore corner cases and other oddities, that would be a good thing. What I am *not* happy with, however, is tests that test an unsupported feature (except to test that it fails with a proper error message, of course). Doing so enshrines the behavior, so that it effectively becomes supported: next time someone makes a change, the test will fail, and they will dutifully make sure that it still passes in the same way. That's the point of having tests, after all. I felt that I made clear that this was my stance, but I think this was partly in the meeting last yesterday and not in this discussion, so I apologize if there was confusion.

Now, here, given the current infrastructure and limitation regarding re-starting the solver after the step limit is reached, the feature set that we can safely commit to support in a coherent and satisfactory way is the use of `(set-option :step-bounds)` in `init` mode (i.e. before any `push`, `pop`, `assert`, etc.). Well, I guess that is not entirely true: we can support the use of `(set-option :step-bounds)` *as long as the step limit is not reached*, and then we can no longer allow increasing the step limit without additional infrastructure -- but that would make the validity of the command depend on internal Alt-Ergo details which is not good user experience.

Given that fact, and given the stance above, the tests should either be in that fragment, or if they are outside of that fragment they should either be errors or contain comments that say "hey we don't do the right thing here currently, it's OK if the test outcome change". But ideally such tests should be kept to a minimum, because it encourages a lot of churn in expected files which leads to paying less attention to changes in test outputs that can be wrong.

I am sorry to be arguing so much on this test, but hopefully my position makes sense here.

I think the options I would be happy with are:
 - Change the test so that the `(set-option)` are in init mode (i.e. before any `push`, `pop`, `declare`, `assert`, etc)
 - Keep the test as is, but change the code so that it *actually fails* instead of silently doing something (we already use an "are we in init mode" test for the "sat-solver" option, it can be reused from there)
 - Keep the test as is, but explicitly add a comment in the test file saying "this is unsupported but we want to make sure we don't crash".

Stevendeo: > Keep the test as is, but change the code so that it actually fails instead of silently doing something (we already use an "are we in init mode" test for the "sat-solver" option, it can be reused from there)

I am OK with this solution, but I think this belongs to another PR (that would slightly change the d_solver_options module).

Stevendeo: By the way, this test is expected to fail in a near future; the first assertion requires mores than 2 steps to be completed, but as it is the option *is* set at initialization because we do not treat SMT instructions imperatively. When we do, this test should fail.
I will add this fact to the test in case I am not assigned to the specific PR that will trigger the error.

bclement-ocp: Thank you, I am OK with this solution. I will approve the changes as soon as the info is added to the test case.

Can you point to the corresponding PR/issue related to making sure the test fails for tracking purposes, I am not sure which one it is.

Stevendeo: Maybe something linked to #899 ? The issue is definitely linked to what we discussed on tuesday about the imperative execution of smtlib instructions.

bclement-ocp (approved):

bclement-ocp:
(Waiting for the benches currently running on `next` to merge)

PR #942, opened 2023-11-16
Make get_value sound by not stating a formula is false when it is absent from the environment

Stevendeo:


bclement-ocp (approved):
Please do not go over 72 characters in the first line of commit messages, that makes them render poorly when using many tools including GitHub.

(Also I guess this is related to `get-assignment` rather than `get-value`?)

Halbaroth (approved):

PR #943, opened 2023-11-16
refactor(BV): Use Z.t in BV solver

bclement-ocp:
The old implementation of the solver relied on the fact that each constant fragment of a bit-vector was either all ones or all zeroes because it defined a constant zero as the smallest bit-vector, a constant one as the biggest bit-vector, and then computed the min and max elements of a set of bit-vectors to check for consistency. With the new solver based on Tarjan's union-find, this limitation is no longer necessary, and it causes the solver to needlessly split bit-vector variables involved in equalities with non-uniform constants [^1].

This patch is a simple refactoring of the bit-vector solver to use integers rather than booleans to represent the constant parts of bit-vectors, hopefully improving performance for bit-vectors with non-uniform constant parts.

[^1]: For instance, solving `x = #b0000` can be done in a single swoop, but solving `x = #b0101` currently first slices `x` into `a @ b @ c @ d` and then assigns values to each of `a`, `b`, `c` and `d`.

### DIFF
--- a/src/lib/reasoners/bitv.ml
+++ b/src/lib/reasoners/bitv.ml
@@ -69,8 +69,7 @@ let pp_sort ppf = function
 
     Note that [tvar] and [defn] values must not be created directly: instead,
     call the [s_var] and [s_cte] helpers. This is important due to the use of
-    physical equality in [union] (in particular, [union] assumes that there is a
-    single [tvar] for the boolean constants [true] and [false]). *)
+    physical equality in [union]. *)
 type tvar = { mutable defn : defn }
 
 and defn =
@@ -87,9 +86,9 @@ and defn =
       (** The sort of this variable. See the type definition for {!sort_var} for
           details. *)
     }
-  | Dcte of bool
-  (** A [Dcte] is a variable that is forced to be equal to either all ones or
-      all zeroes. *)
+  | Dcte of Z.t * int
+  (** A [Dcte] is a variable that is forced to be equal to the specified integer
+      representation in bits. The second argument is the bit width. *)
   | Dlink of tvar
   (** A [Dlink] is a defined variable. All the [Dlink] should be followed to
       arrive either at an unconstrained variable ([Droot]) or a constant
@@ -101,14 +100,15 @@ and defn =
 
 let rec pp_defn ppf = function
   | Droot { id; sorte; _ } -> Fmt.pf ppf "%a_%d" pp_sort sorte id
-  | Dcte b -> Fmt.pf ppf "%d" (if b then 1 else 0)
+  | Dcte (b, w) -> Fmt.pf ppf "%s" (Z.format ("%0" ^ string_of_int w ^ "b") b)
   | Dlink tv -> pp_tvar ppf tv
 
 and pp_tvar ppf { defn } = pp_defn ppf defn
 
 let equal_tvar v1 v2 =
   match v1.defn, v2.defn with
-  | (Droot _ | Dcte _), (Droot _ | Dcte _ ) -> v1 == v2
+  | Droot _, _ | _, Droot _ -> v1 == v2
+  | Dcte (n1, w1), Dcte (n2, w2) -> w1 = w2 && Z.equal n1 n2
   | _ ->
     (* [equal_tvar] should only be used before solving, i.e. before any unions
        are made, and so there should be no [Dlink]. *)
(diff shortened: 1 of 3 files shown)


src/lib/reasoners/bitv.ml
```diff
@@ -1451,11 +1443,17 @@ module Shostak(X : ALIEN) = struct
   let solve r1 r2 pb =
     Sig.{pb with sbt = List.rev_append (solve_bis r1 r2) pb.sbt}
 
-  (* Pop the first bit, raises [Not_found] if there is no first bit *)
+  (* Pop the first bit, raises [Not_found] if there is no first bit.
+
+     Note that the returned bv has an incorrect size. *)
```
hra687261, line 1462: How is the size incorrect?

src/lib/reasoners/bitv.ml
```diff
@@ -1451,11 +1443,17 @@ module Shostak(X : ALIEN) = struct
...
-    | { bv = Cte b; sz } as st :: rst ->
-      Some b, if sz > 1 then { st with sz = sz - 1 } :: rst else rst
+    | ({ bv = Cte n; sz } as st) :: rst ->
+      Some (Z.testbit n (sz - 1)),
+      if sz > 1 then
+        { st with sz = sz - 1 } :: rst
+      else
+        rst
     | { bv = Other _ | Ext _ ; sz } as st :: rst ->
       None, if sz > 1 then { st with sz = sz - 1 } :: rst else rst
```
hra687261, line 1473: Unrelated to the PR, but the comments in lines 1467-1668 should be fixed/reworded, with something like `_ are the values we must (differentiate|distinguish) from (those) that start* with a _`

hra687261 (approved):
LGTM

Halbaroth:
Please don't merge before I did my review ;)

src/lib/reasoners/bitv.ml
bclement-ocp: Because the underlying value is not resized, the size is just changed (for instance we could have `0b1111` with `size = 1` which is "wrong" in other parts of the code). So the high bits must be ignored, which `pop_bit` does.

src/lib/reasoners/bitv.ml
bclement-ocp: Can you clarify what should be fixed? The values in `[t]` all start with a `1`, and the model value must be distinct from all these values; the values in `[f]` all start with a `0` and the model value must also be distinct from all these values. I wouldn't understand that from your reformulation.

hra687261: It's just the wording, the current comments say:
```
    (* [t] are the values we must be distinct from that starts with a '1' *)
    (* [f] are the values we must be distinct from that starts with a '0' *)
```
It's true that what I suggested doesn't help. But I don't get your explanation from the current comments either. 
Putting the explanation you gave as a comment instead of the two comments that are currently present would be better IMO.

bclement-ocp: Right. Would something like "[t] holds the values that (a) we must be distinct from and (b) start with a `1`" be clearer? I will make a PR once this one is merged to avoid conflicts.

hra687261: Yes, it would be clearer.

Halbaroth:
Sorry, I don't know why but my message doesn't appear ;/. I resend it: 
You're PR is great but could you clarify this point please:
>  it defined a constant zero as the smallest bit-vector, a constant one as the biggest bit-vector, and then computed the min and max elements.

Not sure to understand what you mean by min and max elements here and I guess it's important to understand that as you assert this is not the way we check consistency anymore.

Halbaroth (approved):

bclement-ocp:
> Not sure to understand what you mean by min and max elements here and I guess it's important to understand that as you assert this is not the way we check consistency anymore.

The old BV solver (removed as part of #856) created one set for each equivalence class, and stuffed all elements that were equal into the same class. It then needed to check:
 - If there is one constant term in the class, then all the terms in the class must be equal to that constant
 - If there are multiple distinct constant terms in the class, then the problem is unsolvable

It did so by imposing an order on simple terms such that `Cte false < everything < Cte true`, and then checking only for the minimum and maximum elements of the set rather than iterating over all possible elements of the set. This only works because there are exactly two possible constants (`false` and `true`).

The new solver that uses Tarjan's union-find introduced in #856 does not have such a limitation. Instead, the solver always selects constants as the roots of their equivalence classes, and fails if we ever try to merge two equivalence classes with distinct constant roots, which works with an arbitrary number of possible constants.

bclement-ocp:
Note: I want to do a sanity performance check before merging, which is currently blocked by #954

Halbaroth:
Is it ok?

bclement-ocp:
Benches are running.

bclement-ocp:
This is +5-0 on QF_BV with around 3% speedup, merging.

(I did not run benches on BV with quantifiers)

Issue #903, opened 2023-10-18
Improving reasoning for the BV theory


Halbaroth:
This umbrella issue tracks the progression of the BV theory reasoning. Please add related issue here and what we plan to improve in the future.

- [X] Support of `bvnot` reasoning, see #800 and #856.
- [X] Constraint propagation for `bvand`, `bvor`, `bvxor`, etc (see #944)
- [ ] Constraint simplification (from `(bvand (concat a b) (concat c d))` to `(concat (bvand a c) (bvand b d))`)
- [x] Constraint propagation for `bv2nat` and `int2bv`

bclement-ocp:
Some updates here:

 - Doing constraint propagation on `bv2nat` and `int2bv` is not easy to do properly, because we don't have access to the arithmetic domains from `Bitv_rel`
 - This is even more true if we want to do interreductions between arithmetic and bitwise domains
 - The plan is now to instead add separate interval domains for bitvectors to `Bitv_rel` and to perform interreductions following [Sharpening Constraint Programming approaches for
Bit-Vector Theory](https://cea.hal.science/cea-01795779/document)
 - #1040 #1044 #1054 #1055 #1056 do some groundwork to prepare the addition of the new domains, including reworking some of the constraint APIs to be easier to use with multiple domains
 - Additional PRs are currently being tested that actually add the interval domains & arithmetic propagators following the paper above
 - At some point we might want to figure out how to reconcile the BV and Arith interval domains
 - No progress on the "constraint simplification" part yet

Halbaroth:
I think we can remove the milestone on this issue because we do not plan extra works on it before the next release.

bclement-ocp:
Yes, we won't have constraint simplification for 2.6. Let's keep it to track `bv2nat(bvadd)` support.

bclement-ocp:
I wanted to implement support for `bv2nat` applied to `bvadd` etc. operators but did not find the time — or rather, I have several tentative implementations that need to be rebased and compared to figure out which one to use, which I won't have time to do for this release.

We already have many bv improvements, and these changes only add a small additional improvement for the bv2nat problems from our partners (~ 0.2% more problems solved) so it will have to wait for the next release. I'll try to cut a point release once it is integrated.

bclement-ocp:
I said this but then realised we are not even able to prove that `(int2bv (+ (bv2nat x) (bv2nat y))` and `(bvadd x y)` are equal currently, so that's annoying. I will make a small patch for that.

[...]

PR #944, opened 2023-11-16
feat(BV): Word-level propagation for BV logic

bclement-ocp:
**This PR currently includes #943 **

This patch implements constraint propagation for the logical operators (and, or, xor) on bit-vectors.

The goal of this patch is to ensure that Alt-Ergo has a complete understanding of the logical operators on bit-vectors, and in particular that it will never violate the definition of the logical operators when generating models. They are thus removed from the "suspicious" operators (along with `int2bv` and `bv2nat` that are now handled through the "delayed functions" mechanism, even though this patch does not impact these functions).

The patch can be decomposed in changes of the following categories, where each category can mostly be reviewed independently (all in the same patch because they wouldn't really make sense on their own):

 - Some minor refactoring in `Bitv`, `Th_util`, `Expr`, `ModelMap` and `Symbols` to expose the required functions and types, including a `Th_bitv` constructor for bit-vector case-splits and a built-in symbol for `bvxor`

 - The new `Bitlist` module introduces a representation of bit-vector domains as bit-lists, i.e. pairs of `(ones, zeroes)` represented as integers where the bits forced to `1` are set in `ones`, and the bits forced to `0` are set in `zeroes`.

 - There is a new `Congruence` module in `Rel_utils` that is exploited by the BV constraint propagators. The `Congruence` module can be understood as a lighter version of `Use` for semantic values (rather than terms): it provides a mechanism to register "semantic values of interest" (in our case, it will be the arguments of the logical operators and the result of the logical operator themselves) and be notified when the representative of these semantic values of interest are updated. To avoid partially inconsistent states, the congruence module computes the leaves the each semantic values of interest and updates all the semantic values that have a given leaf at once.

   This `Congruence` module is not BV-specific, hence why it is in `Rel_utils`.

 - Finally, the main course is in the `Bitv_rel` module, where the new constraint propagators are added. The constraint propagators are represented using separate `Domains` and `Constraints` modules that are kept normalized (i.e. they only reference current class representatives in `Uf`) through the `Congruence` utility.

   The `Domains` module is simply a map from class representatives to their domain, represented as a `Bitlist`. To avoid unnecessary propagations, the `Domains` module keeps track of the values whose domain has changed since the last propagation, to know that we must propagate the associated constraints again.

   The `Constraints` module is a set of `Constraint`s that keeps track of the semantic values that are involved in each constraint to avoid unnecessary propagations. The `Constraint`s are normalized (so that if we have both `a = b & c` and `a = b & d`, we only keep one once we learn that `c = d`), and we also keep track of the constraints that were never propagated.

   The `Bitv_rel` environment then keeps track of a `Domains` and a `Constraints` set that are updated through `Congruence` whenever a class representative changes (detected through equalities with `Subst` origin). After each change to either the domains or the constraints (either through a class representative change or because we added a new constraint), we propagate the fresh constraints and the constraints involving values whose domain has changed. We maintain the invariant that, outside of the `add` and `assume` functions, all the constraints in the `Constraints` set are satisfied by the domains of the `Domains`.

In the interest of keeping the patch relatively short and facilitate its review, many desirable features have been left out for now, and there are many places where the implementation can be improved. Notably:

 - We do not perform any sort of constraint simplification (outside of the normalization used for de-duplication). This can cause backtracking inefficiencies: if we have a constraint involving the concatenation `a @ b`, and we substitute `b -> c`, then detect a conflict involving `a`, the explanation for the substitution `b -> c` will end up in the conflict even though it had no impact, causing more work than necessary. This can be added in a second step.

 - The constraints are represented in a relational way rather than a functional way. This means that if we have both `a = b & c` and `d = b & c` as constraints, we will not learn that `a = d`. I don't think this has any impact currently, because the only way we can learn a constraint `a = b & c` is if `a` is equal to `b & c` as a term, but could have one in the future if we want to perform simplifications.

 - Only logical operators are supported. In particular, there is no support for combined reasoning with arithmetic, which is a whole can of worms in itself (in the current Alt-Ergo architecture, we do not have access to the interval domains from the bit-vector theory and conversely).

Finally, this patch purposely does not implement any sort of algebraic reasoning on bit-vectors [^1]: while with this patch Alt-Ergo is able to prove theorems such as De Morgan's laws or algebraic properties (associativity, commutativity, neutral elements, inverses, etc.) of the logical operators, this will currently happen through painstaking case-splitting which gets slow on large bit-vector sizes [^2].

[^1]: The equation `x ^ x = 0` happens to be known as a side-effect of the representation of XOR constraints.

[^2]: The value of "large" here is actually fairly small, because proving these laws through propagation often requires exploring the full space, which is exponential.


bclement-ocp:
Note: running benchmarks against `next` require #954 to be merged.

src/lib/reasoners/rel_utils.ml
```diff
@@ -192,3 +192,116 @@ end = struct
...
 end
+
+module Congruence : sig
+  (** The [Congruence] module implements a simil-congruence closure algorithm on
+      semantic values.
+
+      It provides an interface to register some semantic values of interest, and
+      for applying a callback when the representative of those registered values
+      change.
+  *)
```
Halbaroth, line 203: A general question about the necessity of this module. During the last meeting, you suggest that we use expression too deeply in the solver (for instance we keep using expression in SatML for the representation of literals). If we stop doing this and use as soon as possible semantic values, could we use the CC(X) algorithm here?

src/lib/reasoners/rel_utils.ml
```diff
@@ -192,3 +192,116 @@ end = struct
...
+
+      [f] is intended to perform a substitution operation on the type ['a],
+      merging the values associated with [rr] into the values associated with
+      [nrr]. *)
+end = struct
+  module SX = Shostak.SXH
+  module MX = Shostak.MXH
+
+  type t = { parents : SX.t MX.t ; registered : SX.t }
+
```
Halbaroth: Could you add a bit of documentation for `parents`? We call recall it's a similar structure to the `Use` one.

src/lib/reasoners/rel_utils.ml
```diff
@@ -192,3 +192,116 @@ end = struct
...
+  let remove r t =
+    if SX.mem r t.registered then
+      let parents =
+        List.fold_left (fun parents leaf ->
+            MX.update leaf (function
+                | Some deps ->
+                  let deps = SX.remove r deps in
+                  if SX.is_empty deps then None else Some deps
+                | None -> assert false
```
Halbaroth: We may add a comment for this `assert` to explain this cannot occurs as `r` has to be add to the set of terms that use `leaf` in the function `add` above.

src/lib/reasoners/rel_utils.ml
```diff
@@ -192,3 +192,116 @@ end = struct
...
+              ) parents
+          ) t.parents (X.leaves r)
+      in
+      { parents ; registered = SX.remove r t.registered }
+    else
+      invalid_arg "Congruence.remove"
+
+  let subst rr nrr cgr f t =
+    match MX.find rr cgr.parents with
```
Halbaroth, line 288: Is it better to catch the exception of `MX.find` or use `MX.find_opt`? When we don't perform anything in the `Some` branch, I understand we prefer the exception approach rather the optional one, for instance:
```ocaml 
let x = try find u map with Not_found -> (* error handler *) in
```
is a little bit shorter than
```ocaml
let x = match find_opt u map with | Some x -> x | None -> (* error handler *) in
``` 
But here, both looks similar.

src/lib/reasoners/bitlist.mli
```diff
@@ -0,0 +1,126 @@
...
+val value : t -> Z.t
+(** [value b] returns the value associated with the bitlist [b]. If the bitlist
+    [b] is not fully known, then only the known bits (those that are set in
+    [bits_known b]) are meaningful; unknown bits are set to [0]. *)
+
+val subsumes : t -> t -> bool
+(** [subsumes b1 b2] returns [true] if [b1] has more information than [b2].
+
+    More precisely, it checks that all the bits known by [b2] are also known by
+    [b1]; it *does not* check that [b1] and [b2] are consistent. *)
```
Halbaroth: `subsumes` is misleading as the function doesn't check the consistency. When I read `subsumes`, I expect one of the two arguments implies the other one.

src/lib/reasoners/bitlist.ml
```diff
@@ -0,0 +1,158 @@
...
+module Ex = Explanation
+
+exception Inconsistent of Ex.t
+
+(** A bitlist representing the known bits of a bit-vector of width [width].
+
+    Active bits in [bits_set] are necessarily equal to [1].
+    Active bits in [bits_clr] are necessarily equal to [0].
+
+    The explanation [ex] justifies that the bitlist applies. *)
```
Halbaroth, line 26: I'm not sure to understand what you mean by `active`? In my understanding, `bits_set` is the set of bit positions constraint to be `1` and `bits_clr`  is the set of bit positions constraint to be `0`.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,15 +68,730 @@ let dispatch = function
...
+
+  let update ?(init = false) ex r t bl =
+    let changed = ref false in
+    let bitlists =
+      MX.update r (function
+          | Some bl' as o ->
+            (* Need to always call [intersect] to check for consistency *)
+            let bl'' = Bitlist.intersect bl bl' ex in
+            if Bitlist.subsumes bl' bl'' then
+              o
```
Halbaroth: We do this because we don't want to update the domain of `r` with `bl''` if the previous domain `bl'` is finer. As `bl''` is the intersection of `bl` and `bl'`, this can occur only if `bl'' = bl' /\ bl` where `/\ = intersection`.

BUT we manipulate **explained** domains. If `bl'' = bl' /\ bl` as unexplained domains, as explained ones, `bl''` could contain more explanations. So I guess a simpler and correct code would be:
```ocaml 
if Bitlist.equal bl' bl'' then o 
else ...
```
Besides, we should explain that we do this test to avoid flooding the explanation set with useless explanations.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,15 +68,730 @@ let dispatch = function
...
+    match MX.find rr t.bitlists with
+    | bl ->
+      let has_changed = ref (SX.mem rr t.changed) in
+      let bitlists =
+        MX.update nrr (function
+            | Some bl' as o ->
+              (* Need to always call [intersect] to check for consistency *)
+              let bl'' = Bitlist.intersect bl bl' ex in
+              if Bitlist.subsumes bl' bl'' then
+                o
```
Halbaroth: Same remark as above.

We don't need `Bitlist.subsumes`.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,15 +68,730 @@ let dispatch = function
...
+        )
+      ppf t.bitlists
+  let empty = { bitlists = MX.empty ; changed = SX.empty }
+
+  let update ?(init = false) ex r t bl =
+    let changed = ref false in
+    let bitlists =
+      MX.update r (function
+          | Some bl' as o ->
+            (* Need to always call [intersect] to check for consistency *)
```
Halbaroth: I'm not sure this comment is still needed if we remove `Bitlist.subsumes`. We perform `intersect` here because it's the purpose of the function `update`.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,15 +68,730 @@ let dispatch = function
...
+  val empty : t
+  (** Returns an empty domain map. *)
+
+  val update : ?init:bool -> Ex.t -> X.r -> t -> Bitlist.t -> t
+  (** [update ex r d bl] intersects the domain of [r] with bitlist [bl].
+
+      There must already be a domain associated with [r] before calling
+      [update], unless [init] is set to true.
+
+      The explanation [ex] justifies that [bl] applies to [r]. *)
```
Halbaroth: This function may raise `Inconsistent` exception. Its documentation should mention it.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,15 +68,730 @@ let dispatch = function
...
+  val subst : Ex.t -> X.r -> X.r -> t -> t
+  (** [subst ex p v d] replaces all the instances of [p] with [v] in any domain,
+      and merges the corresponding bitlists.
+
+      Use this to ensure that the representation is always normalized.
+
+      [v] must have an existing associated domain (possibly unconstrained)
+      before calling [subst]. Call [update ~init:true] beforehand.
+
+      The explanation [ex] justifies the equality [p = v]. *)
```
Halbaroth: Same here, we should mention `Inconsistent`.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,15 +68,730 @@ let dispatch = function
...
+
+      There must already be a domain associated with [r] before calling
+      [update], unless [init] is set to true.
+
+      The explanation [ex] justifies that [bl] applies to [r]. *)
+
+  val get : X.r -> t -> Bitlist.t
+  (** [get r d] returns the bitlist currently associated with [r] in [d].
+
+      Raises [Not_found] if there is no bitlist associated with [r] in [d]. *)
```
Halbaroth: Nit: `@raises [Not_found] ...`

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,15 +68,730 @@ let dispatch = function
...
+  let get r t = MX.find r t.bitlists
+
+  let subst ex rr nrr t =
+    match MX.find rr t.bitlists with
+    | bl ->
+      let has_changed = ref (SX.mem rr t.changed) in
+      let bitlists =
+        MX.update nrr (function
+            | Some bl' as o ->
+              (* Need to always call [intersect] to check for consistency *)
```
Halbaroth: Same remark, this comment isn't pertinent anymore if we remove `Bitlist.subsumes` below.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,15 +68,730 @@ let dispatch = function
...
+end
+
+module Constraint : sig
+  type repr =
+    | Band of X.r * X.r * X.r
+    (** [Band (x, y, z)] represents [x = y & z] *)
+    | Bor of X.r * X.r * X.r
+    (** [Bor (x, y, z)] represents [x = y | z] *)
+    | Bxor of SX.t
+    (** [Bxor {x₁, ..., xₙ}] represents [x₁ ^ ... ^ xₙ = 0] *)
```
Halbaroth: I'm not sure it's a good idea to use `UTF-8` symbols in the documentation.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,15 +68,730 @@ let dispatch = function
...
+      );
+      tagged
+
+  let tag { tag; _ } = tag
+
+  let pp_repr ppf = function
+    | Band (x, y, z) ->
+      Format.fprintf ppf "%a & %a = %a" X.print y X.print z X.print x
+    | Bor (x, y, z) ->
+      Format.fprintf ppf "%a | %a = %a" X.print y X.print z X.print x
```
Halbaroth: Please use `Fmt` :)

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,15 +68,730 @@ let dispatch = function
...
+    | Bxor xs ->
+      Fmt.(iter ~sep:(any " ^@ ") SX.iter X.print |> box) ppf xs;
+      Fmt.pf ppf " = 0"
+    | Bnot (x, y) ->
+      Format.fprintf ppf "%a = ~%a" X.print x X.print y
```
Halbaroth: Same here :)

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,15 +68,730 @@ let dispatch = function
...
+      [ Uf.LX.mkv_eq lhs one, false, Th_util.CS (Th_util.Th_bitv, Q.one) ]
+    ) else (
+      if Options.get_debug_bitv () then
+        Printer.print_dbg
+          ~module_name:"Bitv_rel" ~function_name:"case_split"
+          "[BV-CS-1] Setting %a to 0" X.print lhs;
+      [ Uf.LX.mkv_eq lhs zero, true, Th_util.CS (Th_util.Th_bitv, Q.of_int 2) ]
+    )
+  | exception Not_found -> []
 let add env uf r t =
```
Halbaroth, line 810: Please add an extra space before this line.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,15 +68,730 @@ let dispatch = function
...
+  val fold_fresh : (Constraint.t -> 'a -> 'a) -> t -> 'a -> t * 'a
+  (** [fold_fresh f cs acc] folds [f] over the fresh constraints in [cs].
+
+      Fresh constraints are constraints that were never propagated yet. *)
+
+  val propagate : t -> X.r -> Domains.t -> Domains.t
+  (** [propagate cs r dom] propagates the constraints associated with [r] in the
+      constraint set [cs] and returns the new domain map after propagation. *)
+end = struct
+  module IM = Map.Make(Int)
```
Halbaroth: Please use `Util.MI`.

bclement-ocp:
Looks like there is unsoundness in some benches — marking as draft until I can dig deeper.

Halbaroth:
Can I submit my review anyway? My remarks are mostly about documentation and some questions.

bclement-ocp:
> Can I submit my review anyway? My remarks are mostly about documentation and some questions.

Yes sure — I have just marked it as draft to make sure it is not merged.

I also have identified the soundness issue (a misunderstanding on my part in the meaning of the `is_cs` return value of `case_split`); a patch is on the way.

bclement-ocp:
The soundness issue is solved through a combination of [Handle case-split conflicts properly](https://github.com/OCamlPro/alt-ergo/pull/944/commits/f9bc3cf43105b1a05363bcba8d57ccc0f7384872) and https://github.com/OCamlPro/alt-ergo/pull/968

Waiting on https://github.com/OCamlPro/alt-ergo/pull/968 and https://github.com/OCamlPro/alt-ergo/pull/969 to be merged to run benches again

Note that there are cases where we used to be able to solve but we no longer are able to solve due to (I believe) the coarse explanations returned by constraint propagation that causes Alt-Ergo to perform exhaustive enumeration rather than actually performing the higher-level reasoning steps that it is able to do without this patch. These should be at least partially fixed by performing constraint simplifications, which was explicitly left out of this PR for now, as explained in the description.

Halbaroth (changes requested):

src/lib/reasoners/bitlist.ml
bclement-ocp: "Active" means "set" or "enabled" here. I will change to state "Bits that are `1` in [bits_set]" to clarify. Note, `bits_set` and `bits_clr` are *bitsets* (i.e. `0b101` represents the set `{0, 2}`).

src/lib/reasoners/bitlist.mli
bclement-ocp: It initially did check that when I called it that way, then it no longer did. Actually this is only used in situations where `equal` can be used instead, so I will just use that and remove `subsumes`.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: `Bitlist.equal` does *not* compare explanations (this is the same for other domains, see e.g. `Intervals.equal`). The test is also to avoid looping by setting `changed` when there were no actual changes, clarified that in a comment.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: This comment was here because `subsumes` did not check for consistency :)

src/lib/reasoners/bitv_rel.ml
bclement-ocp: We are in 2023! And better than to use them in code ;) But fine.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Yeah now that we have actual implementations here no need to keep everything so tightly together, done.

src/lib/reasoners/rel_utils.ml
bclement-ocp: I am not sure it is related to the use of expressions in satML, but yes, ideally CC(X) would have a built-in way of keeping track of this information rather than having each relation re-implement the bits it needs as we currently do. But AE currently doesn't really provide the required infrastructure (I thought about using it, but I believe that would require some non-trivial modifications that would be a full PR on their own, so I decided against it), so at least we can try to share the infrastructure.

src/lib/reasoners/rel_utils.ml
bclement-ocp: Yeah, but `Use` is on terms, and this is on semantic values. Added a comment.

src/lib/reasoners/rel_utils.ml
bclement-ocp: Done.

src/lib/reasoners/rel_utils.ml
bclement-ocp: I tend to prefer using `find` rather than `find_opt` to avoid allocating an `option` (sometimes the compiler is able to elide the allocation with `find_opt`, sometimes not), no other reason.

src/lib/reasoners/bitv_rel.ml

src/lib/reasoners/bitv_rel.ml

src/lib/reasoners/bitv_rel.ml

src/lib/reasoners/bitv_rel.ml

src/lib/reasoners/bitv_rel.ml

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Done (well, `@raise`).

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Good catch, done.

src/lib/reasoners/bitv_rel.ml

src/lib/reasoners/bitv_rel.ml
Halbaroth: Actually, both are correct with the `odoc` syntax which is an extension of the syntax of `ocaml-doc`. Let's keep the universal one.

bclement-ocp:
Test failure is slow f1 again.

bclement-ocp:
This was +876-692 (delta of +184 which is nice) on QF_BV against #943 but the benchmarks lead to the detection of #978 and #979 

Among the 692 regressions there is a certain number of cases where we time out due to being too aggressive in terms of splits; I have implemented support for `Options.get_max_split ()` in the BV domain which seems to help on the few cases I have manually re-run. Rerunning benchmarks on QF_BV with these changes against `next`, with #943 and the fixes for #978 and #979.

Edit: I know I messed up the style, I will fix it once the benches are done.

bclement-ocp:
This is now +879-95 on QF_BV (diff of +784, or +10% compared to what we already solve).

Based on a small sample of the 95 regressions, at least a few seem to be caused by improper splitting, which we can improve on later.

I am going to run benches on BV (not QF) with hopefully similar results; but unless something catastrophic happens there I don't expect much changes, this is ready for round 2 of reviews.

bclement-ocp:
This is +133-2 on BV (around +3%)

src/lib/reasoners/rel_utils.ml
```diff
@@ -192,3 +192,127 @@ end = struct
...
+
+  val remove : X.r -> t -> t
+  (** [remove r t] unregisters the semantic value [r] from the congruence.
+
+      Note that if substitutions have been applied to the congruence after a
+      value has been added, those same substitutions must be applied to the
+      semantic value prior to calling [remove], or [Invalid_argument] will be
+      raised.
+
+      Raised [Invalid_argument] if [r] is not a registered semantic value. *)
```
hra687261:
```suggestion
      @raise [Invalid_argument] if [r] is not a registered semantic value. *)
```

src/lib/reasoners/bitv.mli
```diff
@@ -28,7 +28,31 @@
...
+type 'a simple_term_aux =
+  | Cte of Z.t
+  | Other of 'a signed
+  | Ext of 'a signed * int * int * int (*// id * size * i * j //*)
+
+type 'a simple_term = ('a simple_term_aux) alpha_term
+
+type 'a abstract = 'a simple_term list
+
+(** [extract size i j x] extracts [i..=j] from a composition of size [size].
```
hra687261:
```suggestion
(** [extract size i j x] extracts [i..j] from a composition of size [size].
```

src/lib/reasoners/bitv.ml
```diff
@@ -301,6 +301,35 @@ let fresh_bitv genre sz : bitv =
...
+  if sz = j - i + 1 then
+    [ { bv ; sz } ]
+  else
+    match bv with
+    | Cte b -> [{ bv = Cte (Z.extract b i (j - i + 1)); sz = j - i + 1 }]
+    | Other r -> [{ bv = Ext (r, sz, i, j) ; sz = j - i + 1 }]
+    | Ext (r, sz, k, _) ->
+      [{ bv = Ext (r, sz, i + k, j + k) ; sz = j - i + 1 }]
+
+(* extract i..j from a composition of size size
```
```suggestion
(* extract [i..j] from a composition of size [size]
```

src/lib/reasoners/bitv.ml
```diff
@@ -301,6 +301,35 @@ let fresh_bitv genre sz : bitv =
...
+
+    an element of size [sz] at the head of the composition contains the bits
+    [size - 1 .. size - sz] inclusive *)
+let rec extract size i j = function
+  | [] ->
+    (* We can't extract a bv of length 0! *)
```
```suggestion
    (* We can't extract from a bv of length 0! *)
```

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,17 +68,768 @@ let dispatch = function
...
+
+let is_bv_r r = is_bv_ty @@ X.type_info r
+
+module SX = Shostak.SXH
+module MX = Shostak.MXH
+
+module Domains : sig
+  type t
+  (** The type of domain maps. A domain map maps each representatives (semantic
+      values, of type [X.r]) to their associated domains.
```
hra687261:
```suggestion
  (** The type of domain maps. A domain map maps each representative (semantic
      value, of type [X.r]) to its associated domain.
```
Or
```suggestion
  (** The type of domain maps. A domain map maps representatives (semantic
      values, of type [X.r]) to their associated domains.
```

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,17 +68,768 @@ let dispatch = function
...
+
+      The keys of the domain maps are expected to be current *class
+      representatives*, i.e.  normal forms wrt the `Uf` module, in which case we
+      say the domain map is *normalized*. Use `subst` to ensure that domain maps
+      stay normalized. *)
```
hra687261:
```suggestion
      The keys of the domain maps are expected to be current *class
      representatives*, i.e.  normal forms wrt to the `Uf` module, in which case we
      say the domain map is *normalized*. Use `subst` to ensure that domain maps
      stay normalized. *)
```

src/lib/reasoners/bitlist.mli
```diff
@@ -0,0 +1,120 @@
...
+
+val bits_known : t -> Z.t
+(** [bits_known b] returns the sets of bits known to be either [1] or [0] as a
+    bitmask. *)
+
+val num_unknown : t -> int
+(** [num_unknown b] returns the number of bits unknown in [b]. *)
+
+val is_fully_known : t -> bool
+(** [is_fully_known b] determines there are unknown bits in [b] or not.
```
hra687261:
```suggestion
(** [is_fully_known b] determines if there are unknown bits in [b] or not.
```

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,17 +68,768 @@ let dispatch = function
...
+  module CS = Set.Make(struct
+      type t = Constraint.t
+
+      let compare t1 t2 =
+        Int.compare Constraint.(tag t1.repr) Constraint.(tag t2.repr)
+    end)
+
+  type t = {
+    cs_set : CS.t ;
+    (*** All the contraints currently active *)
```
hra687261:
```suggestion
    (*** All the constraints currently active *)
```

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,17 +68,768 @@ let dispatch = function
...
+        aux x x_sz (lits @ acc) (Bitlist.extract bl 0 (nbits - 1))
+  in
+  fun acc x bl ->
+    aux x (Bitlist.width bl) acc bl
+
+(* Propagate:
+
+   - The constraints that were never propagated since they were added (this
+      includes constraints that changed due to substitutions)
+   - The contraints involving variables whose domain changed since the last
```
hra687261:
```suggestion
   - The constraints involving variables whose domain changed since the last
```

src/lib/reasoners/bitv_rel.ml
```diff
@@ -63,17 +68,768 @@ let dispatch = function
...
+          la
+      in
+      let eqs, domain = propagate constraints domain in
+      if Options.get_debug_bitv () && not (Lists.is_empty eqs) then (
+        Printer.print_dbg
+          ~module_name:"Bitv_rel" ~function_name:"assume"
+          "bitlist domain: @[%a@]" Domains.pp domain;
+        Printer.print_dbg
+          ~module_name:"Bitv_rel" ~function_name:"assume"
+          "bitlist contraints: @[%a@]" Constraints.pp constraints;
```
hra687261:
```suggestion
          "bitlist constraints: @[%a@]" Constraints.pp constraints;
```

hra687261 (approved):
LGTM modulo some minor doc/comment fixes.

bclement-ocp:
@Halbaroth you requested changes in your previous review; do you want to take a second look before I merge?

bclement-ocp:
Oh no the suggestions broke the style check again  :cry:

bclement-ocp:
#903

Halbaroth (approved):

Halbaroth:
I'm ready to merge as soon as the PR is rebased :)

PR #950, opened 2023-11-17
Temporarily disable changing Timeout to I_dont_know in satML

bclement-ocp:
This is a hotfix while we figure out a proper way forward for https://github.com/OCamlPro/alt-ergo/issues/946 so that Alt-Ergo doesn't randomly crash due to timeouts in the meantime.


bclement-ocp:
@Halbaroth @Stevendeo could I get a quick review on this one? Would be great to not have spurious crashes when running benchmarks on `next` (it fills up the sqlite database with garbage).

Longer term the solution is probably along the lines of #936 but for timeouts.

Halbaroth (approved):

PR #952, opened 2023-11-20
Add a SMT-LIB compliant printer for `Expr`

Halbaroth:
We plan to use expression from `Expr` to store model values but we haven't appropriate printer for Symbols in the SMT-LIB format. This commit add an new printer `pp_smtlib` into `Symbols` which prints (as much as possible) symbols using the SMT-LIB standard.

This new printer should be sufficient for printing models as model values couldn't be nonconstant values or even applications of function. Still the printer `pp_smtlib` tries to be as complete as possible and outputs a correct answer for symbols of functions or literals.

I also clean up a bit the AE format printer.

### DIFF
--- a/src/lib/frontend/models.ml
+++ b/src/lib/frontend/models.ml
@@ -312,10 +312,19 @@ module SmtlibCounterExample = struct
     | [Ty.Trecord _r, _arg] -> begin
         match xs_values with
         | [record_name,_] ->
+          (* HOTFIX: this fix is temporary. The current implementation of
+             model generation for records relies on string representants,
+             which means the printer for access symbols has to agree with
+             the name of the field in the type trecord. As the printer
+             [Symbols.print] will always output AE native format, this
+             doesn't agree when the output format is SMT-LIB. But the
+             printer of expression will output the right string if we don't
+             give the arguments of the field. *)
+          let access = Fmt.str "%a" Expr.print (Expr.mk_term f [] ty) in
           add_records_destr
             records
             (asprintf "%a" Expr.print record_name)
-            (Sy.to_string f)
+            access
             rep
         | [] | _ -> records
       end
--- a/src/lib/frontend/models.ml
+++ b/src/lib/frontend/models.ml
@@ -319,7 +319,9 @@ module SmtlibCounterExample = struct
              [Symbols.print] will always output AE native format, this
              doesn't agree when the output format is SMT-LIB. But the
              printer of expression will output the right string if we don't
-             give the arguments of the field. *)
+             give the arguments of the field.
+
+             Issue: https://github.com/OCamlPro/alt-ergo/issues/958 *)
           let access = Fmt.str "%a" Expr.print (Expr.mk_term f [] ty) in
           add_records_destr
             records
--- a/src/lib/structures/expr.ml
+++ b/src/lib/structures/expr.ml
@@ -311,393 +311,374 @@ module F_Htbl : Hashtbl.S with type key = t =
 
 (** pretty printing *)
 
-let print_binders =
-  let print_one fmt (sy, (ty, _)) =
-    Format.fprintf fmt "%a:%a" Sy.print sy Ty.print ty
-  in fun fmt b ->
-    match SMap.bindings b with
-    | [] ->
-      (* can happen when quantifying only on type variables *)
-      Format.fprintf fmt "(no term variables)"
-    | e::l ->
-      print_one fmt e;
-      List.iter (fun e -> Format.fprintf fmt ", %a" print_one e) l
-
-(* let print_list_sep sep pp fmt =
- *   Format.pp_print_list ~pp_sep:(fun fmt _ -> Format.fprintf fmt sep) pp fmt
- *
- * let print_list pp fmt = print_list_sep "," pp fmt *)
-
 module SmtPrinter = struct
+  let pp_binder ppf (var, (ty, _)) =
+    let var =
+      match var with
+      | Sy.Var v -> v
+      | _ -> assert false
+    in
+    Fmt.pf ppf "(%a %a)" Var.print var Ty.pp_smtlib ty
 
-  let rec print_formula fmt form xs bind =
-    let open Format in
+  let pp_binders = Fmt.(iter_bindings ~sep:sp SMap.iter pp_binder)
+
+  let rec pp_formula ppf form xs bind =
     match form, xs, bind with
     | Sy.F_Unit _, [f1; f2], _ ->
-      Format.fprintf fmt "@[(and %a %a)@]" print_silent f1 print_silent f2
+      Fmt.pf ppf "@[(and %a %a)@]" pp_silent f1 pp_silent f2
 
     | Sy.F_Iff, [f1; f2], _ ->
-      fprintf fmt "@[(= %a %a)@]" print_silent f1 print_silent f2
+      Fmt.pf ppf "@[(= %a %a)@]" pp_silent f1 pp_silent f2
 
     | Sy.F_Xor, [f1; f2], _ ->
-      fprintf fmt "@[(not (= %a %a))@]" print_silent f1 print_silent f2
+      Fmt.pf ppf "@[(xor %a %a)@]" pp_silent f1 pp_silent f2
 
     | Sy.F_Clause _, [f1; f2], _ ->
-      fprintf fmt "@[(or %a %a)@]" print_silent f1 print_silent f2
+      Fmt.pf ppf "@[(or %a %a)@]" pp_silent f1 pp_silent f2
 
-    | Sy.F_Lemma, [], B_lemma { user_trs ; main ; name ; binders; _ } ->
-      if Options.get_verbose () then
-        fprintf fmt "(lemma: %s forall %a[%a].@  %a)"
-          name
-          print_binders binders
-          print_triggers user_trs
-          print_silent main
-      else
-        fprintf fmt "(lem %s)" name
+    | Sy.F_Lemma, [], B_lemma { main; name; binders; _ } ->
+      Fmt.pf ppf "(! (forall (%a) %a) :named %s)"
+        pp_binders binders pp_silent main name
 
-    | Sy.F_Skolem, [], B_skolem { main; binders; _ } ->
-      fprintf fmt "(<sko exists %a.> %a)"
-        print_binders binders print_silent main
+    | Sy.F_Skolem, [], B_skolem { main; name; binders; _ } ->
+      Fmt.pf ppf "(! (exists (%a) %a) :named %s)"
+        pp_binders binders pp_silent main name
 
     | _ -> assert false
 
-  and print_lit fmt lit xs =
-    let open Format in
+  and pp_lit ppf lit xs =
     match lit, xs with
     | Sy.L_eq, a::l ->
-      fprintf fmt "(= %a%a)"
-        print a (fun fmt -> List.iter (fprintf fmt " %a" print)) l
-    | Sy.L_neg_eq, [a; b] ->
-      fprintf fmt "(not (= %a %a))" print a print b
+      Fmt.pf ppf "(= %a %a)"
+        pp a (fun ppf -> List.iter (Fmt.pf ppf " %a" pp)) l
 
-    | Sy.L_neg_eq, a::l ->
-      fprintf fmt "(distinct %a%a)"
-        print a (fun fmt -> List.iter (fprintf fmt " %a" print)) l
+    | Sy.L_neg_eq, [a; b] ->
+      Fmt.pf ppf "(not (= %a %a))" pp a pp b
+
+    | Sy.L_neg_eq, _ :: _ ->
+      Fmt.pf ppf "(distinct %a)" Fmt.(list ~sep:sp pp) xs
 
     | Sy.L_built Sy.LE, [a;b] ->
-      fprintf fmt "(<= %a %a)" print a print b
+      Fmt.pf ppf "(<= %a %a)" pp a pp b
 
     | Sy.L_built Sy.LT, [a;b] ->
-      fprintf fmt "(< %a %a)" print a print b
+      Fmt.pf ppf "(< %a %a)" pp a pp b
 
     | Sy.L_neg_built Sy.LE, [a; b] ->
-      fprintf fmt "(> %a %a)" print a print b
+      Fmt.pf ppf "(> %a %a)" pp a pp b
 
     | Sy.L_neg_built Sy.LT, [a; b] ->
-      fprintf fmt "(>= %a %a)" print a print b
+      Fmt.pf ppf "(>= %a %a)" pp a pp b
 
     | Sy.L_neg_pred, [a] ->
-      fprintf fmt "(not %a)" print a
+      Fmt.pf ppf "(not %a)" pp a
 
     | Sy.L_built (Sy.IsConstr hs), [e] ->
-      fprintf fmt "((_ is %a) %a)" Hstring.print hs print e
+      Fmt.pf ppf "((_ is %a) %a)" Hstring.print hs pp e
 
     | Sy.L_neg_built (Sy.IsConstr hs), [e] ->
-      fprintf fmt "(not ((_ is %a) %a))" Hstring.print hs print e
+      Fmt.pf ppf "(not ((_ is %a) %a))" Hstring.print hs pp e
 
     | (Sy.L_built (Sy.LT | Sy.LE) | Sy.L_neg_built (Sy.LT | Sy.LE)
       | Sy.L_neg_pred | Sy.L_eq | Sy.L_neg_eq
       | Sy.L_built (Sy.IsConstr _)
-      | Sy.L_neg_built (Sy.IsConstr _)) , _ ->
+      | Sy.L_neg_built (Sy.IsConstr _)), _ ->
       assert false
 
-  and print_silent fmt t =
-    let open Format in
+  and pp_operator = Symbols.pp_operator ~format:`Smtlib
+
+  and pp_silent ppf t =
     let { f ; xs ; ty; bind; _ } = t in
     match f, xs with
-    (* Formulas *)
-    | Sy.Form form, xs -> print_formula fmt form xs bind
+    | Sy.Form form, xs -> pp_formula ppf form xs bind
+
+    | Sy.Lit lit, xs -> pp_lit ppf lit xs
 
     | Sy.Let, [] ->
       let x = match bind with B_let x -> x | _ -> assert false in
-      fprintf fmt
-        "(let%a %a =@ %a in@ %a)"
-        (fun fmt x -> if Options.get_verbose () then
-            fprintf fmt
-              " [sko = %a]" print x.let_sko) x
-        Sy.print x.let_v print x.let_e print_silent x.in_e
+      Fmt.pf ppf "(let ((%a %a)) %a)"
+        Symbols.print x.let_v
+        pp x.let_e
+        pp_silent x.in_e
 
-    (* Literals *)
-    | Sy.Lit lit, xs -> print_lit fmt lit xs
+    | Sy.(Op Record), _ ->
+      begin
+        match ty with
+        | Ty.Trecord { Ty.lbs = lbs; record_constr; _ } ->
+          assert (List.compare_lengths xs lbs = 0);
+          Fmt.pf ppf "(%a %a)"
+            Hstring.print record_constr
+            Fmt.(list ~sep:sp pp) xs
 
-    | Sy.Op Sy.Get, [e1; e2] ->
-      if Options.get_output_smtlib () then
-        fprintf fmt "(select %a %a)" print e1 print e2
-      else
-        fprintf fmt "%a[%a]" print e1 print e2
-
-    | Sy.Op Sy.Set, [e1; e2; e3] ->
-      if Options.get_output_smtlib () then
-        fprintf fmt "(store %a %a %a)"
-          print e1
-          print e2
-          print e3
-      else
-        fprintf fmt "%a[%a<-%a]" print e1 print e2 print e3
-
-    | Sy.Op Sy.Concat, [e1; e2] ->
-      fprintf fmt "%a@@%a" print e1 print e2
-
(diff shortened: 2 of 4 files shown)


src/lib/structures/symbols.ml
```diff
@@ -317,82 +296,242 @@ let name_to_string =
...
+    | Op Sqrt_real_excess -> Fmt.pf ppf "(_ sqrt_real_excess)"
+    | Op Int_floor -> Fmt.pf ppf "(_ int_floor)"
+    | Op Int_ceil -> Fmt.pf ppf "(_ int_ceil)"
+    | Op Max_real -> Fmt.pf ppf "(_ max_real)"
+    | Op Max_int -> Fmt.pf ppf "(_ max_int)"
+    | Op Min_real -> Fmt.pf ppf "(_ min_real)"
+    | Op Min_int -> Fmt.pf ppf "(_ min_int)"
+    | Op Integer_log2 -> Fmt.pf ppf "(_ integer_log2)"
+    | Op Pow -> Fmt.pf ppf "(_ pow)"
+    | Op Integer_round -> Fmt.pf ppf "(_ integer_round)"
```
bclement-ocp: These should not be indexed (there should be no `_`).

src/lib/structures/symbols.ml
```diff
@@ -317,82 +296,242 @@ let name_to_string =
...
+    match lit with
+    | L_eq -> Fmt.pf ppf "="
+    | L_neg_eq -> Fmt.pf ppf "distinct"
+    | L_built LE -> Fmt.pf ppf "<="
+    | L_built LT -> Fmt.pf ppf "<"
+    | L_neg_built LE -> Fmt.pf ppf ">"
+    | L_neg_built LT -> Fmt.pf ppf ">="
+    | L_neg_pred -> Fmt.pf ppf "not"
+    | L_built (IsConstr hs) -> Fmt.pf ppf "(_ is %a)" Hstring.print hs
+    | L_neg_built (IsConstr hs) -> Fmt.pf ppf "(not (_ is %a))" Hstring.print hs
```
bclement-ocp: This is not a valid smt-lib symbol?

bclement-ocp (changes requested):
There are extraneous indexing in some identifiers that should be removed.

I think in practice we only need to be able to print constant symbols and names (the rest needs to be matched at the expression level to print the correct structure), is that correct?

src/lib/structures/symbols.ml
bclement-ocp: Well except for `Int2BV` which should be `(_ int2bv n)` for `Int2BV n`.

src/lib/structures/symbols.ml
Halbaroth: Which one?

bclement-ocp: The one on the line that I commented on :) `(not (_ is %a))`

src/lib/structures/symbols.ml
Halbaroth: Besides, I remember @Stevendeo used a different name for one them, something as `ae.round`. We should be consistent with that right? We may prefix all of them with `ae.`

bclement-ocp: `Integer_round` is not `ae.round`. `Integer_round` does not exist in the SMT-LIB format. `ae.round` is `Float`.

bclement-ocp: Ah but yes for custom primitives we should prefix with `ae` that is a good idea! BTW, `Int_floor` is `to_int` from the Reals_Ints theory.

Halbaroth:
I completely change the commit:
We plan to use expression from `Expr` to store model values but we
haven't appropriate printer for it in the SMT-LIB format. Indeed, the current
SMT-LIB printer of `Expr` is wrong, both on formulas and simple symbols as
rational numbers or bitvector literals.

This commit changes the SMT-LIB printer of `Expr` in order to always
print SMT-LIB format.

The new SMT-LIB printer has only two limitations:
1. it doesn't print the triggers of lemmas when the verbose flag is up
   because there is no appropriate syntax in the SMT-LIB standard.
2. it doesn't print the detail of the semantic trigger `In`.

Notice that multi-triggers are now printed as DNF formulae.

After discussion, we decide to keep the actual printer of `Symbols` as
we need it for debugging messages. Instead I split it in several
printers in module's implementation and I expose a new printer
`pp_operator` for sake of code factorisation.

Notice that printing names and literal constants should be sufficient to
output models but for sake of completeness, this commit refactors
completely the SMT-LIB printer of `Expr`.

I also clean up the code of both `Symbols` and `Expr` printers in the
following way:
- I use everywhere `Fmt` and in particular `Fmt.list` and `Fmt.iter`.
- I rename some functions but I keep the previous API. I just add a new
  function `pp_operator` to the signature of `Symbols`.
- I remove some useless test with `Options.get_output_smtlib`.

src/lib/frontend/models.ml
```diff
@@ -312,10 +312,19 @@ module SmtlibCounterExample = struct
     | [Ty.Trecord _r, _arg] -> begin
         match xs_values with
         | [record_name,_] ->
+          (* HOTFIX: this fix is temporary. The current implementation of
```
bclement-ocp, line 315: Ideally, there would be a GitHub issue referenced here.

src/lib/structures/expr.ml
```diff
@@ -311,393 +311,374 @@ module F_Htbl : Hashtbl.S with type key = t =
...
+      end
 
-    | Sy.Op Sy.Concat, [e1; e2] ->
-      fprintf fmt "%a@@%a" print e1 print e2
+    | Sy.Op op, [] -> Fmt.pf ppf "%a" pp_operator op
 
-    | Sy.Op Sy.Extract (i, j), [e] ->
-      fprintf fmt "%a^{%d,%d}" print e i j
+    | Sy.Op op, _ :: _ ->
+      Fmt.pf ppf "(%a %a)" pp_operator op Fmt.(list ~sep:sp pp) xs
```
```suggestion
      Fmt.pf ppf "@[(%a@ %a@])" pp_operator op Fmt.(list ~sep:sp pp) xs
```

src/lib/structures/expr.ml
```diff
@@ -311,393 +311,374 @@ module F_Htbl : Hashtbl.S with type key = t =
...
-                              op == Sy.Min_real || op == Sy.Min_int ->
-      fprintf fmt "%a(%a,%a)" Sy.print f print e1 print e2
+    | Sy.Name (n, _, _), [] -> Symbols.pp_name ppf (Hstring.view n)
 
-    (* TODO: introduce PrefixOp in the future to simplify this ? *)
-    | Sy.Op (Sy.Constr hs), ((_::_) as l) ->
-      fprintf fmt
-        "%a(%a)" Hstring.print hs (Util.print_list ~sep:"," ~pp:print) l
+    | Sy.Name (n, _, _), _ :: _ ->
+      Fmt.pf ppf "(%a %a)"
```
```suggestion
      Fmt.pf ppf "@[(%a@ %a)@]"
```

src/lib/structures/expr.ml
```diff
@@ -311,393 +311,374 @@ module F_Htbl : Hashtbl.S with type key = t =
...
+         | MapsTo _ | In _), _ ->
+      (* All the cases have been excluded by the parser. *)
+      assert false
 
-  and print_triggers fmt trs =
-    List.iter (fun { content = l; _ } ->
-        Format.fprintf fmt "| %a@,"  (Util.print_list ~sep:"," ~pp:print) l;
-      ) trs
+  and pp_trigger ppf { content; _ } =
+    Fmt.pf ppf "@[<hv 2>(and %a)@]" Fmt.(list ~sep:sp pp) content
```
```suggestion
    Fmt.pf ppf "@[<hv 2>(%a)@]" Fmt.(list ~sep:sp pp) content
```

src/lib/structures/expr.ml
```diff
@@ -311,393 +311,374 @@ module F_Htbl : Hashtbl.S with type key = t =
...
+
+  and pp_triggers ppf trs =
+    Fmt.pf ppf "@[<hv 2>(or %a)@]" Fmt.(list ~sep:sp pp_trigger) trs
```
```suggestion
    Fmt.pf ppf "@[<hv 2>(%a)@]" Fmt.(list ~sep:sp pp_trigger) trs
```

src/lib/structures/expr.ml
```diff
@@ -311,393 +311,374 @@ module F_Htbl : Hashtbl.S with type key = t =
...
-        fprintf fmt "(lemma: %s forall %a[%a].@  %a)"
-          name
-          print_binders binders
-          print_triggers user_trs
-          print_silent main
-      else
-        fprintf fmt "(lem %s)" name
+    | Sy.F_Lemma, [], B_lemma { main; name; binders; _ } ->
+      Fmt.pf ppf "(! (forall (%a) %a) :named %s)"
+        pp_binders binders pp_silent main name
```
bclement-ocp: User triggers should be printed using `!` and `:pattern` if they exist.

I think we should keep the non-verbose option as well (it can just print the name of the lemma without its definition)

src/lib/structures/expr.ml
```diff
@@ -311,393 +311,374 @@ module F_Htbl : Hashtbl.S with type key = t =
...
 
-    | Sy.F_Skolem, [], B_skolem { main; binders; _ } ->
-      fprintf fmt "(<sko exists %a.> %a)"
-        print_binders binders print_silent main
+    | Sy.F_Skolem, [], B_skolem { main; name; binders; _ } ->
+      Fmt.pf ppf "(! (exists (%a) %a) :named %s)"
+        pp_binders binders pp_silent main name
```
bclement-ocp: Same as above; we should have a non-verbose option (printing just the name).

src/lib/structures/expr.ml
```diff
@@ -311,393 +311,374 @@ module F_Htbl : Hashtbl.S with type key = t =
...
     match lit, xs with
     | Sy.L_eq, a::l ->
-      fprintf fmt "(= %a%a)"
-        print a (fun fmt -> List.iter (fprintf fmt " %a" print)) l
+      Fmt.pf ppf "(= %a %a)"
+        pp a (fun ppf -> List.iter (Fmt.pf ppf " %a" pp)) l
+
     | Sy.L_neg_eq, [a; b] ->
-      fprintf fmt "(not (= %a %a))" print a print b
+      Fmt.pf ppf "(not (= %a %a))" pp a pp b
```
bclement-ocp: We could just remove that case and just always print `(distinct)`, to be fair.

src/lib/structures/expr.ml
```diff
@@ -311,393 +311,374 @@ module F_Htbl : Hashtbl.S with type key = t =
...
-                (Hstring.view field) print e;
+              Fmt.pf ppf "%s%s = %a"  (if first then "" else "; ")
+                (Hstring.view field) pp e;
               false
             ) true lbs xs);
-          fprintf fmt "}";
-        | _ -> assert false
+          Fmt.pf ppf "}";
+        | _ ->
+          (* Exclude by the typechecker. *)
```
```suggestion
          (* Excluded by the typechecker. *)
```

src/lib/structures/expr.ml
```diff
@@ -311,393 +311,374 @@ module F_Htbl : Hashtbl.S with type key = t =
...
+
+  and pp_trigger ppf { content; _ } =
+    Fmt.list ~sep:Fmt.comma pp ppf content
 
-  and print_triggers fmt trs =
-    List.iter (fun { content = l; _ } ->
-        Format.fprintf fmt "| %a@," (Util.print_list ~sep:"," ~pp:print) l;
-      ) trs
+  and pp_triggers ppf trs =
+    Fmt.list ~sep:(fun ppf () -> Fmt.pf ppf "|@,") pp_trigger ppf trs
```
bclement-ocp: Note that this changes the place where the cut occurs, did you test the output? I am not sure cutting immediately after the `|` will do what we want.

(As an aside: `fun ppf () -> Fmt.pf ppf "|@,"` is `Fmt.any "|@,"`)

src/lib/structures/expr.ml
```diff
@@ -311,393 +311,374 @@ module F_Htbl : Hashtbl.S with type key = t =
...
   if Options.get_output_smtlib ()
-  then SmtPrinter.print_triggers fmt
-  else AEPrinter.print_triggers fmt
+  then SmtPrinter.pp_triggers ppf
+  else AEPrinter.pp_triggers ppf
 
-let print_list_sep sep = Util.print_list ~sep ~pp:print
+let print_list_sep sep =
+  if Options.get_output_smtlib () then
+    Fmt.list ~sep:(fun ppf () -> Fmt.pf ppf "%s" sep) SmtPrinter.pp
```
bclement-ocp: Nit: `fun ppf () -> Fmt.pf ppf "%s" sep` is `Fmt.(const string sep)`

src/lib/structures/symbols.ml
```diff
@@ -317,82 +296,209 @@ let name_to_string =
...
+
+    | Not_theory_constant -> Fmt.pf ppf "not_theory_constant"
+    | Is_theory_constant -> Fmt.pf ppf "is_theory_constant"
+    | Linear_dependency -> Fmt.pf ppf "linear_dependency"
+
+    | Optimize {order; is_max=true} ->
+      Fmt.pf ppf "maximize(-,%d)" order
+    | Optimize {order; is_max=false} ->
+      Fmt.pf ppf "minimize(-,%d)" order
+    | Reach -> assert false
```
bclement-ocp: I will admit I have zero clue what this operator is, do you know?

src/lib/structures/symbols.ml
```diff
@@ -317,82 +296,209 @@ let name_to_string =
...
+
+    (* Reals_Ints theory *)
+    | Abs_int | Abs_real -> Fmt.pf ppf "abs"
+    | Real_of_int -> Fmt.pf ppf "to_real"
+    | Real_is_int -> Fmt.pf ppf "is_int"
+    | Integer_round -> Fmt.pf ppf "to_int"
+
+    (* FixedSizedBitVectors theory *)
+    | Extract (i, j) ->
+      Fmt.pf ppf "(_ extract %d %d)" i j
```
bclement-ocp: I believe that Alt-Ergo uses opposite conventions for extract arguments compared to SMT-LIB and that `i` and `j` must be flipped here (see for instance `Expr.Core.BV.extract`)

src/lib/structures/symbols.ml
```diff
@@ -317,82 +296,209 @@ let name_to_string =
...
+
+    (* DT theory *)
+    | Record -> ()
+    | Access s | Constr s | Destruct (s, _) -> Hstring.print ppf s
+
+    (* Float theory *)
+    | Float -> Fmt.pf ppf "ae.round"
+
+    (* Not in the SMT-LIB standard *)
+    | Int2BV _ -> Fmt.pf ppf "(_ int2bv)"
```
bclement-ocp: As I already mentioned:

```suggestion
    | Int2BV n -> Fmt.pf ppf "(_ int2bv %d)" n
```

src/lib/structures/symbols.ml
```diff
@@ -317,82 +296,209 @@ let name_to_string =
...
+    | Not_theory_constant -> Fmt.pf ppf "ae.not_theory_constant"
+    | Is_theory_constant -> Fmt.pf ppf "ae.is_theory_constant"
+    | Linear_dependency -> Fmt.pf ppf "ae.linear_dependency"
+    | Sqrt_real -> Fmt.pf ppf "ae.sqrt_real"
+    | Sqrt_real_default -> Fmt.pf ppf "ae.sqrt_real_default"
+    | Sqrt_real_excess -> Fmt.pf ppf "ae.sqrt_real_excess"
+    | Int_floor -> Fmt.pf ppf "ae.int_floor"
```
bclement-ocp: As I already mentioned:

```suggestion
    | Int_floor -> Fmt.pf ppf "to_int"
```

src/lib/structures/symbols.ml
```diff
@@ -317,82 +296,209 @@ let name_to_string =
...
+    | Fixed | Reach ->
+      (* Unused symbols *)
+      assert false
+
+end
+
+let[@inline always] pp_operator ~(format:[`Ae | `Smtlib]) ppf op =
+  match format with
+  | `Ae -> AEPrinter.pp_operator ppf op
+  | `Smtlib -> SmtPrinter.pp_operator ppf op
```
bclement-ocp: I am not sure the polymorphic variants are very useful here, ``pp_operator ~format:`Ae`` and ``pp_operator ~format:`Smtlib`` are harder to type than separate names such as `pp_ae_operator` and `pp_smt_operator` which also wouldn't require additional parentheses when used with `Fmt.pf`.

Or really we could just expose the `SmtPrinter` and `AEPrinter` modules directly I think.

bclement-ocp (changes requested):
Mostly reasonable, but there are some small issues that should be fixed.

> I completely change the commit:

Ideally, when this is the case, the PR should be closed and a new PR opened instead :)

> it doesn't print the triggers of lemmas when the verbose flag is up because there is no appropriate syntax in the SMT-LIB standard.

As I already mentioned on Zulip, this is `!` with `:pattern`. See page 32 of the standard.

> Notice that multi-triggers are now printed as DNF formulae.

This seems weird; triggers are not booleans. I'd suggest using the standard `:pattern` syntax.

(As an aside: commit messages should be written using imperative style — "Add a SMT-LIB compliant printer for `Expr`" rather than "Adding a SMT-LIB compliant printer for `Expr`" etc. — and generally describe things from a point of view of the change set, i.e. "Use `Fmt` everywhere" rather than "I use `Fmt` everywhere". This helps focus the attention on the changes rather than the author.)

src/lib/frontend/models.ml
Halbaroth: Sure, actually it will be fix by the new representation of model values because we will use `Expr.t` here instead of `string`.

src/lib/structures/symbols.ml
Halbaroth: For me both solutions are equivalent.

src/lib/structures/symbols.ml
Halbaroth: Actually I checked. It's completely unused. Both legacy and Dolmen parser never produces such symbol. Same situation for the symbol `Fixed`. It was supposed to be used by the arithmetic theory in `arith.ml`:
```ocaml
    | Sy.Op Sy.Fixed, _ ->
      (* Fixed-Point arithmetic currently not implemented *)
      assert false
```

src/lib/structures/symbols.ml
Halbaroth: Uhm, so `Integer_round` and `Int_floor` are both `to_int`?

src/lib/structures/expr.ml
Halbaroth: I guess you meant `"@[(%a@ %a)@]"` to be consistent with other printers?

src/lib/structures/symbols.ml
bclement-ocp: `Integer_round` is not `to_int`, I don't think I said that? `Integer_round` uses the fpa rounding mode for rounding to integers (it has type `rounding_mode -> real -> int`).

(If PR prints `Integer_round` as `to_int`, I missed that during review but it is wrong)

src/lib/structures/expr.ml
bclement-ocp: Actually I meant what I wrote because I don't like closing boxes after the parentheses (it leads to more unreadable `)))))))))`) but that would work as well as long as we don't print everything as a unit.

src/lib/frontend/models.ml
Halbaroth: It's done #958.

src/lib/structures/expr.ml
Halbaroth: After a quick check in the SMT-LIB references, it seems patterns of multi-triggers are separated by `:pattern`.

bclement-ocp: Ah yes, I messed up the structure — this should be `Fmt.pf ppf "@[<hv 2>:pattern@ (%a)@]" Fmt.(list ~sep:sp pp_trigger) trs`

src/lib/structures/symbols.ml
bclement-ocp: Right well, we should probably remove these then.

Halbaroth: I plan to do later ;). I open an issue #959.

src/lib/structures/symbols.ml
Halbaroth: I misread, fixed.

src/lib/structures/expr.ml
Halbaroth: I tested and it wasn't pretty good. I fixed it.

src/lib/frontend/models.ml
bclement-ocp: The comment should mention a link to #958 — the point of having an issue is so that we can track what we have to do, and the point of having the link in the comment is so that we know which issue to update if the problem ends up fixed by an unrelated refactor.

src/lib/structures/expr.ml
```diff
@@ -470,10 +474,10 @@ module SmtPrinter = struct
       assert false
 
   and pp_trigger ppf { content; _ } =
-    Fmt.pf ppf "@[<hv 2>(and %a)@]" Fmt.(list ~sep:sp pp) content
+    Fmt.pf ppf ":pattern @[<hv 2>(%a@])" Fmt.(list ~sep:sp pp) content
 
   and pp_triggers ppf trs =
-    Fmt.pf ppf "@[<hv 2>(or %a)@]" Fmt.(list ~sep:sp pp_trigger) trs
+    Fmt.pf ppf "@[%a@]" Fmt.(list ~sep:sp pp_trigger) trs
```
bclement-ocp, line 480: Nit: this is now `Fmt.(list ~sep:sp (box pp_trigger)) trs`

bclement-ocp (changes requested):
Looks good now — please just include a backlink to #958 in the corresponding comment

src/lib/frontend/models.ml
Halbaroth: Done :)

bclement-ocp (approved):

Issue #937, opened 2023-11-16
`get-assignment` is wrong with CDCL solver


bclement-ocp:
Consider the test `tests/models/bool/bool3.models.smt2` which we only test using the Tableaux solver (this is an issue, we should be testing model generation on the default solver now that this is supported).

```smt-lib
(set-logic ALL)
(set-option :produce-assignments true)
(set-option :produce-models true)
(declare-const x Bool)
(declare-const y Bool)
(assert (or (! (and x y) :named foo) (! (and (not x) (not y)) :named bar)))
(check-sat)
(get-model)
(get-assignment)
```

Alt-Ergo with the (default -- using `alt-ergo tests/models/bool/bool3.models.smt2`) CDCL solver finds the following model and assignment:
```smt-lib
unknown
(
  (define-fun x () Bool false)
  (define-fun y () Bool false)
)
((foo false)
 (bar false))
```

This is wrong: `bar` should be `true`, not `false` (`(and (not false) (not false))` is clearly `true`).

I believe this should be fixed by switching to the new Dolmen implementation of `:named` terms, since that will automatically declare `foo` and `bar` as actual constants.

bclement-ocp:
> this is an issue, we should be testing model generation on the default solver now that this is supported

Before someone beats me to the punch: PR incoming for that.

Stevendeo:
Can we close this issue?

bclement-ocp:
Looks like this is fixed in #961 and we forgot to link it, yes.

PR #957, opened 2023-11-20
Dolmen named formulae

### LINKED ISSUE #956: Switch to proper support of `:named` in Dolmen
See: https://github.com/Gbury/dolmen/pull/199

(Assigning @Stevendeo who wrote the first implementation but feel free to assign me instead)

Stevendeo:
Updates the handling of named formulas with the latest Dolmen update


Gbury:
If I read the code correctly, you currently treat all implicit definitions as named definitions, which happens to be the case currently, but may change in the future. It would be better to lookup the `named` tag on the id being defined to determine whether it comes from a `:named` annotation.

src/bin/common/solving_loop.ml
```diff
@@ -97,44 +97,18 @@ let enable_maxsmt b =
...
-      (List.fold_left get_named_of_term acc l1)
-      l2
-  | _ -> acc
+  | `Defs [`Term_def (id, _, _, _, t)] ->
+    begin
+      match DStd.Id.name id with
+      | Simple n -> Util.MS.add n t acc
+      | _ -> assert false (* Named terms are expected to be simple *)
+    end
+  | _ -> assert false (* Implicit definitions are epected to be definitions *)
```
```suggestion
  | _ -> assert false (* Implicit definitions are expected to be definitions *)
```

bclement-ocp (changes requested):
Looks good; please just update the code with @Gbury's remark

bclement-ocp:
Actually it looks like there are still issues with the assignments/models that need to be investigated.

Stevendeo:
> It would be better to lookup the `named` tag on the id being defined to determine whether it comes from a `:named` annotation.

Okay I get it, thanks! Why does the string attached to the `:named` tag does not hold the term name? Or in other words, why does `DStd.Expr.Id.get_tag id DStd.Expr.Tags.named` returns `Some ""` for named formulae?

Gbury:
> > 
> 
> Okay I get it, thanks! Why does the string attached to the `:named` tag does not hold the term name? Or in other words, why does `DStd.Expr.Id.get_tag id DStd.Expr.Tags.named` returns `Some ""` for named formulae?

Well, because the term name is already in the name of the constant. Part of the reason/problem is that other languages (ae's native iirc) have a slightly different notion of `named` which attach a string to an expression, which is why historically, the `named` tag uses a string. However, in the case of smtlib, the `named` annotation creates a new constant, and it seemed better to use an empty string to prevent people from wrongly assuming that they should use the string payload, where instead they should use the constant's name.

That begin said, this is admittedly not ideal, and it could be changed.

Stevendeo:
> Actually it looks like there are still issues with the assignments/models that need to be investigated.

Fixed, but it required some changes on how `get_value` is called.

nix/sources.json
```diff
@@ -8,7 +8,7 @@
         "rev": "baaf1f92ccd473294679dd9025c3ae9a441a82fa",
         "sha256": "0qpkp1av2jk96zx5lgk9h29ak9klp32g8m7hddyzj251jrkxcbmk",
```
bclement-ocp:
```suggestion
        "rev": "b14eb8a2400c3dd2c34cded212c2135bbce1a57c",
        "sha256": "0r1q1byys41snf7bydscp9raj406ypjz7rwri96hsswx3d7jd08b",
```

src/bin/common/solving_loop.ml
```diff
@@ -944,6 +931,19 @@ let main () =
...
   in
+  let handle_stmts all_context st l =
+    let rec aux named_map st = function
+      | [] -> st
+      | [main_stmt] ->
+        let st = handle_stmt all_context st main_stmt in
+        State.set named_terms named_map st
+      | named_stmt :: tl ->
+        let st = handle_stmt all_context st named_stmt in
+        let named_map = get_named_of_stmt ~acc:named_map named_stmt in
```
bclement-ocp, line 918: Should we maybe only do this if `named_stmt.implicit` is true? Although I think all terms but the last one are implicit currently that could change — I think it's best to dispatch on whether `implicit` is `true/false` for calling `get_named_of_stmt` rather than if it is the last term or not, might save us some surprises down the line (or not, but hey).

tests/models/bool/bool1.models.expected
```diff
@@ -1,7 +1,8 @@
 
 unknown
 (
-  (define-fun q () Bool false)
+  (define-fun p () Bool true)
+  (define-fun q () Bool true)
 )
 ((notp unknown)
  (notnq unknown))
```
bclement-ocp, line 8: Hmm, weird that we don't find these — we should find them now.

bclement-ocp (changes requested):
We should check `implicit` rather than depend on the fact that all statements but the last are implicit (or at the very least, if we do depend on that, there should be a comment explaining this).

Also it looks like there are weird things happening with `get_value` — shouldn't we always be able to get the value of the assignments now that there are actual terms associated with them? But it looks like we still reply `unknown` all the time, which is not very useful — can you investigate? This might get out of scope for this PR, so feel free to open an issue instead.

Stevendeo:
> But it looks like we still reply unknown all the time, which is not very useful — can you investigate?

I already did actually, and my conclusion is that these assignments are not always put in the `gamma` map as is, which is the actual data `get_value` works with. @Halbaroth is working on some stuff about the model's API and I wanted to wait for his work before further investigation.

bclement-ocp:
> I already did actually, and my conclusion is that these assignments are not always put in the `gamma` map as is, which is the actual data `get_value` works with. @Halbaroth is working on some stuff about the model's API and I wanted to wait for his work before further investigation.

This should be unrelated to the model API stuff. I just looked at the code; it should be correct for the `Fun_sat` solver, but looking in `gamma` for `Satml_frontend` is not correct (the two solvers are structured very differently), we need to call `SAT.boolean_model` and extract the assignment from there. So really this is just #937 and unrelated to this PR then.

bclement-ocp (approved):
Please rebase to fix the merge conflicts and then we should be ready to merge.

Stevendeo:
> This should be unrelated to the model API stuff.

The function `get_value` is not named `get_assignments` because I had in mind this function would handle the `get-value` instruction which requires the models.

bclement-ocp (approved):

PR #961, opened 2023-11-21
Fixed get-value

Stevendeo:
Uses the boolean model of `satml_frontend` in `get_value`.

### DIFF
--- a/src/lib/reasoners/satml_frontend.ml
+++ b/src/lib/reasoners/satml_frontend.ml
@@ -1399,7 +1399,7 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
     | Ty.Tbool ->
       begin
         let bmodel = SAT.boolean_model env.satml in
-        List.find_map
+        Stdcompat.List.find_map
           (fun Atom.{lit; neg = {lit=neglit; _}; _} ->
              if E.equal t lit then
                Some E.vrai
--- a/src/lib/reasoners/satml_frontend.ml
+++ b/src/lib/reasoners/satml_frontend.ml
@@ -1398,14 +1398,17 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
     match E.type_info t with
     | Ty.Tbool ->
       begin
-        match ME.find_opt t env.gamma with
-        | None ->
-          begin
-            match ME.find_opt (E.neg t) env.gamma with
-            | None -> None
-            | Some _ -> Some E.faux
-          end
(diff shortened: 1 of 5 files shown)


src/lib/reasoners/satml_frontend.ml
```diff
@@ -1398,14 +1398,17 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
-        | None ->
-          begin
-            match ME.find_opt (E.neg t) env.gamma with
-            | None -> None
-            | Some _ -> Some E.faux
-          end
-        | Some _ -> Some E.vrai
+        let bmodel = SAT.boolean_model env.satml in
+        List.find_map
+          (fun Atom.{lit; neg = {lit=neglit; _}; _} ->
```
bclement-ocp, line 1403: Nit: it would be better to call `Atom.literal` and `Expr.neg` here instead; the structure of atoms should be treated as an abstract type as much as possible.

bclement-ocp (approved):
LGTM. Were the changes in solving_loop forgotten in #957?

src/lib/reasoners/satml_frontend.ml
Stevendeo: Is there something preventing us from defining Atom.t as an abstract type?

Stevendeo:
> Were the changes in solving_loop forgotten in https://github.com/OCamlPro/alt-ergo/pull/957?

Some of them yes, I had already started to write the fix on some changes that I did not commit.

src/lib/reasoners/satml_frontend.ml
bclement-ocp: I believe both `Satml_types` and `Satml` modules actually sort of need to know its internals currently, so nothing major as far as I know. Making it an abstract type is not entirely trivial refactoring, but we can try to preemptively not make it too hard on ourselves for when we actually do it :)

Halbaroth: Please open an low-priority issue on this point ;)

Halbaroth (approved):

src/lib/reasoners/satml_frontend.ml
bclement-ocp: Sure (but I want to be clear that this is a refactoring I don't think we should do it for the sake of doing it, it is more something to keep in mind if we can include into architectural changes to `Satml` — this change would probably be quite a bit of churn that "for the sake of abstract types" alone is probably too weak a reason).

Issue #634, opened 2023-06-08
Switching to ocamlformat


Halbaroth:
The current style used in the codebase is very inconsistent. Basile and I want to switch to ocamlformat. I know some of you dislike it but we can configure it to produce a better output. 

If you have some preferences for the configuration, let me know :)

As I explained there #628, I prefer merging OptimAE before using ocamlformat.

PR #962, opened 2023-11-21
Imperative API 

### LINKED ISSUE #899: Imperative API for FunSAT

Stevendeo:
Tackles #899 by making both Sat API imperative.
- The imperative is made... well, imperative; it still relies on `Satml_frontend` for properly wrapping calls to `Satml`. We should be able to merge them, but I choosed not to in this PR to avoid spaghettification.
- The functional sat now has an imperative frontend upadting a reference to the environment.

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -136,10 +136,10 @@ let main () =
           Options.Time.set_timeout (Options.get_timelimit ());
         end;
       SAT.reset_refs ();
-      let ftdn_env =
-        List.fold_left
-          (FE.process_decl ~hook_on_status:Frontend.print_status)
-          (FE.init_env used_context)
+      let ftdn_env = FE.init_env used_context in
+      let () =
+        List.iter
+          (FE.process_decl ~hook_on_status:Frontend.print_status ftdn_env)
           cnf
       in
       if Options.get_timelimit_per_goal() then
@@ -152,8 +152,9 @@ let main () =
          we have to drop the partial model in order to prevent
          printing wrong model. *)
       match ftdn_env.FE.res with
-      | `Sat partial_model | `Unknown partial_model ->
+      | `Sat | `Unknown ->
         begin
+          let partial_model = ftdn_env.sat_env in
           let mdl = Model ((module SAT), partial_model) in
           if Options.(get_interpretation () && get_dump_models ()) then begin
             let ur = SAT.get_unknown_reason partial_model in
--- a/src/lib/frontend/frontend.ml
+++ b/src/lib/frontend/frontend.ml
@@ -129,20 +129,20 @@ module type S = sig
   type sat_env
 
   type res = [
-    | `Sat of sat_env
-    | `Unknown of sat_env
+    | `Sat
+    | `Unknown
     | `Unsat
   ]
 
-  type env = {
+  type env = private {
     used_context : used_context;
     consistent_dep_stack: (res * Explanation.t) Stack.t;
     sat_env : sat_env;
-    res : res;
-    expl : Explanation.t
+    mutable res : res;
+    mutable expl : Explanation.t
   }
 
-  type 'a process = ?loc : Loc.t -> env -> 'a -> env
+  type 'a process = ?loc : Loc.t -> env -> 'a -> unit
 
   val init_env : ?sat_env:sat_env -> used_context -> env
 
@@ -162,7 +162,7 @@ module type S = sig
     ?hook_on_status: (sat_env status -> int -> unit) ->
     env ->
     sat_tdecl ->
-    env
+    unit
 
   val print_model: sat_env Fmt.t
 end
@@ -213,8 +213,8 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
   type sat_env = SAT.t
 
   type res = [
-    | `Sat of sat_env
-    | `Unknown of sat_env
+    | `Sat
+    | `Unknown
     | `Unsat
   ]
 
@@ -222,18 +222,18 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
     used_context : used_context;
     consistent_dep_stack: (res * Explanation.t) Stack.t;
     sat_env : sat_env;
-    res : res;
-    expl : Explanation.t
+    mutable res : res;
+    mutable expl : Explanation.t
   }
 
-  type 'a process = ?loc : Loc.t -> env -> 'a -> env
+  type 'a process = ?loc : Loc.t -> env -> 'a -> unit
 
   let init_env ?(sat_env=SAT.empty ()) used_context =
     {
       used_context;
       consistent_dep_stack = Stack.create ();
       sat_env;
-      res = `Unknown (SAT.empty ());
+      res = `Unknown;
       expl = Explanation.empty
     }
 
@@ -255,10 +255,11 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
         (Ex.print_unsat_core ~tab:false) dep;
     try
       let pb = E.Set.elements (Ex.formulas_of dep) in
-      let env =
-        List.fold_left
-          (fun env f ->
-             SAT.assume env
+      let satenv = SAT.empty () in
+      let () =
+        List.iter
+          (fun f ->
+             SAT.assume satenv
                {E.ff=f;
                 origin_name = "";
                 gdist = -1;
@@ -272,10 +273,11 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
                 from_terms = [];
                 theory_elim = true;
                } Ex.empty
-          ) (SAT.empty ()) pb
+          )
+          pb
       in
       ignore (SAT.unsat
-                env
+                satenv
                 {E.ff=E.vrai;
                  origin_name = "";
                  gdist = -1;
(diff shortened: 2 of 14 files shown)


src/bin/common/solving_loop.ml
```diff
@@ -136,10 +136,10 @@ let main () =
...
-      let ftdn_env =
-        List.fold_left
-          (FE.process_decl ~hook_on_status:Frontend.print_status)
-          (FE.init_env used_context)
+      let ftdn_env = FE.init_env used_context in
+      let () =
+        List.iter
+          (FE.process_decl ~hook_on_status:Frontend.print_status ftdn_env)
           cnf
       in
```
bclement-ocp, line 144:
```suggestion
      List.iter
        (FE.process_decl ~hook_on_status:Frontend.print_status ftdn_env)
        cnf;
```

src/lib/frontend/frontend.ml
```diff
@@ -213,27 +213,27 @@ module Make(SAT : Sat_solver_sig.S) : S with type sat_env = SAT.t = struct
...
     | `Unsat
   ]
 
   type env = {
     used_context : used_context;
     consistent_dep_stack: (res * Explanation.t) Stack.t;
     sat_env : sat_env;
-    res : res;
-    expl : Explanation.t
+    mutable res : res;
```
bclement-ocp, line 225: I think this is duplicated with the info in the sat now?

tests/smtlib/testfile-get-info1.dolmen.expected
```diff
@@ -1,7 +1,7 @@
 
 unknown
 (
- :steps 8)
+ :steps 7)
```
bclement-ocp, line 4: Hm, that is surprising… This may indicate that the change is not as innocuous as we thought — unless this is the weak table somehow?

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1222,21 +1206,20 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
     try
       unsat_rec env ~first_call
     with
-    | I_dont_know env as e ->
+    | I_dont_know as e ->
       begin
         try analyze_unknown_for_objectives env e unsat_rec_prem
```
bclement-ocp: Probably can remove `e` from the arguments

Stevendeo:
There is a test that does not pass on the CI, but it works on my side...

src/lib/reasoners/satml_frontend.ml
```diff
@@ -791,33 +786,28 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
-              env,
               {acc with
                activate = FF.Map.add ff None acc.activate;
                updated = true}
           else
             let ff_abstr,new_proxies,proxies_mp, new_vars =
               FF.cnf_abstr env.ff_hcons_env ff env.proxies acc.new_vars
             in
-            let env = {env with proxies = proxies_mp} in
+            let () = env.proxies <- proxies_mp in
```
```suggestion
            env.proxies <- proxies_mp;
```

Stevendeo:
The test does not seem to be executed with `make runtest`, when I play it manually it does not pass on my side as well.

bclement-ocp (commented):
This looks reasonable except for the change in step limit which is a bit worrying. Maybe we should run benches on ae-format just to check that everything is A-OK before merging?

src/bin/common/solving_loop.ml
Stevendeo: I don't like the `;` for large (i.e., 2 lines or more) chunks of code.

src/lib/frontend/frontend.ml
Stevendeo: Where is the information in the SAT? You mean by calling `get_unknown_reason`?

bclement-ocp: Yes. But it's orthogonal to this PR anyways.

src/bin/common/solving_loop.ml
bclement-ocp: And I don't like `let () =` ever so where does that lead us :)

We don't have a style guide so I won't bikeshed, but you use the exact syntax I suggested in `worker_js.ml` so please at least be consistent.

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1116,9 +1104,10 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
         (* The answer for the first objective is infinity. We stop here as
            we cannot go beyond infinity and the next objectives with lower
            priority cannot be optimized in presence of infinity values. *)
-        raise unknown_exn;
+        raise I_dont_know;
 
       | (e, tv, is_max, is_le) :: l ->
+        env.objectives <- Some obj;
```
bclement-ocp: Can you explain why this is moved here and why it is correct? Is it related to the test failure you observed?

Stevendeo: In the old version, if `acc` is empty, the objective update will not be taken into account because an `I_dont_know` exception is raised. It only is otherwise that the objectives must be updated.

bclement-ocp: In the old version, `env.objectives` is a reference that is shared across all versions of `env` (including the one in `unknown_exn`), so in fact the objectives should still get updated even though that is confusing control flow. Is this the change that fixes the test failure?

Stevendeo: Oh you're right I forgot it was a reference; and no, it is not what fixes the test failure.

Halbaroth: Actually the goal of this reference isn't to be able to retrieve objectives (currently, objectives are stored in a struct from `Models.ml` and the reference `last_saved_model` gives us access to them) but to keep in mind we're exploring branches beyond SAT because we're looking for a better model during the optimization. If we got an `unsat` answer  in this context, it means the last attempt to improve the model failed but the previous one is correct and we have to return `unknown` for the frontend. The critical code is here:
 ```ocaml
    | IUnsat (env, _) as e ->
      if !(env.objectives) == None then raise e;
      (* TODO: put the correct objectives *)
      raise (I_dont_know env)
```

File: src/bin/common/solving_loop.ml (hunk 1 above)
Comment (Halbaroth) [reply src/bin/common/solving_loop.ml:144]: I agree that we should write style guide (not too long). I prefer using `let () =` too but it doesn't really matter.

File: src/lib/reasoners/fun_sat.mli [hunk 7]
```diff
@@ -28,4 +28,37 @@
 (*                                                                        *)
 (**************************************************************************)
 
-include Sat_solver_sig.SatContainer
+(** A functional implementation of the solver. *)
```
Halbaroth: The comment is a bit confusing because we don't what `the solver` means here? Another SAT solver? I suggest:

```suggestion
(** A functional SAT solver implementation. *)
```

src/bin/common/solving_loop.ml
Stevendeo: > We don't have a style guide so I won't bikeshed, but you use the exact syntax I suggested in worker_js.ml so please at least be consistent.

You do realize all I want now is to update `worker_js.ml`, do you? :rofl:

Stevendeo: More seriously, my first message was just explaining why I actually chose to use a `let ()` syntax, and I am sorry if you took it personally. I met people who liked `;` and others who liked `let ()`, and I can't guess what will be the opinion of my reviewer on the matter in advance.

If you want Alt-Ergo to follow a style guide for that matter, I will be happy to create it with you actually!

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1277,37 +1259,34 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
         Steps.push_steps ();
-        {acc with guards =
-                    {acc.guards with
-                     current_guard = expr_guard;} }
+        env.guards.current_guard <- expr_guard
       )
       ~max:to_push
       ~elt:()
-      ~init:env
+      ~init:()
```
Halbaroth, line 1271: Probably the worst way to write for loops now :) but I agree that getting ride of this heresy is out of the scope of this PR.

src/bin/common/solving_loop.ml
bclement-ocp: > You do realize all I want now is to update worker_js.ml, do you? 🤣

I would like you to use the same format in both cases but as I said: we don't have a style guide currently, and so I won't bikeshed the alternative you choose — update either `worker_js.ml` or `solving_loop.ml` I won't care.

> If you want Alt-Ergo to follow a style guide for that matter, I will be happy to create it with you actually!

Having a style good sure would be a good thing (I guess #634 should be part of that discussion as well)! The existing code is a bit of a mess and mixes different styles in somewhat arbitrary ways. Opened #967

bclement-ocp:
> Oh you're right I forgot it was a reference; and no, it is not what fixes the test failure.

Well don't leave us hanging — what did, then?

Halbaroth (approved):

Stevendeo:
> Well don't leave us hanging — what did, then?

Just a `let env =` that I forgot to replace by its proper side effect; the heuristic mechanism that forces the Normal strategy depends on a counter (`last_forced_normal`) that is reinitialized every time it triggers. The counter was not reinitialized, and the force normal behavior was looping. I am not sure of the details regarding the raised exception though, maybe there is a subtle bug hiding behind this. All I managed to find is that the culprit was the first assertion of `Satml.update_lazy_cnf`, which triggers when trying to add twice the same formula in the environment (also, there is a generic exception catcher in satml_frontend that breaks my callstack and I really want to remove it).

bclement-ocp:
Ah, you must have changed it in a force push then, which is why I didn't see it. The question was mostly to know if it was a simple omission or if there was some trickery involved, looks like it is the first case if there was a `let env = { env with xxx } in` lying around. Please run ae-format benches then we should be good to merge!

> there is a generic exception catcher in satml_frontend that breaks my callstack and I really want to remove it

Yeah, this `assert false` is regularly annoying, we should just re-raise the exception there instead I guess, would be happy to accept a PR that does this.

src/bin/common/solving_loop.ml
Stevendeo: Here is a proposition: as long as there is no rule about something, the decision goes to the person who wrote the code. We can't afford to lose hours of reflexion on stuff that are not specified.

Stevendeo:
> Yeah, this assert false is regularly annoying, we should just re-raise the exception there instead I guess, would be happy to accept a PR that does this.

Wouldn't that break the callstack as well? I would rather remove this catch (or replace it with the actual exception that are raised by the wrapped function).

bclement-ocp:
> Wouldn't that break the callstack as well? I would rather remove this catch (or replace it with the actual exception that are raised by the wrapped function).

No, OCaml should properly preserves the call stack when re-raising exceptions.

Stevendeo:
Now I'm thinking of it, maybe the exception was changed to an assert false is due to the fact that it may raise other exception (I_dont_know, Sat) that could be catched at a higher level. If we are sure it is not the case, why not just removing this code?

src/bin/common/solving_loop.ml
bclement-ocp: I am OK with that. I am not trying to force your hand in any way, and I have stated repeatedly that *I am OK with keeping `let () =` here*. I am simply asking that whatever style is used, the code at least be consistent within a single PR, is that really too much to ask?

bclement-ocp:
> Now I'm thinking of it, maybe the exception was changed to an assert false is due to the fact that it may raise other exception (I_dont_know, Sat) that could be catched at a higher level. If we are sure it is not the case, why not just removing this code?

As a matter of fact I don't think we are sure that is not the case (it shouldn't be the case but… bugs exist I guess). In this case, the proper thing to do would be to raise an error (from the `Errors` module) instead (raising from within an exception handler, rather than asserting, should also preserve the call stack).

Or we could keep the `assert false` but add the backtrace to the debug message (using `Printexc.get_backtrace ()`), which maybe is the better solution.

src/bin/common/solving_loop.ml
Stevendeo: > I am not trying to force your hand in any way... 

You are.

bclement-ocp: I think there is clearly a misunderstanding somewhere. I am sorry if this is the way your perceive what I am saying, because that is definitely not the intent.

Here is how the conversation went from my side:

 - I make a *suggestion* to change from `let () = ..` to using the sequence operator
 - You refuse to do it because you do not like the sequence operator for long expressions. That is fine!
 - I *accept that*, but point out that you did use the sequence operator in an exact copy of the same code, and ask you to be consistent (which is more important than any specific style being used)
 - You say "now I just want to update the other file!"
 - *I am OK with that*, which I state. I thought that we agreed, then, and that it was the end of it
 - The discussion switches to "what should we do when something is not specified" and to let the author make decisions
 - I agree with that (most of the time, style suggestions are that, *suggestions*), but there is a change that I thought we agreed on making (see above) and that was not done, which I point out, a bit sharply I admit, I am sorry for that.

To reiterate:
 - I have no strong feelings on the specific style being used
 - I do believe that it is important to be consistent to improve maintainability and clarity of the code
 - I want the PR to move forward based on a mutual understanding between author and reviewer
 - I am not going to hold up the PR for this (the only reason I have not approved yet is that I am waiting for the bench results)
 - If you prefer not to make any change here, please let me know (I guess that is the case, but again — this is not what I initially understood from the conversation); I am happy to move the PR forward based on mutual understanding.

I hope this helps clarify things, happy to have a chat with you one-on-one to clear things up if needed.

I understand that style preferences are subjective, and I respect your perspective on the matter. Looking forward to working with you on a style guide that can streamline these issues in the future :)

Stevendeo: Okay, I think I see where the misunderstanding is.

> You do realize all I want now is to update worker_js.ml, do you? 

This was a joke (which is what I meant in my next message starting with 'More seriously, ...') based on the fact that I believe the choice of `let ()` and `;` syntax can coexist without reducing readibility and I was making fun of my contradiction spirit :wink: I did not plan on updating `worker_js.ml` as I said, for that would be very aggressive way to express my stubbornness.

This conversation went a little too far :sweat_smile: sorry about that!

bclement-ocp: Ah, I understand better now, jokes are not always easy to tell apart in text format! Glad we could clear that up.

> I did not plan on updating worker_js.ml as I said, for that would be very aggressive way to express my stubbornness.

I don't think it would have been aggressive, given that it was one of the two alternatives I was initially asking for.

bclement-ocp:
Benches look good — let's merge. Thanks!

PR #965, opened 2023-11-22
Fix the locked file

Halbaroth:
We should never pin dependencies in `opam` files or their templates. This commit removes all the pins in `alt-ergo-lib.opam` and `alt-ergo-lib.opam.template` and update the locked file.


bclement-ocp:
> We should never pin dependencies in opam files or their templates.

I disagree, and I think that this is just a way of specifying a dev version of a dependency since `opam` doesn't support that. But at least there will be only one file in the diffs, I guess.

(The build issue needs to be fixed, though)

bclement-ocp (approved):

Halbaroth:
It's temporary. We could discuss about this later ;)

PR #968, opened 2023-11-23
fix(BV): Fix unsoundness in BV model generation

bclement-ocp:
The BV model generation was using an incorrect understanding of the [is_cs] flag that lead to unsoundness because it assumed that it was fine to set this flag to [false] when the value is forced *in the current context* whereas it is only valid to do so when the value is forced *in all contexts*.

This PR fixes that, and adds copious amounts of documentation to explain the [is_cs] flag so that we don't make the same mistake again.

Unfortunately the fix leads to a test that no longer passes, but for good reason as the reasoning used for it was unsound. Hopefully this should be fixed in a follow-up to
https://github.com/OCamlPro/alt-ergo/pull/944

I have also added a new test that is actually satisfiable but would previously return [unsat]. It times out for now, but is solved properly in https://github.com/OCamlPro/alt-ergo/pull/944


bclement-ocp:
Test failure in macOS looks to be spurious timeout on `f1.ae` (our timeout is very close to the time it actually takes AE to solve this formula — this one and `everything.ae`).

Halbaroth:
> Test failure in macOS looks to be spurious timeout on `f1.ae` (our timeout is very close to the time it actually takes AE to solve this formula — this one and `everything.ae`).

We got this issue several times with Steven, we should increase the timeout limit or create a category for slow tests. But we could check if the test wasn't faster on v2.5.x also.

Halbaroth (approved):
LGTM

bclement-ocp:
> We got this issue several times with Steven, we should increase the timeout limit or create a category for slow tests. But we could check if the test wasn't faster on v2.5.x also.

Yeah, I get that on my own machine from time to time as well; we could make it a slow test. I don't think it was faster on 2.5.x tho, I recall this popping up sporadically since I started working on AE last year, but we could check.

PR #970, opened 2023-11-23
Simplify smart constructors for literal numbers

Halbaroth:
We made a lot of conversion between string and Zarith representation for literals numbers in `Expr`. This commit simplifies the situation.

Now we perform the conversion of the string representation into Zarith one once in both legacy and Dolmen frontend.

The motivation for this modification comes from the fact we'll use expressions to store model values. If we use string here, it's not easy to normalize terms like `-(- 0 x)` to `x`.

With the Zarith representation, the library does the normalization for us.

Remove the smart constructors `int` and `real` in `Expr` which aren't useful anymore.

### DIFF
--- a/src/lib/frontend/cnf.ml
+++ b/src/lib/frontend/cnf.ml
@@ -83,9 +83,12 @@ let rec make_term up_qv quant_basename t =
         E.mk_term s [t1; t2] ty
       end
 
-    | TTprefix ((Sy.Op Sy.Minus) as s, n) ->
-      let t1 = if ty == Ty.Tint then E.Ints.of_int 0 else E.Reals.of_int 0 in
-      E.mk_term s [t1; mk_term n] ty
+    | TTprefix (Sy.Op Sy.Minus, n) ->
+      if ty == Ty.Tint then
+        E.Ints.(~- (mk_term n))
+      else
+        E.Reals.(~- (mk_term n))
+
     | TTprefix _ ->
       assert false
 
@@ -253,10 +256,8 @@ and make_form up_qv name_base ~toplevel f loc ~decl_kind : E.t =
                 {c = {tt_ty = Ty.Tint;
                       tt_desc = TTconst(Tint "1")}; annot = t1.annot} in
               let tt2 =
-                E.mk_term (Sy.Op Sy.Minus)
-                  [make_term up_qv name_base t2;
-                   make_term up_qv name_base one]
-                  Ty.Tint
+                E.Ints.((make_term up_qv name_base t2)
+                        - (make_term up_qv name_base one))
               in
               E.mk_builtin ~is_pos:true Sy.LE
                 [make_term up_qv name_base t1; tt2]
--- a/src/lib/frontend/cnf.ml
+++ b/src/lib/frontend/cnf.ml
@@ -59,9 +59,9 @@ let rec make_term up_qv quant_basename t =
     | TTconst Tvoid ->
       E.void
     | TTconst (Tint i) ->
-      E.int i
+      E.Ints.of_Z (Z.of_string i)
     | TTconst (Treal n) ->
-      E.real (Numbers.Q.to_string n)
+      E.Reals.of_Q n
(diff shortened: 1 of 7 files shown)


Stevendeo (approved):

Stevendeo (changes requested):
Revoking my approval until tests are performed to check if this PR has broken anything

bclement-ocp (approved):
Finally no longer transforming `-1` into `0 - 1` :tada: 

(Crossing my fingers that it won't have to be reverted due to regressions in term matching)

> Revoking my approval until tests are performed to check if this PR has broken anything

FYI the repo is configured so that approval *and* tests passing are required for merging (unless you click on the big scary checkmark that says "I am being naughty")

src/lib/frontend/cnf.ml
```diff
@@ -84,7 +84,7 @@ let rec make_term up_qv quant_basename t =
       end
 
     | TTprefix ((Sy.Op Sy.Minus) as s, n) ->
-      let t1 = if ty == Ty.Tint then E.int "0" else E.real "0"  in
+      let t1 = if ty == Ty.Tint then E.Ints.of_int 0 else E.Reals.of_int 0 in
```
hra687261:
```suggestion
      let t1 = if ty == Ty.Tint then E.Ints.of_Z Z.zero else E.Reals.of_Q Q.zero in
```
Can be applied everywhere else

hra687261 (approved):
LGTM, I would only suggest using `of_(Z|Q) (Z|Q).(one|zero)` instead of `(Ints|Reals).of_int (1|0)`

src/lib/frontend/cnf.ml
Halbaroth: Is it better?

Stevendeo:
> FYI the repo is configured so that approval and tests passing are required for merging (unless you click on the big scary checkmark that says "I am being naughty")

I was talking about Marvin tests :)

Halbaroth:
> Finally no longer transforming `-1` into `0 - 1` 🎉
> 
> (Crossing my fingers that it won't have to be reverted due to regressions in term matching)
> 
> > Revoking my approval until tests are performed to check if this PR has broken anything
> 
> FYI the repo is configured so that approval _and_ tests passing are required for merging (unless you click on the big scary checkmark that says "I am being naughty")

I cross my fingers too :D and I'm naughty.

src/lib/frontend/cnf.ml
hra687261: Marginally, if at all. I guess it's more a matter of taste than performance.

bclement-ocp:
> The motivation for this modification comes from the fact we'll use expressions to store model values. If we use string here, it's not easy to normalize terms like -(- 0 x) to x.

I left the tab opened and just noticed this remark — do models actually see values like `-(- 0 x)`? This doesn't look like a proper semantic value and sounds like a bug if that is the case.

Halbaroth:
Don't worry, it's not the case. Actually the issue came from the smart constructors `int` and `real` of the `Expr` module. They introduced expressions of the form `0 + x`. I had the issue during optimization because we use these smart constructors to build the optimized value and now I don't use semantic values to store these optimized values, thus the reduction `0 + x -> x` isn't perform anymore.

Halbaroth:
As I explained offline, this PR is a bit risky. After running some benchmarks on both our internal set of tests and SMT-LIB, it seems to be okay. Changing expressions `0 - x` into `(- x)` for literal numbers `x`, decreases the depth of some expressions and so the order of expressions. We observe both regressions and improvements. 

Still, this PR introduces another important change and I'm sure this code is safe. Basically, when we manipulate an expression of the form `0 - x` with `x` a variable and substitute `x` by a literal, say `5`, we got a term `Sy.Int (-5)`, which isn't a term already knew by AE. In the previous code, we got a term of the form `0 - 5`. 

I will move the preprocess computation directly in the SMT printers.

Halbaroth:
I think we can close this PR as we don't need to perform this preprocessing phase.

PR #971, opened 2023-11-23
Smtlib mode mechanism

Stevendeo:
As discussed in #936 ,  there is currently no mechanism to check whether an option can be updated or not given the analysis context. [The standard](https://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf#52) describes the different configuration in which each option can be updated or not, and even more.

This PR adds a mode mechanism and implements a check on different instructions (set-logic, set-option, etc.).


tests/cram.t/run.t
```diff
@@ -43,44 +43,52 @@ combinations of produce-models et al.
...
-  $ echo '(get-model)' | alt-ergo --sat-solver Tableaux -i smtlib2 -o smtlib2 --continue-on-error
+  $ echo '(set-logic ALL)(check-sat)(get-model)' | alt-ergo --sat-solver Tableaux -i smtlib2 -o smtlib2 --continue-on-error 2> /dev/null
+  
+  unknown
   (error "Model generation disabled (try --produce-models)")
 
 The messages above mention `--produce-models`, but we can also use
 `set-option`.
 
-  $ echo '(get-model)' | alt-ergo --produce-models -i smtlib2 -o smtlib2 --continue-on-error
```
bclement-ocp, line 57: Hm, these should have been broken since the switch to Dolmen as default frontend, are they skipped somehow?

Anyways, this is no longer the same test as before, and the existing test (adding `(set-logic)`) should still be valid. This is also true for the next test below; please do not remove them.

src/lib/frontend/d_state_option.ml
```diff
@@ -106,6 +133,6 @@ let options_requiring_initialization = [
 
 let init st =
   List.fold_left
-    (fun st (module S : S) -> S.reset st)
+    (fun st (module S : S) -> S.set (S.get (S.reset st)) st)
```
bclement-ocp: Nit: `set (get (reset x))` is confusing to read. Maybe `reset` should be called `clear`?

src/bin/common/solving_loop.ml
```diff
@@ -97,6 +97,13 @@ let enable_maxsmt b =
   else
     DStd.Extensions.Smtlib2.(disable maxsmt)
 
+let cmd_on_modes st modes cmd =
+  if O.get_input_format () = Some Options.Smtlib2 then begin
+    let curr_mode = DO.Mode.get st in
+    if List.for_all (Fun.negate (Util.equal_mode curr_mode)) modes then
```
```suggestion
    if not (List.exists (Util.equal_mode curr_mode)) modes then
```

bclement-ocp (changes requested):
Looks reasonable; the tests for calling `(get-model)` before `(check-sat)` should not be removed.

tests/cram.t/run.t
Stevendeo: > are they skipped 

No they are not skipped, they just result in an error; why do you think they should have been broken?

> the existing test (adding (set-logic)) should still be valid

I am not sure which test you are referring to. `(set-logic ALL)(get-model)` is invalid, as `(get-model)` can only be called in `Sat` mode which happens after a `check-sat`; or did I miss something?

Stevendeo: Or are you saying we want this test to show that alt-ergo will fail on bad smt files? This sounds relevant.

bclement-ocp (changes requested):
Please restore the tests for the "No model produced" error that you removed, as I asked in the previous round of review.

Stevendeo:
> Please restore the tests for the "No model produced" error that you removed, as I asked in the previous round of review.

For some reason my comments were pending; you should see them now

tests/cram.t/run.t
bclement-ocp: > No they are not skipped, they just result in an error; why do you think they should have been broken?

I thought that `(get-model)` would have returned an error about `set-logic` not being present rather than an error about the model not being here, but it looks like Dolmen does not complain about a missing `set-logic` for `get-model`.

> Or are you saying we want this test to show that alt-ergo will fail on bad smt files? This sounds relevant.

Yes, that was what the test was doing (although the error was different; it was "No model produced" and now it should be "Invalid command in start mode").

(Now thinking that we should probably also have a test for `(set-logic ALL)(assert false)(check-sat)(get-model)` for the "No model produced" error, I don't think we have that one)

tests/smtlib/testfile-steps-bound1.dolmen.err.expected
```diff
@@ -0,0 +1 @@
+(error "Invalid action during mode Util.Assert: Set option steps_bound")
```
bclement-ocp: "mode Util.Assert" is not very nice; if that's what `@@deriving show` does we should write our own printer instead.

tests/cram.t/run.t
Gbury: > I thought that `(get-model)` would have returned an error about `set-logic` not being present rather than an error about the model not being here, but it looks like Dolmen does not complain about a missing `set-logic` for `get-model`.

Not by default indeed. The situation is that, the dolmen functions to typecheck things would raise a missing set-logic error if called, but in the case of `get-model`, nothing needs to be typed. For that, dolmen has a `Flow` module, whose role is to check that the statements respect the smtlib constraints, see https://github.com/Gbury/dolmen/pull/135

bclement-ocp: Ah! Interesting! I thought for a moment that we could maybe use that in this PR, but it looks like it is not designed to be extensible — it doesn't look like we can access the current mode from the state, or otherwise configure custom mode-specific filters (e.g. for custom options that can only be set in start mode), is that correct?

Gbury: That's correct currently, but I'd accept a PR to make it extensible.

bclement-ocp: OK. I think if Dolmen already had everything we needed it would have made sense to use that, but given that we would need to modify the `Flow` module to suit our needs and that we have an already working implementation inside Alt-Ergo in this PR, I am fine with either solution.

Stevendeo: > (Now thinking that we should probably also have a test for (set-logic ALL)(assert false)(check-sat)(get-model) for the "No model produced" error, I don't think we have that one)

This would return a different error: after check-sat we would be in `Unsat` mode, which is an invalid mode for `get-model`

bclement-ocp: Another error is fine; what I meant is that this is a common scenario that should fail and that (I believe) we don't test.

src/bin/common/solving_loop.ml
```diff
@@ -649,10 +684,13 @@ let main () =
   let handle_custom_statement id args st =
     match id, args with
     | Dolmen.Std.Id.{name = Simple "minimize"; _}, [term] ->
+      cmd_on_modes st [Assert] "minimize";
       handle_minimize_term term st
     | Dolmen.Std.Id.{name = Simple "maximize"; _}, [term] ->
+      cmd_on_modes st [Assert] "maximize";
       handle_maximize_term term st
     | Dolmen.Std.Id.{name = Simple "get-objectives"; _}, args ->
+      cmd_on_modes st [Sat] "maximize";
```
```suggestion
      cmd_on_modes st [Sat] "get-objectives";
```

src/lib/structures/errors.ml
```diff
@@ -254,3 +271,9 @@ let report fmt = function
   | Dolmen_error (code, descr) ->
     Format.fprintf fmt "Error %s (code %i)" descr code;
   | Warning_as_error -> ()
+  | Mode_error (mode, merr) ->
+    Format.fprintf
+      fmt
+      "Invalid action during mode %a: %a"
```
bclement-ocp:
```suggestion
      "Invalid action during %a mode: %a"
```

bclement-ocp (approved):
LGTM; just two small typos I missed on first read.

PR #973, opened 2023-11-24
fix(build): Use opam install --locked for Docker build

bclement-ocp:
Missed in #965


MuSSF (approved):

Halbaroth:
It's strange, I didn't suggest this modification because I thought it wouldn't work. The version of `opam` shipped with the docker is `2.1` which means the bug we got on other workflows should occur here, even if we use the flag `--locked`. It seems it doesn't :)

PS: please don't merge PR before I validate them.

bclement-ocp:
The bug is with the setup-ocaml action. The docker container does not use the setup-ocaml action.

bclement-ocp:
> PS: please don't merge PR before I validate them.

Are you really berating me for a one-word PR that fixes the build?

Issue #500, opened 2022-02-25
AE seems stuck on a trivial test case


pictavien:
Hi,

I'm using alt-ergo (2.4.0). AE fails to prove, at least in a reasonable amount of time, the
following simple case generated by Why3 (the goal is an axiom). 
After inlining the function `f`, it works fine.

```
(* this is the prelude for Alt-Ergo, version >= 2.4.0 *)
(* this is a prelude for Alt-Ergo integer arithmetic *)
type string

logic match_bool : bool, 'a, 'a -> 'a

axiom match_bool_True :
  (forall z:'a. forall z1:'a. (match_bool(true, z, z1) = z))

axiom match_bool_False :
  (forall z:'a. forall z1:'a. (match_bool(false, z, z1) = z1))

axiom CompatOrderMult :
  (forall x:int. forall y:int. forall z:int. ((x <= y) -> ((0 <= z) ->
  ((x * z) <= (y * z)))))

function f(x: int) : int = (((x * x) * x) - x)

axiom f_diff :
  (forall x:int. (exists y:int. ((f((x + 1)) - f(x)) = (3 * y))))

goal f_diff2 :
  (forall x:int. (exists y:int. ((f((x + 1)) - f(x)) = (3 * y))))
```

Gbury:
Sorry for the delay, I'll try and take a look at this as soon as possible.

bclement-ocp:
Looks like this is AC(X) exploding by trying to saturate way too many equations involving `f(x)`, `f(x + 1)` and variants. Either inlining `f` or abstracting it (making it uninterpreted) makes AltRg answer immediately because it no longer spends time with irrelevant congruences involving `f`.

After discussing with @Gbury there are at least 4 underlying issues that contribute to the problem:

 - SAT should be able to figure out very early that we have both `f_diff` and its negation in the context and conclude immediately, but it doesn't.
 - ~Arith receives terms like `f((x - 1) + 1)` which don't get normalized to `f(x)` (we suspect that `f((x - 1) + 1)` gets created by substiting `x → x - 1` in `f(x + 1)` and it doesn't get renormalized~ This has either been fixed or was an incorrect reading of the Uf output from me (I suspect this is a byproduct of #566 )
 - Trying to fix the above showed that we sometime create the term `1 + x`, and substituting `x + 1` instead allows AltRg to solve quickly, so this is an instance where commutative matching would help (@Halbaroth experimented with this at some point)
[...]

PR #974, opened 2023-11-24
fix(AC): Also abstract non-AC arguments of AC symbols

### LINKED ISSUE #964: Stack overflow
The following file produces a stack overflow. Note that I don't know if it is provable or not, this is a reduced example.
[bug.txt](https://github.com/OCamlPro/alt-ergo/files/13435641/bug.txt)

bclement-ocp:
Alt-Ergo uses an abstraction mechanism for the arguments of AC symbols to ensure termination of the induced rewrite ordering without resorting to a recursive path ordering.

This is described in section 6 of [this paper][1], except that the implementation doesn't do exactly what is described there -- in particular, it only abstracts nested AC symbols, which is an issue when an argument is not an AC symbol but a semantic value that contains another AC symbol.

This patch changes the `abstract2` function in `ac.ml`, where this abstraction mechanism is implemented, to also introduce abstracted constants for non-constant terms.

Fixes #964

Note: while this is ready for review, I want to run benches prior to actually merging — this changes the way reasoning proceeds in the presence of AC symbols, which can have an impact on any NLA problems (see also #500).

[1]: https://arxiv.org/abs/1207.3262

### DIFF
--- a/src/lib/reasoners/ac.ml
+++ b/src/lib/reasoners/ac.ml
@@ -166,7 +166,16 @@ module Make (X : Sig.X) = struct
   let abstract2 sy t r acc =
     match X.ac_extract r with
     | Some ac when Sy.equal sy ac.h -> r, acc
-    | None -> r, acc
+    | None -> (
+        if X.is_constant r then r, acc
+        else
+          match Expr.term_view t with
+          | { Expr.xs = []; bind = B_none; _ } -> r, acc
+          | { ty; _ } ->
+            let k = Expr.fresh_name ty in
+            let eq = Expr.mk_eq ~iff:false k t in
+            X.term_embed k, eq::acc
+      )
     | Some _ -> match Expr.term_view t with
       | { Expr.f = Sy.Name (hs, Sy.Ac, _); xs; ty; _ } ->
         let aro_sy = Sy.name ("@" ^ (HS.view hs)) in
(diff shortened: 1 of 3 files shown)


bclement-ocp:
Damn, -11 on AE-Format, looks like this is too aggressive :(

> This patch changes the abstract2 function in ac.ml, where this abstraction mechanism is implemented, to also introduce abstracted constants for non-constant terms.

Following discussion at the dev meeting yesterday: this is not actually what happens in the paper, as noticed by @Halbaroth 
Closing, will make a separate PR with a better fix.

Halbaroth:
Did you try to implement naively the paper? I mean you explain to us that your patch abstracts some interpreted symbols to avoid a bunch of new abstract variables. Did you try the naive way? Is it terrible?

bclement-ocp:
> Did you try to implement naively the paper? I mean you explain to us that your patch abstracts some interpreted symbols to avoid a bunch of new abstract variables. Did you try the naive way? Is it terrible?

I think it would be problematic (strictly speaking, the paper requires introducing one variable for every AC and uninterpreted symbol appearing in a semantic value, so there would never `f(x) + 1` but only `k + 1` and `f(x) = k`), but as explained in #990, the code is not written in a way that makes this easy. The location in the code (in `Ac.make`) where we perform the abstraction is not the same as the place where the paper describes the abstraction happening (the paper describes it as a global processing; my understanding is that it would roughly map to `X.solve`), and implementing the paper the naïve way would require re-architecturing some parts of the AC(X) implementation, which I would rather avoid (at least until #989 is completed), especially given that it's not clear how well it would work.

PR #975, opened 2023-11-24
Print unknown reason in SMT-LIB syntax

### LINKED ISSUE #966: Invalid unknown reason syntax
This is not a valid syntax. This should be `(:reason-unknown s)` where `s` is a valid s-expression, "Step limit: 3" is not a valid s-expression. It should be something similar to `(:reason-unknown (:step-limit 3))`.

_Originally posted by @bclement-ocp in https://github.com/OCamlPro/alt-ergo/pull/936#discussion_r1400607672_

Stevendeo:
> The old format is preserved for legacy messages in the Alt-Ergo format, but the SMT-LIB format is the default one for pretty-printers, since we are moving towards preferring the SMT-LIB format.
Fixes https://github.com/OCamlPro/alt-ergo/issues/966

( cf https://github.com/Stevendeo/alt-ergo/pull/1 )

### DIFF
--- a/src/lib/reasoners/sat_solver_sig.mli
+++ b/src/lib/reasoners/sat_solver_sig.mli
@@ -40,7 +40,7 @@ type unknown_reason =
   | Timeout of timeout_reason
 
 (** Prints the unknown reason in the default SMT-LIB format *)
-val pp_unknown_reason: unknown_reason Fmt.t
+val pp_smt_unknown_reason: unknown_reason Fmt.t
 
 (** Prints an optional unknown reason in Alt-Ergo format *)
 val pp_ae_unknown_reason_opt : unknown_reason option Fmt.t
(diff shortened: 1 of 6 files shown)


src/lib/reasoners/sat_solver_sig.mli
```diff
@@ -39,8 +39,11 @@ type unknown_reason =
   | Step_limit of int
   | Timeout of timeout_reason
 
+(** Prints the unknown reason in the default SMT-LIB format *)
 val pp_unknown_reason: unknown_reason Fmt.t
```
hra687261:
```suggestion
val pp_smt_unknown_reason: unknown_reason Fmt.t
```
For consistency with other function names, but it not necessary

hra687261 (approved):

Halbaroth (approved):

PR #976, opened 2023-11-24
Replace `Symbols.t` by `Var.t` in binders and substitutions

Halbaroth:
Alt-Ergo embeds variables into symbols then expressions. It's a good design but at many places in the codebase, we use symboles where variables are clearly sufficient.

For instance, we shouldn't use symbols in substitutions or binders because we always bind a variable to an expression or substitute a variable by an expression.

Let's use the simpler type `Var.t` as much as possible.

This commit should be safe because the comparison functions of `Symbols.t` and `Var.t` agree on variables.

### DIFF
--- a/src/lib/frontend/cnf.ml
+++ b/src/lib/frontend/cnf.ml
@@ -117,6 +117,13 @@ let rec make_term up_qv quant_basename t =
     | TTlet (binders, t2) ->
       let binders =
         List.rev_map (fun (s, t1) ->
+            (* Remark: the parser ensures that binders' symbol are always
+               variables. We could modify the typechecker module to
+               produce the appropriate type for binders, that is `Var.t list`,
+               but this requires a large amount of modifications in the
+               typechecker, which is a legacy part of our codebase.
+
+               See PR: https://github.com/OCamlPro/alt-ergo/pull/976. *)
             match s with Sy.Var v -> v, mk_term t1 | _ -> assert false)
           (List.rev binders)
       in
--- a/src/lib/frontend/cnf.ml
+++ b/src/lib/frontend/cnf.ml
@@ -116,12 +116,13 @@ let rec make_term up_qv quant_basename t =
 
     | TTlet (binders, t2) ->
       let binders =
-        List.rev_map (fun (s, t1) -> s, mk_term t1)
+        List.rev_map (fun (s, t1) ->
+            match s with Sy.Var v -> v, mk_term t1 | _ -> assert false)
           (List.rev binders)
       in
       List.fold_left
-        (fun acc (sy, e) ->
-           E.mk_let sy e acc
+        (fun acc (v, e) ->
+           E.mk_let v e acc
            [@ocaml.ppwarning "TODO: should introduce fresh vars"]
         )(mk_term t2) binders
 
@@ -339,8 +340,8 @@ and make_form up_qv name_base ~toplevel f loc ~decl_kind : E.t =
     | TFlet(_,binders,lf) ->
       let binders =
         List.rev_map
-          (fun (sy, (e : _ Typed.tlet_kind))  ->
-             sy,
+          (fun (v, (e : _ Typed.tlet_kind))  ->
+             v,
              match e with
              | TletTerm t -> make_term up_qv name_base t
              | TletForm g -> mk_form up_qv ~toplevel:false g.c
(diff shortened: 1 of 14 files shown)


src/lib/frontend/cnf.ml
```diff
@@ -116,12 +116,13 @@ let rec make_term up_qv quant_basename t =
 
     | TTlet (binders, t2) ->
       let binders =
-        List.rev_map (fun (s, t1) -> s, mk_term t1)
+        List.rev_map (fun (s, t1) ->
+            match s with Sy.Var v -> v, mk_term t1 | _ -> assert false)
```
Stevendeo, line 127: Is this invariant really always true? In that case, could we replace the type of binders by a list of `Var.t`s?

src/lib/structures/var.ml
```diff
@@ -59,6 +59,10 @@ let equal a b = compare a b = 0
 
 let hash { id; _ } = id
 
+let underscore =
+  Random.self_init ();
+  of_string @@ Format.sprintf "_%d" (Random.int 1_000_000)
```
bclement-ocp, line 64: Wait, I was not aware we were doing that — this does not seem safe at all! We should use a counter instead (there are no guarantees that two consecutive `Random.int` calls won't return the same value, which would be problematic as `forall x, y : int. x + y` would become `forall _1, _1: int. _1 + _1` which… is not the same at all.

bclement-ocp (approved):

src/lib/structures/var.ml
bclement-ocp: Opened #977

Halbaroth: ah ah :) you don't know the best part of our mighty codebase I see :) 
This snippet is pure gold. I shared it one year ago on zulip, best joke of my short software developer career.

src/lib/structures/expr.ml
```diff
@@ -2569,14 +2570,22 @@ module Purification = struct
...
+        in
+        mk_term (Sy.Var fresh_var) [] t.ty , add_let fresh_var t lets
 
       | _ -> (* detect ITEs *)
         match t.xs with
         | [_;_;_] when is_ite t.f ->
-          let fresh_sy = Sy.fresh_skolem ~is_var:true "Pur-Ite" in
-          mk_term fresh_sy [] t.ty , add_let fresh_sy t lets
+          let fresh_var =
+            match Sy.fresh_skolem ~is_var:true "Pur-Ite" with
```
Stevendeo: I think it is better to add a `Symbols.fresh_skolem_name` instead of creating a var and destroying it with an `assert false` (or simple export the SkolemId from symbols)? I wonder if the current `fresh_skolem` function is still relevant: is there still cases in your PR where `fresh_skolem` is called with without creating a `var`?

src/lib/frontend/cnf.ml
Halbaroth: Actually I'm checking it right now. I forgot to do it before publishing the PR. I hope the legacy parser enforce this invariant (the new one does).

src/lib/structures/expr.ml
Halbaroth: I'll check. Good catch.

src/lib/structures/var.ml
bclement-ocp: Ah! But what I said above is not true — underscore is a global variable, not a function. OK, that makes things better — there can "just" be conflicts with a variable named `_n` for some `n` that changes each time Alt-Ergo is run. We should just call this `"._"` instead, no?

src/lib/frontend/cnf.ml
Halbaroth: Actually, before submitting this PR, I tried to change the type of the binders in the legacy typechecker. This will requires to change many types, mostly because the typechecker aggregates the symbols in a context, so embed variables into symbols makes the design simpler here. I decided to keep the code as before. The only place where we produce a new `TTlet` term is here:
https://github.com/OCamlPro/alt-ergo/blob/7363cd05a0b3af10a6a6b98738ee0810899052b2/src/lib/frontend/typechecker.ml#L913-L934

As you can see line 926, we store new variables in the environment of the typechecker and we retrieve them at the line 932 to produce the list of binders. Thus, the binders are always variables (embedded in symbols).

src/lib/structures/expr.ml
Halbaroth: The function `Symbols.fresh_skolem` isn't necessary anymore. I create a two functions `fresh_skolem_string` and `fresh_skolem_name`. It would be better to extract the module of identifiers from the `Symbols` module and create a function `fresk_skolem` in `Var` but it's a small detail.

src/lib/frontend/cnf.ml
Stevendeo: I suggest you add to the assert false a TODO stating that we could actually simplify but it would require to update some legacy code which was out of the scope of the PR adding this assert false (and a reference to typechecker.ml).

Stevendeo (approved):

PR #978, opened 2023-11-27
fix(BV): Apply C-substitutions recursively

bclement-ocp:
Turns out that https://github.com/OCamlPro/alt-ergo/pull/696 was not enough, because `uniforme_slice` can call `slice_vars` recursively due to B-variables being split, which in turn can still cause recursive splitting in the C-substitution.


src/lib/reasoners/bitv.ml
```diff
@@ -1103,13 +1101,15 @@ module Shostak(X : ALIEN) = struct
...
-        |[] -> []
-        |v::r ->
+        | [] -> []
+        | v::r ->
           match assoc_var v subs with
-          | vl -> vl @ f_aux r
+          | vl ->
+            (* Substitutions may be recursive and need to be applied again to
+               their result. *)
+            f_aux (vl @ r)
```
hra687261, line 1112: Kinda out of the scope of the PR, but I wonder if it would be possible to apply the substitutions inside `subs`?
So if [`x -> [x1; y; x1]`;`y -> [y1; y2]`] then we can have [`x -> [x1; y1; y2; x1]`;`y -> [y1; y2]`], but I'm not sure that the risk/cost/benefit is worth it.

hra687261 (approved):
LGTM.

src/lib/reasoners/bitv.ml
bclement-ocp: We could, but given that C variables occur exactly twice (and one occurrence was eagerly substituted when the substitution was created) I don't think it would bring much benefit.

PR #979, opened 2023-11-27
fix(BV): Allow substituting constants

bclement-ocp:
Substituting a constant should be a no-op (just like substituting a variable that is not in the substitution). Somehow this case was missed in the new version of `apply_sub`. It is fairly hard to trigger this case however: it can only happen in `apply_subs`, and requires a C-substitution to be created, which in turn requires a C-variable to be split during B-variable slicing, and then the C-substitution must be applied to an equation rooted in a distinct term that contains a constant. This patch includes a test case where this happens.

Note: currently includes #978

### DIFF
--- a/src/lib/reasoners/bitv.ml
+++ b/src/lib/reasoners/bitv.ml
@@ -921,11 +921,14 @@ module Shostak(X : ALIEN) = struct
 
     (* [assoc_var v sub] is the same as [assoc_var_id] but takes a variable.
 
-       Requires: [v] is an unconstrained root variable.
+       Requires: [v] is an unconstrainted root variable or a constant.
+       Raises: [Not_found] if [v] is not in the substitution, or if [v] is a
+         constant.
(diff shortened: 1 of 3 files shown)


src/lib/reasoners/bitv.ml
```diff
@@ -923,11 +921,15 @@ module Shostak(X : ALIEN) = struct
 
     (* [assoc_var v sub] is the same as [assoc_var_id] but takes a variable.
 
-       Requires: [v] is an unconstrained root variable.
+
+       Requires: [v] is an unconstrainted root variable or a constant.
+       Raises: [Not_found] if [v] is not in the substitution, or if [v] is a
+         constant.
```
```suggestion
       Requires: [v] is an unconstrained root variable or a constant.
       Raises: [Not_found] if [v] is not in the substitution, or if [v] is a
         constant.
```

hra687261 (approved):
LGTM modulo minor edits to a comment

Halbaroth (approved):
Please rebase the PR and I merge it :)

PR #980, opened 2023-11-27
fix(build): Correct setup-ocaml option name

bclement-ocp:
This uses the correct `allow-prerelease-opam` option name instead of `allow-prelease-opam` which does not exist and was typo'd in https://github.com/OCamlPro/alt-ergo/pull/965

Hopefully this fixes the build, which is broken again.


bclement-ocp:
> Hopefully this fixes the build, which is broken again.

Looks like it does :tada:

Halbaroth (approved):

PR #983, opened 2023-11-28
Use a fixed name for Var.underscore

bclement-ocp:
There is no good reason to use a random number in the name of the `underscore` variable; this can cause reproducibility issues and potentially accidental unsoundness (I have not looked into details into how `underscore` is used, possibly it would lead to crashes rather than unsoundness, but both are bad anyways).

This patch always uses `._` instead, which is a solver-reserved name.


Gbury:
Slight clarification: identifiers starting with a dot, such as `._` are solver-reserved in the SMT-LIB language, but other languages do not have a concept of solver-reserved identifiers. That begin said, most languages (at least from what I remember) do not allow identifiers to start with a dot, so that should be relatively safe (ultimately, I would very strongly advise against using string equality as a way to determine equality of symbols in a solver after typechecking, but I know it would be complicated to change that in alt-ergo).

bclement-ocp:
That is correct; Alt-Ergo officially supports only SMT-LIB and the native Alt-Ergo language as inputs (even though in practice other languages will work with the Dolmen frontend, they are not officially supported), and both of those either forbid identifiers starting with a dot or reserve them for solver usage.

I also agree on avoiding string comparison, but it is not completely trivial to change — and as long as we are using string comparison, we can at least avoid adding an additional layer of potential failure by using randomness on top!

Gbury:
It's certain that removing uses of `Random` is a good thing (even more so when the randomness is useless).

Halbaroth:
It seems that the variable `underscore` is used by the matching engine. Basically, we can use underscore symbols in our patterns:
https://github.com/OCamlPro/alt-ergo/blob/7bddee4351ba99966168ce82fac977ca22b4b370/src/lib/reasoners/matching.ml#L376-L387

Underscores can appear in generated triggers too.

Besides, this feature isn't supported by the new frontend in user-defined triggers.

To be safe, we have to ensure that we never produce a variable `._` in patterns of triggers but for the underscore variable. I'm pretty sure the user can produce declared variable whose the name will be `._` but we have to check it's not the case in the `Triggers` module of `Expr`.

bclement-ocp:
> we can use underscore symbols in our patterns
> Underscores can appear in generated triggers too.

I don't understand how we can create `Var.underscore` outside of through calling `Var.underscore`. The previous definition is a *random name* that is generated at runtime.

> I'm pretty sure the user can produce declared variable whose the name will be ._

Names starting with a `.` are invalid in the Alt-Ergo format, and reserved for solver use in the SMT-LIB format, users can't use it.

Issue #834, opened 2023-09-22
Printing unknown reason


Halbaroth:
We should support the SMT-LIB syntax `(get-info :reason-unknown)` to be able to print the reason why Alt-Ergo answers `timeout`. 

See https://github.com/OCamlPro/alt-ergo/pull/829#discussion_r1332945479

The PR #837 partially solves this issue.

Stevendeo:
I have been testing something before closing this issue, and something annoying happens.

- If we set a global timeout, we stop right after the timeout so the get-info :reason-unknown is not reached
- If we set a timeout per goal and return unknown, `(get-info :reason-unknown)` returns an error.

Stevendeo:
Ah, actually with the latest `next` (after Fix timeout) it is worse with timelimit-per-goal. As the assertion fails, it returns a first unknown before reaching `check-sat` which also returns `unknown`...

bclement-ocp:
Works now.

PR #984, opened 2023-11-28
Fix timelimit per goal

Stevendeo:
While testing for #834, I noticed PR #982 broke a few things with the timelimit_per_goal option. When such a timeout is reached during an assertion happening before a check-sat, alt-ergo printed the `unknown` answer twice: one for the assert, and once for the check-sat. This PR fixes it and adds a corresponding test.


bclement-ocp (approved):

bclement-ocp:
Why is this still complaining that the branch is out-of-date with the base branch, I thought @Halbaroth disabled that?

bclement-ocp:
> Why is this still complaining that the branch is out-of-date with the base branch, I thought @Halbaroth disabled that?

Apparently not, I did since that's what we agreed on in the dev meeting.

Halbaroth:
I didn't disable it. I just said it wasn't very convenient ;)

PR #986, opened 2023-11-29
cleanup: Use proper ADT for variable classes

bclement-ocp:
This patch cleans up the way we use the '?' (question mark) and '_' special variables by introducing special constructors in `Symbols.var_bnd` and `Var.t` for their corresponding purposes.

Note that it looks like this fixes a subtle bug in `IntervalCalculus`, which assumes that all variables whose name starts with a '?' are "local" variables from semantic triggers, while that is not true when using the SMT-LIB format. However, since that piece of code only runs during semantic matching, it would require defining theory extensions in the SMT-LIB format to trigger the bug, and since that is not supported, there is in fact no bug.

Following discussions in #983 and at the dev meeting. Note that there was no possible confusion in #983 — we collectively forgot that variables are identified by a globally unique identifier (and `Var.underscore` always has ID 1).

### DIFF
--- a/src/lib/frontend/d_cnf.ml
+++ b/src/lib/frontend/d_cnf.ml
@@ -47,12 +47,6 @@ let unsupported msg =
     (fun str -> Errors.(run_error (Unsupported_feature str)))
     msg
 
-module Shared = struct
-  (* Shared constants to avoid allocations*)
-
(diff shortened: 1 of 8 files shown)


Halbaroth (approved):
LGTM

PR #990, opened 2023-11-29
fix(AC): Abstract AC arguments that contain other AC symbols

### LINKED ISSUE #964: Stack overflow

bclement-ocp:
This patch implements a variation of the abstraction mechanism described in the AC(X) paper. This should hopefully ensure that we can't create substitution cycles due to improper term ordering.

Note that the issue is related to comparing distinct AC symbols: when the AC symbols are identical, the AC theory uses a multiset ordering on its argument, which prevent substitution cycles. But when the AC symbols are different, only the symbols are compared, not the arguments, and so we must rely on the abstraction mechanism to prevent cycles.

This is a do-over of https://github.com/OCamlPro/alt-ergo/pull/974 that should be closer in spirit to the implementation of the paper and without the associated regressions (based upon a manual inspection; will re-run the benchmarks).

Fixes #964

### DIFF
--- a/src/lib/reasoners/ac.ml
+++ b/src/lib/reasoners/ac.ml
@@ -192,7 +192,13 @@ module Make (X : Sig.X) = struct
      correctness of the corresponding abstraction process has not been proven.
      See also https://github.com/OCamlPro/alt-ergo/issues/989
 
-     [1]: https://arxiv.org/pdf/1207.3262.pdf *)
+     [1]: Canonized Rewriting and Ground AC Completion Modulo Shostak Theories:
+            Design and Implementation.
+          Sylvain Conchon, Evelyne Contejean, Mohamed Iguernelala.
+          lmcs:1034 - Logical Methods in Computer Science, September 14, 2012,
+            Volume 8, Issue 3.
+          doi:10.2168/LMCS-8(3:16)2012
+          https://arxiv.org/pdf/1207.3262.pdf *)
   let abstract2 sy t r acc =
     if List.exists (is_other_ac_symbol sy) (X.leaves r) then
       match X.ac_extract r, Expr.term_view t with
--- a/src/lib/reasoners/ac.ml
+++ b/src/lib/reasoners/ac.ml
@@ -163,25 +163,60 @@ module Make (X : Sig.X) = struct
   let fold_flatten sy f =
     List.fold_left (fun z (rt,n) -> flatten sy ((f rt),n) z) []
 
-  let abstract2 sy t r acc =
+  let is_other_ac_symbol sy r =
     match X.ac_extract r with
-    | Some ac when Sy.equal sy ac.h -> r, acc
-    | None -> r, acc
-    | Some _ -> match Expr.term_view t with
-      | { Expr.f = Sy.Name { hs; kind = Sy.Ac; _ }; xs; ty; _ } ->
+    | Some ac -> not (Sy.equal sy ac.h)
+    | None -> false
+
+  (* This implements a variant of the term abstraction process described in
+     section 6 of the AC(X) paper [1].
+
+     The abstraction process given in the paper requires to abstract all AC
+     symbols appearing in a non-AC context, but the implementation does not
+     know about the context at the time the abstraction is performed (note that
+     rules Abstract1 and Abstract2 are concerned with *equations* while the
+     abstraction process here occurs at the time of building semantic values,
+     which is earlier). Further, the implementation seems to implicitly rely on
+     term ordering (older terms are ordered before newer terms, so in
+     particular subterms are always smaller than terms that contains them) to
+     cheaply prevent loops rather than introducing all the abstracted variables
+     that the theoretical presentation in the paper would require.
+
+     So the implementation below of the Abstract2 rules deviates from the
+     presentation in the paper to accomodate those differences, globally,
+     between the implementation and theoretical description of AC(X).
+
+     More precisely, `abstract2` will abstract terms that *contain* AC leaves
+     when they appear as argument of an AC symbol. This ensures that AC terms
+     satisfy the T_AC definition from page 22 of the paper, although
+     correctness of the corresponding abstraction process has not been proven.
+     See also https://github.com/OCamlPro/alt-ergo/issues/989
+
+     [1]: https://arxiv.org/pdf/1207.3262.pdf *)
+  let abstract2 sy t r acc =
+    if List.exists (is_other_ac_symbol sy) (X.leaves r) then
+      match X.ac_extract r, Expr.term_view t with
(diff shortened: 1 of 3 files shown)


bclement-ocp:
> without the associated regressions

Confirmed: this patch is +3-0 on AE-format.

bclement-ocp:
@Gbury @Halbaroth ping (PR rebased onto next)

bclement-ocp:
@Gbury ping

Halbaroth:
I will review this PR this afternoon :) Thanks for your patience!

src/lib/reasoners/ac.ml
```diff
@@ -163,26 +163,61 @@ module Make (X : Sig.X) = struct
...
+     presentation in the paper to accomodate those differences, globally,
+     between the implementation and theoretical description of AC(X).
+
+     More precisely, `abstract2` will abstract terms that *contain* AC leaves
+     when they appear as argument of an AC symbol. This ensures that AC terms
+     satisfy the T_AC definition from page 22 of the paper, although
+     correctness of the corresponding abstraction process has not been proven.
+     See also https://github.com/OCamlPro/alt-ergo/issues/989
+
+     [1]: https://arxiv.org/pdf/1207.3262.pdf *)
```
Halbaroth: Nit: we should use DOI as much as possible to have permanent link:
https://doi.org/10.2168/LMCS-8(3:16)2012

src/lib/reasoners/ac.ml
```diff
@@ -163,26 +163,61 @@ module Make (X : Sig.X) = struct
...
+    | None -> false
+
+  (* This implements a variant of the term abstraction process described in
+     section 6 of the AC(X) paper [1].
+
+     The abstraction process given in the paper requires to abstract all AC
+     symbols appearing in a non-AC context, but the implementation does not
+     know about the context at the time the abstraction is performed (note that
+     rules Abstract1 and Abstract2 are concerned with *equations* while the
+     abstraction process here occurs at the time of building semantic values,
```
Halbaroth, line 178: If I understand well, in the paper we first define a set of terms `T` and equations `s = t` with `s, t \in T` and we consider a subset of equations (named abstracted equations (technically it's not a subset as we introduce abstract variables from a fresh set `K`)) on which the multiset order should ensure the termination of the `AC(X)` procedure. 

In the implementation, we never produce the superset of equations because the terms (named semantic values in the implementation) are abstracted by construction?

src/lib/reasoners/ac.ml
```diff
@@ -163,26 +163,61 @@ module Make (X : Sig.X) = struct
...
+
+     More precisely, `abstract2` will abstract terms that *contain* AC leaves
+     when they appear as argument of an AC symbol. This ensures that AC terms
+     satisfy the T_AC definition from page 22 of the paper, although
+     correctness of the corresponding abstraction process has not been proven.
+     See also https://github.com/OCamlPro/alt-ergo/issues/989
+
+     [1]: https://arxiv.org/pdf/1207.3262.pdf *)
+  let abstract2 sy t r acc =
+    if List.exists (is_other_ac_symbol sy) (X.leaves r) then
```
Halbaroth, line 203: I  guess this is the only important line in this PR :)

Let's `u` and `v` two distinct AC symbols (for sake of simplicity, say that `u` and `v` are binary integer operators).
Consider the expressions `A := u(v(x, y), z)` and `B := u(v(x, y) + t, z)`. 

In the previous implementation, we abstracted the subterm `v(x, y)` in the former case but we don't in the latter one. 

In the new implementation, we'll abstract both of them because `v(x, y)` is now a leaf of an argument of `u` and `u` is distinct of `v`. 

If we replace the operator `+` by a non-ac uninterpreted symbol, we don't abstract the subterm `v(x, y)`, right? For instance we don't abstract `v(x, y)` in the term `u(h(v(x, y), t), z)` where `h` is a non-ac uninterpreted operator.

I guess the notation Σ_∅ in 𝒯_∅ in the paper denotes the interpreted symbols? 

We don't need to abstract uninterpreted symbols here because, by definition they are already abstracted, so there is no relation between `h(x, y)` and `h(y, x)` that can induce an infinity path in the reduction algorithm?

Halbaroth (approved):

src/lib/reasoners/ac.ml
bclement-ocp: I generally don't use DOI links because, in my experience, they more often than not point to sources that are not open access — and they typically point to some sort of publisher page rather than to the actual article, introducing friction. I'd argue arXiv (and institutional archives such as HAL) links are about as permanent as DOIs.

That said, the paper title should definitely be there; I will update the comment with the paper title and DOI for completeness (but keep the arXiv link as it is more convenient for the reader).

src/lib/reasoners/ac.ml
bclement-ocp: I think it is something like that, yes — in the paper we first build the equations and then perform term abstraction until all equations are in "abstracted form" (note that being abstracted is a condition on the equation, not on the term). But in the implementation we abstract terms when building semantic values, and so we must make sure that *all* possible equations between arbitrary semantic values are in abstracted form.

In particular, this implies (definition of T_AC) that AC symbols that appear at any depth inside the arguments of another AC symbol must be abstracted, which is what this PR enforces.

There might be additional properties that we need to enforce to ensure completeness; for instance, it is still possible to build equalities between uninterpreted/AC terms which should not be allowed as per the paper (one at least must be abstracted as per Definition 6.1). I think that this turns out OK because comparison between AC semantic values/terms starts by comparing the head symbol, and everything behaves as if the term with smallest head symbol was always abstracted in a consistent way — but I have done no proof and could easily be wrong.

src/lib/reasoners/ac.ml
bclement-ocp: Yes exactly — uninterpreted terms do not take part in this process because with semantic values we never look inside uninterpreted terms to see that they actually contain AC symbols. Instead, if we merge AC terms, say we have `h(v(x, v(y, z)))` and `h(v(v(x, y), z))`, the AC theory will normalize `v(x, v(y, z))` and `v(v(x, y), z)` (as semantic values) to the same representative, it will be picked up by CC(X) to prove equality of the two `h` terms. But we never actually look inside or indeed rewrite inside uninterpreted terms.

src/lib/reasoners/ac.ml
Halbaroth: I agree that arXiv links should be stable and adding the title of the article should sufficient to retrieve the article in future.

PR #992, opened 2023-11-29
cleanup: Remove one layer of unique variable identifiers

bclement-ocp:
Since 17877b6fd and the introduction of the `Var` module, variables already have globally unique identifiers, we don't need a second globally unique identifier associated with each globally unique identifier.


Halbaroth (approved):
LGTM

PR #995, opened 2023-11-29
doc(CDCL): Document the CDCL-Tableaux solver

bclement-ocp:
This is a documentation-only patch for the CDCL-Tableaux solver. There is not much additional documentation for the pure CDCL solver, except regarding the use of the `tatoms_queue` as an intermediate place to hold decided and (clause) propagated atoms before performing theory propagations in bulk. The CDCL solver is heavily based on miniSAT, and hence http://www.decision-procedures.org/handouts/MiniSat.pdf is probably a better documentation at a high level than anything we could write.

### DIFF
--- a/src/lib/reasoners/satml.ml
+++ b/src/lib/reasoners/satml.ml
@@ -28,6 +28,42 @@
 (*                                                                        *)
 (**************************************************************************)
 
+(** This module implements a CDCL solver, as well as some extensions based on a
+    combination of CDCL with ideas from Tableaux solver (the solver using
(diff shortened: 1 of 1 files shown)


bclement-ocp:
Build failure is infamous `f1.ae` failures. Which is good that it is spurious, because it would be worrying to have real failures on a PR that touches no code.

bclement-ocp:
@Halbaroth can we merge this?

Halbaroth (approved):

PR #996, opened 2023-11-30
Box at toplevel in SMT-LIB printer

Halbaroth:
Add a box at the toplevel of the SMT-LIB printer to get a decent output. Replace `hv` box by simple box in the same printer and remove some `Fmt.sp`.

### DIFF
--- a/src/lib/structures/expr.ml
+++ b/src/lib/structures/expr.ml
@@ -350,7 +350,7 @@ module SmtPrinter = struct
   and pp_lit ppf lit xs =
     match lit, xs with
     | Sy.L_eq, a::l ->
-      Fmt.pf ppf "(= %a %a)"
+      Fmt.pf ppf "@[<2>(= %a %a@])"
         pp a (fun ppf -> List.iter (Fmt.pf ppf " %a" pp)) l
 
     | Sy.L_neg_eq, _ :: _ ->
@@ -481,8 +481,8 @@ module SmtPrinter = struct
   and pp_verbose ppf t = pp_silent ppf t
 
   and pp ppf t =
-    if Options.get_debug () then Fmt.box ~indent:2 pp_verbose ppf t
-    else Fmt.box ~indent:2 pp_silent ppf t
+    if Options.get_debug () then pp_verbose ppf t
+    else pp_silent ppf t
 
 end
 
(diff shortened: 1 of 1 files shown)


bclement-ocp (approved):
LGTM.

We might want `(Fmt.box pp_silent)` instead of `pp_silent` everywhere as well?

Halbaroth:
Uhm it's strange that this PR improves the situation for the SMT-LIB format. Indeed, I didn't put boxes at the top level of the SMT-LIB printer but the AE one. I fixed the issue (and I amend the previous commit because it was it's very purpose, which means you cannot see the wrong commit anymore...)

Maybe the situation is better because I removed the `hv` boxes and some space hints.

Besides, `pp_silent` opens a box at the top level of most of its branches (there are few cases in which we don't want to open a box, for instance for rational numbers). 

Let's imagine we have two printers `pp1` and `pp2`. `pp2` always opens a box at the top level but `pp1` doesn't. Instead, `pp1` always calls `pp2`. We need to open a box in `pp1` too?

bclement-ocp:
> Let's imagine we have two printers `pp1` and `pp2`. `pp2` always opens a box at the top level but `pp1` doesn't. Instead, `pp1` always calls `pp2`. We need to open a box in `pp1` too?

No, that should not be necessary!

Halbaroth:
So we don't need to open a new box in `pp` or `pp_verbose` because we always do it in `pp_silent` but for some leaves of the AST, right?

bclement-ocp:
> So we don't need to open a new box in `pp` or `pp_verbose` because we always do it in `pp_silent` but for some leaves of the AST, right?

`pp_silent` does not always open a box (e.g. for literals there are no boxes). In any case, opening too many boxes is less of an issue than opening too few.

bclement-ocp:
> for literals there are no boxes

Actually, for equalities specifically

Halbaroth:
> > for literals there are no boxes
> 
> Actually, for equalities specifically

I didn't make much effort here as we plan to remove the n-ary equality.

Halbaroth:
I add a `pp_boxed` wrapper around `pp_silent` and use the wrapper version everywhere. `pp_boxed` opens a box to be sure we never print an expression without using a box at the top level ;)

bclement-ocp (approved):
That also works!

PR #997, opened 2023-12-04
WIP: Nix for CI?

bclement-ocp:


bclement-ocp:
CI tests took about 2m28s (3m14s with an empty cache) vs 4m30s using setup-ocaml — almost twice faster.

macOS still slow as molasses, but we gain a bit as well — 10min with a primed cache instead of 13-15min.

Issue #993, opened 2023-11-29
Optimization and `push` and `pop`


Halbaroth:
The current implementation of `SatML` doesn't apply properly `push` and `pop` in presence of objectives. Indeed, the following input (thanks @bclement-ocp):
```smt
(set-option :produce-models true)
(set-logic ALL)
(declare-const x Int)
(assert (<= 0 x 10))
(push 1)
(maximize x)
(pop 1)
(minimize x)
(check-sat)
(get-model)
(get-objectives)
```
produces the output:
```
unknown
(
  (define-fun x () Int 10)
)
(objectives 
  (x 10)
  (x 10)
) 
```
But we expect to get `0` for the second objective constraint on `x` in the first-order model and `-oo` in the objective model.
The problem occurs after #921 because we don't store objectives as tagged subformulae in the SAT solver but a list in the `Theory` module.

Gbury:
Note: in the response to `(get-objectives)`, I would only expect a single objective constraint, because the first one appearing in the problem (the `maximize` on line 6), has effectively been popped/removed by the `pop` operation, and therefore when we get to the `check-sat`, there is only one constraint in scope (the one from line 8).

Halbaroth:
Both bugs should be solve by pushing case splits in SAT solvers. The current implementation doesn't restore completely the environment of the module `Theory` after a pop statement.

Halbaroth:
@bclement-ocp I am not sure the solution you suggested can works well. 
As I mentioned it, the objectives are registered in the theory environments. These environments are backtracked but all the objectives are registered at the decision level 0! This is the reason why we have two bugs here:
1. We do not produce the best value in incremental mode.
2. We continue to print objectives that are not in the context after popping. 

For instance, for the following input file:
```smt2 
(set-option :produce-models true)
(set-logic ALL)
(declare-const x Int)
(declare-const y Int)
(assert (<= 0 x 10))
(push 1)
(maximize x)
(assert (<= 0 y 10))
(pop 1)
(minimize y)
(check-sat)
(get-objectives)
```
Alt-Ergo outputs:
```
(objectives
  (x 10)
  (y (- oo))
)
```

A very simple patch consists in saving the last theory environment in `push` and restore it in `pop`. Last week, I tried this solution and it works fine (solved both bugs) but it is awkward to save all the theory environment just because we save objectives here. We should save objectives directly in `SatML` as we do not plan to support optimization for `FunSAT`.

[...]

PR #1000, opened 2023-12-04
Incremental version of the solver

Stevendeo:
The target of this PR: the first part before getting rid of the Commands module. As the legacy frontend still relies on it, this PR only cuts all the dependencies of the dolmen loop (solving_loop.ml)

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -121,6 +121,8 @@ let add_if_named
             names. *)
     acc
 
+type stmt = Typer_Pipe.typechecked D_loop.Typer_Pipe.stmt
+
 (** Translates dolmen locs to Alt-Ergo's locs *)
 let dl_to_ael dloc_file (compact_loc: DStd.Loc.t) =
   DStd.Loc.(lexing_positions (loc dloc_file compact_loc))
@@ -1139,7 +1141,20 @@ let main () =
     set_partial_model_and_mode solve_res st
   in
 
-  let handle_stmt :
+  let handle_reset st =
+    let () = Steps.reset_steps () in
+    st
+    |> State.set partial_model_key None
+    |> State.set solver_ctx_key empty_solver_ctx
+    |> State.set is_decision_env false
+    |> DO.Mode.clear
+    |> DO.Optimize.clear
+    |> DO.ProduceAssignment.clear
+    |> DO.init
+    |> State.set named_terms Util.MS.empty
+  in
+
+  let handle_stmt_full_incremental :
     Frontend.used_context ->
     State.t ->
     [< Typer_Pipe.typechecked | `Check of 'a ] D_loop.Typer_Pipe.stmt ->
@@ -1248,17 +1263,7 @@ let main () =
             st
           end
 
-      | {contents = `Reset; _} ->
-        let () = Steps.reset_steps () in
-        st
-        |> State.set partial_model_key None
-        |> State.set solver_ctx_key empty_solver_ctx
-        |> State.set is_decision_env false
-        |> DO.Mode.clear
-        |> DO.Optimize.clear
-        |> DO.ProduceAssignment.clear
-        |> DO.init
-        |> State.set named_terms Util.MS.empty
+      | {contents = `Reset; _} -> handle_reset st
 
       | {contents = `Exit; _} -> raise Exit
 
@@ -1319,15 +1324,85 @@ let main () =
           "Ignoring statement: %a" Typer_Pipe.print td;
         st
   in
-  let handle_stmts all_context st l =
+
+  let handle_stmts_full_incremental all_context st l =
     let rec aux named_map st = function
       | [] -> State.set named_terms named_map st
       | stmt :: tl ->
-        let st = handle_stmt all_context st stmt in
+        let st = handle_stmt_full_incremental all_context st stmt in
         let named_map = add_if_named ~acc:named_map stmt in
         aux named_map st tl
     in
     aux (State.get named_terms st) st l
+  in
+
+  let current_path_key : stmt list State.key =
+    State.create_key ~pipe:"" "current_path"
+  in
+  let pushed_paths_key : stmt list Vec.t State.key =
+    State.create_key ~pipe:"" "pushed_paths"
+  in
+  let state_get key ~default st =
+    try State.get key st with State.Key_not_found _ -> default
+  in
+  let get_current_path = state_get current_path_key ~default:[] in
+  let get_pushed_paths = state_get pushed_paths_key ~default:(Vec.create ~dummy:(Obj.magic 0))
+  in
+  let handle_stmt_pop_reinit all_context st l =
+    let rec aux named_map st = function
+      | [] -> State.set named_terms named_map st
+      | (stmt: stmt) :: tl ->
+        begin
+          let current_path = get_current_path st in
+          match stmt with
+          | {contents = `Push n; _} ->
+            let pushed_paths = get_pushed_paths st in
+            for _ = 0 to n - 1 do
+              Vec.push pushed_paths current_path
+            done;
+            aux named_map st tl
+          | {contents = `Pop n; _} ->
+            let pushed_paths = get_pushed_paths st in
+            Format.eprintf "Vec size: %i@." pushed_paths.sz;
+            let rec pop_until until res =
+              if until = 0 then res else pop_until (until - 1) (Vec.pop pushed_paths)
+            in
+            let prefix = pop_until (n - 1) (Vec.pop pushed_paths) in
+            let st = handle_reset st in
+            let whole_path = List.rev_append prefix tl in
+            Format.eprintf "Restarting with %i instructions@." (List.length whole_path);
+            aux (State.get named_terms st) st whole_path
+          | {contents; _ } ->
+            let st =
+              let new_current_path =
+                match contents with
+                | `Clause _
+                | `Decls _
+                | `Defs _
+                | `Hyp _
+                | `Other _
+                | `Reset
+                | `Reset_assertions
+                | `Set_info _
+                | `Set_logic _
+                | `Set_option _ -> stmt :: current_path
+                | _ -> current_path
+              in
+              State.set current_path_key new_current_path st
+            in
+            let st = handle_stmt_full_incremental all_context st stmt in
+            let named_map = add_if_named ~acc:named_map stmt in
+            aux named_map st tl
+        end
+    in
+    aux (State.get named_terms st) st l
+  in
+  let handle_stmts all_context st l =
+    begin
+      if Options.get_imperative_mode ()
+      then handle_stmts_full_incremental
+      else handle_stmt_pop_reinit
+    end all_context st l
   in
   let d_fe filename =
     let logic_file, st = mk_state filename in
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -1174,7 +1174,6 @@ let main () =
 
   (* TODO: reset options to their initial value. *)
   let handle_reset st =
-    let () = Steps.reset_steps () in
     st
     |> State.set partial_model_key None
     |> State.set solver_ctx_key empty_solver_ctx
@@ -1408,6 +1407,8 @@ let main () =
       in
       let rev_prefix = pop_until (n - 1) (Vec.pop pushed_paths) in
       let st = handle_reset st in
+      (* Part of the reset, the current path must be reinitialized as well. *)
+      let st = State.set current_path_key [] st in
       replay st (List.rev rev_prefix)
     in
 
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -1173,7 +1173,7 @@ let main () =
   in
 
   (* TODO: reset options to their initial value. *)
-  let handle_reset st =
+  let reset_state st =
     st
     |> State.set partial_model_key None
     |> State.set solver_ctx_key empty_solver_ctx
@@ -1296,7 +1296,7 @@ let main () =
 
       | {contents = `Reset; _} ->
         let () = Steps.reset_steps () in
-        handle_reset st
+        reset_state st
 
       | {contents = `Exit; _} -> raise Exit
 
@@ -1415,7 +1415,7 @@ let main () =
           pop_until (until - 1) (Vec.pop pushed_paths)
       in
       let rev_prefix = pop_until (n - 1) (Vec.pop pushed_paths) in
-      let st = handle_reset st in
+      let st = reset_state st in
       (* Part of the reset, the current path must be reinitialized as well. *)
       let st = State.set current_path_key [] st in
       replay st (List.rev rev_prefix)
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -1150,26 +1150,26 @@ let main () =
   let handle_solve =
     let goal_cnt = ref 0 in
     fun st id contents loc attrs ->
-    let module Api = (val DO.SatSolverModule.get st) in
+      let module Api = (val DO.SatSolverModule.get st) in
       let file_loc = (State.get State.logic_file st).loc in
-    let id =
-      match (State.get State.logic_file st).lang with
-      | Some (Smtlib2 _) ->
-        DStd.Id.mk DStd.Namespace.term @@
-        "g_" ^ string_of_int (incr goal_cnt; !goal_cnt)
-      | _ -> id
-    in
-    let contents =
-      match contents with
-      | `Solve (hyps, []) -> `Check hyps
-      | `Solve ([], [t]) -> `Goal t
-      | _ ->
-        let loc = DStd.Loc.loc file_loc loc in
-        fatal_error "%a: internal error: unknown statement"
-          DStd.Loc.fmt loc
-    in
-    (* Performing the query *)
-    handle_query st id loc attrs contents
+      let id =
+        match (State.get State.logic_file st).lang with
+        | Some (Smtlib2 _) ->
+          DStd.Id.mk DStd.Namespace.term @@
+          "g_" ^ string_of_int (incr goal_cnt; !goal_cnt)
+        | _ -> id
+      in
+      let contents =
+        match contents with
+        | `Solve (hyps, []) -> `Check hyps
+        | `Solve ([], [t]) -> `Goal t
+        | _ ->
+          let loc = DStd.Loc.loc file_loc loc in
+          fatal_error "%a: internal error: unknown statement"
+            DStd.Loc.fmt loc
+      in
+      (* Performing the query *)
+      handle_query st id loc attrs contents
   in
 
   (* TODO: reset options to their initial value. *)
@@ -1204,15 +1204,15 @@ let main () =
         contents = (`Goal _) as contents;
         implicit = _;
       } ->
-        (* In the non imperative mode, the Solve instruction is handled differently
-           (i.e. no pop/push). *)
+        (* In the non imperative mode, the Solve instruction is handled
+           differently (i.e. no pop/push). *)
         assert (not (Options.get_imperative_mode ()));
         cmd_on_modes st [Assert; Sat; Unsat] "goal";
         let st = pop_if_post_query st in
         (* Pushing the environment once. This allows to keep a trace of the old
            environment in case we want to assert afterwards.
-           The `pop` instruction is handled by the hook on the mode: when we assert
-           anything, we must make sure to go back to `Assert` mode. *)
+           The `pop` instruction is handled by the hook on the mode: when we
+           assert anything, we must make sure to go back to `Assert` mode. *)
         let st = push_before_query st in
         handle_query st id loc attrs contents
 
@@ -1249,15 +1249,15 @@ let main () =
       (* When the next statement is a goal, the solver is called and provided
          the goal and the current context *)
       | { id; contents = (`Solve _ as contents); loc ; attrs; implicit=_ } ->
-        (* In the non imperative mode, the Solve instruction is handled differently
-           (i.e. no pop/push). *)
+        (* In the non imperative mode, the Solve instruction is handled
+           differently (i.e. no pop/push). *)
         assert (not (Options.get_imperative_mode ()));
         cmd_on_modes st [Assert; Unsat; Sat] "check-sat";
         let st = pop_if_post_query st in
         (* Pushing the environment once. This allows to keep a trace of the old
            environment in case we want to assert afterwards.
-           The `pop` instruction is handled by the hook on the mode: when we assert
-           anything, we must make sure to go back to `Assert` mode. *)
+           The `pop` instruction is handled by the hook on the mode: when we
+           assert anything, we must make sure to go back to `Assert` mode. *)
         let st = push_before_query st in
         handle_solve st id contents loc attrs
 
@@ -1419,8 +1419,8 @@ let main () =
       State.set is_decision_env true st
     in
 
-    (* The pop corresponding to the previous push. It must be applied everytime the
-       mode goes from Sat/Unsat to Assert. *)
+    (* The pop corresponding to the previous push. It must be applied everytime
+       the mode goes from Sat/Unsat to Assert. *)
     let rec pop_if_post_query st =
       if State.get is_decision_env st
       then pop 1 st (aux Util.MS.empty)
@@ -1450,7 +1450,9 @@ let main () =
             let st = pop_if_post_query st in
             let st = push_before_query st in
             handle_query st id loc attrs contents
-          | { id; contents = (`Solve _ as contents); loc ; attrs; implicit=_ } ->
+          | {
+            id; contents = (`Solve _ as contents); loc ; attrs; implicit=_
+          } ->
             cmd_on_modes st [Assert; Unsat; Sat] "check-sat";
             let st = pop_if_post_query st in
             let st = push_before_query st in
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -431,21 +431,16 @@ let main () =
 
   (* Before each query, we push the current environment. This allows to keep a
      fresh one for the next assertions. *)
-  let internal_push st =
-    if State.get is_decision_env st then
-      (* We already performed a check-sat *)
-      st
-    else
-      begin
-        let module Api = (val (DO.SatSolverModule.get st)) in
-        Api.FE.push 1 Api.env;
-        State.set is_decision_env true st
-      end
+  let push_before_query st =
+    assert (not (State.get is_decision_env st));
+    let module Api = (val (DO.SatSolverModule.get st)) in
+    Api.FE.push 1 Api.env;
+    State.set is_decision_env true st
   in
 
   (* The pop corresponding to the previous push. It is applied everytime the
      mode goes from Sat/Unsat to Assert. *)
-  let internal_pop st =
+  let pop_if_post_query st =
     if State.get is_decision_env st then
       begin
         let module Api = (val (DO.SatSolverModule.get st)) in
@@ -641,7 +636,7 @@ let main () =
     DO.Mode.add_hook
       (fun _ ~old:_ ~new_ st ->
          match new_ with
-         | Assert -> internal_pop st
+         | Assert -> pop_if_post_query st
          | _ -> st
       )
   in
@@ -1070,12 +1065,12 @@ let main () =
      the hook on D_state_option.Mode. *)
   let handle_query st id loc attrs contents =
     let module Api = (val (DO.SatSolverModule.get st)) in
-    let st = internal_pop st in
+    let st = pop_if_post_query st in
     (* Pushing the environment once. This allows to keep a trace of the old
        environment in case we want to assert afterwards.
        The `pop` instruction is handled by the hook on the mode: when we assert
        anything, we must make sure to go back to `Assert` mode. *)
-    let st = internal_push st in
+    let st = push_before_query st in
     let st_loc =
       let file_loc = (State.get State.logic_file st).loc in
       dl_to_ael file_loc loc
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -1199,9 +1199,12 @@ let main () =
         in
         st
 
-      | {contents = `Decls l; _} ->
+      | {contents = `Decls l; loc; _} ->
         let st = DO.Mode.set Util.Assert st in
-        D_cnf.cache_decls l;
+        let decls = D_cnf.make_decls l in
+        let module Api = (val DO.SatSolverModule.get st) in
+        let loc = dl_to_ael file_loc loc in
+        List.iter (fun decl -> Api.FE.declare ~loc decl Api.env) decls;
         st
 
       (* When the next statement is a goal, the solver is called and provided
@@ -1371,7 +1374,7 @@ let main () =
             let st = handle_reset st in
             let whole_path = List.rev_append prefix tl in
             Format.eprintf "Restarting with %i instructions@." (List.length whole_path);
-            aux (State.get named_terms st) st whole_path
+            aux Util.MS.empty st whole_path
           | {contents; _ } ->
             let st =
               let new_current_path =
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -488,7 +488,10 @@ let main () =
       recoverable_error "%t" msg; st
     | Util.Timeout ->
       Printer.print_status_timeout None None None None;
-      exit_as_timeout ()
+      if (not(Options.get_timelimit_per_goal ())) then
+        exit_as_timeout ()
+      else
+        st
     | Errors.Error e ->
       recoverable_error "%a" Errors.report e;
       st
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -1294,7 +1294,9 @@ let main () =
             st
           end
 
-      | {contents = `Reset; _} -> handle_reset st
+      | {contents = `Reset; _} ->
+        let () = Steps.reset_steps () in
+        handle_reset st
 
       | {contents = `Exit; _} -> raise Exit
 
@@ -1386,6 +1388,13 @@ let main () =
      this, but it is not the case yet. *)
   let handle_stmt_pop_reinit all_context st l =
 
+    let has_timeout st =
+      let module Api = (val DO.SatSolverModule.get st) in
+      match Api.SAT.get_unknown_reason Api.env.sat_env with
+      | Some (Timeout _) -> true
+      | _ -> false
+    in
+
     (* Pushes n times the current path. *)
     let push n st =
       let current_path = get_current_path st in
@@ -1419,7 +1428,6 @@ let main () =
       push 1 st;
       State.set is_decision_env true st
     in
-
     (* The pop corresponding to the previous push. It must be applied everytime
        the mode goes from Sat/Unsat to Assert. *)
     let rec pop_if_post_query st =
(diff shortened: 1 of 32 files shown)


src/lib/reasoners/satml_frontend.ml
```diff
@@ -1303,7 +1303,7 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
 
   let assume env gf _dep =
     (* dep currently not used. No unsat-cores in satML yet *)
-    assert (SAT.decision_level env.satml == 0);
+    (* assert (SAT.decision_level env.satml == 0); *)
```
Stevendeo: I commented these assertions because one test failed and I forgot to put them back; the test `./alt-ergo -o smtlib2 --prelude tests/cram.t/prelude.ae tests/cram.t/postlude.smt2` still fails because of it

bclement-ocp: Yeah because there are `push`'d level now. We should probably be able to expose the "push level" and check that instead (i.e. assertion that the decision level is equal to the length of the increm_guard). Probably as a `SAT.can_assume env.satml` check or some such.

src/bin/common/solving_loop.ml
```diff
@@ -230,7 +279,7 @@ let main () =
         { state with solver_ctx = { state.solver_ctx with global = cnf; }}
       | Typed.TAxiom (_, s, _, _) when Ty.is_local_hyp s ->
         let cnf = Cnf.make state.solver_ctx.local td in
-        { state with solver_ctx = { state.solver_ctx with local = cnf; }}
+        { state with solver_ctx = {state.solver_ctx with local = cnf; }}
```
Halbaroth: Is it an artifact?

tests/cram.t/prelude.ae
```diff
@@ -10,5 +10,5 @@ logic f : real -> real
 
 theory F_theory extends FPA =
   axiom f_triggers :
-    forall x : real [ int_floor(x), x in ]0., ?] ]. my_leq(f(x), x)
+    forall x: real [ f(x), f(x) in ]0., ?] ]. my_leq(f(x), x)
```
Halbaroth, line 13: Why did you change this test?

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1303,8 +1302,8 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
 
   let assume env gf _dep =
     (* dep currently not used. No unsat-cores in satML yet *)
-    assert (SAT.decision_level env.satml == 0);
-    try ignore (assume_aux ~dec_lvl:0 env [add_guard env gf])
+
+    try ignore (assume_aux ~dec_lvl:(SAT.decision_level env.satml) env [add_guard env gf])
```
Halbaroth: We should benchmark this PR to be sure this modification is ok :)

src/lib/reasoners/satml_frontend.ml
Stevendeo: Would it be correct not to assert at the decision level `0`, but at `decision_level env`?

tests/cram.t/prelude.ae
Stevendeo: With the changes of this PR, the `int_floor` was triggered during the first `check-sat` ; I changed the trigger to make sure it was independent from the first push/pop.

src/lib/reasoners/satml_frontend.ml
Stevendeo: >expose the "push level"

Both my solution and yours may be incorrect actually. 

```
(assert f) ; decision level = 0, push level = 0
(push 42) ; decision level = 0, push level = 42
(check-sat)
```

Stevendeo: Maybe a counter for every query?

bclement-ocp: Yeah I realized that after (because for the first `check-sat` it will be `0` as it used to be). The correct test I believe is `decision_level <= push_level` (to also capture both `(check-sat)(push)(check-sat)` and `(push)(check-sat)(pop)(check-sat)`).

src/bin/common/solving_loop.ml
```diff
@@ -121,13 +121,62 @@ let add_if_named
...
+        "This file is known to be Sat but Alt-Ergo return Unsat";
+      Errors.warning_as_error ()
+    end
+  | Sat _ ->
+    if known_status == Status_Unsat then begin
+      Printer.print_wrn
+        "This file is known to be Unsat but Alt-Ergo return Sat";
+      Errors.warning_as_error ()
+    end
+  | Unknown _ -> assert false
```
bclement-ocp: Instead of having an `assert false` here, you could have two separate functions `check_unsat_status` and `check_sat_status`.

src/bin/common/solving_loop.ml
```diff
@@ -834,79 +937,298 @@ let main () =
...
+
+      (* Goal definition *)
+      | {
+        id; loc; attrs;
+        contents = (`Goal _) as contents;
+        implicit = _;
+      } ->
+        cmd_on_modes st [Assert; Sat; Unsat] "goal";
+        (* Setting the mode is done by handle_query. *)
+        handle_query st id loc attrs contents
```
bclement-ocp: I believe this cannot happen anymore as the `` `Goal`` case is created by Alt-Ergo and never produced by Dolmen.

src/bin/common/solving_loop.ml
```diff
@@ -524,6 +595,20 @@ let main () =
...
+     to the Sat or Unsat mode we end up in. If we start asserting again, we must
+     do it in the previous environment.
+  *)
+  let () =
+    DO.Mode.reset_hooks ();
+    DO.Mode.add_hook
+      (fun _ ~old ~new_ st ->
+         match old, new_ with
+         | (Sat | Unsat), Assert -> internal_pop st
+         | _ -> st
```
bclement-ocp: I am quite uncomfortable with this approach. Implicitly popping the state when we switch mode seems brittle, because we need to be very careful with the way state transitions are handled. We kind of have to anyways, but I'd very much rather have code like `prepare_for_query st` or `pop_model st` or the like to explicitly perform the check rather than an innocuous-looking `DO.Mode.set Assert st` that needs accompanying comments to clarify it is only there to trigger hooks. Just call the code.

src/bin/common/solving_loop.ml
```diff
@@ -354,6 +408,23 @@ let main () =
...
+    Steps.apply_without_step_limit (fun _ -> Api.FE.push 1 Api.env);
+    st
+  in
+
+  (* The pop corresponding to the previous push. It is applied everytime the
+     mode goes from Sat/Unsat to Assert. *)
+  let internal_pop st =
+    assert (List.mem (DO.Mode.get st) [Sat; Unsat]);
+    let module Api = (val (DO.SatSolverModule.get st)) in
+    Steps.apply_without_step_limit (fun _ -> Api.FE.pop 1 Api.env);
```
bclement-ocp: Do we need to remove the step limit here? I think it was not done previously for `pop`.

src/bin/common/solving_loop.ml
```diff
@@ -354,6 +408,23 @@ let main () =
     State.create_key ~pipe:"" "is_incremental"
   in
 
+  (* Before each query, we push the current environment. This allows to keep a
+     fresh one for the next assertions. *)
+  let internal_push st =
+    let module Api = (val (DO.SatSolverModule.get st)) in
+    Steps.apply_without_step_limit (fun _ -> Api.FE.push 1 Api.env);
```
bclement-ocp: I believe `FE.push` already does `apply_without_step_limit`.

src/bin/common/solving_loop.ml
```diff
@@ -682,61 +767,79 @@ let main () =
...
-  let handle_optimize_stmt ~is_max loc id (term : DStd.Expr.Term.t) st =
-    let contents = `Optimize (term, is_max) in
-    let stmt = { Typer_Pipe.id; contents; loc; attrs = []; implicit = false } in
-    let cnf =
-      D_cnf.make (State.get State.logic_file st).loc
-        (State.get solver_ctx_key st).ctx stmt
-    in
-    (* Using both optimization and incremental mode may be wrong if
-       some optimization constraints aren't at the toplevel.
-       See issue: https://github.com/OCamlPro/alt-ergo/issues/993. *)
```
bclement-ocp, line 700: Please don't remove comments pointing to issues that are not solved.

src/bin/common/solving_loop.ml
```diff
@@ -682,61 +767,79 @@ let main () =
...
-      handle_optimize_stmt ~is_max:true loc id term st
+      let expr =
+        D_cnf.make_expr
+          ~loc
+          ~name_base
+          ~toplevel:true
+          ~decl_kind:Expr.Dobjective
+          term
+      in
+      handle_optimize_stmt ~is_max:true loc expr st
```
bclement-ocp: I guess these two cases can be factorized with `("minimize" | "maximize") as name_base` and `let is_max = match name_base with "minimize" -> false | "maximize" -> true | _ -> assert false in`.

src/bin/common/solving_loop.ml
```diff
@@ -834,79 +937,298 @@ let main () =
       assignments
   in
 
+  (* Copied from D_cnf, this treats the `Hyp case. *)
```
bclement-ocp, line 974: Was this copied from `D_cnf` or moved from there? If it was copied, can we expose it from `D_cnf` with the appropriate refactoring if needed to avoid duplication?

src/bin/common/solving_loop.ml
```diff
@@ -834,79 +937,298 @@ let main () =
...
       let file_loc = (State.get State.logic_file st).loc in
-      let solver_ctx = State.get solver_ctx_key st in
       match td with
+      (* Set logic *)
       | { contents = `Set_logic _; _} ->
         cmd_on_modes st [Start] "set-logic";
         DO.Mode.set Util.Assert st
+
+      (* Push and Pop commands *)
+      | { contents = `Pop n; loc; _ } ->
```
bclement-ocp: This still needs to update the `incremental_depth` for the optimization compatibility check (see #993 ).

src/lib/reasoners/satml.ml
```diff
@@ -999,9 +1004,11 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
   let print_aux fmt hc =
     Format.fprintf fmt "%a@," Atom.pr_clause hc
 
+  let is_unsat env : unit = env.is_unsat <- true
```
bclement-ocp: I know we have the confusing `is_mine` already but let's not add functions called `is_XXX` that do anything else than return whether their argument satisfies the "XXX" property. If this function is really needed, call it `set_unsat` (but I find `env.is_unsat <- true` to be more clear; having a function for that kind of implies that there is something more going on).

src/lib/reasoners/fun_sat.ml
```diff
@@ -1654,14 +1642,29 @@ module Make (Th : Theory.S) = struct
...
           in
+          let gamma =
+            (* If we made a check-sat and we want to work again on the
+               environment, we have to remove the guard from gamma. *)
+            match ME.find_opt guard_to_neg acc.gamma with
+            | None -> acc.gamma
+            | Some (_, _, dlvl, plvl) ->
+              ME.filter
+                (fun _ (_, _, dlvl', plvl') ->
+                   not (dlvl = dlvl' && plvl' >= plvl))
```
bclement-ocp, line 1685: Shouldn't we also remove things at later decision levels? I'd have expected `not (dlvl' > dlvl || (dlvl' = dlvl && plvl' >= plvp))`.

src/lib/reasoners/fun_sat.ml
```diff
@@ -1637,13 +1632,6 @@ module Make (Th : Theory.S) = struct
       ~init:env
 
   let pop env to_pop =
-    if Options.get_tableaux_cdcl () then
-      Errors.run_error
-        (Errors.Unsupported_feature
-           "Incremental commands are not implemented in \
-            Tableaux(CDCL) solver ! \
-            Please use the Tableaux or CDLC SAT solvers instead"
-        );
```
bclement-ocp, line 1660: I thought this was still not working — just commenting here to make sure we don't merge until either this works with Tableaux(CDCL) (probably not worth it, waiting on @Halbaroth investigation whether we should remove Tableaux(CDCL)) or to keep the assert back in.

src/lib/frontend/d_cnf.mli
```diff
@@ -44,22 +44,71 @@ val dty_to_ty : ?update:bool -> ?is_var:bool -> D_loop.DStd.Expr.ty -> Ty.t
...
+val cache_decls : D_loop.Typer_Pipe.decl list -> unit
+
+(* val make :
+ *   D_loop.DStd.Loc.file ->
+ *   Commands.sat_tdecl list ->
+ *   [< D_loop.Typer_Pipe.typechecked
+ *   | `Goal  of Dolmen.Std.Expr.term
+ *   | `Check of Dolmen.Std.Expr.term list
+ *          > `Hyp ] D_loop.Typer_Pipe.stmt ->
+ *   Commands.sat_tdecl list *)
```
bclement-ocp:
```suggestion
```

src/lib/frontend/d_cnf.mli
```diff
@@ -44,22 +44,71 @@ val dty_to_ty : ?update:bool -> ?is_var:bool -> D_loop.DStd.Expr.ty -> Ty.t
...
+      simple definitions (transformed in `Assume);
+    - Type aliases (filtered out);
+    - Statements used in models (they must not be in the input list, otherwise
+      this function fails). *)
+val make_defs :
+  D_loop.Typer_Pipe.def list ->
+  Loc.t ->
+  [> `Assume of string * Expr.t  | `PredDef of Expr.t * string ] list
+
+(** [mk_expr ~loc ~name_base ~toplevel ~decl_kind term]
```
bclement-ocp:
```suggestion
(** [make_expr ~loc ~name_base ~toplevel ~decl_kind term]
```

src/lib/frontend/d_cnf.mli
```diff
@@ -44,22 +44,71 @@ val dty_to_ty : ?update:bool -> ?is_var:bool -> D_loop.DStd.Expr.ty -> Ty.t
       individually.
 *)
 
+(** [ make_defs dlist loc ]
```
bclement-ocp:
```suggestion
(** [make_defs dlist loc]
```

src/bin/common/solving_loop.ml
bclement-ocp: (I believe it was moved, in which case "This treats the `Hyp case" is enough here)

src/lib/reasoners/fun_sat_frontend.ml
```diff
@@ -39,6 +39,8 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
 
   let empty () = ref (FS.empty ())
 
+  let cdcl_tab_push_pop_stack = Stack.create ()
```
bclement-ocp: This stack cannot be implemented as a global variable; it needs to be part of the environment.

(Also, I don't think it will work since the Tableaux(CDCL) solver uses a mutable CDCL solver internally)

bclement-ocp:
Looks like there are conflict, rebase would be appreciated :)

src/bin/common/solving_loop.ml
Stevendeo: This code comes from a copy/paste, do you think we should do an update in this PR?

src/bin/common/solving_loop.ml
Stevendeo: It is harder to make sure the `pop` is called each time we change mode than adding a hook. Besides, the exact semantic of the internal push/pop is precisely to be called whenever we switch mode. Of course we will trigger errors when we forget to change modes, but that is IMO other issues. I have tried to document the code as well as possible about this.

Another solution I have thought about is to add another mode: `Computing` between `Assert` and `Sat`/`Unsat`. When going from `Assert` to `Computing`, we perform an internal push ; when we go from `Sat`/`Unsat` to `Assert`, we perform an internal pop. That way, all calls to internal pop/push are centralized in hooks, and we can even add a flag in the state to enforce the consistency.

src/bin/common/solving_loop.ml
Stevendeo: I had an issue when testing the push/pop test in the smtlib tests, I can retry to check if this was indeed triggered by this pop or not.

src/bin/common/solving_loop.ml
bclement-ocp: If it is copied it might be fine but minor cleanups on copied code is fine as well, your call.

src/bin/common/solving_loop.ml
bclement-ocp: Nope that is not true, sorry, I confused `` `Goal`` and `` `Check`` I believe.

src/lib/reasoners/satml_frontend.ml
Stevendeo: Here's another solution: at the end of a query, keep track of the current decision level and set it as the latest known decision level on the `satml_frontend` side. I find it hard to justify keeping the invariant `decision_level <= push_level` for a consistency check (in other words, if we ever raise this assertion during a pr, I'm not sure how relevant it would be for debugging).

bclement-ocp: > I find it hard to justify keeping the invariant decision_level <= push_level for a consistency check (in other words, if we ever raise this assertion during a pr, I'm not sure how relevant it would be for debugging).

This is not just a consistency check; I am not sure the solver will do the right thing in this case. In particular, I am not sure the assumed assertions cannot be forgotten upon later backtracking — in fact, I am pretty sure they will be forgotten (partially) for the default CDCL-Tableaux solver given my current understanding of `lazy_cnf`.

src/bin/common/solving_loop.ml
bclement-ocp: > the exact semantic of the internal push/pop is precisely to be called whenever we switch mode

I disagree: there is now an additional transition to the `Assert` mode when calling `(check-sat)` which is not present in the SMT-LIB automaton, which muddies the semantics. My issue is exactly that the internal push/pop thing doesn't map properly to the existing state machine, which is the part that makes me uncomfortable (I should have commented there rather than here, sorry).

I kind of like having a separate `Computing` or `Check` state because it makes the semantics clear (if we are in `Computing` or `Check` we have the internal push, otherwise we don't — as you say, it would make the internal push/pops actually follow the state transitions), but I also don't really like introducing a new mode for this because it also makes the mode deviate from the SMT-LIB automaton it is supposed to represent.

I am aware that not calling `DO.Mode.set Assert` there would by risky and could cause double pops. As an alternative to the `Computing` state, what do you think of having a separate field `has_local_state`? We'd set it to `true` when an internal push is needed (due to the presence of local hypotheses or assertions in `Solve` — but we wouldn't for `` `Solve ([], [])`` which is `(check-sat)`); if it is true when entering `Assert` mode we do an `internal_pop` and set it to `false` (so that it is always `false` in `Assert` mode); if it is true in `handle_query`, we do an `internal_pop` and set it to `false` as well. This way, the hooks are used to enforce changes that are part of the SMT-LIB automaton, and explicit calls to `internal_push` and `internal_pop` are made otherwise.

src/lib/reasoners/satml_frontend.ml
bclement-ocp: (Note that for decisions level between `0` and `push_level` forgetting assumptions upon backtracking is OK because that is exactly the semantics we want for `push` and `pop`)

src/bin/common/solving_loop.ml
Stevendeo: Tests told you were right ^^ maybe we should add tests for alt-ergo files with dolmen?
I believe `Goal can still be created when Dolmen parses .ae files, doesn't it?

bclement-ocp: All tests are run with Dolmen already, including `.ae` files.

I believe that Dolmen now creates `` `Solve`` entries, not `` `Goal`` entries, for Alt-Ergo native inputs — but `` `Goal`` entries are created for other input formats (TPTP I think). Which we don't officially support so… I guess we could remove (on the other hand if it is 3 lines to support… why not).

src/bin/common/solving_loop.ml
Stevendeo: > I disagree: there is now an additional transition to the Assert mode when calling (check-sat) which is not present in the SMT-LIB automaton, which muddies the semantics. 

You're right, I shouldn't do this transition and have a flag that states whether I already have pushed or not; which is equivalent to the `has_local_state` field you are describing.

Stevendeo:
I'm not sure why, but the tests do not pass on the CI, but `make runtest-ci` do work on my side.

bclement-ocp:
> I'm not sure why, but the tests do not pass on the CI, but `make runtest-ci` do work on my side.

Weird. I can't reproduce either. But on the CI it seems surprisingly reproducible — all runners on all OSes fail. Maybe a cache problem, somehow?

Stevendeo:
I'm going to push some commits to this branch to test what's wrong with the tests on the CI, please don't take them into account

Stevendeo:
Here's the CI's result : 
```
Fatal error: exception File "src/lib/reasoners/satml.ml", line 1990, characters 9-15: Assertion failed
```

There is no line 1990 in `satml.ml`...

Stevendeo:
Looks like CI tests were made on a hybrid version of this branch and the next branch, with a custom merge performed on the two branches...

Stevendeo:
Rebased! Not sure we should merge it as it is though; or as we said conditionally with a new option. I'm not sure it is a good idea though, or we will need to test alt-ergo twice every time (once without the option and once with) to make sure we don't break anything.

bclement-ocp:
Thanks! Will do another round of review tomorrow if I can find the time and next week otherwise.

> I'm not sure it is a good idea though, or we will need to test alt-ergo twice every time (once without the option and once with) to make sure we don't break anything.

I don't think we need to test both with and without the option. The way I see it:

 - This is a feature that we definitely want and will ultimately include by default once we solve the performance issues
 - If we don't merge this, the code will rot in a separate branch and when we want to pick up this work and solve the performance issues, we will get lost in many merge conflicts. In addition to the merge conflicts, we will have to fix any bugs that have been introduced.
 - If we do merge this, even without testing it, we might miss bugs in the incremental solver that we will have to fix those bugs before making it the default, but at least there will be no code drift and no painful merge conflict.

So even without testing both options having the code in the main branch will make things easier for us later, and it also allows us to implement features that depend on incrementality working (e.g. #1032 ) internally even if we don't use incrementality in the main solving loop.

Stevendeo:
Do you prefer we have two different functions for the main solving loop or have the two modes interlaced in the same function ?

bclement-ocp:
I think I'd rather avoid the two modes interlaced in the same function if possible, that sounds very hard to maintain (although maybe it would motivate us to fix the performance issues faster :sweat_smile: ).

The best would be to have a common incremental interface with an implementation that is actually incremental, and another implementation that resets on `pop`. I don't remember the PR well enough to know if that's easy to do or not. If not we can duplicate the function but wouldn't that be a lot of duplicated code? Otherwise maybe we can keep only the various fixes from this PR but not the solving loop changes, if it's not too much work to separate them?

Stevendeo:
As I spent some time rebasing this PR, I'd rather be sure of what we're going to do. I don't mind keeping it updated while we extract the bugfixes we want.
I think I can go for a incremental version with the pop reinitialization (which would be similar to the command version), but it is possible we still need the Commands version in case the pop-reset one triggers unexpected bugs/regressions.

bclement-ocp:
> As I spent some time rebasing this PR, I'd rather be sure of what we're going to do. I don't mind keeping it updated while we extract the bugfixes we want.

I agree (and thanks for doing the rebase work). I think that this boils down to how much time/complexity we estimate each approach would take.

> I think I can go for a incremental version with the pop reinitialization (which would be similar to the command version), but it is possible we still need the Commands version in case the pop-reset one triggers unexpected bugs/regressions.

Would the pop reinitialization be more complex than pushing the `Commands` (more precisely the `solver_ctx` type from the solving loop) down into that interface? I think the risk of having performance regressions due to that should be low (although we are never safe from unexpected bugs), so I'd say we go for this option unless you think it's too complex/risky.

That said I am now thinking: if I remember correctly the main reason for not merging was performance of the incremental implementation. Did we try simply clearing the set of learned clauses when `pop` is called? This might be enough to eliminate the performance differences with the incremental implementation?

(We lose some of the benefits of incrementality but that's OK, and easy to make an option for)

Stevendeo:
> Would the pop reinitialization be more complex than pushing the `Commands` (more precisely the `solver_ctx` type from the solving loop) down into that interface? 

The Commands version would only require to take the old function and put it back I think, so it would be quite straightforward. In the case of reinitialization, I would need to implement a stmt memorisation mechanism similar to the Commands one, but that should be quite easy (and it would actually be interesting to see if this new version actually breaks something).

> That said I am now thinking: if I remember correctly the main reason for not merging was performance of the incremental implementation. Did we try simply clearing the set of learned clauses when `pop` is called? This might be enough to eliminate the performance differences with the incremental implementation?

I don't remember actually. I think the performance issue came from the push/pop involving extra implications of guards.

Stevendeo:
Okay I finished working on the pop-reinitialization version. What it does now is to rework all the analysis (removing the smt commands we don't want to be executed twice) when a pop occurs. I will add some documentation to explain what's going on, but at least tests are OK on my machine.

bclement-ocp:
> What it does now is to rework all the analysis (removing the smt commands we don't want to be executed twice) when a pop occurs

Ah so this uses a `stmt list` instead of `command list`? That sounds smart.

bclement-ocp:
> I will add some documentation to explain what's going on, but at least tests are OK on my machine.

You marked this as ready for review but also noted this; should I wait for that documentation to be written before doing a review or do you want a pass now?

Stevendeo:
I thought I was ready modulo documentation, but as I wrote documentation I noticed a few issues remaining (cf CI). Ready for review in a minute

Stevendeo:
Ready! For real this time

src/lib/util/options.mli
```diff
@@ -660,6 +663,10 @@ val get_type_only : unit -> bool
 val get_type_smt2 : unit -> bool
 (** Default to [false] *)
 
+(** [true] if the program shall stop after SMT2 typing. *)
```
bclement-ocp: Copy-paste gone wrong

src/lib/util/options.mli
```diff
@@ -298,6 +298,9 @@ val set_triggers_var : bool -> unit
 (** Set [type_smt2] accessible with {!val:get_type_smt2} *)
 val set_type_smt2 : bool -> unit
 
+(** Set [type_smt2] accessible with {!val:get_type_smt2} *)
```
bclement-ocp: Copy-paste gone wrong

src/lib/reasoners/satml.ml
bclement-ocp: Please call this function `set_unsat` or `set_is_unsat` :)

src/lib/reasoners/satml.ml
```diff
@@ -1027,10 +1039,27 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
            assert (ta.is_true);
            assert (ta.var.level >= 0);
            if ta.var.level = 0 then begin
-             incr nb_f;
```
bclement-ocp: Should not be removed

src/lib/reasoners/satml.ml
```diff
@@ -1027,10 +1039,27 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
...
              (ta.lit, Th_util.Other, Ex.empty, 0, env.cpt_current_propagations)
              :: acc
            end
+           (* This was first added by PR 1000, but it triggers a bug in
+              tests/cc/testfile-ac_cc002.ae with option
+              --no-tableaux-cdcl-in-instantiation *)
```
bclement-ocp, line 1044: If this code is removed then I believe the `unit_tenv_queue` field should also be removed (in fact I think it should be removed in any case, it was a quick-and-dirty hack to test an hypothesis if I recall — same goes for this code).

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1309,8 +1310,12 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
       Inst.add_terms env.inst
         (E.max_ground_terms_rec_of_form gf.E.ff) gf;
     try
-      assert (SAT.decision_level env.satml == 0);
-      let _updated = assume_aux ~dec_lvl:0 env [gf] in
+      let _updated =
```
bclement-ocp, line 1313: Should we assert that `decision_level <= push_level` here as well?

(If it's not possible I think we need to do something because we will forget the assertion on backtracking)

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1309,8 +1310,12 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+        assume_aux
+          ~dec_lvl:(SAT.decision_level env.satml)
```
bclement-ocp, line 1315: I think ideally this should be either `dec_lvl:0` or `dec_lvl:(SAT.push_level env.satml)` (the 2nd is probably better) here, but alas SatML would not support this with `--no-minimal-bj`.

src/lib/reasoners/fun_sat.ml
```diff
@@ -1651,13 +1655,15 @@ module Make (Th : Theory.S) = struct
...
+    (* It does not work with Tableaux(CDCL); right now it is (unsoundly)
+       handled by fun_sat_frontend but we should fix it. *)
+    (* if Options.get_tableaux_cdcl () then
```
bclement-ocp, line 1660: Please uncomment this code.

src/lib/reasoners/fun_sat.ml
```diff
@@ -1625,20 +1626,23 @@ module Make (Th : Theory.S) = struct
...
```
bclement-ocp, line 1631: Please uncomment this code.

src/lib/reasoners/fun_sat.ml
bclement-ocp: My comment above still stands. Is the simpler test enough because we pop levels one by one?

src/bin/common/solving_loop.ml
bclement-ocp: This still applies, please do not remove comments pointing to issues that are not solved.

src/lib/frontend/d_cnf.ml
```diff
@@ -176,6 +176,8 @@ type _ DStd.Builtin.t +=
   | Integer_round
   | Abs_real
   | Sqrt_real
+  | Int2BV of int
+  | BV2Nat of int
```
bclement-ocp, line 180: These have been removed (probably a rebase artifact)

src/bin/common/solving_loop.ml
```diff
@@ -996,43 +1324,136 @@ let main () =
...
+          | {contents = `Pop n; _} ->
+            let pushed_paths = get_pushed_paths st in
+            let rec pop_until until res =
+              if until = 0 then
+                res
+              else
+                pop_until (until - 1) (Vec.pop pushed_paths)
+            in
+            let prefix = pop_until (n - 1) (Vec.pop pushed_paths) in
+            let st = handle_reset st in
```
bclement-ocp: I don't think it is right to use `handle_reset` here; some things should not be reset (such as the step counter — we did do the previous steps and we should count them, even if we are redoing them).

src/bin/common/solving_loop.ml
```diff
@@ -356,6 +409,53 @@ let main () =
...
+  let pushed_paths_key : stmt list Vec.t State.key =
+    State.create_key ~pipe:"" "pushed_paths"
+  in
+
+  (* Before each query, we push the current environment. This allows to keep a
+     fresh one for the next assertions. *)
+  let internal_push st =
+    if State.get is_decision_env st then
+      (* We already performed a check-sat *)
+      st
```
bclement-ocp: If I am reading things correctly, I think `State.get is_decision_env st` can never be `true` here and we should `assert (not (State.get is_decision_env st))` here.

src/bin/common/solving_loop.ml
```diff
@@ -356,6 +409,53 @@ let main () =
...
+
+  (* Key used only in handle_stmt_pop_reinit, saves the commands executed
+     when a push is executed and replays them when a pop is met. *)
+  let pushed_paths_key : stmt list Vec.t State.key =
+    State.create_key ~pipe:"" "pushed_paths"
+  in
+
+  (* Before each query, we push the current environment. This allows to keep a
+     fresh one for the next assertions. *)
+  let internal_push st =
```
bclement-ocp: I think this could use a more explicit name such as `push_before_query` (especially since the `Frontend` module also has an `internal_push` function that behaves very differently).

src/bin/common/solving_loop.ml
```diff
@@ -356,6 +409,53 @@ let main () =
...
+      begin
+        let module Api = (val (DO.SatSolverModule.get st)) in
+        Api.FE.push 1 Api.env;
+        State.set is_decision_env true st
+      end
+  in
+
+  (* The pop corresponding to the previous push. It is applied everytime the
+     mode goes from Sat/Unsat to Assert. *)
+  let internal_pop st =
```
bclement-ocp: Same as above, could be `pop_after_query`

tests/adts/simple_1.ae
```diff
@@ -49,13 +49,15 @@ goal g_valid_4_2:
   e ? B ->
   false
 
-(**)
+(* Isse 1008: this goal fail when encapsulated by a push/pop. *)
```
bclement-ocp, line 52: Should we reinstate this test?

bclement-ocp (commented):
I did a first pass; I will do a second more thorough one later this week but there is still actionable feedback. I wish we had to extract less stuff from `D_cnf` but oh well.

We need to run benchmarks on both internal datasets (AE format and SMT format) in both modes, will queue them tonight

src/lib/reasoners/satml.ml
Stevendeo: I think it was important to restore the old unit_tenv from before a push (cf Satml.pop)

src/bin/common/solving_loop.ml
Stevendeo: Should `(reset)` reset the Steps counter actually?

bclement-ocp: Yes, the specification says:

> (reset) resets the solver completely to the state it had after it was started and before it started
reading commands.
> This allows the user to reset the state of the solver without paying the cost of restarting it.

Basically the goal is that calling alt-ergo on a script that does `<script1>(reset)<script2>` should behave identically as calling alt-ergo on `<script1>` then a separate instance on `<script2>`.

(It is not the case currently because we don't reset counters and weak tables, but in any case that is a different specification than what is needed for replaying statements in this PR and so the same function should not be used)

src/lib/reasoners/satml.ml
bclement-ocp: I believe this should only be the case with the code below uncommented — without this code, Alt-Ergo should only ever add to `unit_tenv` facts that are true at level `0` (otherwise it is probably a soundness bug) which shouldn't be impacted by push/pop.

(If I recall correctly, I added this code and the `unit_tenv` stack because `unit_tenv` with the specification of only adding things known at level `0` is less useful with incremental solving and we were trying to solve the performance issues)

Stevendeo:
I can't figure out why there is a test not working on the CI that works on my side. I guess this is because the CI is slow, but I'm not sure where the error is triggered...

bclement-ocp:
> I can't figure out why there is a test not working on the CI that works on my side. I guess this is because the CI is slow, but I'm not sure where the error is triggered...

The test is `testfile-qfbv-timeout.dolmen.smt2` so that is probably because this PR somehow breaks the timelimit-per-goal option.

PR #1004, opened 2023-12-07
refactor(BV): Track domains for uninterpreted leaves only

bclement-ocp:
This patch changes the internal representation of domains in the bitvector relations. Without the patch, we are keeping track of domains for *all* bit-vector terms, which is correct but wasteful. With the patch, we only track the domains of (currently) uninterpreted class representatives, and rebuild the domain associated with other terms on-the-fly before propagation.

This avoids carrying around huge domain maps with a potentially large amount of duplication, and allow to share propagations between terms that include identical sub-terms immediately without a round-trip through Uf (without the patch, if we learn that `0 @ x -> 01??`, it does not immediately propagate that `1 @ x -> 11??`; we first have to assert `x<2,2> = 1` which gets solved into the substitution `x |-> 1 @ x'` and we only learn `11 @ x' -> 11??` when performing the substitution). On the other hand, it means that `Domains.get` and `Domains.update` are a bit more expensive now.

Since the `Congruence` is now only used for constraints, and not for domains, we only need to add the arguments of the constraints to the `Congruence` (which should also be a slight performance boost).  Other changes to the use of the `Congruence` module are needed to fix the mismatch between what is stored in the `Domains` (whose keys are now uninterpreted terms) and the `Constraints` (the actual arguments of the constraints in the original problem). There is one exception: we now add the arguments of `Distinct` to the congruence, which is required for completeness because the terms that we generate during case splits are not guaranteed to already exist (in fact, in general they don't, since we create 1-bit extractions that have no reason to be in the original problem).

Note that there is also a small bugfix in `Congruence`: when performing a substitution we need to remove the old terms, otherwise we will forget dependencies if we ever add it back later.

Note: this is +13-4 on BV (benches are running on QF_BV); the regressions seem to be due to poor interaction with `bv2nat` which will be improved upon in the future.

### DIFF
--- a/src/lib/reasoners/bitv_rel.ml
+++ b/src/lib/reasoners/bitv_rel.ml
@@ -111,9 +111,6 @@ module Domains : sig
 
       Use this to ensure that the representation is always normalized.
 
-      [v] must have an existing associated domain (possibly unconstrained)
-      before calling [subst]. Call [update] beforehand.
-
       The explanation [ex] justifies the equality [p = v].
 
       @raise Bitlist.Inconsistent if this causes any domain in [d] to become
@@ -131,7 +128,12 @@ module Domains : sig
 end = struct
   type t =
     { bitlists : Bitlist.t MX.t
-    (** Mapping from semantic values to their bitlist domain. *)
+    (** Mapping from semantic values to their bitlist domain.
+
+        Note: this mapping only contains domain for leaves (i.e. uninterpreted
+        terms or AC symbols); domains associated with more complex semantic
+        values are rebuilt on-the-fly using the structure of said semantic
+        values. *)
     ; changed : SX.t
     (** Elements whose domain has changed since last propagation. *)
     }
@@ -144,7 +146,7 @@ end = struct
       ppf t.bitlists
   let empty = { bitlists = MX.empty ; changed = SX.empty }
 
-  let update ex r t bl =
+  let update_leaf ex r t bl =
     let changed = ref false in
     let bitlists =
       MX.update r (function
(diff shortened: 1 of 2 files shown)


bclement-ocp:
This is +36-179 on QF_BV, which is less good. From a cursory look at the problems, this seems to be (counter-intuitively) due to an improvement: because we have better propagators (now actually complete on ground inputs) Alt-Ergo degrades into full backtracking while doing splitting! This is confirmed by checking that the regressions (at least the 3-4 random ones I selected) are solved by doing either `--case-split-policy after-matching` or `--case-split-policy before-matching`.

I have opened #1005; in the meantime, I think this PR can be reviewed independently.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -171,45 +173,67 @@ end = struct
...
+        let bl = Bitlist.extract bl mid (Bitlist.width bl - 1) in
+
+        match bv with
+        | Bitv.Cte z ->
+          (* Nothing to update, but still check for consistency! *)
+          ignore @@ Bitlist.intersect bl (Bitlist.exact sz z Ex.empty) ex;
+          t, bl_tail
+        | Other r -> update_signed ex r t bl, bl_tail
+        | Ext (r, r_size, i, j) ->
+          (* r<i, j> = bl -> r = ?^(r_size - j - 1) @ bl @ ?^i *)
```
Halbaroth, line 197: I guess you mean: 
if the value `r<i, j>` is in the domain `bl` then `r` is in the domain 
`?^(r_size - j - 1) @ bl @ ?^i` ?

Halbaroth (approved):
LGTM but we shouldn't merge before solving #1005.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -171,45 +173,67 @@ end = struct
...
+
+  let update ex r t bl =
+    fst @@ List.fold_left (fun (t, bl) { Bitv.bv; sz } ->
+        (* Extract the bitlist associated with the current component *)
+        let mid = Bitlist.width bl - sz in
+        let bl_tail =
+          if mid = 0 then Bitlist.empty else
+            Bitlist.extract bl 0 (mid - 1)
+        in
+        let bl = Bitlist.extract bl mid (Bitlist.width bl - 1) in
```
Halbaroth, line 188: I don't understand why we extract the end of the bistlist here instead of a prefix of it. I would expect to use a prefix of `bl` as domain to intersect with the current domain of `bv`.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Yes, I kind abuse notations and use `bl` to denote an element of the domain here. I think it's a bit easier to think in terms of equality rather than set inclusion but I can change the `=` to `in`.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Bit-vectors are little endian, so if you have for instance `#b11110000` you might think of `#b1111` as being a prefix but the `#b1111` is actually the slice `4..7` (and `#b0000` is the slice `0..3`).

OTOH we represent compositions in the same order as for `concat` and so `(concat #b1111 #b0000) = #b11110000` is represented as the list `[#b1111; #b0000]` as a semantic value (well except that the `#b1111` and `#b0000` are actually merged during normalization but you get the idea).

bclement-ocp:
Following the dev meeting: since this is still a net positive compared to last release we should merge in spite of #1005 (partially because we want to merge #1010).

PR #1006, opened 2023-12-11
Make the `Heap` module polymorphic

bclement-ocp:
This is a rewriting of the `Iheap` module to apply to arbitrary elements (in practice, atoms). The implementation is intrusive: elements need to store their current index, and thus can only be part of a single heap.

The end result is more or less identical to the [Heap](https://github.com/Gbury/mSAT/blob/master/src/core/Heap.ml) module in mSAT.


bclement-ocp:
Note: I called the heap in SatML "Aheap" but really it should be "Vheap"; I will change the name and force-push with that single change.

bclement-ocp:
Note; the PR includes a couple cleanup changes in addition to the general `Heap` change that are somewhat related (they were written at the same time) but actually independent. These in their own commits; and I'd recommend reviewing each commit separately. I can make separate PRs for these small cleanups if needed; I only didn't do it because the commits are tiny.

src/lib/util/heap.ml
```diff
@@ -0,0 +1,183 @@
...
+  let[@inline] increase s x = assert (in_heap x); percolate_down s x
+
+  let filter ({ heap } as s) filt =
+    let j = ref 0 in
+    let lim = Vec.size heap in
+    for i = 0 to lim - 1 do
+      if filt (Vec.get heap i) then begin
+        Vec.set heap !j (Vec.get heap i);
+        Rank.set_index (Vec.get heap i) !j;
+        incr j;
```
Halbaroth: The previous code factorises the call `Vec.get`. Is there good reason to do three calls? As `Vec.get` doesn't perform bound checks (with assertions disabled), it doesn't matter probably but I'm not sure the compiler can optimise this code since it has to ensure the internal state of the vector `heap` doesn't change between each calls.

src/lib/util/heap.mli
```diff
@@ -28,48 +28,79 @@
...
+
+  val insert : t -> elt -> unit
+  (** Inset a new element in the heap. *)
+
+  val grow_to_by_double: t -> int -> unit
+  (** Grow the size of the heap by multiplying it by 2
+      until it is at least the size specified. *)
+
+  val remove_min : t -> elt
+  (** Remove the minimum element from the heap and return it. *)
```
Halbaroth: Now `remove_min` will raise `Not_found`. We should mention it in its documentation.

src/lib/util/heap.ml
```diff
@@ -0,0 +1,183 @@
...
+    begin
+    if in_heap s n then
+    begin
+    percolate_up cmp s (Vec.get s.indices n);
+    percolate_down cmp s (Vec.get s.indices n)
+    end
+    else insert cmp s n
+    end;
+    assert (heap_property cmp s)
+  *)
```
Halbaroth: I know it's out of the scope of this PR but do you know what's the purpose of this snippet? We may get rid of it?

src/lib/util/heap.mli
```diff
@@ -28,48 +28,79 @@
...
 
-val size : t -> int
-(** Returns the current size of the heap. *)
+  val compare : t -> t -> int
+  (** A total ordering function over the set elements.
+      This is a two-argument function [f] such that
+      [f e1 e2] is zero if the elements [e1] and [e2] are equal,
+      [f e1 e2] is strictly negative if [e1] is smaller than [e2],
+      and [f e1 e2] is strictly positive if [e1] is greater than [e2]. *)
+end
```
Halbaroth, line 64: As you taught me, doing arrays of pairs isn't perform in both memory and access as well as doing pairs of arrays. We should benchmark this code.

src/lib/reasoners/satml_frontend.ml
```diff
@@ -539,26 +539,6 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
-    else
-      (* higher weight is better hence compare w2 w1 *)
-      let res = Stdlib.compare w2 w1 in
-      if res <> 0 then res
-      else
-        (* lower index is better *)
-        compare i1 i2
-
-  let max a b = if cmp_tuples a b > 0 then a else b
-
```
Halbaroth, line 561: Please, can you explain what was the purpose of this code and why we don't need it anymore?

bclement-ocp: I have no idea what was the purpose of this code but it is only used to compute some value in `atoms_from_sat_branch` which is then ignored (see `Some (_, l) -> ...` line 628).

src/lib/util/heap.mli
bclement-ocp: I am not sure this matters much here because this is not about arrays of pairs vs pairs of arrays, it is about replacing an indirection through an array (`indices.(i)`) by a direct access (`a.hindex`) so there is only one level of indirection in each case, the data is just stored differently (and we also don't iterate linearly on the array). I am OK with running benchmarks tho.

src/lib/util/heap.ml
bclement-ocp: Looks like this is code to change the value associated with an element but that we use the more precise `decrease` and `increase` instead, we can probably remove it.

src/lib/util/heap.ml
bclement-ocp: No good reason, I think I made the change to convince myself of correctness. Will revert.

src/lib/util/heap.mli
bclement-ocp: True, thanks for noticing!

bclement-ocp:
This inadvertently flipped the comparison order so that we were trying variables with low activity first, I have pushed a fix (see [here](https://github.com/OCamlPro/alt-ergo/compare/ab56c5db07652470f73183a4a2d68b015949c67f..c7d82700c96aba9430f46ce025911c8446fc2bf8)). Rerunning benches.

src/lib/util/heap.ml
Halbaroth: We can remove it.

src/lib/util/heap.mli
Halbaroth: I think you should run benchmarks at least to check we don't break an implicit invariant. Maybe another part of the code used the fact that vectors were sparse. It's unlikely but still let's check it :)

Halbaroth (approved):
Let's merge after benchmarking the PR :)

src/lib/util/heap.ml
bclement-ocp: This is done, sorry I wasn't clear.

bclement-ocp:
This is +1-0 (after correction) w/ no perf regressions, merging.

Stevendeo:
There is (at least) one undocumented assertion added by this PR:  in Satml.pop,  `assert (env.next_dec_guard = 0)`.
@bclement-ocp can you give some insights about it?

bclement-ocp:
That is weird, it shouldn't have made it to this PR. I don't remember if
that is an actual assertion or a debug artifact - I am at a working group
today but will check tomorrow. It should be true on next anyways but feel
free to comment it in #1000 in the meantime.

On 2023-12-14 13:50 +01:00 CET, "Stevendeo" ***@***.***>
wrote:
> There is (at least) one undocumented assertion added by this PR: in
Satml.pop, `assert (env.next_dec_guard = 0)`.

bclement-ocp:
OK, this is an assertion that ensures that the stack of guards and the stack of decisions are in the same order, essentially, but it does so too eagerly for #1000 — I will make a PR with a correct & documented version.

Issue #556, opened 2023-04-02
Get rid of the legacy front-end?


hra687261:
Alt-Ergo should only have one front-end, it's very annoying to have to do any changes to the front-end twice, once in Alt-Ergo, once in Dolmen.
If more testing needs to be done, then it should be a priority to do it and fix the eventual issues that would come up with switching to Dolmen.

Halbaroth:
I totally agree that we should get rid of it as soon as possible. But we should be sure the actually parser in Dolmen behaves closely like the legacy one. We have very few files in the native input format, which means we didn't test it properly. 
Maybe, we could produce some tests for Dolmen before removing the legacy parser?

We should also check the support of `psmt` in Dolmen.

hra687261:
We can probably generate a good amount of them from Why3's test-suite.
Also once printers become available in Dolmen we can translate some of them.
But I think it should be a priority, because it significantly slows down development whenever someones touches the frontend.

bclement-ocp:
Blocked by #585 and #586

hra687261:
Closing this issue since the legacy front end was removed with OCamlPro/alt-ergo#1251

PR #1009, opened 2023-12-12
Wrong destructor for mutually recursive ADT

Halbaroth:
The new Dolmen frontend didn't create the correct destructor in presence of mutually recursive ADT. For instance the input:
```smt2
(set-logic ALL)
(declare-datatypes ((Data1 0) (Data2 0))
  (
    ((cons1 (d1 Int) (d2 Data2)))
    (
      (cons2 (d3 Data1))
      (cons3)
    )
  )
)
(declare-const a Data1)
(declare-const b Data2)
(declare-const x Int)
(assert ((_ is cons1) a))
(assert ((_ is cons3) b))
(assert (= (d1 a) x))
(check-sat)
```
raised an assertion in the record theory.
The presence of a flaw in a test of `d_cnf` to discriminate between records and ADT produced `Sy.Access` destructors intead of `Sy.Destructor` destructors on some terms whose the type is a mutually recursive ADT. For instance `d1 a` was translated into `mk_term Sy.Access [a] ...`.

The fix consists in using the cache of AE types in `d_cnf` to discriminate between records and ADT.

Notice that the legacy frontend parses the previous file correctly.

### DIFF
--- a/src/lib/frontend/d_cnf.ml
+++ b/src/lib/frontend/d_cnf.ml
@@ -1081,18 +1081,19 @@ let rec mk_expr
 
           | B.Destructor { case; field; adt; _ }, [x] ->
             begin match DT.definition adt with
-              | Some (Adt { cases; record; _ }) ->
+              | Some (Adt { cases;  _ }) ->
                 begin match cases.(case).dstrs.(field) with
                   | Some { path; _ } ->
                     let name = get_basename path in
                     let ty = dty_to_ty term_ty in
                     let e = aux_mk_expr x in
                     let sy =
-                      if record || Array.length cases = 1
-                      then
+                      match Cache.find_ty (DE.Ty.Const.hash adt) with
+                      | Trecord _ ->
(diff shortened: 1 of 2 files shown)


bclement-ocp (approved):
I thought we were using the `record` flag to determine whether to create an ADT or a record. I guess we don't do it if they are in a mutually recursive nest because the legacy typechecker didn't either, and so the `record` flag is no longer correct there:

https://github.com/OCamlPro/alt-ergo/blob/f7ae6f78e2c6f28133464aa7e133c241b2276c89/src/lib/frontend/d_cnf.ml#L766-L768

I think the fix is sound but we might want to revisit this choice — it doesn't make much sense to ever use the ADT theory for types that are records (or enums, for that matter), even if they are mutually recursive with ADTs. Probably should wait until we remove the legacy frontend (#556) to avoid parity issues.

Halbaroth:
I noticed this comment. I agree it will be simpler to modify this after removing the legacy frontend.

Issue #1005, opened 2023-12-11
Case split policy may be too aggressive


bclement-ocp:
As observed in #1004, there are issues with the BV theory where once we have split enough to determine the exact value of a bit-vector, we find a contradiction. But we do so because the problem is unsat, and since the propagators are complete (should be, at least) on ground inputs, we end up doing an expensive exhaustive search in `look_for_sat`, which times out quickly.

Possible solutions:
 - #901 (might be worth looking at using #901 only for BV in a first time and migrating the other theories in a second step)
 - use different split policies for different theories
 - tweak the use of `--case-split-policy after-theory-propagate` (only split every N propagations?)

PR #1010, opened 2023-12-12
refactor(BV): Merge `Congruence` and `Constraints`

bclement-ocp:
In https://github.com/OCamlPro/alt-ergo/pull/944, the `Congruence` module was added to simplify handling dependences for both domains and constraints, because we tracked domains separately for all terms.

After https://github.com/OCamlPro/alt-ergo/pull/1004, we now track domains only for uninterpreted leaves, and the `Congruence` module is only used for `Constraints`. This leads to a sort of double indirection: the `Congruence` module keeps track of reverse uninterpreted leaves -> class representative dependencies, and then the `Constraints` module keeps track of reverse class representative -> constraint dependencies.

This patch removes the `Congruence` module entirely; instead, the `Constraints` module is now keeping track of reverse uninterpreted leaves -> constraint dependencies directly.

In addition, the `Constraints` module is made into a functor and moved to the `Rel_utils` module, since it is now a generic module that can be useful for any relational theory (like `Congruence` was before).

I recommend reviewing the PR commit by commit to better be able to follow the changes: the commit that does the actual changes in implementation of the `Constraints` module is the one titled "Fold `Congruence` into `Constraints`"; the others are mostly shuffling code around, which makes the global diffs hard to follow.

**Note**: This currently includes #1004.


bclement-ocp:
This is +1-0 on smtlib/BV; not sure it is worth running on QF_BV for this change since that takes a lot of time (maybe we should build a small subset of QF_BV).

Edit: +1-0 compared to #1004

src/lib/reasoners/rel_utils.ml
```diff
@@ -193,126 +193,179 @@ end = struct
...
+      {b Note}: The comparison function is arbitrary and has no semantic
+      meaning. You should not depend on any of its properties, other than it
+      defines an (arbitrary) total order on constraint representations. *)
 
-      It provides an interface to register some semantic values of interest, and
-      for applying a callback when the representative of those registered values
-      change.
-  *)
+  val subst : Explanation.t -> X.r -> X.r -> t -> t
+  (** [subst ex p v cs] replaces all the instaces of [p] with [v] in the
```
```suggestion
  (** [subst ex p v cs] replaces all the instances of [p] with [v] in the
```

src/lib/reasoners/rel_utils.ml
```diff
@@ -193,126 +193,179 @@ end = struct
     env, { Sig_rel.assume = assume_nontrivial_eqs eqs la; remove = [] }
 end
 
-module Congruence : sig
-  (** The [Congruence] module implements a simil-congruence closure algorithm on
-      semantic values.
+module type Constraint = sig
+  type t
+  (** The type of bit-vector constraints.
```
Halbaroth: I guess we expect to use this module for constraints in other theories? For instance, we should use it in ADT theory.

src/lib/reasoners/rel_utils.ml
```diff
@@ -193,126 +193,179 @@ end = struct
...
+
+  val pp : t Fmt.t
+  (** Pretty-printer for constraints. *)
+
+  val compare : t -> t -> int
+  (** Comparison function for constraints.
+
+      The comparison only depends on the constraint representation: two
+      constraints with identical representations but different explanations will
+      compare equal.
```
Halbaroth: Do you mean two identical representations with different explanations **have** to be equal for this comparison function?

src/lib/reasoners/rel_utils.ml
```diff
@@ -193,126 +193,179 @@ end = struct
...
+      assert (not (MX.mem rr cs_map));
+      { cs_set ; cs_map ; fresh }
+    | exception Not_found -> bcs
 
-          (* Propagate the substitution *)
-          cgr, f r r' t
-        ) rr_deps (cgr, t)
-    | exception Not_found -> cgr, t
+  let add bcs c =
+    if CS.mem c bcs.cs_set then
```
Halbaroth, line 359: We use the polymorphic comparison here. It means the explanation is updated by this function but the `Constraint.compare` is supposed to ignore the explanation? Is it intentional?

We should at least mention in the documentation of `add` that the structure is updated if the explanation changed.

src/lib/reasoners/rel_utils.ml
```diff
@@ -193,126 +193,179 @@ end = struct
...
-    ; registered : SX.t
-    (** The set of terms we care about. If [x] is in [registered],
-        then [x] is also in [parents(y)] for each [y] in [leaves(x)]. *)
-    }
+  module CS = Set.Make(Constraint)
 
-  let empty = { parents = MX.empty ; registered = SX.empty }
+  type t = {
+    cs_set : CS.t ;
+    (*** All the constraints currently active *)
```
Halbaroth, line 301: Nit: why did you use three asterisks?

src/lib/reasoners/rel_utils.ml
```diff
@@ -193,126 +193,179 @@ end = struct
...
+
+      The explanation [ex] justifies the equality [p = v]. *)
+
+  val fold_leaves : (X.r -> 'a -> 'a) -> t -> 'a -> 'a
+
+  type domain
+
+  val propagate : t -> domain -> domain
+  (** [propagate c dom] propagates the constraints [c] in [d] and returns the
+      new domains. *)
```
```suggestion
      new domain. *)
```

src/lib/reasoners/rel_utils.ml
```diff
@@ -193,126 +193,179 @@ end = struct
...
+module type Constraint = sig
+  type t
+  (** The type of bit-vector constraints.
+
+      Constraints are associated with a justification as to why they are
+      currently valid. The justification is only used to update domains,
+      identical constraints with different justifications will otherwise behave
+      identically (and, notably, will compare equal).
+
+      Bit-vector constraints contains semantic values / term representatives of
```
Halbaroth: Same here.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -131,7 +127,12 @@ module Domains : sig
 end = struct
   type t =
     { bitlists : Bitlist.t MX.t
-    (** Mapping from semantic values to their bitlist domain. *)
+    (** Mapping from semantic values to their bitlist domain.
+
+        Note: this mapping only contains domain for leaves (i.e. uninterpreted
+        terms or AC symbols); domains associated with more complex semantic
+        values are rebuilt on-the-fly using the structure of said semantic
+        values. *)
```
Halbaroth: It seems this PR is rebased on #1004, which means we cannot merge it before solving #1005 too.

src/lib/reasoners/rel_utils.ml
```diff
@@ -193,126 +193,179 @@ end = struct
...
-  val subst : X.r -> X.r -> t -> (X.r -> X.r -> 'a -> 'a) -> 'a -> t * 'a
-  (** [subst p v t f x] performs a local congruence closure of the
-      substitution [p -> v].
+  val fold_fresh : (Constraint.t -> 'a -> 'a) -> t -> 'a -> t * 'a
+  (** [fold_fresh f cs acc] folds [f] over the fresh constraints in [cs].
 
-      More precisely, it will fold [f] over the pairs [(rr, nrr)] such that:
-      - [rr] was registered in the congruence
-      - [nrr] is [X.subst p v rr]
+      Fresh constraints are constraints that were never propagated yet. *)
```
Halbaroth: We should mention that the fresh constraint queue is flushed by this iterator.

src/lib/reasoners/rel_utils.ml
```diff
@@ -193,126 +193,179 @@ end = struct
...
+  module CS = Set.Make(Constraint)
 
-  let empty = { parents = MX.empty ; registered = SX.empty }
+  type t = {
+    cs_set : CS.t ;
+    (*** All the constraints currently active *)
+    cs_map : CS.t MX.t ;
+    (*** Mapping from semantic values to the constraints that involves them *)
+    fresh : CS.t ;
+    (*** Fresh constraints that have never been propagated *)
```
Halbaroth, line 305: I don't understand the meaning of this field. We add a constraint to it in `add` and we flush it in `fold_fresh` but we don't touch it in `propagate`.

Halbaroth (changes requested):

src/lib/reasoners/bitv_rel.ml
bclement-ocp: It is explicitly mentioned in the PR description that it includes #1004 :)

src/lib/reasoners/rel_utils.ml
bclement-ocp: Yeah, I intended to remove reference to bit-vectors in this module but forgot some :)

src/lib/reasoners/rel_utils.ml
bclement-ocp: They don't strictly have to, but they typically should. Actually I am not sure if it is very useful to bundle the explanations into the constraints, we might be able to have a single implementation for explanations (over a non-explained domain type). I'd rather do that in a separate PR to keep this one simple; in the meantime I will reformulate; this was written this way prior to being converted to a functor.

src/lib/reasoners/rel_utils.ml
bclement-ocp: Where do we use polymorphic comparison? `CS` is defined as `Set.Make(Constraint)`, `CS.mem` uses `Constraint.compare`.

src/lib/reasoners/rel_utils.ml
bclement-ocp: Because `propagate` propagates the constraints associated with a single representative `r`. The `fresh` field is cleared by `fold_fresh` which is intended to be used as `Constraints.fold_fresh Constraint.propagate` (see usage in `bitv_rel.ml`). I will refactor to expose `propagate_fresh` instead of `fold_fresh` which makes more sense.

src/lib/reasoners/rel_utils.ml
bclement-ocp: It is a typo that was already present in #944, I will fix it

src/lib/reasoners/rel_utils.ml
bclement-ocp: I will rename this function as `propagate_fresh` and only expose `fold_fresh Constraint.propagate` and mention this in the doc of that function.

src/lib/reasoners/rel_utils.ml
bclement-ocp: The `domain` type is expected to contain domains for many variables; I will rename it (and add a short documentation as well).

bclement-ocp:
Rebased after merging #1004; should be ready for a 2nd round @Halbaroth

Halbaroth (approved):

PR #1011, opened 2023-12-12
fix(BV): Actually propagate freshness change

bclement-ocp:
The `Constraints` module has a mechanism to remember fresh constraints that have never been propagated, so that we don't need to always re-propagate constraints that have already been propagated if their arguments have not changed. Unfortunately, we forget the change to this set of fresh constraints, and hence always re-propagate all active constraints, which renders the optimization useless.

This patch actually takes into consideration the changes to the set of fresh constraints.


bclement-ocp:
This is `+0-0` (no changes) on BV, with comparable total times but slightly less timeouts (45 vs 54)

Halbaroth (approved):

PR #1013, opened 2023-12-14
Documentation of the ADT

Halbaroth:
This PR documents and cleans a bit the ADT theory.

Replace the PR #544 but this old PR contains more documentation on Shostak so I didn't close it yet.

### DIFF
--- a/src/lib/reasoners/adt.ml
+++ b/src/lib/reasoners/adt.ml
@@ -34,9 +34,18 @@ module Hs = Hstring
 
 type 'a abstract =
   | Constr of { c_name : Hs.t ; c_ty : Ty.t ; c_args : (Hs.t * 'a) list }
+  (* [Cons { c_name; c_ty; c_args }] reprensents the application of the
+     constructor [c_name] of the ADT [ty] with the arguments [c_args]. *)
+
   | Select of { d_name : Hs.t ; d_ty : Ty.t ; d_arg : 'a }
-  | Tester of { t_name : Hs.t ; t_arg : 'a } (* tester is currently not used *)
+  (* [Select { d_name; d_ty; d_arg }] represents the destructor [d_name] of
+     the ADT [d_ty] on the ADT value [d_arg]. *)
+
+  | Tester of { t_name : Hs.t ; t_arg : 'a }
+  (* Tester is currently not used. *)
+
   | Alien of 'a
+  (* [Alien r] represents a uninterpreted ADT semantic value. *)
 
 module type ALIEN = sig
   include Sig.X
@@ -53,15 +62,14 @@ let constr_of_destr ty dest =
       "ty = %a" Ty.print ty;
   match ty with
   | Ty.Tadt (s, params) ->
-    let bdy = Ty.type_body s params in
-    begin match bdy with
-      | Ty.Adt cases ->
-        try
-          List.find
-            (fun { Ty.destrs; _ } ->
-               List.exists (fun (d, _) -> Hstring.equal dest d) destrs
-            )cases
-        with Not_found -> assert false (* invariant *)
+    begin
+      let Ty.Adt cases = Ty.type_body s params in
+      try
+        List.find
+          (fun { Ty.destrs; _ } ->
+             List.exists (fun (d, _) -> Hstring.equal dest d) destrs
+          ) cases
+      with Not_found -> assert false (* invariant *)
     end
   | _ -> assert false
 
@@ -81,7 +89,12 @@ module Shostak (X : ALIEN) = struct
     not (Options.get_disable_adts ()) &&
     match sy, ty with
     | Sy.Op (Sy.Constr _), Ty.Tadt _ -> true
-    | Sy.Op Sy.Destruct (_,guarded), _ -> not guarded
+    | Sy.Op Sy.Destruct (_, guarded), _ ->
+      (* A guarded destructor isn't interpreted by the ADT theory.
+         If we assume the tester of the constructor associated with
+         this destructor, we produce release the guard. See the
+         documentation of [env.selectors] in [Adt_rel]. *)
+      not guarded
     | _ -> false
 
   let embed r =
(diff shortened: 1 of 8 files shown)


src/lib/reasoners/adt.ml
```diff
@@ -81,35 +89,36 @@ module Shostak (X : ALIEN) = struct
     not (Options.get_disable_adts ()) &&
     match sy, ty with
     | Sy.Op (Sy.Constr _), Ty.Tadt _ -> true
-    | Sy.Op Sy.Destruct (_,guarded), _ -> not guarded
+    | Sy.Op Sy.Destruct (_, guarded), _ ->
+      (* A guarded destructor isn't interpreted by the ADT theory.
+         If we assume the tester of the constructor associated with
+         this destructor, we produce release the guard. See the
```
bclement-ocp: Not sure I understand what "produce release the guard" means, can you clarify?

src/lib/structures/ty.mli
```diff
@@ -114,8 +114,9 @@ module Set : Set.S with type elt = t
 
 
 val assoc_destrs : Hstring.t -> adt_constr list -> (Hstring.t * t) list
-(** returns the list of destructors associated with the given consturctor.
-    raises Not_found if the constructor is not in the given list *)
+(** [assoc_destrs cons cases] returns the list of destructors associated with
+    the consturctor [cons] and the body [cases].
```
bclement-ocp:
```suggestion
    the constructor [cons] in the ADT defined by [cases].
```

src/lib/reasoners/adt_rel.ml
```diff
@@ -55,17 +55,83 @@ type t =
...
+       destructor equations are classified by their corresponding
+       constructor.
+
+       Consider [d] a guarded destructor and [c] its associated constructor.
+
+       - When we register the destructor application [d t] in [add_aux], we
+         produce an equation [d t = d' t] where [d'] is the non-guarded
+         version of [d]. If we don't already know that the value of [t] has
+         to use the constructor [c], we put the equation [d t = d' t] in this
+         map. Otherwise we propagate the equation to CC(X).
```
bclement-ocp, line 102: Can you clarify the mapping? I assume it is `d t -> c -> [d t = d' t, ...]`?

src/lib/reasoners/adt_rel.ml
```diff
@@ -55,17 +55,83 @@ type t =
...
     pending_deds :
       (r Sig_rel.literal * Ex.t * Th_util.lit_origin) list
+      (* Pending queue of facts discovered by the theory. Facts are pending
+         in the following situations:
+
+         - The domain of a uninterpreted semantic value becomes a singleton.
+           In this case, we deduce the constructor of the semantic value and
+           we add this equation to the pending queue using [deduce_is_constr].
+
+         - When we add a guarded destructor for the first time, we add a new
```
bclement-ocp: Do you mean when the guard is activated (i.e. when we learn the tester)?

src/lib/reasoners/adt_rel.ml
```diff
@@ -180,7 +239,17 @@ let seen_tester r hs env =
...
+     (cons (d1 ty1) ... (dn tyn))
+   this function adds the new equation to [eqs]:
+     [t = cons x1 ... xn]
+   where x1, ..., xn are fresh names of type respectively ty1, ..., tyn
+   and [t] is the term associated with the uninterpreted semantic value [r].
+
+   If the tester [(_ cons) t] hasn't been already seen, we also add it to
+   the equations [eqs].
 
+   Assume that [r] is a semantic value of type [Ty.Adt]. *)
```
bclement-ocp, line 251:
```suggestion
   Assumes that [r] is a semantic value of type [Ty.Adt]. *)
```

src/lib/reasoners/adt_rel.ml
```diff
@@ -180,7 +239,17 @@ let seen_tester r hs env =
...
 
+(* For a uninterpreted semantic value [r] and [h] a constructor of the form:
+     (cons (d1 ty1) ... (dn tyn))
+   this function adds the new equation to [eqs]:
+     [t = cons x1 ... xn]
+   where x1, ..., xn are fresh names of type respectively ty1, ..., tyn
+   and [t] is the term associated with the uninterpreted semantic value [r].
+
+   If the tester [(_ cons) t] hasn't been already seen, we also add it to
+   the equations [eqs].
```
bclement-ocp, line 249: I assume that "seen" here means assumed, so maybe the following would be more clear:

> If the literal [(_ cons) t] was not already assumed, we add it to [eqs].

src/lib/reasoners/adt_rel.ml
```diff
@@ -373,20 +440,15 @@ let add_aux env (uf:uf) (r:r) t =
       { env with seen_access = SE.add t env.seen_access }
 
     | Sy.Op Sy.Destruct _, _ ->
-      assert false (* not possible *)
+      (* This case is excluded by the parser. *)
```
bclement-ocp: I am not sure I understand what the parser has to do with this, can you clarify?

src/lib/reasoners/adt_rel.ml
```diff
@@ -398,69 +460,131 @@ let count_splits env la =
...
+        (declare-const z Data)
+        (assert ((_ is cons1) x))
+        (assert ((_ is cons1) y))
+        (assert ((_ is cons1) z))
+        (assert (distinct x y z))
+        (check-sat)
+       ```
+       The expected answer is `unsat`. If we remove the second constructor
+       [cons2], Alt-Ergo will promote the ADT `Data` to a record and we got the
+       correct answer. *)
```
bclement-ocp: As noted in #1014 this doesn't seem to be true.

bclement-ocp (changes requested):
This is a PR that documents the ADT theory, but also includes code simplifications, which makes it hard to read as it requires many mental context switches. The code simplifications look correct, and more documentation is always good, however there are some points in the documentation that need clarification (questions inline).

I also think that it would be better to add a link to #1014 rather than copying the description of that bug in a comment.

src/lib/reasoners/adt_rel.ml
Halbaroth: Technically, it's not an invariant of the parser but the arity of the operator `Destruct` is `1`. I suggest: 
```ocaml
(* The arity of the [Sy.Destruct] operator is 1. *)
```

Currently, we can produce a destructor with any number of arguments because the internal structure of expressions isn't sufficiently opaque. You can produce such terms with `Expr.mk_term` for instance. Builtin operators shouldn't be build by `mk_term` in fact. We should reserve `mk_term` for uninterpreted functions.

src/lib/reasoners/adt_rel.ml
Halbaroth: Yes, `seen` means `assumed`. In fact, we can rename the field `seen_testers` to `assumed_testers`. I guess the old name comes from a previous implementation of the theory.

src/lib/reasoners/adt.ml
Halbaroth: I meant "we release the guard" but it's a conventional usage. I'll reformulate.

src/lib/reasoners/adt_rel.ml
Halbaroth: I guess my comment is unclear but the mapping is not as you think. 
For some representative (semantic value) `r`, the map contains a map of constructors of `X.type_info r` to destructor equations `d t = d' t` where `t` is a term of the class of `r`.

src/lib/reasoners/adt_rel.ml
Halbaroth: It's the opposite situation. We know that the domain is a singleton, thus we call `deduce_is_constr`. This function propagate two literals to CC(X) through the queue `pending_deds`:
- an equation of the form `t = cons x_1 ... x_n`
- the literal `(_ cons t)`.

Thus in this situation we activate any guard on destructors of `cons` because the domain is a singleton. It makes sense as a singleton domain means that the only consistent way to produce a value for `t` is to use the constructor `cons`, so we can active the guard of any destructor of `cons`.

bclement-ocp: This explanation corresponds to my understanding (although I think you meant `((_ is cons) t)`), but I still don't understand what "when we add a guarded destructor for the first time" means. Can you clarify the second list item?

bclement-ocp (changes requested):

src/lib/reasoners/adt_rel.ml
Halbaroth: If we have already assumed the tester of the associated constructor of the destructor `d` and we add `d` to the theory, we don't create a pending equation in `selectors`.

bclement-ocp (approved):
This is more clear now, thanks!

PR #1015, opened 2023-12-15
fix(CDCL): Fix and document `pop` assertion

bclement-ocp:
All guards are decided at the beginning of the search process, and so when popping we should be in one of two states: either all guards have been decided (we started the search) or none are (we did not start the search). This is what the assertion tries to ensure (if it is not the case, there is probably a bug).

However, with incremental solving, the assertion is not right: if we do push, push, solve, pop, push, and finally pop, the first pop will reset the solver to a state where the first guard is decided, and after the last push, the first guard is decided and the second is not.

This would be valid behavior, and so this patch modifies (and documents) the assertion to instead check that the next guard is actually a guard (in range of the `increm_guards` vector).


Stevendeo (approved):
LGTM

PR #1016, opened 2023-12-15
Removing exhaustive exception catch in satml_frontend

Stevendeo:
This catch breaks the call stack tracking; the frontend and the binary already handles the exceptions, no need to to it at the reasoner level.


bclement-ocp (approved):
Yeah that one is super annoying and the exception printing is not really working very well… I can't count the number of times I have replaced that `assert` with `raise e`. I am OK with removing this (although I guess keeping the print + `raise e` would be slightly safer as it leaves a trace in the logs if the exception is incorrectly caught at a higher level, but I am OK with just removing everything).

Halbaroth (approved):

PR #1020, opened 2024-01-03
Clean up documentation of the SAT API

Halbaroth:
The documentation of the SAT API was a bit outdated and don't use some tags of `odoc`.

I also merge `empty` and `empty_with_inst` into a single function. Currently, the second function is only used by the JS worker. 
The argument of this constructor is supposed to filter ground facts generated by the instantiation engine but the JS worker uses it only to register all these facts.

### DIFF
--- a/src/lib/reasoners/sat_solver_sig.ml
+++ b/src/lib/reasoners/sat_solver_sig.ml
@@ -72,58 +72,78 @@ module type S = sig
   exception Unsat of Explanation.t
   exception I_dont_know
 
-  (* the empty sat-solver context *)
-  val empty : unit -> t
-  val empty_with_inst : (Expr.t -> bool) -> t
+  val empty : ?selector:(Expr.t -> bool) -> unit -> t
+  (** [empty ~selector ()] creates an empty environment.
 
-  (** [push env n] add n new assertion levels.
-      A guard g is added for every expression e assumed at the current
-      assertion level.
-      Ie. assuming e after the push will become g -> e,
-      a g will be forced to be true (but not propagated at level 0) *)
+      The optional argument [selector] is used to filter ground facts
+      discovered by the instantiation engine. *)
+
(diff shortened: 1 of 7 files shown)


src/lib/reasoners/sat_solver_sig.ml
```diff
@@ -72,58 +72,78 @@ module type S = sig
...
 
-  (* [pred_def env f] assume a new predicate definition [f] in [env]. *)
   val pred_def : t -> Expr.t -> string -> Explanation.t -> Loc.t -> unit
+  (** [pred_def env f] assumes a new predicate definition [f] in [env]. *)
 
-  (** [optimize env ~is_max o] registers the expression [o]
-      as an objective function. *)
   val optimize : t -> is_max:bool -> Expr.t -> unit
+  (** [optimize env ~is_max o] registers the expression [o] as an objective
+      function.
```
bclement-ocp, line 111: Should briefly mention the meaning of `is_max` (maximize vs minimize).

src/lib/reasoners/sat_solver_sig.ml
```diff
@@ -72,58 +72,78 @@ module type S = sig
...
+
+      After optimization, the value of this objective is returned by
+      [get_objectives]. *)
 
-  (* [unsat env f size] checks the unsatisfiability of [f] in
-     [env]. Raises I_dont_know when the proof tree's height reaches
-     [size]. Raises Sat if [f] is satisfiable in [env] *)
   val unsat : t -> Expr.gformula -> Explanation.t
+  (** [unsat env f size] checks the unsatisfiability of [f] in [env].
```
bclement-ocp: There is no `size` argument? `unsat` has only two arguments; this seems outdated.

bclement-ocp (approved):
The doc for `unsat` seems outdated (but now with odoc markers), otherwise looks good :+1:

Issue #1023, opened 2024-01-08
Invalid SAT environment while optimizing


Halbaroth:
The current implementation of `unsat_rec_prem` is wrong in presence of optimization if we try to use the SAT API directly. Consider the input problem:
```smt2 
(set-option :produce-models true)
(set-logic ALL)
(declare-const x Int)
(assert (<= x 10))
(maximize x)
(check-sat)
```
This problem is SAT and we expect to be able to reuse the environment after the `(check-sat)` statement. If we use the Dolmen frontend, it's okay because we throw away the previous environment and produce a fresh environment with all the assertions before each `(check-sat)`.

If we use the imperative SAT API, we want to write:
```ocaml
let env = SAT.empty () in 
SAT.declare env `x`;
SAT.assume env `x <= 10`;
SAT.optimize env ~is_max:true `x`;
SAT.unsat env;
SAT.assume env `x <= 10`;
```
But the current implementation will raise the exception `Unsat` during the last statement. The reason is simple: in `analyze_unknown_for_objectives` we call again the solver until getting an unsatisfiable context. In other words, in presence of optimization, the environment `env` after calling `SAT.unsat` is always `unsatisfiable`. In the previous example, it will contain both the assertions `x <= 10` and `x > 10`.

A naive solution consists in protecting the penultimate environment by using the push/pop mechanism and retrieve the model and objectives values produced before popping. Notice that we shouldn't produce an environment with extra push levels, otherwise the behaviour of the `pop` statement will be wrong after calling `SAT.unsat`.

To implement this solution, we can save the new assertions discovered in `analyze_unknown_for_objectives` in a stack, perform a `(push 1)` and a `(pop 1)` around the `assume` and `unsat` in `unsat_rec_prem` and after getting an `unsat` exception, get the tail of the stack and assert it without push/pop around it.

If you have a better solution, I'll be glad to implement it ;)

Halbaroth:
Another solution (a bit better in terms of performance I guess) consists in using a different kind of guards to protect the environments during the exploration for optimization. These guards won't appear in the `stack_guard`. In this way, the `pop` statement will behave correctly after calling `SAT.unsat` and we can use an internal `push/pop` mechanism to protect the environments in `unsat_rec_prem`.

bclement-ocp:
[...]

PR #1025, opened 2024-01-08
fix(arith): Do not overly tighten bounds over infinite domains

### LINKED ISSUE #1022: Unsoundness in model generation
On the following *sat* problem from the SMT-LIB (QF_LIA/20180326-Bromberger/unbd-sage/unbd010v15c/unbd-sage0):

```smt-lib
(set-info :smt-lib-version 2.6)
(set-logic QF_LIA)
(set-info :source |
Generated by: Martin Bromberger
Generated on: 2018-01-06
Generator: https://www.mpi-inf.mpg.de/fileadmin/inf/rg1/Documents/RandomUnboundedSageScript.tar.gz
Application: This is a satisfiable problem with bounded and unbounded directions and, therefore, a good test for the termination/completeness of branch-and-bound solvers.
Publications: M. Bromberger. A Reduction from Unbounded Linear Mixed Arithmetic Problems into Bounded Problems. (Work in progress.)

The benchmarks in this class were randomly generated by a sagemath script and by way of construction they are all satisfiable.
Target solver: CVC4 Mathsat SPASS-IQ YICES Z3
|)
(set-info :license "https://creativecommons.org/licenses/by/4.0/")
(set-info :category "random")
(set-info :status sat)
(declare-fun x0 () Int)
(declare-fun x1 () Int)
(declare-fun x2 () Int)
(declare-fun x3 () Int)
(declare-fun x4 () Int)
(declare-fun x5 () Int)
(declare-fun x6 () Int)
(declare-fun x7 () Int)
(declare-fun x8 () Int)
(declare-fun x9 () Int)

(assert (and 
(<= (- 7) (+ (* (- 2) x0) (+ (* (- 1) x2) (+ (* 1 x3) (+ (* 1 x5) (+ (* (- 1) x6) (+ (* 1 x8) (* 1 x9) ))))))) 
(<= 1 (+ (* (- 2) x3) (+ (* (- 1) x4) (+ (* (- 2) x7) (* (- 1) x9) )))) 
(<= 2 (+ (* 2 x0) (+ (* 2 x1) (+ (* 2 x2) (+ (* (- 2) x3) (+ (* 1 x4) (+ (* 2 x6) (* (- 2) x8) ))))))) 
(<= (- 2) (+ (* (- 1) x0) (+ (* 1 x1) (+ (* (- 2) x2) (+ (* 2 x4) (+ (* 2 x5) (+ (* (- 1) x7) (* (- 2) x9) ))))))) 
(<= 15 (+ (* (- 1) x0) (+ (* 1 x2) (+ (* (- 2) x3) (+ (* 2 x5) (+ (* 2 x6) (+ (* 2 x7) (* (- 2) x8) ))))))) 
(<= 8 (+ (* (- 2) x4) (+ (* 2 x5) (+ (* 2 x6) (+ (* (- 2) x8) (* 2 x9) ))))) 
(<= (- 11) (+ (* (- 2) x0) (+ (* 2 x2) (+ (* (- 1) x3) (+ (* 2 x4) (+ (* 1 x5) (+ (* (- 2) x6) (* (- 2) x7) ))))))) 
(<= 9 (+ (* (- 2) x1) (+ (* 2 x3) (+ (* (- 2) x4) (+ (* 2 x7) (* 1 x9) ))))) 
(<= 15 (+ (* 2 x0) (+ (* (- 1) x1) (+ (* (- 1) x3) (+ (* 2 x4) (+ (* 2 x6) (+ (* 1 x7) (+ (* (- 2) x8) (* 1 x9) )))))))) 
(<= 37 (+ (* 9 x0) (+ (* (- 9) x1) (+ (* (- 7) x2) (+ (* 7 x3) (+ (* (- 14) x4) (+ (* (- 15) x5) (+ (* 1 x6) (+ (* 9 x7) (* 7 x8) ))))))))) 
(<= (- 159) (+ (* (- 15) x0) (+ (* 28 x1) (+ (* 16 x2) (+ (* (- 8) x3) (+ (* 36 x4) (+ (* 28 x5) (+ (* (- 7) x6) (+ (* (- 24) x7) (+ (* (- 11) x8) (* (- 4) x9) )))))))))) 
(<= 56 (+ (* 20 x0) (+ (* (- 20) x1) (+ (* (- 13) x2) (+ (* 12 x3) (+ (* (- 32) x4) (+ (* (- 39) x5) (+ (* (- 2) x6) (+ (* 16 x7) (+ (* 20 x8) (* 4 x9) )))))))))) 
(<= 300 (+ (* (- 1) x0) (+ (* (- 35) x1) (+ (* (- 10) x2) (+ (* (- 15) x3) (+ (* (- 26) x4) (+ (* (- 2) x5) (+ (* 19 x6) (+ (* 27 x7) (+ (* (- 9) x8) (* 2 x9) )))))))))) 
(<= (- 45) (+ (* 10 x0) (+ (* (- 1) x1) (+ (* (- 3) x2) (+ (* 19 x3) (+ (* (- 15) x4) (+ (* (- 17) x5) (+ (* (- 6) x6) (+ (* 7 x7) (+ (* 12 x8) (* (- 3) x9) )))))))))) 
(<= 55 (+ (* 17 x0) (+ (* (- 17) x1) (+ (* (- 14) x2) (+ (* 1 x3) (+ (* (- 15) x4) (+ (* (- 30) x5) (+ (* 1 x6) (+ (* 7 x7) (+ (* 13 x8) (* 7 x9) ))))))))))))
(check-sat)
(exit)
```

Alt-Ergo replies:

```console
$ alt-ergo unbd-sage2.smt2 
                                       
; File "unbd-sage2.smt2", line 43, characters 1-12: I don't know (0.0729) (37 steps) (goal g_1)

unknown
```

That's good (the problem is SAT).

If we request model generation, however, we have an issue:

```console
$ alt-ergo --produce-model unbd-sage2.smt2
                                       
; File "/tmp/imreg/unbd-sage2.smt2", line 43, characters 1-12: Valid (0.0794) (37 steps) (goal g_1)

unsat

```

Some more info:

 - This does not only happen for this specific problem but also for a buch of problems in `QF_LIA/20180326-Bromberger/unbd-sage` and other SAT problems with integers (e.g. `QF_LIA/dilig/15-17.smt2`).
 - The exception comes from [this line](https://github.com/OCamlPro/alt-ergo/blob/214ea7c7fe7063458d0f6485ed8b48dcb8211f16/src/lib/reasoners/intervalCalculus.ml#L1983) and in [this call](https://github.com/OCamlPro/alt-ergo/blob/214ea7c7fe7063458d0f6485ed8b48dcb8211f16/src/lib/reasoners/intervalCalculus.ml#L2026) to `model_from_simplex`
 - This is because the `int_sim` environment we create with "FM-Simplex encoding" (I do not know what that is) [here](https://github.com/OCamlPro/alt-ergo/blob/214ea7c7fe7063458d0f6485ed8b48dcb8211f16/src/lib/reasoners/intervalCalculus.ml#L2021) is actually unsat
 - The issue does not occur with the fm-simplex plugin (so this is an interaction between model generation and the inequalities module)
 - Using the "regular" `env.int_sim` here works, at least in this specific case, but presumably that is not acceptable to use in the general case (?). I do not know why we need this "FM-Simplex" encoding (yet), but it seems that we are computing it wrong, somehow.

bclement-ocp:
Model generation for arithmetic constraints proceeds in two stages:

 - In the first stage, we only choose a value for variables that are within a finite domain (i.e. integer variables bounded from both above and below). This is the behavior of regular case splits, and happens in function `default_case_split`.

 - In the second stage, which we only reach if model generation is enabled, we must select a value for all variables. In this case, we select one interval for variables that can have multiple intervals (recall that we model arithmetic domains using unions of intervals). This happens in function `case_split_union_of_intervals`. If the selected interval is finite, we go back to the first stage in the next case split.

 - In the final stage, all remaining variables have infinite intervals as domains, and we must now pick a value for these. This happens in `model_from_unbounded_domains` -- even though for rational variables, the domain can still be bounded. For integers however, infinite intervals mean that they lack either a finite lower or upper bound (or both). For integers, we create a new simplex environment in `fm_simplex_unbounded_integers_encoding` where we adjust the bounds by a factor of half the sum of the absolute values of the coefficients. This means that an inequality `0 < 3x + 4y` becomes `0 + (3 + 4) / 2 < 3x + 4y`, i.e. `3.5 < 3x + 4y`. This removes `x = 1, y = 0` from the domain. On its own, this would be correct (albeit strange) because after removing a finite number of elements from an infinite domain there is still an infinite number of elements to choose from. But within a larger system this can (and is) problematic.

This patch removes the "adjustment" of the bounds which is not justified and actually causes unsoundness.

Fixes #1022

### DIFF
--- a/src/lib/reasoners/intervalCalculus.ml
+++ b/src/lib/reasoners/intervalCalculus.ml
@@ -1921,37 +1921,19 @@ let fm_simplex_unbounded_integers_encoding env uf =
            "case-split over unions of intervals is needed !!!";
          assert false
 
-       | [(mn, ex_mn), (mx, ex_mx)] ->
+       | [(lb, ex_lb), (ub, ex_ub)] ->
          let l, c = P.to_list p in
-         let l = List.rev_map (fun (c, x) -> x, c) (List.rev l) in
-         assert (Q.sign c = 0);
-         let cst0 =
-           List.fold_left (fun z (_, c) -> Q.add z (Q.abs c))Q.zero l
-         in
-         let cst = Q.div cst0 (Q.from_int 2) in
-         assert (mn == None || mx == None);
-         let min =
-           let mn =
-             match mn with
-             | None -> None
-             | Some (q, q') -> Some (Q.add q cst, q')
-           in
-           bnd_to_simplex_bound (mn, ex_mn)
-         in
-         let max =
-           let mx =
-             match mx with
-             | None -> None
-             | Some (q, q') -> Some (Q.sub q cst, q')
-           in
-           bnd_to_simplex_bound (mx, ex_mx)
-         in
+         assert (Q.is_zero c);
+         assert (lb == None || ub == None);
+         let min = bnd_to_simplex_bound (lb, ex_lb) in
+         let max = bnd_to_simplex_bound (ub, ex_ub) in
          match l with
          | [] -> assert false
-         | [x, c] ->
+         | [c, x] ->
            assert (Q.is_one c);
            Sim.Assert.var simplex x ?min ?max |> fst
          | _ ->
+           let l = List.rev_map (fun (c, x) -> x, c) (List.rev l) in
            let xp = alien_of p in
            let sim_p =
              match Sim.Core.poly_of_slake simplex xp with
@@ -2009,7 +1991,7 @@ let model_from_simplex sim is_int env uf =
       )[] (List.rev main_vars)
 
 
-let model_from_unbounded_domains =
+let model_from_infinite_domains =
   let mk_cs acc (x, v, _ex) =
     ((LR.view (LR.mk_eq x v)), true,
      Th_util.CS (Th_util.Th_arith, Q.from_int 2)) :: acc
@@ -2040,7 +2022,7 @@ let case_split env uf ~for_model =
     else
       begin
         match case_split_union_of_intervals env uf with
-        | [] -> model_from_unbounded_domains env uf
+        | [] -> model_from_infinite_domains env uf
         | l -> l
       end
   | _ -> res
--- a/tests/issues/777.models.expected
+++ b/tests/issues/777.models.expected
@@ -1,5 +1,5 @@
 
 unknown
 (
-  (define-fun i () Int 2)
+  (define-fun i () Int 1)
 )
--- a/tests/models/uf/uf2.models.expected
+++ b/tests/models/uf/uf2.models.expected
@@ -1,7 +1,7 @@
 
 unknown
 (
-  (define-fun f ((arg_0 Int)) Int (ite (= arg_0 (- 2)) 0 2))
+  (define-fun f ((arg_0 Int)) Int (ite (= arg_0 (- 1)) 0 1))
   (define-fun a () Int 0)
-  (define-fun b () Int (- 2))
+  (define-fun b () Int (- 1))
 )


bclement-ocp:
Great, we don't have the same model for OCaml 4.14.0 and 5.0.1… or with/without the weak table :/

@Halbaroth any progress on #790 ? :sweat_smile: 

(Although I don't really understand what is impacted by the weak table here… there are no disjunctions or quantifiers in the problem, and yet we choose `x1` first with `--disable-weak` and `x3` first without?)

Halbaroth:
#790 is stuck because we use custom function as `ae.round` in our models and Dolmen rejects them because it doesn't know this value.

I also implemented `get-value` but it doesn't work with optimization because of #1023 :D

bclement-ocp:
> #790 is stuck because we use custom function as `ae.round` in our models and Dolmen rejects them because it doesn't know this value.

There are no `ae.round` in the expected models though, so that shouldn't be an issue?

> I also implemented `get-value` but it doesn't work with optimization because of #1023 :D

:+1: although that shouldn't block using Dolmen to check the models on CI

(I am now thinking that maybe I did not link the correct issue, I was thinking of the one about using Dolmen to check models on the CI instead of comparing models textually)

bclement-ocp:
OK, the main cause of the discrepancy is because we use `hash_cmp` to define the simplex variables [here](https://github.com/OCamlPro/alt-ergo/blob/214ea7c7fe7063458d0f6485ed8b48dcb8211f16/src/lib/reasoners/intervalCalculus.ml#L93) which in turn means that the weak table impact the solution we get from the simplex, which can cause large discrepancies in the output.

Using `str_cmp` here stabilizes the result for this new test case, and it might improve stability generally — hopefully without impacting performance too much. I will do a test run with `str_cmp` for the simplex variables to see what happens.

Halbaroth:
> There are no `ae.round` in the expected models though, so that shouldn't be an issue?

You're right. I meant there is such functions in the input problem, thus Dolmen cannot evaluate them to check the model because it doesn't know an interpretation of these functions.

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -1921,37 +1921,19 @@ let fm_simplex_unbounded_integers_encoding env uf =
...
-             match mx with
-             | None -> None
-             | Some (q, q') -> Some (Q.sub q cst, q')
-           in
-           bnd_to_simplex_bound (mx, ex_mx)
-         in
+         assert (Q.is_zero c);
+         assert (lb == None || ub == None);
+         let min = bnd_to_simplex_bound (lb, ex_lb) in
+         let max = bnd_to_simplex_bound (ub, ex_ub) in
```
Halbaroth, line 1929: I agree that `lb` for `lower bound` and `ub` for `upper bound` are better names here as the interval isn't always closed. We should keep this convention for the results of `bnd_to_simplex_bound`. May be `sim_lb` and `sim_up` are better?

Halbaroth:
I think we should rename `model_from_unbounded_domains` to `model_from_infinite_domains` ;)

> But within a larger system this can (and is) problematic.

Can you explain? It look sound to remove finitely many elements of an infinite set. Besides, do you have any clue on the reason of this strange code?

Halbaroth (changes requested):

src/lib/reasoners/intervalCalculus.ml
bclement-ocp: I kept the names `min` and `max` here because those are the names of the labelled arguments of the functions `Sim.Assert.var` and `Sim.Assert.poly` (also they are technically minimum and maximum because a strict lower bound of value `n` is represented as the pair `(n, +1)` and a strict upper bound of value `n` as the pair `(n, -1)` which are the actual minimum/maximum values).

I can change them but I think it's more clear to use the names from the API in this case.

bclement-ocp (commented):
> I think we should rename model_from_unbounded_domains to model_from_infinite_domains ;)

Done.

> Can you explain? It look sound to remove finitely many elements of an infinite set. Besides, do you have any clue on the reason of this strange code?

The propagators are not complete. Assume you have `x > 0` and also `x ** 1 = 1`; the exponentiation is not a linear function and so we cannot reason about it within the simplex; removing `1` from the domain of `x` would be unsound because `x = 1` is the only solution of `x ** 1 = 1`.

bclement-ocp:
> do you have any clue on the reason of this strange code?

Also: not really, unfortunately :( We discussed it with @Gbury but could not figure out either a logical justification or practical interest for doing this, it just looks like a misguided attempt to do an (incorrect) simplifying cut.

bclement-ocp:
> do you have any clue on the reason of this strange code?

Also: not really, unfortunately :( We discussed it with @Gbury but could not figure out either a logical justification or practical interest for doing this, it just looks like a misguided attempt to do an (incorrect) simplifying cut.

Halbaroth:
Ok, I'll merge your PR if the benchmarks are okay after changing the comparison function ;)

bclement-ocp:
Removed the additional tests after the discussion at the dev meeting (see also [here](https://github.com/OCamlPro/alt-ergo/issues/563#issuecomment-1883423674)); should be ready for a 2nd round.

Halbaroth (approved):

PR #1027, opened 2024-01-18
feat(CDCL): Add ability to decide on semantic literals

bclement-ocp:
This patch introduces the ability for the CDCL solver to deal with semantic literals. This is part of work towards #901 although it does not (yet) achieves the goals of #901.

More precisely:

 - We add a new `Literal` module in `structures` to represent literals that are either syntaxic (represented by expressions) or semantic (represented using the `Xliteral` module).

 - The new `Literal` module is used to represent the literals associated with the `Atom`s in the CDCL solver [1]

 - The CDCL solver has a special handling of these semantic literals, which are treated as "local" literals: they are not added in the variable ordering, and will not be decided upon by the solver. Instead, a new mechanism inspired by mSAT (and, transitively, mcSAT) is provided, where the theory is able to mark certain literals as "immediate decisions". When literals are marked this way, the solver will decide on one of the marked literals the next time a decision must be made. Literals marked as immediate decisions are part of a transient solver state, and are forgotten upon backtracking.

 - The theory uses this marking mechanism to communicate case splits (and a similar mechanism for optimization objectives, which is separate because the value of the objective is sent back to the theory once the decision is made) to the solver.

This mechanism allows to introduce splits that are, in a way, "local", and do not  pollute other branches of the solver. It can be seen as an intermediate step towards a mcSAT approach, but where we are still deciding on boolean literals and hence try to avoid adding a large number of splits globally into the solver -- although it does mean that the splits are not incorporated into the activity heuristic.

The functionality is currently gated behind a new flag `--enable-sat-cs`.

Note that with `--enable-sat-cs` this change is +46-13 on ae-format (and +8-7 on benchtop-tests) or a net positive of +33 on ae-format, so we might want to enable it by default.

[1] : It could also be used for the `Bj`s in the Tableaux solver, if we
      wanted to suport similar features in the Tableaux solver.

### DIFF
--- a/src/lib/reasoners/intervalCalculus.ml
+++ b/src/lib/reasoners/intervalCalculus.ml
@@ -1996,9 +1996,10 @@ let case_split env uf ~for_model =
   | _ -> res
 
 (* Helper function used in [optimizing_objective] to pick a value
-   for the polynomial [p] in its interval. We use this function in the case
-   the value produced by the optimization procedure doesn't satisfy some
-   constraints that involve strict inequalities or the problem is unbounded. *)
+   for the polynomial [p] in its interval. We use this function to split the
+   search space and bias it towards the optimum in the case the value produced
+   by the optimization procedure doesn't satisfy some constraints that involve
+   strict inequalities or the problem is unbounded. *)
 let middle_value env ~is_max ty p bound =
   let interval =
     match MP0.find_opt p env.polynomes, bound with
--- a/src/lib/reasoners/intervalCalculus.ml
+++ b/src/lib/reasoners/intervalCalculus.ml
@@ -552,7 +552,7 @@ module Debug = struct
   let implied_equalities l =
     if get_debug_fm () then
       let pp_literal ppf = function
-        | Sig_rel.LTerm e -> Expr.print ppf e
+        | Literal.LTerm e -> Expr.print ppf e
         | LSem ra -> LR.print ppf (LR.make ra)
       in
       let print fmt (ra, ex, _) =
@@ -2070,11 +2070,16 @@ let optimizing_objective env uf Objective.Function.{ e; is_max; _ } =
             Objective.Value.Minfinity
         in
         (* As the problem is unbounded, we need to produce a value for the
-           objective function. In this case, we pick a value in the domain
-           of the polynomial [p]. *)
+           objective function. Rather than picking a value directly, we add a
+           constraint to let the case split mechanism pick a "large" (or
+           "small") value in the interval (otherwise, later objectives are
+           computed while assuming that the middle value is forced, which is
+           incorrect). *)
         let case_split =
-          LR.mkv_eq r1 (middle_value env ~is_max ty p None), true, Th_util.CS
-            (Th_util.Th_arith, Q.one)
+          LR.mkv_builtin false LT
+            [r1; (middle_value env ~is_max ty p None)],
+          true,
+          Th_util.CS (Th_util.Th_arith, Q.one)
         in
         Some Th_util.{ value; case_split }
 
--- a/src/lib/reasoners/intervalCalculus.ml
+++ b/src/lib/reasoners/intervalCalculus.ml
@@ -2077,7 +2077,7 @@ let optimizing_objective env uf Objective.Function.{ e; is_max; _ } =
            incorrect). *)
         let case_split =
           LR.mkv_builtin false LT
-            [r1; (middle_value env ~is_max ty p None)],
+            [r1; middle_value env ~is_max ty p None],
           true,
           Th_util.CS (Th_util.Th_arith, Q.one)
         in
--- a/src/lib/reasoners/satml.ml
+++ b/src/lib/reasoners/satml.ml
@@ -1393,7 +1393,7 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
            Note that if the [fuip] is itself a semantic literal, it will act as
            a (local) learned clause (see [semantic_expansion]). *)
         let has_split = Vec.exists is_semantic lclause.atoms in
-        if has_split then (
+        if not has_split then (
           Vec.push env.learnts lclause;
           attach_clause env lclause;
           clause_bump_activity env lclause
--- a/src/lib/reasoners/satml.ml
+++ b/src/lib/reasoners/satml.ml
@@ -1382,9 +1382,10 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
           Atom.make_clause name learnt vraie_form true history
         in
         (* Do not learn clauses involving semantic literals (but still record
-           them as a reason for the propagated literal), otherwise we could
-           learn stuff about these semantic literals in a scope where they are
-           no longer valid.
+           them as a reason for the propagated literal). Learned clauses are
+           valid at level 0, but semantic literals can contain terms created by
+           the Shostak module that are only useful in a portion of the search
+           tree, and in particular would not exist (in Uf) at lower levels.
 
            Note that we try to avoid semantic literals as much as possible in
            [conflict_analyze_aux], but we can still have semantic decisions
--- a/src/lib/reasoners/satml.ml
+++ b/src/lib/reasoners/satml.ml
@@ -666,6 +666,7 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
       Queue.clear env.tatoms_queue;
       Queue.clear env.th_tableaux;
       env.tenv <- Vec.get env.tenv_queue lvl; (* recover the right tenv *)
+      env.nchoices <- Vec.get env.nchoices_stack lvl;
       if Options.get_cdcl_tableaux () then begin
         env.lazy_cnf <- Vec.get env.lazy_cnf_queue lvl;
         env.relevants <- Vec.get env.relevants_queue lvl;
--- a/src/lib/reasoners/satml.ml
+++ b/src/lib/reasoners/satml.ml
@@ -91,6 +91,10 @@ module type SAT_ML = sig
   type t
 
   val solve : t -> unit
+  val compute_concrete_model :
+    declared_ids:Id.typed list ->
+    t ->
+    Models.t Lazy.t * Objective.Model.t
 
   val set_new_proxies : t -> FF.proxies -> unit
 
@@ -113,7 +117,7 @@ module type SAT_ML = sig
     t -> FF.hcons_env -> Satml_types.Atom.Set.t
   val current_tbox : t -> th
   val set_current_tbox : t -> th -> unit
-  val empty : unit -> t
+  val create : Atom.hcons_env -> t
 
   val assume_th_elt : t -> Expr.th_elt -> Explanation.t -> unit
   val decision_level : t -> int
@@ -154,6 +158,11 @@ module Vheap = Heap.Make(struct
     let compare (a : t) (b : t) = Stdlib.compare b.weight a.weight
   end)
 
+let is_semantic (a : Atom.atom) =
+  match Shostak.Literal.view a.lit with
+  | LTerm _ -> false
+  | LSem _ -> true
+
 module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
 
   module Matoms = Atom.Map
@@ -161,6 +170,8 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
   type th = Th.t
   type t =
     {
+      hcons_env : Atom.hcons_env;
+
       (* si vrai, les contraintes sont deja fausses *)
       mutable is_unsat : bool;
 
@@ -186,6 +197,15 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
 
       (* une pile qui pointe vers les niveaux de decision dans trail *)
       mutable trail_lim : int Vec.t;
+
+      mutable nchoices : int ;
+      (** Number of semantic choices (case splits) that have been made. Semantic
+          literals are not counted as assignments but are still part of the
+          trail; we keep track of this number to properly compute the number of
+          assignments to boolean literals in [nb_assigns]. *)
+
+      mutable nchoices_stack : int Vec.t;
+      (** Stack for [nchoices] values. *)
 
       (* Tete de la File des faits unitaires a propager.
          C'est un index vers le trail *)
@@ -368,13 +388,31 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
       mutable increm_guards : Atom.atom Vec.t;
 
       mutable next_dec_guard : int;
+
+      mutable next_decisions : Atom.atom list;
+      (** Literals that must be decided on before the solver can answer [Sat].
+
+          These are added by the theory through calls to [acts_add_decision]. *)
+
+      mutable next_objective :
+        (Objective.Function.t * Objective.Value.t * Atom.atom) option;
+      (** Objective functions that must be optimized before the solver can
+          answer [Sat].
+
+          The provided [Atom.atom] correspond to a decision forcing the
+          [Function.t] to the corresponding [Value.t] (or nudging the model
+          towards "more optimal" values if the objective is not reachable); once
+          the decision is made by the solver, the optimized value is sent back
+          to the theory through [Th.add_objective]. *)
     }
 
   exception Conflict of Atom.clause
   (*module Make (Dummy : sig end) = struct*)
 
-  let empty () =
+  let create hcons_env =
     {
+      hcons_env;
+
       is_unsat = false;
 
       unsat_core = None;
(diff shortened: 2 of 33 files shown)


bclement-ocp:
I forgot to mention but: the `--enable-sat-cs` flag currently only applies to the Arrays theory, because for some theories (e.g. integers) it leads to regressions, probably due to polluting the main environment too much with irrelevant decisions. Further work towards #901 should investigate strategies to gradually move the other theories to this mechanism without introducing regressions (the enum and bit-vector theories should be prime candidates).

src/lib/structures/literal.ml
```diff
@@ -0,0 +1,111 @@
...
+  let pp = pp_view Sem.print
+  let hash = hash_view Sem.hash
+  let equal = equal_view Sem.equal
+  let compare = compare_view Sem.compare
+  let neg = neg_view Sem.neg
+
+  let normal_form = function
+    | LTerm e ->
+      (* XXX do better *)
+      (* Note: I don't know what "better" is. *)
```
Halbaroth, line 85: Ah ah :D I know exactly where you got this comment and I don't know what it means too.

src/lib/structures/literal.mli
```diff
@@ -0,0 +1,85 @@
...
+  (** [neg t] is the boolean negation of [t]. *)
+
+  val normal_form : t -> t * bool
+  (** [normal_form l] returns the normal form of [l]. The normal form of [l] is
+      a pair [l', is_neg] such that:
+
+      - [l] is [neg l'] if [is_neg] is [true]
+      - [l] is [l'] if [is_neg] is [false]
+      - [normal_form l'] is [l', false] *)
+
```
Halbaroth: As `l'` is the returned value of `normal_form l`, it's easier to understand these conditions in terms of `l`:
```suggestion
      - [l'] is [neg l] if [is_neg] is [true]
      - [l'] is [l] if [is_neg] is [false]
      - [normal_form l'] is [l', false] *)
```
and `is_neg = is_neg l`.

src/lib/structures/objective.mli
```diff
@@ -90,7 +90,7 @@ module Model : sig
   (** [functions mdl] returns the list of objective functions of the model
       [mdl] in decreasing order of priority. *)
 
-  val next_unknown : for_model:bool -> t -> (Function.t * int) option
+  val next_unknown : t -> Function.t option
   (** [next_unknown ~for_model mdl] returns the next optimization in
```
Halbaroth, line 94: We have to update the documentation too ;)

src/lib/structures/satml_types.mli
```diff
@@ -111,6 +110,8 @@ module type ATOM = sig
   val hash_atom  : atom -> int
   val tag_atom   : atom -> int
 
+  val add_lit_atom :
+    hcons_env -> Shostak.Literal.t -> var list -> atom * var list
```
Halbaroth: I'm not very fond of this name. We may call it `add_sem_atom`?

src/lib/reasoners/satml.ml
```diff
@@ -386,13 +406,31 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
       mutable increm_guards : Atom.atom Vec.t;
 
       mutable next_dec_guard : int;
+
+      mutable next_decisions : Atom.atom list;
+      (** Literals that must be decided on before the solver can answer [Sat].
+
+          These are added by the theory through calls to [acts_add_decision]. *)
```
Halbaroth, line 395: We may say that these decisions are always semantic.

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1057,7 +1092,7 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
             | Limit (_, v) ->
               raise (Give_up ((e, v, is_max, false) :: acc))
             | Unknown ->
-              assert false
```
Halbaroth: Is it an oversight or this case can occur?

src/lib/reasoners/theory.ml
```diff
@@ -689,32 +631,138 @@ module Main_Default : S = struct
...
+         For the [Real] or [Int] expressions, we have two cases:
+         - If the objective function is a linear functions of variables, the
+           decision procedure implemented in Ocplib-simplex cannot fail to
+           optimize the split. For instance, if we try to maximize the
+           expression:
+             5 * x + 2 * y + 3 where x and y are real variables,
+           the procedure will success to produce the upper bound of [x] and
+           [y] modulo the other constraints on it.
+         - If the objective function isn't linear, the nonlinear part of the
+           expression have seen as uninterpreted term of the arithemic theory.
```
Halbaroth: :D I didn't open an PR for such a small mistake 

```suggestion
           expression is seen as uninterpreted term of the arithmetic theory.
```

src/lib/reasoners/theory.ml
```diff
@@ -689,32 +631,138 @@ module Main_Default : S = struct
...
+            match L.view a with
+            | Literal.LTerm a when E.Set.mem a assumed_set -> accu
+            | aview ->
+              let gamma, aview =
+                match aview with
+                | LTerm t -> gamma, Literal.LTerm t
+                | LSem a ->
+                  (* When assuming a semantic literal (typically a case split),
+                     we may end up in cases where they contain terms that have
+                     not been added. *)
```
Halbaroth: We may mention here the cause of this issue, that is some semantic literals can be repushed on the trail of SatML after backtracking the environment of the union-find ;)

src/lib/reasoners/theory.ml
```diff
@@ -689,32 +631,138 @@ module Main_Default : S = struct
...
+                  let gamma = SE.fold (fun t gamma ->
+                      fst @@ CC_X.add_term gamma facts t Ex.empty
+                    ) trms gamma in
+                  gamma, LSem aview
+              in
+              CC_X.add_fact facts (aview, ex, o);
+              let assumed_set =
+                match aview with
+                | LTerm a -> E.Set.add a assumed_set
+                | _ -> assumed_set
```
Halbaroth, line 757: It seems that `assumed_set` is used for two things:
- it's a field of the record `Models.t` which is used in a deprecated feature.
- If we use a tableaux-like method for theories (which is the default), we try to instantiate free variables of the atoms of `assumed_set` in order to produce new facts. I guess the idea is to reach an eventual contradiction by using the instantiation engine here. With the current implementation, we cannot match semantic values, so it makes sense you ignore semantic literals here, but I wonder if it would be relevant if we could do it? I believe it's not relevant because such contradictions should be found in the relation theories before emitting the case-split.

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2102,11 +2102,16 @@ let optimizing_objective env uf Objective.Function.{ e; is_max; _ } =
             Objective.Value.Minfinity
         in
         (* As the problem is unbounded, we need to produce a value for the
-           objective function. In this case, we pick a value in the domain
-           of the polynomial [p]. *)
+           objective function. Rather than picking a value directly, we add a
+           constraint to let the case split mechanism pick a "large" (or
+           "small") value in the interval (otherwise, later objectives are
+           computed while assuming that the middle value is forced, which is
+           incorrect). *)
```
Halbaroth, line 2078: We should update the documentation of `middle_value` as we don't use this function anymore to force a specific value.

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -2102,11 +2102,16 @@ let optimizing_objective env uf Objective.Function.{ e; is_max; _ } =
...
         let case_split =
-          LR.mkv_eq r1 (middle_value env ~is_max ty p None), true, Th_util.CS
-            (Th_util.Th_arith, Q.one)
+          LR.mkv_builtin false LT
+            [r1; (middle_value env ~is_max ty p None)],
```
Halbaroth:
```suggestion
            [r1; middle_value env ~is_max ty p None],
```

bclement-ocp:
So it turns out that I forgot to reset the number of choices when backtracking, which means that we reply "sat" from satML too early, which in turns means we are more aggressive with instantiation, and that seems to be beneficial somehow, since we get an improvement… but doing what should be the right thing (i.e. actually counting semantic decisions properly) causes large regressions (`+69-137`). I will investigate next week.

bclement-ocp:
> So it turns out that I forgot to reset the number of choices when backtracking, which means that we reply "sat" from satML too early, which in turns means we are more aggressive with instantiation, and that seems to be beneficial somehow, since we get an improvement… but doing what should be the right thing (i.e. actually counting semantic decisions properly) causes large regressions (`+69-137`). I will investigate next week.

Turns out I made a typo while rebasing: https://github.com/OCamlPro/alt-ergo/pull/1027/commits/d708f3315ca9b0bc250c90d5f81160cf85724905

Rerunning the benches.

bclement-ocp:
The benches now say `+49-14` with the correct code rebased on e18c2d17

src/lib/reasoners/th_util.mli
```diff
@@ -107,3 +107,36 @@ type optimized_split = {
...
+    Inspired by mSAT's equivalent type [1].
+
+    [1] :
+      https://github.com/Gbury/mSAT/blob/ \
+      1496a48bc8b948e4d5a2bc20edaec33a6901c8fa/src/core/Solver_intf.ml#L104 *)
+type 'literal acts = {
+  acts_add_decision_lit : 'literal -> unit ;
+  (** Ask the SAT solver to decide on the given formula before it can answer
+      [SAT]. The order of decisions is among multiple calls of
+      [acts_add_decision_lit] is unspecified.
```
Halbaroth:
```suggestion
      [SAT]. The order of decisions among multiple calls of
      [acts_add_decision_lit] is unspecified.
```

src/lib/reasoners/theory.ml
```diff
@@ -898,22 +956,17 @@ module Main_Default : S = struct
 
   let get_assumed env = env.assumed_set
 
-  let optimize env ~is_max f =
+  let get_objectives env = env.objectives
+
+  let add_objective env fn value =
     (* We have to add the term [f] and its subterms as the MaxSMT
        syntax allows to optimize expressions that aren't part of
        the current context. *)
```
Halbaroth:
```suggestion
    (* We have to add the term [fn] and its subterms as the MaxSMT
       syntax allows to optimize expressions that aren't part of
       the current context. *)
```

src/lib/reasoners/theory.ml
```diff
@@ -494,33 +450,21 @@ module Main_Default : S = struct
...
     and aux env sem_facts acc_choices new_choices =
       Options.tool_req 3 "TR-CCX-CS-Case-Split";
       match new_choices with
       | [] ->
-        begin match Objective.Model.next_unknown env.objectives ~for_model with
-          | Some (obj, order) ->
-            optimizing_objective env sem_facts acc_choices obj order
-          | None ->
-            generate_choices env sem_facts acc_choices
-        end
```
Halbaroth, line 518: After this modification, I believe the `aux` dispatcher isn't useful anymore. If `generate_choices` found new choices, we can call `propagate_choices` directly in this function. In `propagate_choices`, we call `generate_choices` if there is no more choices to propagate. Finally we replace the top-level call of `aux` in `look_for_sat` by:
```ocaml
propagate_choices env sem_facts (List.rev env.choices) new_choices
```

Notice we mention `aux` in the documentation of this function too ;)

src/lib/reasoners/theory.ml
```diff
@@ -611,14 +554,6 @@ module Main_Default : S = struct
...
               let uf =  CC_X.get_union_find t.gamma in
               let filt_choices = filter_choices uf t.choices in
-              let filt_choices =
-                if  Objective.Model.is_empty t.objectives ||
-                    Objective.Model.has_no_limit t.objectives
-                    (* otherwise, we may be unsound because infty optims
-                       are not propagated *)
-                then filt_choices
-                else []
-              in
```
Halbaroth, line 623: From my understanding, this piece of code ensures we don't reuse previous choices in presence of optimization after backtracking in the SAT solver. Is it correct to remove it because we entail optimization splits in the SAT environment in this PR?

src/lib/reasoners/satml.ml
```diff
@@ -1318,15 +1388,34 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
...
         let lclause =
           Atom.make_clause name learnt vraie_form true history
         in
-        Vec.push env.learnts lclause;
-        attach_clause env lclause;
-        clause_bump_activity env lclause;
+        (* Do not learn clauses involving semantic literals (but still record
+           them as a reason for the propagated literal), otherwise we could
+           learn stuff about these semantic literals in a scope where they are
+           no longer valid.
```
Halbaroth: I'm not sure to understand this comment. Do you avoid semantic literals here for sake of soundness or efficiency? From my understanding, if we learn something in the CDCL framework, this fact have to be entailed by the problem. It may be irrelevant after backtracking but still correct. So I guess you meant we should avoid learning semantic facts because the case-split mechanism could produce a lot of them and we don't want to increase too much the size of the clause database?

Halbaroth (approved):

src/lib/structures/objective.mli
bclement-ocp: Done.

src/lib/structures/satml_types.mli
bclement-ocp: I don't like it either, but I do not have a better solution. `add_sem_atom` would be confusing IMO because `Shostak.Literal.t` can contain either "syntaxic" literals (`LTerm`) or "semantic" literals (`LSem`).

Maybe renaming this one to `add_atom` and the current `add_atom` to `add_expr_atom`?

src/lib/reasoners/satml.ml
bclement-ocp: I'd rather not; while it is true that currently only semantic decisions are added here, the implementation doesn't actually care.

src/lib/reasoners/satml_frontend.ml
bclement-ocp: Oversight; fixed.

src/lib/reasoners/satml.ml
bclement-ocp: Clauses are learned at level 0 but semantic literals can includes terms created by the Shostak module that depend on assertions at greater levels, and hence do not exist (in the union-find etc.) at level 0 (what I mean by "no longer valid" here). I have updated and clarified the comment.

src/lib/reasoners/theory.ml
bclement-ocp: Done.

src/lib/reasoners/theory.ml
bclement-ocp: Yes, now that optimizations are applied in the main environment, nothing special needs to be done in the case-split environment, it just fits the existing framework.

src/lib/reasoners/theory.ml
bclement-ocp: Clarified the comment.

src/lib/reasoners/theory.ml
bclement-ocp: For the first point: the `prop_model` in model generation needs the value of the boolean variables of the original problem; semantic literals are not in the original problem and are not needed there.

For the second point: theoretically I guess we could, however I believe that the `assumed_set` is only used to expand user-defined predicates, so it would not really be useful here.

src/lib/reasoners/intervalCalculus.ml
bclement-ocp: Done.

bclement-ocp (commented):
Should be ready for round 2.

(Note: the changes are fairly involved and so I want to re-run the benches with the exact code that will get merged prior to merging; please don't merge when approving)

Halbaroth:
Looks good to me, you can merge the PR.

bclement-ocp:
There is an additional bit of code that I believe is needed in `try_to_backjump_further` (we could end up taking a semantic decision from the "fake" `lazy_cnf` that we shouldn't have taken); I will fix this and rerun benchmarks before merging.

bclement-ocp:
> There is an additional bit of code that I believe is needed in `try_to_backjump_further` (we could end up taking a semantic decision from the "fake" `lazy_cnf` that we shouldn't have taken); I will fix this and rerun benchmarks before merging.

Pushed the aforementioned change as well as separation of `next_decisions` into `next_decisions` (mandatory, keep all) and `next_split` (optional, keep only one). This is because otherwise in the initial processing phase (before actually solving) we will add the same split over and over again to `next_decisions`. I have kept `next_decisions` even though it is no longer used; I'm not too sure about this — it can be useful in the near future but also could be re-introduced at that point.

Re-running benchmarks before merging; although we still need to decide if `--enable-sat-cs` should be on by default (I'd say yes, but I'm also OK waiting until we can enable it for all theories w/o regressions).

bclement-ocp:
The current version is +50-8 on AE-format.

bclement-ocp:
I will go ahead and merge this now with the flag disabled — I think it is acceptable to have it as an experimental flag (and this simplifies the optimization code somewhat) but we should have a comprehensive story for all theories before enabling the flag by default.

PR #1028, opened 2024-01-19
cleanup: Use proper namespacing for names

bclement-ocp:
This introduces proper namespacing for variable names instead of relying on string prefixes. In order to disturb the ordering as little as possible, names are pre-mangled using their namespace (e.g. "x" in the `Internal` namespace is stored as ".!x") and further operations (hashing, comparison, etc.) are performed on the mangled names.


TODO
```diff
@@ -0,0 +1 @@
+fake_XX unused since aebf5ac7437b57aacc6aa6389e237c6968a3a84f
```
Halbaroth: Is it an artifact?

Halbaroth (approved):

TODO
bclement-ocp: Oops yes that is a rebase artifact :sweat_smile: will remove before merging

PR #1036, opened 2024-02-07
Fix the `Dune` rule to generate the manpage

### LINKED ISSUE #1035: Alt-Ergo 2.5.2 doesn't compile with `Dune 3.13`
Last week, I noticed that `Alt-Ergo 2.5.2` cannot be installed on Marvin with `opam` anymore. 
It was a fresh installation (the switch was completely new), so I guessed the issue comes from one of the dependencies of `Alt-Ergo`.

I didn't find a way to produce any valuable logs with `Dune` but I got the same bug on my laptop this afternoon so I started a long bisection with `opam` and `dune`. 

My first guess was correct: the issue comes from the last version of `Dune`. It seems this combination of rules produces a deadlock somewhere in `Dune 3.13` but everything is fine with older versions:
```dune
(executable
  (name Main_text)
  (public_name alt-ergo)
  (package alt-ergo)
  (libraries alt_ergo_common)
  (link_flags (:include flags.dune))
  (modules Main_text)
  (promote (until-clean)))

(rule
  (target alt-ergo.1)
  (action (with-stdout-to %{target} (run alt-ergo --help=groff))))

(install
 (files alt-ergo.1)
 (section man)
 (package alt-ergo))
```

If I removes the rules to generate and install the manpage, `Alt-Ergo` compiles. I believe `dune` is stuck in a deadlock.

To solve this issue I have several options:
- For the release `2.5.2`, we can amend the package on `opam-repository` to add an upper bound for the `dune` version.
- We can remove the rule completely. Honestly, no one cares about the man page as we get a similar result with `alt-ergo --help` and I'm not fond of running `Alt-Ergo` while installing it. 

It seems the compilation works if we specify the dependency:
```dune 
(rule
  (deps (:bin Main_text.exe))
  (target alt-ergo.1)
  (action (with-stdout-to %{target} (run %{bin} --help=groff))))
```

I don't know if it's a bug of `Dune`. I open an issue https://github.com/ocaml/dune/issues/9944 for it.

Halbaroth:
`Dune 3.13` seems to be stuck if we don't specify explicitly the dependency between the executable `Main_text.exe` and the rule which produces the manpage.

Add an explicit dependency clause to prevent this issue.

This commit has to be backport on `v2.5.x` too.

### DIFF
--- a/src/bin/text/dune
+++ b/src/bin/text/dune
@@ -12,8 +12,9 @@
 
 ; Rule to generate a man page for alt-ergo
 (rule
+  (deps (:bin Main_text.exe))
   (target alt-ergo.1)
-  (action (with-stdout-to %{target} (run alt-ergo --help=groff))))
+  (action (with-stdout-to %{target} (run %{bin} --help=groff))))
 
 ; Install the man page
 (install


bclement-ocp:
Please use `%{bin:alt-ergo}` instead of using then internal `Main_text.exe`, as suggested [here](https://github.com/OCamlPro/alt-ergo/issues/1035#issuecomment-1932574424).

Halbaroth:
Actually, I tried this but it doesn't work with this patch:
```
(rule
  (target alt-ergo.1)
  (action (with-stdout-to %{target} (run %{bin:alt-ergo} --help=groff))))
```
I got an infinity loop. I also tried 
```
(rule
  (deps (:bin alt-ergo))
  (target alt-ergo.1)
  (action (with-stdout-to %{target} (run %{bin} --help=groff))))
```
but `Dune` doesn't know a rule to produce `alt-ergo`.

bclement-ocp:
> Actually, I tried this but it doesn't work with this patch:
> 
> ```
> (rule
>   (target alt-ergo.1)
>   (action (with-stdout-to %{target} (run %{bin:alt-ergo} --help=groff))))
> ```
> 
> I got an infinity loop. I also tried
> 
> ```
> (rule
>   (deps (:bin alt-ergo))
>   (target alt-ergo.1)
>   (action (with-stdout-to %{target} (run %{bin} --help=groff))))
> ```
> 
> but `Dune` doesn't know a rule to produce `alt-ergo`.

Given that this looks to be a (fixed) [dune bug](https://github.com/ocaml/dune/issues/9944#issuecomment-1933563549) I don't think we should do anything (I was wrong about `run alt-ergo` not following dependencies, as per [the dune doc](https://dune.readthedocs.io/en/stable/concepts/variables.html): "Note that (run %{bin:program} ...) and (run program ...) behave in the same way.").

emillon:
FYI, you can also infer the target and use the short form:

```
(rule
 (with-stdout-to
  alt-ergo.1
  (run alt-ergo --help=groff)))
```

PR #1039, opened 2024-02-07
Use a wrapper to set timers

Halbaroth:
We wrap some important functions with timers in order to profile their calls but we don't use a dedicated function to do this job everywhere in the codebase. In particular, there is some places where we don't ensure the timer is correctly paused even if the wrapped function raises an exception.

This PR adds a new wrapper `with_timer` to perform this job correctly.

Add timers for the ADT and Bitv theories too.

### DIFF
--- a/src/lib/reasoners/relation.ml
+++ b/src/lib/reasoners/relation.ml
@@ -127,6 +127,7 @@ let try_query (type a) (module R : Sig_rel.RELATION with type t = a) env uf a
   | None -> k ()
 
 let query env uf a =
+  Options.exec_thread_yield ();
   try_query (module Rel1) env.r1 uf a @@ fun () ->
   try_query (module Rel2) env.r2 uf a @@ fun () ->
   try_query (module Rel3) env.r3 uf a @@ fun () ->
--- a/src/lib/reasoners/relation.ml
+++ b/src/lib/reasoners/relation.ml
@@ -46,6 +46,8 @@ module Rel6 : Sig_rel.RELATION = Adt_rel
 
 module Rel7 : Sig_rel.RELATION = Ite_rel
 
+let timer = Timers.M_None
+
 type t = {
   r1: Rel1.t;
   r2: Rel2.t;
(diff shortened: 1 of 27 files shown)


bclement-ocp:
Instead of adding a `Rel_utils.Instrumentation` module that is intended to be used in exactly the same way in all relation modules (and silently providing no errors if not used), I'd rather centralize the timer management in the `Shostak` and `Relations` module as much as possible by:

 - Adding `val timer : Timers.ty_module` to `Sig.SHOSTAK` and to the relations (we can either add it to the `RELATION` signature and use `M_None` for the `Relation` module or go for the ugly-but-practical `let timer1 = Timers.M_Arith` in `Relation`)
 - Wrap theory calls with the appropriate timer directly in the `Shostak` and `Relation` modules

What do you think?

src/lib/reasoners/relation.ml
```diff
@@ -103,26 +119,21 @@ let assume_th_elt env th_elt dep =
...
-        match Rel4.query env.r4 uf a with
-        | Some _ as ans -> ans
-        | None ->
-          match Rel5.query env.r5 uf a with
-          | Some _ as ans -> ans
-          | None ->
-            match Rel6.query env.r6 uf a with
-            | Some _ as ans -> ans
-            | None -> Rel7.query env.r7 uf a
+  try_query (module Rel1) env.r1 uf a @@ fun () ->
```
bclement-ocp, line 132: I am not sure it actually matters but let's keep the `Options.exec_thread_yield ()`.

src/lib/reasoners/shostak.ml
```diff
@@ -383,19 +383,33 @@ struct
...
     | false , false , false , false, true , false ->
+      Timers.with_timer Timers.M_Adt Timers.F_make @@ fun () ->
       ADT.make t
+
     | false , false , false , false, false, true  ->
+      Timers.with_timer Timers.M_AC Timers.F_make @@ fun () ->
       AC.make t
+
     | false , false , false , false, false, false ->
+      Timers.with_timer Timers.M_None Timers.F_make @@ fun () ->
```
bclement-ocp: I don't think `M_None` should actually be used; let's keep `term_embed` untimed

src/lib/reasoners/shostak.ml
```diff
@@ -544,14 +558,31 @@ struct
...
+
           (*| Ty.Tunit           -> pb *)
+
           | Ty.Tadt _ when not (Options.get_disable_adts()) ->
+            Timers.with_timer ADT.timer Timers.F_solve @@ fun () ->
             ADT.solve ra rb pb
-          | _                  -> solve_uninterpreted ra rb pb
+
+          | _                  ->
+            Timers.with_timer Timers.M_None Timers.F_solve @@ fun () ->
```
bclement-ocp: Same as above; `M_None` should not be used.

bclement-ocp (commented):
The code looks nicer now :)

It does add quite a bit of indirections to a couple of hot paths and may have an impact on performance (which I hope is minimal… but maybe not); can you do a benchmark run (with no profiling)?

(As an aside: I wonder how much we actually need to embed profiling into release builds, in which case we could replace all of this infrastructure with the `landmarks` library instead)

src/lib/reasoners/relation.ml
Halbaroth: oups! `exec_thread_yield` was here for the GUI but I agree we have to keep it as it tells use it's a hot path.

Halbaroth:
> It does add quite a bit of indirections to a couple of hot paths and may have an impact on performance (which I hope is minimal… but maybe not); can you do a benchmark run (with no profiling)?

I thought the same thing but I guess it should be fine. I'll benchmark.

Halbaroth:
> (As an aside: I wonder how much we actually need to embed profiling into release builds, in which case we could replace all of this infrastructure with the `landmarks` library instead)

I worked hard this week to restore the profiling feature, so no...

src/lib/reasoners/shostak.ml
Halbaroth: Or we can create a new module name: `M_Combine`.

bclement-ocp:
> I worked hard this week to restore the profiling feature, so no...

What is broken? I used it yesterday and it seemed to work. Isn't it `--bottom-classes` that is broken (which we should probably have an issue for)?

bclement-ocp:
Ah I guess you mean the "we don't properly pause" thing, gotcha.

src/lib/reasoners/shostak.ml
bclement-ocp: Sure, that works too

bclement-ocp: Although `term_embed` is so trivial that I am not sure the timing overhead will not be more expensive than the function itself

Halbaroth: Right, for this one is useless.

bclement-ocp:
> I thought the same thing but I guess it should be fine. I'll benchmark.

Looks to be a wash, even a tad faster on this PR which I don't really understand but I'll take it — let's remove the use of the `M_None` timer and merge :tada:

bclement-ocp (approved):

src/lib/reasoners/shostak.ml
Halbaroth: I removed it in this case.

PR #1040, opened 2024-02-08
cleanup(CP): Simplify constraint handling

bclement-ocp:
This patch rewrites the `Constraints_make` module in order to make it more flexible (this is preparatory step for dealing with arithmetic bit-vector constraints, see #903).

More precisely:

 - Constraints no longer need to keep track of their own explanations, this is now entirely done by the `Constraints_make` functor. This makes it simpler to write `Constraint` modules, and avoid repeating boilerplate code to deal with explanation storage. Instead, the explanations are provided to the `Constraint` module in its `propagate` function.

 - The `Constraints_make` functor no longer need to know about constraint propagation. Instead, it simply keeps track of constraints that need to be propagated (pending constraints), and provides an API to iterate (and remove) the set of constraints to be propagated, letting the user take care of propagation proper.

 - The `Constraints_make` functor now tracks separately the constraint arguments and the leaves of said arguments. The leaves are used to know which constraints need to be updated during a substitution, and the arguments are used to mark as pending all constraints that apply to a given representative when its domain is updated (note that, for the bit-list domains, we actually store the domains at the leaves, so the arguments mapping is not used -- but this still makes the module more flexible in general, and in particular will allow to introduce arithmetic domains that need to be stored for all values, not only leaves, for the purpose of #903 in a future PR)

The new design should make it easier to write `Constraint` modules. It also fixes a bug in the contract between the `Constraint` and `Domain` modules regarding substitution: the `Domain` modules was written under the assumption that constraints applying to `changed` domains would always be marked as pending upon substitution, but that is not true because we use functional representations where such updates are delayed, and hence the `changed` flag needs to be propagated after substitution (if appropriate).

### DIFF
--- a/src/lib/reasoners/rel_utils.ml
+++ b/src/lib/reasoners/rel_utils.ml
@@ -197,182 +197,210 @@ module type Constraint = sig
   type t
   (** The type of constraints.
 
-      Constraints are associated with a justification as to why they are
-      currently valid. The justification is only used to update domains,
-      identical constraints with different justifications will otherwise behave
-      identically (and, notably, will compare equal).
-
-      Constraints contains semantic values / term representatives of type
-      [X.r]. We maintain the invariant that the semantic values used inside the
-      constraints are *class representatives* i.e. normal forms wrt the `Uf`
-      module, i.e. constraints have a normalized representation. Use `subst` to
-      ensure normalization. *)
+      Constraints apply to semantic values of type [X.r] as arguments. *)
 
   val pp : t Fmt.t
   (** Pretty-printer for constraints. *)
 
   val compare : t -> t -> int
-  (** Comparison function for constraints.
+  (** Comparison function for constraints. The comparison function is
+      arbitrary and has no semantic meaning. You should not depend on any of
+      its properties, other than it defines an (arbitrary) total order on
+      constraint representations. *)
 
-      Constraints typically include explanations, which should not be included
-      in the comparison function: code working with constraints expects
-      constraints with identical representations but different explanations to
-      compare equal.
+  val fold_args : (X.r -> 'a -> 'a) -> t -> 'a -> 'a
+  (** [fold_args f c acc] folds function [f] over the arguments of constraint
+      [c].
 
-      {b Note}: The comparison function is arbitrary and has no semantic
-      meaning. You should not depend on any of its properties, other than it
-      defines an (arbitrary) total order on constraint representations. *)
+      During propagation, the constraint {b MUST} only look at (and update)
+      the domains associated of its arguments; it is not allowed to look at
+      the domains of other semantic values. This allows efficient updates of
+      the pending constraints. *)
 
-  val subst : Explanation.t -> X.r -> X.r -> t -> t
-  (** [subst ex p v cs] replaces all the instances of [p] with [v] in the
+  val subst : X.r -> X.r -> t -> t
+  (** [subst p v cs] replaces all the instances of [p] with [v] in the
       constraint.
 
-      Use this to ensure that the representation is always normalized.
-
-      The explanation [ex] justifies the equality [p = v]. *)
-
-  val fold_leaves : (X.r -> 'a -> 'a) -> t -> 'a -> 'a
-
-  type domain
-  (** The type of domains.
-
-      This is typically a mapping from variables to their own domain, but no
-      expectations is made upon the actual structure of that type. *)
-
-  val propagate : t -> domain -> domain
-  (** [propagate c dom] propagates the constraints [c] in [d] and returns the
-      new domain. *)
-
+      Substitution can perform constraint simplification. *)
 end
 
-module Constraints_Make(Constraint : Constraint) : sig
+type 'a explained = { value : 'a ; explanation : Explanation.t }
+
+let explained ~ex value = { value ; explanation = ex }
+
+module Constraints_make(Constraint : Constraint) : sig
   type t
-  (** The type of constraint sets. A constraint sets records a set of
-      constraints that applies to semantic values, and remembers which
-      constraints are associated with each semantic values.
+  (** The type of constraint sets. A constraint set records a set of
+      constraints that applies to semantic values, and remembers the relation
+      between constraints and semantic values.
 
-      It is used to only propagate constraints involving semantic values whose
-      associated domain has changed.
+      The constraints associated with specific semantic values can be notified
+      (see [notify]), which is used to only propagate constraints involving
+      semantic values whose domain has changed.
 
-      The constraint sets are expected to keep track of *class representatives*,
-      i.e.  normal forms wrt the `Uf` module, in which case we say the
-      constraint set is *normalized*. Use `subst` to ensure normalization. *)
+      The constraints that have been notified are called "pending
+      constraints", and the set thereof is the "pending set". These are
+      constraints that need to be propagated, and can be recovered using
+      [next_pending]. *)
 
   val pp : t Fmt.t
   (** Pretty-printer for constraint sets. *)
 
   val empty : t
-  (** Returns an empty constraint set. *)
+  (** The empty constraint set. *)
 
-  val subst : Explanation.t -> X.r -> X.r -> t -> t
-  (** [subst ex p v cs] replaces all the instances of [p] with [v] in the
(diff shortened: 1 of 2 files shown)


bclement-ocp:
I have pushed a small modification to a comment stating a restriction on constraint substitutions that was true in an early implementation of this PR, but is no longer a restriction.

For more details, the restriction was about preventing constraint substitution from preventing simplifications, which was intended to avoid having to mark the new constraint as pending after substitution as an optimization. However, it turned out to be too tricky to do this properly, and so I have removed the optimization — and hence the restriction — but forgot to update that comment.

bclement-ocp:
> it turned out to be too tricky to do this properly

Actually just realized that this is because in the current version the `bvxor` constraint does perform simplifications and it was never going to work (unlike the context where the PR was originally written, with a version of `bvxor` that does not perform simplifications) (:

src/lib/reasoners/rel_utils.ml
```diff
@@ -197,182 +197,210 @@ module type Constraint = sig
...
+    { active; args_map; leaves_map; pending }
+
+  let subst ~ex rr nrr t =
+    match MX.find rr t.leaves_map with
+    | cs ->
+      CS.fold (fun c t ->
+          let pending = CS.mem c t.pending in
+          let t = remove c t  in
+          let ex = Explanation.union ex c.explanation in
+          add ~pending ~ex (Constraint.subst rr nrr c.value) t
```
Halbaroth: I believe the flag `pending` is useless here. Indeed, the constraint `c` is removed in `remove c t` (as the comparison function of the module `CS` ignores the explanation, even if the underlying explanation of `c` changed during this function iteration, we remove the only copy of `c` in the map), thus the function `CS.find` will always raise `Not_found` in the function `add`.

src/lib/reasoners/rel_utils.ml
```diff
@@ -197,182 +197,210 @@ module type Constraint = sig
...
+      @raise Not_found if there are no pending constraints. *)
 end = struct
-  module IM = Util.MI
   module MX = Shostak.MXH
 
-  module CS = Set.Make(Constraint)
+  module CS = Set.Make(struct
+      type t = Constraint.t explained
+
+      let compare a b = Constraint.compare a.value b.value
```
Halbaroth, line 296: I don't know if it was like this before but the fact we ignore explanations here means we cannot store several times the same constraint with not comparable explanations (by not comparable explanations, I mean there is no one of them which is subsumed by the other one).

We should document this point.

src/lib/reasoners/rel_utils.ml
```diff
@@ -197,182 +197,210 @@ module type Constraint = sig
...
+        | None -> Some (CS.singleton c)
       ) cs_map
 
-  let cs_remove cs r cs_map =
+  let fold_leaves f c acc =
+    Constraint.fold_args (fun r acc ->
+        List.fold_left (fun acc r -> f r acc) acc (X.leaves r)
+      ) c acc
+
+  let add ?(pending = false) ~ex c t =
```
Halbaroth: We can remove the flag `pending` as we don't use it anymore.

src/lib/reasoners/rel_utils.ml
```diff
@@ -197,182 +197,210 @@ module type Constraint = sig
...
-    (*** Mapping from semantic values to the constraints that involves them *)
-    fresh : CS.t ;
-    (*** Fresh constraints that have never been propagated *)
+    args_map : CS.t MX.t ;
+    (** Mapping from semantic values to constraints involving them *)
+
+    leaves_map : CS.t MX.t ;
+    (** Mapping from semantic values to constraints they are a leaf of *)
+
+    active : CS.t ;
```
Halbaroth, line 306: I guess `active` means  constraint already propagated with its explanation to `CC(X)`? `propagated` may be a better name?

src/lib/reasoners/rel_utils.ml
```diff
@@ -197,182 +197,210 @@ module type Constraint = sig
...
+    let active = CS.remove c t.active in
+    let args_map =
+      Constraint.fold_args (cs_remove c) c.value t.args_map
+    in
+    let leaves_map = fold_leaves (cs_remove c) c.value t.leaves_map in
+    let pending = CS.remove c t.pending in
+    { active; args_map; leaves_map; pending }
+
+  let subst ~ex rr nrr t =
+    match MX.find rr t.leaves_map with
```
Halbaroth, line 371: Can you explain why is it correct to look up in `t.leaves_map` instead of `t.args_map`?
I guess it's because we perform substitution on leaves?

Halbaroth (changes requested):

src/lib/reasoners/rel_utils.ml
bclement-ocp: The result of `Constraint.subst rr nrr c.value` is not identical to `c`. It is possible that the resulting constraint was already there for another reason (e.g. suppose we have `C1(x, y)` that was pending and `C1(x, x)` that was already present but not pending, then we apply the substitution `y → x` and `C1(x, y)` becomes `C1(x, x)` but we want to preserve the fact that it was pending and needs to be propagated.

Although if this is the case and `C1(x, x)` was not pending then presumably it has already been propagated? This is probably a relic that was necessary for the optimization I mentioned in another comment but is no longer required without that optimization; I think we can indeed remove the flag here.

src/lib/reasoners/rel_utils.ml
bclement-ocp: This was already the case before, it was explained in the documentation of `Constraint.compare`. This is now enforced by `Constraints_make` rather than `Constraint`, so I removed it from the documentation of `Constraint.compare`; I will mention it in the documentation of `Constraints_make`.

src/lib/reasoners/rel_utils.ml
bclement-ocp: `propagated` would be confusing because this set can contain constraints that have actually never been propagated.

`active` means constraints that are, well, active (the CP community would say "posted" but I don't like that term) — they must be true in the model, and will be propagated when the domain of their arguments change.

src/lib/reasoners/rel_utils.ml
bclement-ocp: By definition, substitutions apply on leaves (terms and AC applications). It would be incorrect to look in `t.args_map`: for instance, if C is the constraint `r = x & (y ++ z)` (where `++` is BV concatenation) then the `args_map` is `r → C; x → C; (y ++ z) → C` but substitutions can only apply to `r`, `x`, `y` and `z`.

bclement-ocp: I will clarify this because it's not strictly correct: Alt-Ergo does propagate `Subst` rewritings for superterms; for instance, if we substitute `z → z'` then the UF module will generate the substitution `y ++ z → y ++ z'`. However, the Shostak modules only supports substituting leaves, and we ultimately call `X.subst` here.

Now that I think about it more, I believe this was an error in the design of the original `Constraint` module — it should not call `X.subst` but rather simply perform a comparison on the result of congruence closure (i.e. the thing that gets substituted). This is probably also true of the `Domains` module.

bclement-ocp: After some more investigation: using comparison instead of `X.subst` here doesn't work for a reason related to #474: when class representatives change from `rr` to `nrr`, we *only* get the substitution `rr → nrr` if `rr` was a semantic value for a term of the initial problem (iow, if it is in `repr`). So for instance if we have substitutions `x → 0 @ .k0` (where `.k0` is a fresh constant) then `.k0 → 0 @ .k1`, the relations will only see `x → 00 @ .k1` and `.k0 → 0 @ .k1` — but not `0 @ .k0 → 00 @ .k1`.

Halbaroth: My connection is terrible, I will try again to comment...
So, as you changed the design of these modules in recent PR, I believe we can merge this one now?

bclement-ocp:
>  So, as you changed the design of these modules in recent PR, I believe we can merge this one now?

The PR is still marked as "Changes requested" so I was waiting for your 2nd round of review & the green checkmark :) I will merge taking this message as approval.

PR #1049, opened 2024-03-01
feat(plugins): Use dune-site for inequalities plugins

### LINKED ISSUE #1047: Tests failure with OCaml 5.2
The tests for alt-ergo currently fails when using OCaml 5.2
```
#=== ERROR while compiling alt-ergo.2.5.2 =====================================#
# context              2.2.0~beta2~dev | linux/x86_64 | ocaml-variants.5.2.0+trunk | file:///home/opam/opam-repository
# path                 ~/.opam/5.2/.opam-switch/build/alt-ergo.2.5.2
# command              ~/.opam/5.2/bin/dune build -p alt-ergo -j 1 --promote-install-files=false @install @runtest
# exit-code            1
# env-file             ~/.opam/log/alt-ergo-2395-562a0c.env
# output-file          ~/.opam/log/alt-ergo-2395-562a0c.out
### output ###
# File "tests/cram.t/run.t", line 1, characters 0-0:
# /usr/bin/git --no-pager diff --no-index --color=always -u _build/.sandbox/2230f8f3c94abeb6211ea04264c94d64/default/tests/cram.t/run.t _build/.sandbox/2230f8f3c94abeb6211ea04264c94d64/default/tests/cram.t/run.t.corrected
# diff --git a/_build/.sandbox/2230f8f3c94abeb6211ea04264c94d64/default/tests/cram.t/run.t b/_build/.sandbox/2230f8f3c94abeb6211ea04264c94d64/default/tests/cram.t/run.t.corrected
# index 938d571..b9a9b6d 100644
# --- a/_build/.sandbox/2230f8f3c94abeb6211ea04264c94d64/default/tests/cram.t/run.t
# +++ b/_build/.sandbox/2230f8f3c94abeb6211ea04264c94d64/default/tests/cram.t/run.t.corrected
# @@ -1,6 +1,6 @@
#    $ echo '(check-sat)' | alt-ergo --inequalities-plugin does-not-exist -i smtlib2 -o smtlib2 2>&1 >/dev/null | sed -E 's/\(\\".*\\"\)//'
#    alt-ergo: ; Fatal Error: [Dynlink] Loading the 'inequalities' reasoner (FM module) plugin in "does-not-exist" failed!
# -            >> Failure message: error loading shared library: Dynlink.Error (Dynlink.Cannot_open_dll "Failure")
# +            >> Failure message: error loading shared library: Failure("$TESTCASE_ROOT/does-not-exist: cannot open shared object file: No such file or directory")
#  
#  Now we will have some tests for the models. Note that it is okay if the format
#  changes slightly, you should be concerned with ensuring the semantic is
```
This change in the compiler is expected and comes from https://github.com/ocaml/ocaml/pull/12213 but it would be nice to have the tests fixed at some point so that we can make sure that all the packages are working as expected.

bclement-ocp:
We already use dune-site to manage plugin files. With this patch, we now also use the dune-site mechanism to manage plugin loading for the inequalities plugin. This should make alt-ergo work with bytecode-only compilers again (see [1]).

We still use Dynlink directly for additional parsers. Support for parser plugins is deprecated as it is not supported with the Dolmen frontend and will be removed in a future release, so the churn doesn't seem worth it.

Fixes #1047

[1] : https://github.com/ocaml/opam-repository/pull/25294

### DIFF
--- a/docs/sphinx_docs/Usage/index.md
+++ b/docs/sphinx_docs/Usage/index.md
@@ -240,8 +240,10 @@ The Fm-Simplex plugin can be used as follows:
 
         $ alt-ergo --inequalities-plugin fm-simplex [other-options] file.<ext>
 
-The inequalities plugins are [Dune-site plugins] registered in the `(alt-ergo
-plugins)` site. The `fm-simplex` plugin comes built-in with Alt-Ergo.
+The `fm-simplex` inequality plugin comes built-in with Alt-Ergo and no further
+installation is required. External inequality plugins are supported through the
+[Dune-site plugins] mechanism; developers of these plugins must register them
+in the `(alt-ergo plugins)`.
 
 ### Preludes
 
--- a/docs/sphinx_docs/Usage/index.md
+++ b/docs/sphinx_docs/Usage/index.md
@@ -238,16 +238,10 @@ See the [AB-Why3 README] file for the documentation of the AB-Why3 plugin
 
 The Fm-Simplex plugin can be used as follows:
 
-        $ alt-ergo --inequalities-plugin fm-simplex-plugin.cmxs [other-options] file.<ext>
-        $ alt-ergo --inequalities-plugin some-path/fm-simplex-plugin.cmxs [other-options] file.<ext>
+        $ alt-ergo --inequalities-plugin fm-simplex [other-options] file.<ext>
 
-   Alt-Ergo will try to load a local plugin called
-   "fm-simplex-plugin.cmxs". If this fails, Alt-Ergo tries to load it
-   from the default plugins directory (run `alt-ergo --where plugins`
-   to see its absolute path). You can also provide a relative or an
-   absolute path as shown by the second command above. Also, you
-   should replace ".cmxs" by ".cma" if you are working with bytcode
-   binaries.
+The inequalities plugins are [Dune-site plugins] registered in the `(alt-ergo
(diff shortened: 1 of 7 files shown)


docs/sphinx_docs/Usage/index.md
```diff
@@ -339,3 +333,4 @@ A small example of how to use the Alt-Ergo web worker can be build with the comm
 [API documentation]: ../API/index.md
 [AB-Why3 README]: ../Plugins/ab_why3.md
 [Input section]: ../Input_file_formats/index
+[Dune-site plugins]: https://dune.readthedocs.io/en/stable/sites.html#plugins
```
Halbaroth, line 338: Not sure to understand why we need to add this pointer in the documentation. It's completely transparent for the user, right? If I understand well, the user has to install the `opam` package `fm-simplex-plugin` and use the option `--inequalities-plugin fm-simplex` to activate it.

We may mention it in the documentation.

Halbaroth (approved):

docs/sphinx_docs/Usage/index.md
bclement-ocp: Right, I was not very clear — the `fm-simplex-plugin` is bundled with alt-ergo and does not need to be installed separately; however this note is intended for users that wish to build their own inequality plugin: they need to follow the dune-site documentation.

I will reword the doc to clarify these points.

PR #1054, opened 2024-03-13
feat(BV,CP): Use a FIFO queue for constraint propagation

bclement-ocp:
This patch changes the bit-vector propagator to use a FIFO queue to select which propagator to run. Using a FIFO queue gives fairness guarantees on the constraints that get propagated, and does not depend on hash-consing shenanigans as in the current implementation.

The implementation is a simple wrapper around the `Queue` module from the stdlib and a hash table to ensure uniqueness of constraints in the queue.

*Note*: This depends on (and includes) #1044; only the last commit titled "feat(BV, CP): Use a FIFO queue for constraint propagation" is new.

### DIFF
--- a/src/lib/reasoners/bitv_rel.ml
+++ b/src/lib/reasoners/bitv_rel.ml
@@ -398,6 +398,8 @@ let add_eqs =
 type any_constraint =
   | Constraint of Constraint.t Rel_utils.explained
   | Structural of X.r
+  (** Structural constraint associated with [X.r]. See
+      {!Rel_utils.Domains.structural_propagation}. *)
 
 module QC = Uqueue.Make(struct
     type t = any_constraint
--- a/src/lib/reasoners/bitv_rel.ml
+++ b/src/lib/reasoners/bitv_rel.ml
@@ -135,6 +135,10 @@ module Domains = Rel_utils.Domains_make(Domain)
 module Constraint : sig
   include Rel_utils.Constraint
 
+  val equal : t -> t -> bool
+
+  val hash : t -> int
+
   val bvand : X.r -> X.r -> X.r -> t
   (** [bvand x y z] is the constraint [x = y & z] *)
 
@@ -237,6 +241,10 @@ end = struct
   let pp ppf { repr; _ } = pp_repr ppf repr
 
   let compare { tag = t1; _ } { tag = t2; _ } = Stdlib.compare t1 t2
+
+  let equal t1 t2 = Int.equal t1.tag t2.tag
+
+  let hash t1 = Hashtbl.hash t1.tag
 
   let subst rr nrr c =
     hcons @@ subst_repr rr nrr c.repr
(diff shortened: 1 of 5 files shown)


src/lib/reasoners/rel_utils.ml
```diff
@@ -564,31 +572,28 @@ module Constraints_make(Constraint : Constraint) : sig
...
   val fold_args : (X.r -> 'a -> 'a) -> t -> 'a -> 'a
   (** [fold_args f t acc] folds [f] over all the term representatives that are
       arguments of at least one constraint. *)
-
-  val next_pending : t -> Constraint.t explained * t
-  (** [next_pending t] returns a pair [c, t'] where [c] was pending in [t] and
-      [t'] is identical to [t], except that [c] is no longer a pending
-      constraint.
-
-      @raise Not_found if there are no pending constraints. *)
```
Halbaroth, line 591: You mention this function in the documentation of the type `t` of `Constraints_make`. I cannot point out the appropriate line :/

src/lib/reasoners/rel_utils.ml
```diff
@@ -564,31 +572,28 @@ module Constraints_make(Constraint : Constraint) : sig
...
+  val clear_pending : t -> t
+  (** [clear_pending t] returns a copy of [t] except that no constraints are
+      marked as pending. *)
 
-      This function should be used when the domain of [r] is updated, if
-      domains are tracked for leaves only. *)
+  val has_pending : t -> bool
+  (** [has_pending t] returns [true] if there is any constraint marked as
+      pending, in which case [iter_pending] and [clear_pending] are guaranteed
+      to be no-ops. Should only be used for optimization. *)
```
Halbaroth: I believe you meant that `iter_pending` and `clear_pending` are guaranteed to be no-ops if `has_pending` returns `false`?

src/lib/reasoners/bitv_rel.ml
```diff
@@ -387,6 +395,24 @@ let add_eqs =
   fun acc x bl ->
     aux x (Bitlist.width bl) acc bl
 
+type any_constraint =
+  | Constraint of Constraint.t Rel_utils.explained
+  | Structural of X.r
```
Halbaroth, line 400: We may mention the meaning of `Structural` here because the precise definition is in `Rel_utils`.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -395,23 +421,39 @@ let add_eqs =
 
    Iterate until fixpoint is reached. *)
 let propagate =
-  let rec propagate changed bcs dom =
-    match Constraints.next_pending bcs with
-    | { value; explanation = ex }, bcs ->
+  let touch bcs queue r =
+    QC.push queue (Structural r);
+    Constraints.iter_parents (fun c -> QC.push queue (Constraint c)) r bcs
```
Halbaroth, line 428: Is there any reason to motivate this order in the propagation? :)

Halbaroth (changes requested):

src/lib/reasoners/rel_utils.ml
bclement-ocp: Indeed, will fix

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Not really, I don't think there is a particularly significant order between constraints here. I did put the `Structural` one first because it is the strongest for the bit-vector domain, in particular when a leaf is updated (any bit known in the leaf will be known in the semantic values that contain the leaves) and so the most likely to trigger future propagations, so putting it first probably is slightly less likely to cause duplicate propagations later, but I am not sure it actually matters.

bclement-ocp:
Changes applied, ready for round 2.

(Note that this should not be merged until #1044 has been reviewed)

Halbaroth (approved):

Halbaroth:
I'll let you merge this sequence of PR ;)

PR #1055, opened 2024-03-13
feat(BV,CP): Ephemeral API for domain updates

bclement-ocp:
This patch adds a new ephemeral (imperative) interface to the `Domains` API, and uses it for constraint propagation. The conversion between the persistent and ephemeral APIs is provided by a function `edit` that returns an ephemeral domain (implemented using a hash table) from a persistent one, and a `snapshot` function that performs the inverse operation.

Ephemeral domains are only used internally during constraint propagation so that the external API of the `Bitv_rel` module stays persistent.

This leads to more readable code for the constraints, and should be more efficient on large problems as we do not keep looking up the domains in a `Map`.

**Note**: This depends on (and includes) #1044 and #1054; only the last commit titled "feat(BV,CP): Ephemeral API for domain updates" is new.

### DIFF
--- a/src/lib/reasoners/rel_utils.ml
+++ b/src/lib/reasoners/rel_utils.ml
@@ -313,8 +313,7 @@ module type Domains = sig
         @raise Domain.Inconsistent if the intersection is empty. *)
 
     type t
-    (** The type of ephemeral domains, a mutable mapping of semantic values to
-        [domain]s. *)
+    (** Mutable mappings from semantic values to [domain]s. *)
 
     val handle : t -> X.r -> handle
     (** [handle t r] returns the [handle] associated with [r].
@@ -322,8 +321,9 @@ module type Domains = sig
         There is a unique handle associated with each semantic value [r] that is
         created on-the-fly when [handle t r] is called for the first time.
 
-        The domain associated with the handle is initialized with [default r]
-        the first time it is created, and updated with [update]. *)
+        The domain associated with the handle is initialized from the
+        underlying persistent domain the first time it is accessed, and updated
+        with [update]. *)
 
     val structural_propagation : t -> X.r -> unit
     (** Perform structural propagation for the given representative.
--- a/src/lib/reasoners/rel_utils.ml
+++ b/src/lib/reasoners/rel_utils.ml
@@ -17,6 +17,7 @@
 module X = Shostak.Combine
 module MX = Shostak.MXH
 module SX = Shostak.SXH
+module HX = Shostak.HX
 module L = Xliteral
 module LR = Uf.LX
 module SR = Set.Make(
@@ -273,15 +274,6 @@ module type Domains = sig
   val get : X.r -> t -> elt
   (** [get r t] returns the domain currently associated with [r] in [t]. *)
 
-  val update : X.r -> elt -> t -> t
-  (** [update r d t] intersects the domain of [r] in [t] with the domain [d].
-
-      {b Soundness}: The domain [d] must already include the justification
-      that it applies to [r].
-
-      @raise Inconsistent if this causes the domain associated with [r] to
-      become empty. *)
-
   val fold_leaves : (X.r -> elt -> 'a -> 'a) -> t -> 'a -> 'a
   (** [fold f t acc] folds [f] over all the domains in [t] that are associated
       with leaves. *)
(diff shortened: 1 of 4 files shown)


src/lib/reasoners/rel_utils.ml
```diff
@@ -295,33 +287,78 @@ module type Domains = sig
...
+        [domain]s. *)
+
+    val handle : t -> X.r -> handle
+    (** [handle t r] returns the [handle] associated with [r].
+
+        There is a unique handle associated with each semantic value [r] that is
+        created on-the-fly when [handle t r] is called for the first time.
+
+        The domain associated with the handle is initialized with [default r]
+        the first time it is created, and updated with [update]. *)
```
Halbaroth: Where is the definition of the function `default`? It doesn't appear in the implementation of `handle`.

src/lib/reasoners/rel_utils.ml
```diff
@@ -295,33 +287,78 @@ module type Domains = sig
...
+        the [handle]'s representative.
+
+        If this changes the domain associated with the handle, the handle is
+        marked as changed.
+
+        @raise Domain.Inconsistent if the intersection is empty. *)
+
+    type t
+    (** The type of ephemeral domains, a mutable mapping of semantic values to
+        [domain]s. *)
```
Halbaroth: In my opinion, `The type of xs` means the values of this type are the `xs`, so here I understand that `type t` is the type whose the inhabitants are single ephemeral domains. The second part of the comment is clearer by far. So  I suggest:
``` 
The type of mutable mappings of semantic values to [domain]s. 
```

src/lib/reasoners/rel_utils.ml
```diff
@@ -480,27 +518,105 @@ struct
...
+        HX.add t.handles r handle;
+        handle
+
+    let structural_propagation t r =
+      (* Structural propagation is always correct and does not require
+         explanations because it follows the structure of the semantic value
+         itself. *)
+      let get r = !!(handle t r) in
+      let update r d = update ~ex:Explanation.empty (handle t r) d in
+      if X.is_a_leaf r then
```
Halbaroth, line 581: I know the function `X.is_a_leaf` isn't a part of this PR but I would like to discuss a bit about it. I'm not a fond of its name. 

Commonly the terminology `leaf` denotes nodes of trees without children. The name `X.is_a_leaf` sticks this naming convention (at least for non-ac symbols).

The papers about `CC(X)` defined another notion of leaves: a leaf of a semantic value `r` (relatively to a subtheory `T`) is a maximal subterm of `r` that isn't interpreted in `T`. The call `X.leaves r` returns the list of the leaves in this sense (for the theory of the top symbol of `r`).

The elements of `X.leaves` doesn't satisfy the predicate `X.is_a_leaf` in general.

We should at least mention this difference in the documentation of `X.is_a_leaf` or rename some functions. I'm not sure what is the best solution because the papers about Alt-Ergo used the second convention and keeping an implementation close to papers is a good thing.

Halbaroth: We may rename `X.leaves` to `X.aliens` for instance.

Halbaroth (changes requested):

src/lib/reasoners/rel_utils.ml
bclement-ocp: > The papers about CC(X) defined another notion of leaves: a leaf of a semantic value r (relatively to a subtheory T) is a maximal subterm of r that isn't interpreted in T. The call X.leaves r returns the list of the leaves in this sense (for the theory of the top symbol of r).
>
> The elements of X.leaves doesn't satisfy the predicate X.is_a_leaf in general.

I don't think that is true? `X.leaves` can only return uninterpreted terms and AC symbols, which is also what is tested by `is_a_leaf`. I think that this corresponds to the CC(X) definition (or rather the AC(X) definition which allows AC symbols to be leaves).

If `r` is a semantic value whose top interpreted symbol is a symbol in `T`, then the maximal subterms of `r` that is not interpreted in `T` is necessarily uninterpreted; if it was interpreted in another theory (except AC) we could not write a solver.

Halbaroth: You are right. I had a misconception about leaves in `CC(X)`. I have been fooled by this sentence in the paper about `CC(X)`:
``` 
Its role is to return the set
of maximal uninterpreted values a given semantic value consists of. Its
behaviour is left undefined, but is constrained by axioms given below.
```
Here `uninterpreted` means uninterpreted for the supertheory `X` and indeed, after checking the implementation of `Arith.leaves`, we recursively call `X.leaves` on the maximal uninterpreted values for Arith in `Arith.leaves`.

bclement-ocp:
Changes applied; ready for round 2.

Halbaroth (approved):

PR #1056, opened 2024-03-13
feat(BV,CP): Explicit constraint simplification

bclement-ocp:
This patch exposes constraint simplification as an explicit primitive, instead of performing simplification implicitly during substitutions.

This is preparatory work to represent constraints in a more generic way, allowing to limit boilerplate when adding new constraints (e.g. for arithmetic operations on bit-vectors), which is easier to do if substitution and simplification are separate operations.

Note that constraint simplification (both the existing implementation and the new one in this patch) can currently only occur based on the underlying semantic values; the domains are not available at the time simplification is performed. Allowing to perform simplification based on domains is left to a future patch.

**Note**: This depends on (and includes) https://github.com/OCamlPro/alt-ergo/pull/1044, https://github.com/OCamlPro/alt-ergo/pull/1054, and #1055; only the last commit titled "feat(BV,CP): Explicit constraint simplification" is new.

### DIFF
--- a/src/lib/reasoners/rel_utils.ml
+++ b/src/lib/reasoners/rel_utils.ml
@@ -619,6 +619,17 @@ struct
       ) t.Ephemeral.dirty_cache t.persistent
 end
 
+(** The ['c acts] type is used to register new facts and constraints in
+    [Constraint.simplify]. *)
+type 'c acts =
+  { acts_add_lit_view : X.r L.view -> unit
+  (** Assert a semantic literal. *)
+  ; acts_add_eq : X.r -> X.r -> unit
+  (** Assert equality between two semantic values. *)
+  ; acts_add_constraint : 'c -> unit
+  (** Assert a new constraint. *)
+  }
+
 module type Constraint = sig
   type t
   (** The type of constraints.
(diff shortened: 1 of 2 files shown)


src/lib/reasoners/rel_utils.ml
```diff
@@ -815,4 +849,44 @@ end = struct
     MX.fold (fun r _ acc ->
         f r acc
       ) c.args_map acc
+
+  let simplify_pending =
+    (* Recursion needed because adding new constraints changes the pending set
+       and they also need to be simplified *)
+    let rec simplify_aux eqs t to_simplify =
+      let eqs = ref eqs in
```
Halbaroth, line 858: We accumulate all the discovered equations by the simplification pass in this list but this list could contain duplications. We may construct a set and extract the list of its elements later?

Halbaroth (approved):

src/lib/reasoners/rel_utils.ml
bclement-ocp: I believe equalities found here ultimately end up in [Ccx.make_unique](https://github.com/OCamlPro/alt-ergo/blob/fb0293a28be8e29e6f9df6aba8ac23933857c62b/src/lib/reasoners/ccx.ml#L448) which does deduplicate. I think usually we won't find too much duplicates here (at most the number of constraints anyways) so that it shouldn't be an issue; I'd rather update the `Relation` API to use a set everywhere rather than do it locally. I will open an issue for this.

Halbaroth (approved):

PR #1058, opened 2024-03-13
 feat(BV): Interval domains for bit-vectors 

bclement-ocp:
This patch adds interval domains to the Bitv_rel module, as well as
interreductions between the bitlist and interval domains following:

  Sharpening Constraint Programming approaches for Bit-Vector Theory.
  Zakaria Chihani, Bruno Marre, François Bobot, Sébastien Bardin.
  CPAIOR 2017. International Conference on AI and OR Techniques in
    Constraint Programming for Combinatorial Optimization Problems, Jun
    2017, Padova, Italy.

More precisely:

 - The `Intervals` module is extended to support the `extract` operation, which is used to propagate between bit-vector compositions and their components;

 - The interreductions are implemented using the new `Bitlist.increase_lower_bound`, `Bitlist.decrease_upper_bound`, and the new `shared_msb` helper in `Bitv_rel`;

 - Propagations are performed by alternating propagations until fixpoint in each domain, followed by interreductions and propagations until fixpoint in the other domain, until reaching a fixpoint for the whole procedure. It is not clear that this is the best strategy; the goal is to try and limit interreductions since they are relatively expensive but we should revisit this once more operations are supported.

For now, only the `bvule`, `bvult`, `bvugt` and `bvuge` primitives are supported as built-in bit-vector operations; other operations such as `bvadd` are still encoded using `bv2nat`. These operations will be migrated to bit-vector primitives in a follow-up PR.

Finally, there are some tests for the tricky bits (`Intervals.extract` and the interreduction primitives) using QCheck.


bclement-ocp:
Note: this is +107-515 or a net regression of about 400 problems on our QF_BV subset. This is expected: this PR uses bit-vector domains for `bvule` / `bvult` / etc. but does not change the use of `bv2nat` / `int2bv` round-trip for other arithmetic operators. This means that we are prevented from doing many interesting propagations.

A follow-up PR (should be ready sometime next week) fixes that by implementing arithmetic propagators for the BV interval domains: it is currently +1156-209 (or an improvement of almost 1000 problems) on that same subset compared to `next`, at least in part due to interreduction between arithmetic and bitwise constraints.

src/lib/reasoners/intervals.ml
```diff
@@ -1200,6 +1200,150 @@ let pick ~is_max { ints; is_int; _ } =
...
+(*****************)
+(*   Bit-vector  *)
+(*****************)
+
+(** Apply function [f] to the interval.
+
+    [f] *MUST* be monotone (either increasing or decreasing depending on the
+    value of the [increasing] labelled argument)
+
+    The [minfty] and [pinfty] optional arguments provide values for [map_bigint]
```
Halbaroth: What's `map_bigint` here? I guess you meant `f` and `map_bigint` is a deprecated reference to `map_increasing_bigint` ?

src/lib/reasoners/intervals.ml
```diff
@@ -1200,6 +1200,150 @@ let pick ~is_max { ints; is_int; _ } =
...
+          | Large (ub, ex) -> Large (f ub, ex)
+          | Pinfty -> get_infty pinfty
+          | Strict _ | Minfty -> assert false
+        in
+        let lb, ub = if increasing then f_lb, f_ub else f_ub, f_lb in
+        (lb, ub) :: l
+      ) [] uints.ints
+  in
+  let res = union_intervals { uints with ints } in
+  assert (res.ints != []);
```
Halbaroth: I don't understand why we need to normalize the result before returning it. If `f` is an increasing function, it sends an union of disjoint intervals to an union of disjoint intervals. If `f` is decreasing, we only need to reverse the list `ints`. I'm wrong?

bclement-ocp:
After reading and thinking a lot (and starting a small Coq developement) about the explanations for intervals:

 - I am quite confident that the `map_monotone` function introduced in this PR, and the `map2_monotone` function introduced in #1085, are correct
 - I am reasonably confident that the other interval functions introduced in this PR and following PRs are correct when called on a single interval, with the exception maybe of `extract`
 - I am *not* convinced that the other interval functions are correct when called on an union of intervals.

That is not to say that I don't think they are correct (in particular, I am confident that all of these functions return the right values for the bounds), but that they may forget necessary explanations in some cases, causing unsoundness. I think the cases where we forget explanations are quite rare (or may be made impossible currently) so I will rewrite these functions to try to be less clever and add more explanations. I will also write a documentation for the intervals module, now that i understand the invariants better.

This does not impact the rest of the PR or the rest of the PRs in the stack.

tests/bitv/bitlist-interval001.smt2
```diff
@@ -0,0 +1,5 @@
+(set-logic ALL)
```
Halbaroth, line 1: Nit: the logic `QF_BV` is sufficient here?

src/lib/reasoners/bitv_rel.ml
```diff
@@ -77,6 +77,94 @@ let is_bv_ty = function
...
+    in
+    assert (j = -1);
+    acc
+
+  let map_signed f { Bitv.value; negated } sz =
+    if negated then lognot sz (f value) else f value
+
+  let map_leaves f r =
+    List.fold_left (fun ival { Bitv.bv; sz } ->
+        let ival = Intervals.scale (Q.of_bigint @@ Z.(~$1 lsl sz)) ival in
```
Halbaroth: I'm not sure to understand this line. My understanding is as follows:
`map_leaves` builds an interval domain for the bitvector `r` by building interval domains of its leaves. For instance, says that `r` is the bitvector `[|10101|] + x + (extract i j k y)`. We build an interval domain for `[|10101|]`, `x` and `(extract i j k y)` and build the interval domain of `r` from them.

The function `Intervals.add i_1 i_2` computes the sum set `i_1 + i_2`, that is the direct image of `(i_1, i_2)` by `(x, y) |--> x + y`. 

The function `Intervals.scale s i` computes the direct image of the union of intervals `i` by the homothety  `x |--> s * x`.

We starts the construction with the interval singleton `0`, which makes sense as this set is the neutral element for the operator `Intervals.add`  on the union of intervals.

We have to make room for the interval domain of `{ Bitv.bv; sz }` in `ival` and I guess it's the purpose of the call to `Intervals.scale` here. It works if we fold on `r` from the most significant bit to the least one. Is it the case? I cannot remember if the bitvector are represented from the most significant bit to the least one in `bitv`?

src/lib/reasoners/bitv_rel.ml
```diff
@@ -149,11 +237,17 @@ module Constraint : sig
...
+
+  val bvugt : X.r -> X.r -> t
+
   val propagate : ex:Ex.t -> t -> Domains.Ephemeral.t -> unit
   (** [propagate ~ex t dom] propagates the constraint [t] in domain [dom].
 
       The explanation [ex] justifies that the constraint [t] applies, and must
       be added to any domain that gets updated during propagation. *)
+
+  val propagate_interval : ex:Ex.t -> t -> Interval_domains.Ephemeral.t -> unit
```
Halbaroth, line 249: I know it's the same documentation of `propagate` but we should copy it or add a pointer to it.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -424,74 +642,284 @@ module Any_constraint = struct
...
+
+    (* First, we propagate the pending constraints to both domains. Changes in
+       the bitlist domain are used to shrink the interval domains. *)
+    touch_pending queue;
+    propagate_bitlist queue touched bcs bdom;
+    assert (QC.is_empty queue);
+
+    touch_pending queue;
+    HX.iter (fun r () ->
+        HX.replace bitlist_changed r ();
```
Halbaroth, line 852: Why not using a set of semantic values instead of a map?

src/lib/reasoners/bitv_rel.ml
```diff
@@ -77,6 +77,94 @@ let is_bv_ty = function
...
+          (* sz = j - i + 1 => i = j - sz + 1 *)
+          let int = Intervals.extract int (j - sz + 1) j in
+
+          let acc = match bv with
+            | Bitv.Cte z ->
+              (* Nothing to update, but still check for consistency *)
+              ignore @@
+              intersect ~ex:Ex.empty int
+                (Intervals.point (Q.of_bigint z) Tint Ex.empty);
+              acc
```
Halbaroth: I know it's not a part of this PR but I wonder if we could implement a better iterator:
1. When I see a fold iterator raising an exception, I expect that the exception is raised by the iterated function, not the iterator (or the iterator may raise a very specific exception as `List.fold_left2` for instance).

2. We only call `Domain.fold_leaves` to update leaves' domain of a semantic value `r` and to detect inconsistency during the structural propagation:
```ocaml
Domain.fold_leaves (fun r d () -> update r d) r (get r) ()
``` 
As we don't store bitlists for constant bitvectors, `fold_leaves` performs the update step (which consists in checking the consistency) when it sees a constant bitvector in `r`.

It means that `fold_leaves` isn't a generic iterator and we have to read the implementation of both `fold_leaves` and `f` to understand what's happened.

src/lib/reasoners/intervals.ml
Halbaroth: If it's to be sure the returned domain is normalized, we may replace this by a heavy assert.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -424,74 +642,284 @@ module Any_constraint = struct
...
+  if nshared > 0 then
+    let ex = Ex.union inf_ex sup_ex in
+    let shared_bl =
+      Bitlist.exact nshared (Z.extract inf (sz - nshared) nshared) ex
+    in
+    update ~ex bv @@
+    Bitlist.concat shared_bl (Bitlist.unknown (sz - nshared) Ex.empty)
+
+(* These are helpers for readability because [Intervals.mk_closed] makes my
+   brain melt *)
```
Halbaroth: :D

bclement-ocp:
> I am quite confident that the map_monotone function introduced in this PR, and the map2_monotone function introduced inhttps://github.com/OCamlPro/alt-ergo/pull/1085, are correct

Turns out they are not (: I thought I had replicated the behavior of the `add` function properly, but turns out I missed a crucial detail.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -424,74 +642,284 @@ module Any_constraint = struct
...
+    Intervals.new_borne_inf ex lb ~is_le:(is_large ineq)
+      (Intervals.undefined Tint)
+  | Bounded (lb, lineq), Bounded (ub, uineq) ->
+    Intervals.mk_closed
+      lb ub
+      (is_large lineq) (is_large uineq)
+      ex ex Tint
+
+(* Algorithm 1 from "Sharpening Constraint Programming approaches for
+   Bit-Vector Theory". *)
```
Halbaroth: We should add an example as it does in the article. I know we can read the article to find an example but it's more convenient to inline it here.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -424,74 +642,284 @@ module Any_constraint = struct
...
+(* If m and M are the minimal and maximal values of an union of intervals, the
+   longest sequence of most significant bits shared between m and M can be fixed
+   in the bit-vector domain; see "Is to BVs" in section 4.1 of
+
+   Sharpening Constraint Programming approaches for Bit-Vector Theory.
+   Zakaria Chihani, Bruno Marre, François Bobot, Sébastien Bardin.
+   CPAIOR 2017. International Conference on AI and OR Techniques in
+   Constraint Programming for Combinatorial Optimization Problems, Jun
+   2017, Padova, Italy.
+   https://cea.hal.science/cea-01795779/document *)
```
Halbaroth: Same remark, we should add an example here.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -424,74 +642,284 @@ module Any_constraint = struct
...
+  update ~ex int @@
+  List.fold_left (fun acc ((lb, _ex_lb), (ub, _ex_ub)) ->
+      (* Note: the bounds adjustment is expressed by removing specific
+         intervals that are forbidden by the bitlist domain. This is always
+         correct independently of the bounds' explanations: we only use the
+         bounds to guide us on *which* intervals to remove. Hence, we just
+         ignore the bounds' explanations. *)
+      let acc =
+        match lb with
+        | None -> assert false
```
Halbaroth: I guess it cannot happen because our interval domains are always subsets of the interval `[0, n)` for some integer `n`.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -424,74 +642,284 @@ module Any_constraint = struct
...
+            (* no change *)
+            assert (Z.equal new_lb_z lb_z);
+            acc
+          | exception Not_found ->
+            (* No value larger than lb matches the bit-pattern *)
+            remove acc @@ interval ~ex (large lb) infinite
+      in
+      let acc =
+        match ub with
+        | None -> assert false
```
Halbaroth: Same remark.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -424,74 +642,284 @@ module Any_constraint = struct
...
+let constrain_interval_from_bitlist int bv =
+  let open Interval_domains.Ephemeral in
+  let ex = Bitlist.explanation bv in
+  (* [exclude i1 i2] being [i2 \ i1] also makes my brain hurt *)
+  let remove i2 i1 = Intervals.exclude i1 i2 in
+  update ~ex int @@
+  List.fold_left (fun acc ((lb, _ex_lb), (ub, _ex_ub)) ->
+      (* Note: the bounds adjustment is expressed by removing specific
+         intervals that are forbidden by the bitlist domain. This is always
+         correct independently of the bounds' explanations: we only use the
```
Halbaroth: This comment is a bit confusing. I understood that we drop these explanations in the new interval domain, which is clearly incorrect.

Halbaroth:
I started a review of the code? Should I discard it?

bclement-ocp:
Everything except for changes to the `Intervals.ml` module is still valid and can be reviewed. There are subtleties in the interval module (I am starting to believe that there are pre-existing hard-to-trigger soundness bugs lurking, although the module is very unclear about its invariants so maybe it is actually impossible), so I would hold off reviewing changes to this file specifically. But the rest should not be affected — again, it is just implementation details of the way the Intervals module stores explanations.

Halbaroth (changes requested):

Halbaroth:
> Everything except for changes to the `Intervals.ml` module is still valid and can be reviewed. There are subtleties in the interval module (I am starting to believe that there are pre-existing hard-to-trigger soundness bugs lurking, although the module is very unclear about its invariants so maybe it is actually impossible), so I would hold off reviewing changes to this file specifically. But the rest should not be affected — again, it is just implementation details of the way the Intervals module stores explanations.

Honestly, when I read this module months ago, my conclusion was we should rewrite it completely. It's not a priority but I plan to work on it later.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Yeah, `fold_leaves` is poorly named, I agree. It is indeed not supposed to be a generic iterator; it should have a name that references (structural) propagation. I propose to change that in another PR though.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: The `map_leaves` function operates on the structural representation of a bit-vector, so I guess `+` here means `concat`?

If in terms of concatenation we have `[| 10101 |] ++ x ++ (extract i j y)`, in terms of addition we have `[| 10101|] * 2^(sz x + j - i + 1) + x * 2^(j - i + 1) + (extract i j y)`. The scale is to bring the value in the appropriate range.

Bit-vectors are indeed stored most significant bit first (but indexed by least-significant bit first, which is always easy to mess up).

src/lib/reasoners/bitv_rel.ml
bclement-ocp: What is happening is:

 - We have a bitlist domain that has an explanation [ex].
 - We build some intervals that the bitlist domain proves are impossible (for instance, if the bitlist domain is `00??`, then the interval `[4, 15]` is impossible because no value in the interval `[4, 15]` starts with `00`).
 - We are removing those intervals from the existing interval domain using `exclude`

The intervals that we build from the bitlist domain only need the explanation from the bitlist domain to be valid, so we don't need to incorporate the explanations from the old bounds into it. The `exclude` and `intersect` functions from the intervals module takes care of appropriately (hopefully… but if not it is a bug in `exclude`, not in this code) updating the bounds to take into accound both explanations.

Is that more clear?

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Yep. I guess I could treat this case as `Some (0, 0)`, but I'd rather be strict. I will add a comment explaining.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Because access to a set is `O(log n)` while a hash map is `O(1)`. There is no hash set in OCaml.

src/lib/reasoners/intervals.ml
bclement-ocp: Yeah, that should be `f`, thanks (the name of the argument to this function used to be `map_bigint` instead of `f`).

src/lib/reasoners/intervals.ml
bclement-ocp: We need to normalize the result because `f` is not guaranteed to be *strictly* increasing/decreasing.

So for instance maybe we have `[10, 12] U [14, 20]` but `f(x) = 10` for any `x ≥ 10` so after the call we get `[10, 10] U [10, 10]` which is not properly normalized.

We need to support this because `shr` is not strictly increasing (all sufficiently small values map to `0`).

(interestingly enough, I believe this is the case where `union_intervals` currently has a subtle soundness bug (: )

tests/bitv/bitlist-interval001.smt2
bclement-ocp: I tend to default to using `(set-logic ALL)` now because I copy-paste tests to test things during development or debugging and Dolmen complains if you use `QF_BV` but later use quantifiers or integers.

(Also Why3 always uses `(set-logic ALL)` so we might as well test with this)

bclement-ocp:
> Honestly, when I read this module months ago, my conclusion was we should rewrite it completely. It's not a priority but I plan to work on it later.

I am about 99% sure there currently are soundness bugs that can be triggered by properly crafted formulas so I am actually looking into it now. Hopefully I will not need a full rewrite to fix them :/

src/lib/reasoners/intervals.ml
Halbaroth: In English `increasing`means `strictement croissant`. So we should write [non-decreasing](https://fr.wiktionary.org/wiki/non-decreasing) :)

src/lib/reasoners/bitv_rel.ml
Halbaroth: > Bit-vectors are indeed stored most significant bit first (but indexed by least-significant bit first, which is always easy to mess up).

I think we should mention this detail of implementation in the documentation of `Bitv`.

src/lib/reasoners/bitv_rel.ml
Halbaroth: oh sorry, I was thinking about hash set but there is no such thing!

src/lib/reasoners/intervals.ml
bclement-ocp: > In English increasingmeans strictement croissant. So we should write [non-decreasing](https://fr.wiktionary.org/wiki/non-decreasing) :)

Unfortunately the usage varies and `increasing` is used for both. Monotone is the non-strict version and this is what the function is called. I will clarify in the comment

src/lib/reasoners/bitv_rel.ml
bclement-ocp: It is mentioned in the documentation of `extract` already but we could move it to the documentation of `abstract` I suppose (not in this PR tho).

src/lib/reasoners/intervals.ml
Halbaroth: `strictly increasing` is acceptable ;)

bclement-ocp:
I have adapted the code on top of #1108  (unfortunately I messed up the rebase so the history is a bit weird, apologies for that!).

bclement-ocp:
> I am about 99% sure there currently are soundness bugs that can be triggered by properly crafted formulas so I am actually looking into it now. Hopefully I will not need a full rewrite to fix them :/

FWIW: We were in the 1% of chances, but the code in this PR for `Intervals.extract` was still wrong. Fairly confident the new one is right (and if wrong that we have a chance to catch it by reading the code).

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Done

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Done

src/lib/reasoners/bitv_rel.ml
bclement-ocp: This code has changed and now uses the new `finite_lower_bound` and `finite_upper_bound` helpers w/o assertions.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Same as above

src/lib/reasoners/bitv_rel.ml
bclement-ocp: This comment is no longer needed now that the iteration on the union-of-interval (correctly) do not expose the bounds after #1108

Halbaroth:
As soon as you rebase it, I will do a last pass on this PR but I think it's already good.

bclement-ocp:
Rebased the whole stack.

src/lib/reasoners/rel_utils.ml
```diff
@@ -267,7 +267,7 @@ module type Domain = sig
 
       @raise Inconsistent if [r] cannot possibly be in the domain of [t]. *)
 
-  val map_leaves : (X.r -> 'a -> t) -> X.r -> 'a -> t
+  val map_leaves : (X.r -> t) -> X.r -> t
   (** [map_leaves f r acc] is the "inverse" of [fold_leaves] in the sense that
       it rebuilds a domain for [r] by using [f] to access the domain for each
       of [r]'s leaves. *)
```
Halbaroth: This documentation is outdated.

src/lib/reasoners/intervals.mli
```diff
@@ -30,6 +30,8 @@
 
 open Intervals_intf
 
+val map_bound : ('a -> 'b) -> 'a bound -> 'b bound
```
Halbaroth, line 30: Please add a documentation :)

src/lib/reasoners/intervals_intf.ml
```diff
@@ -653,6 +653,10 @@ module type Ring = sig
   (** [add u1 u2] evaluates to {m \{ x + y \mid x \in S_1, y \in S_2 \}} when
       [u1] evaluates to {m S_1} and [u2] evaluates to {m S_2}. *)
 
+  val scale : value -> t -> t
+  (** [scale v u] evaluates to {m \{ v \times x \mid x \in S \}} when [u]
+      evaluates to {m S}. *)
```
Halbaroth: This function raises an assertion if the scalar is zero. We should mention it in its documentation.

src/lib/reasoners/intervals.mli
```diff
@@ -43,11 +45,36 @@ module Real : AlgebraicField
...
+      range [[0, n)] ([0] inclusive, [n] exclusive). *)
+
+  val lognot : t -> t
+  (** Bitwise logical negation. [lognot u] always returns [-u - 1]. *)
+
+  val extract : t -> ofs:int -> len:int -> t
+  (** [extract s i j] returns the bits of [s] from position [i] to [j],
+      inclusive.
+
+      Represents the function [fun x -> floor(x / 2^i) % 2^(j - i + 1)].
```
Halbaroth, line 69: As this function is clear not semialgebraic, we cannot expect that the image of union of intervals is again an union of intervals. We should mention that this map compute an estimation of this image by an union of intervals.

Do you know if we compute the best approximation, that is the smallest union that contains the image?

src/lib/reasoners/bitv_rel.ml
```diff
@@ -428,55 +643,256 @@ module Any_constraint = struct
...
+  in
+  try
+    while true do
+      Interval_domains.Ephemeral.iter_changed touch dom;
+      Interval_domains.Ephemeral.clear_changed dom;
+      Any_constraint.propagate
+        Constraint.propagate_interval
+        Interval_domains.Ephemeral.structural_propagation
+        (QC.pop queue) dom
+    done
```
Halbaroth: These function are almost the same. We can factorize them as follows:
```ocaml 
let propagate (type a) (module Eph : Rel_utils.Ephemeral with type t = a)
  propagator queue touched bcs (dom : a) =
  let touch_c c = QC.push queue (Constraint c) in
  let touch r =
    HX.replace touched r ();
    QC.push queue (Structural r);
    Constraints.iter_parents touch_c r bcs
  in
  try
    while true do
      Eph.iter_changed touch dom;
      Eph.clear_changed dom;
      Any_constraint.propagate
        propagator Eph.structural_propagation
        (QC.pop queue) dom
    done
  with QC.Empty -> ()

let propagate_bitlist =
  propagate (module Domains.Ephemeral) Constraint.propagate

let propagate_intervals =
  propagate (module Interval_domains.Ephemeral) Constraint.propagate_interval
``` 
(Assuming that we have extract the signature of the `Ephemeral` in the module type `Rel_utils.Domains`).

src/lib/reasoners/bitv_rel.ml
```diff
@@ -149,11 +236,17 @@ module Constraint : sig
...
 
+  val bvule : X.r -> X.r -> t
+
+  val bvugt : X.r -> X.r -> t
+
   val propagate : ex:Ex.t -> t -> Domains.Ephemeral.t -> unit
   (** [propagate ~ex t dom] propagates the constraint [t] in domain [dom].
 
       The explanation [ex] justifies that the constraint [t] applies, and must
       be added to any domain that gets updated during propagation. *)
```
Halbaroth: I think we should rename this function into `propagate_bitlist`

src/lib/reasoners/bitv_rel.ml
```diff
@@ -75,6 +75,93 @@ let is_bv_ty = function
 
 let is_bv_r r = is_bv_ty @@ X.type_info r
 
+module Interval_domain : Rel_utils.Domain with type t = Intervals.Int.t = struct
```
Halbaroth, line 78: For sake of uniformity, we should rename `Domains` into `Bitlist_domains`.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -428,55 +643,256 @@ module Any_constraint = struct
...
+    let bitlist_changed = HX.create 17 in
+    let touched = HX.create 17 in
+    let bdom = Domains.edit bdom in
+    let idom = Interval_domains.edit idom in
+
+    (* First, we propagate the pending constraints to both domains. Changes in
+       the bitlist domain are used to shrink the interval domains. *)
+    touch_pending queue;
+    propagate_bitlist queue touched bcs bdom;
+    assert (QC.is_empty queue);
```
Halbaroth, line 848: It would be cleaner to move the creation of the queue in `propagate_bitlist`. The assertion can be remove after this modification. So the `propagate` function looks like this:
```ocaml 
let propagate (type a) (module Eph : Rel_utils.Ephemeral with type t = a) propagator =
  let queue = QC.create 17 in
  let touch_c c = QC.push queue (Constraint c)
  in fun touched bcs (dom : a) ->
  QC.clear queue;
  Constraints.iter_pending touch_c bcs;
  let touch r =
    HX.replace touched r ();
    QC.push queue (Structural r);
    Constraints.iter_parents touch_c r bcs
  in
  try
    while true do
      Eph.iter_changed touch dom;
      Eph.clear_changed dom;
      Any_constraint.propagate
        propagator Eph.structural_propagation
        (QC.pop queue) dom
    done
  with QC.Empty -> ()
```
I put the queue in the closure to avoid doing a lot of allocations in the hot loop.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -428,55 +643,256 @@ module Any_constraint = struct
...
+          Domains.Ephemeral.(!!(handle bdom r))
+      ) touched;
+    HX.clear touched;
+    propagate_intervals queue touched bcs idom;
+    assert (QC.is_empty queue);
+
+    (* Now the interval domain is stable, but the new intervals may have an
+       impact on the bitlist domains, so we must shrink them again when
+       applicable. We repeat until a fixpoint is reached. *)
     let bcs = Constraints.clear_pending bcs in
```
Halbaroth, line 864: As we flush the pending constraints here, we can use the new `propagate` function in this loop again.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -428,55 +643,256 @@ module Any_constraint = struct
...
+    touch_pending queue;
+    HX.iter (fun r () ->
+        HX.replace bitlist_changed r ();
+        constrain_interval_from_bitlist
+          Interval_domains.Ephemeral.(handle idom r)
+          Domains.Ephemeral.(!!(handle bdom r))
+      ) touched;
+    HX.clear touched;
+    propagate_intervals queue touched bcs idom;
+    assert (QC.is_empty queue);
```
Halbaroth: Same here.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -428,55 +643,256 @@ module Any_constraint = struct
...
+
+   Sharpening Constraint Programming approaches for Bit-Vector Theory.
+   Zakaria Chihani, Bruno Marre, François Bobot, Sébastien Bardin.
+   CPAIOR 2017. International Conference on AI and OR Techniques in
+   Constraint Programming for Combinatorial Optimization Problems, Jun
+   2017, Padova, Italy.
+   https://cea.hal.science/cea-01795779/document
+
+   This function is a wrapper calling [Bitlist.increase_lower_bound] and
+   [Bitlist.decrease_upper_bound] on all the constituent interavals of an union;
```
Halbaroth:
```suggestion
   [Bitlist.decrease_upper_bound] on all the constituent intervals of an union;
```

Halbaroth (changes requested):

src/lib/reasoners/bitv_rel.ml
```diff
@@ -428,55 +643,256 @@ module Any_constraint = struct
...
-      QC.push queue (Structural r);
-      Constraints.iter_parents touch_c r bcs
+    while HX.length touched > 0 do
+      HX.iter (fun r () ->
+          constrain_bitlist_from_interval
+            Domains.Ephemeral.(handle bdom r)
+            Interval_domains.Ephemeral.(!!(handle idom r))
+        ) touched;
+      HX.clear touched;
+      propagate_bitlist queue touched bcs bdom;
```
Halbaroth, line 872: I don't understand why the queue isn't necessary empty here? We have flushed `bcs` before this loop and `bcs` is immutable and the queue is empty before the loop too.

bclement-ocp: The queue is empty here. If `constrain_bitlist_from_interval` changes the domain of some bitlists, it will be tracked in the `changed` field inside `bdom`, which is used by `propagate_bitlist` to refill the queue.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: > I put the queue in the closure to avoid doing a lot of allocations in the hot loop.

This is the reason I avoided doing something like this — there is enough imperative things and mutable state going on in that part of the code that I really don't want to hide side effects in partial applications. In a situation like this (lots of mutable structures) I think the more repetitive/less elegant code is better because things are laid out more clearly in a sequential way.

Ultimately, the goal is to find the right abstraction and move this code in a shared module that all relations can communicate through. As long as the general domains architecture is still in a bit of flux, I'd really prefer to keep the constraint propagation part as pedestrian as possible.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: I simplify this in the patch that I mentioned at the dev meeting (deduplicating domains); I'd prefer merging the these functions in that patch (also to avoid painful rebase) if that's OK.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Agreed — actually did this in a later patch in the series.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: I wanted this patch to avoid touching the bitlist part as much as possible. Renamed it.

src/lib/reasoners/intervals.mli
bclement-ocp: I am not sure I understand your point about semialgebraicity. The image of an (integer) interval by this function is an union of (integer) intervals (in fact either a single interval or an union of two disjoint intervals), so the image of an union of intervals is also an union of intervals.

src/lib/reasoners/rel_utils.ml
bclement-ocp: Done (took me a minute to realize you meant the `acc` argument no longer exists).

src/lib/reasoners/intervals_intf.ml
bclement-ocp: Done.

src/lib/reasoners/intervals.mli
bclement-ocp: Done.

src/lib/reasoners/bitv_rel.ml

bclement-ocp:
@Halbaroth as discussed prior to my vacation this is ready for a second pass.

Halbaroth (approved):
Let's merge this one :)

PR #1062, opened 2024-03-19
Update `Lib_usage.ml`

Halbaroth:
We have changed a bit the API of the AE library but the `Lib_usage.ml` isn't up to date. 

I don't know why I prevented `Dune` from compiling this file, so I restore the compilation and the test of this file.

### DIFF
--- a/dune
+++ b/dune
@@ -1,1 +1,0 @@
-(data_only_dirs examples)
--- a/examples/dune
+++ b/examples/dune
@@ -6,6 +6,7 @@
 
 (rule
  (alias runtest)
+ (package alt-ergo-lib)
  (action
   (ignore-stdout
    (ignore-stderr
--- a/examples/lib_usage.ml
+++ b/examples/lib_usage.ml
@@ -84,24 +84,18 @@ let typed, _env = Typechecker.type_file parsed
 
 let pbs = Typechecker.split_goals_and_cnf typed
 
-module SAT = Fun_sat.Make(Theory.Main_Default)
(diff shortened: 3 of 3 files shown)


Halbaroth:
Actually, @bclement-ocp  suggested to remove the compilation of `examples`: https://github.com/OCamlPro/alt-ergo/pull/796#issuecomment-1708518619

Do you have any reason? I can move the `Lib_usage.ml` file.

bclement-ocp:
I admit I don't remember the details but I believe it was actually a `dune` configuration problem and the examples were introducing bogus dependencies on the OPAM CI; it's not a problem to build the examples in general.

I am not sure why I suggested to not build the examples; I think the issue is lack of a `(package)` stanza (probably `(package alt-ergo-lib)`).

Halbaroth:
```dune
(executable
 (name Lib_usage)
 (public_name lib-usage)
 (package alt-ergo-lib)
 (libraries AltErgoLib)
 (modules Lib_usage)
)
``` 
?

bclement-ocp:
I am not sure; I think we need to make sure that all of the following are OK (again my memory is fuzzy but if I recall correctly the examples were run for the parsers):

```console
$ dune clean && dune build -p alt-ergo-lib @runtest
$ dune clean && dune build -p alt-ergo-parsers @runtest
$ dune clean && dune build -p alt-ergo @runtest
```

If all of these pass and the `Lib_usage` is run only in one of them (probably not `parsers`) then we should be fine.

bclement-ocp (approved):
LGTM provided that the aforementioned CLI commands succeed :+1:

PR #1064, opened 2024-03-20
fix(incremental): Reset to `Assert` mode after `push`/`pop` command

### LINKED ISSUE #1024: Mode is not properly updated on push/pop

bclement-ocp:
Fixes #1024

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -415,21 +415,23 @@ let main () =
           "The output format %s is not supported by the Dolmen frontend."
           s
   in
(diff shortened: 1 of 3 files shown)


Halbaroth (approved):
Thanks :)

PR #1045, opened 2024-02-22
feat: Statically linked binary releases

bclement-ocp:
This patch modifies the Docker workflow to build a statically linked portable binary of Alt-Ergo, following [1].

Building the Linux binaries is fairly quick and hence performed on push/pull_request.

[1] : https://ocamlpro.com/blog/2021_09_02_generating_static_and_portable_executables_with_ocaml/

PR #1067, opened 2024-03-21
WIP: macOS static binaries

bclement-ocp:


bclement-ocp:
This works -- closing. See #1045

PR #1068, opened 2024-03-21
Logs

Halbaroth:
This PR starts to use logs to print debug messages.
I tried to keep it as reasonable as possible, so I didn't change all the
printers.

- Every named modules and functions are located in a new module `Self`.
  I created this module because we share these named locations between
  printers, timers and the profiling module;
- Each named modules expose a `Logs` source (see `Self.get_source`);
- The CLI option `--dd` changes the level of these sources;
- I use the tag feature of `Logs` to print function names;
- I clean and use Logs to print debug messages in `Shostak`, `Satml`,
  `Satml_frontend` and the `enum` theory;
- The new Logs reporter `Printer.reporter` chooses the appropriate output
  according to the source.


Halbaroth:
This PR adds a pin on Ocplib-simplex (see issue #1077). We have to add this pin because the latest release of Ocplib-simplex prints some debugging messages at the `App` level.

bclement-ocp:
@Halbaroth what is the status of this PR?

bclement-ocp:
(Superseded by #1206 )

PR #1070, opened 2024-03-22
Sort trigger patterns before caching

Halbaroth:
As we noticed a long time ago, the `Matching` module of Alt-Ergo sorts patterns of multi-triggers during each iteration of the matching round. It's inefficient as the weight used for sorting doesn't change.

I thought we should perform these sorts during the generation of triggers in `Expr` but actually, Alt-Ergo can generate new triggers later depending on the matching configuration (see `Util.matching_env`).

In this PR, we perform the sorts just before caching the generated triggers in `Matching`.

### DIFF
--- a/src/lib/reasoners/matching.ml
+++ b/src/lib/reasoners/matching.ml
@@ -678,6 +678,8 @@ module Make (X : Arg) : S with type theory = X.t = struct
              | Util.Backward -> backward_triggers q, "Backward"
              | Util.Forward  -> forward_triggers q, "Forward"
            in
+           assert (List.for_all
+                     (fun E.{content; _} -> Lists.is_sorted pat_weight content) tgs);
            if Options.get_debug_triggers () then
              Printer.print_dbg
                ~module_name:"Matching" ~function_name:"add_triggers"
--- a/src/lib/reasoners/matching.ml
+++ b/src/lib/reasoners/matching.ml
@@ -482,10 +482,20 @@ module Make (X : Arg) : S with type theory = X.t = struct
     Debug.match_pats_modulo pat lsubsts;
     List.fold_left (match_one_pat mconf env tbox pat) [] lsubsts
 
+  let pat_weight s t =
+    let sf = (Expr.term_view s).f in
+    let tf = (Expr.term_view t).f in
+    match sf, tf with
+    | Symbols.Name _, Symbols.Op _ -> -1
+    | Symbols.Op _, Symbols.Name _ -> 1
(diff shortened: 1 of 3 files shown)


bclement-ocp:
I believe there are still places where we can create unsorted triggers, namely from `backwards_triggers`, `forward_triggers` and `make` itself.

I am a bit concerned about the potential for subtle bugs if unsorted triggers make their way to the `matching` function; I'd like to at least have a debug assertion that the triggers are sorted at that point.

Halbaroth:
Good catch! I didn't notice they created new triggers!

src/lib/reasoners/matching.ml
```diff
@@ -672,6 +678,8 @@ module Make (X : Arg) : S with type theory = X.t = struct
              | Util.Backward -> backward_triggers q, "Backward"
              | Util.Forward  -> forward_triggers q, "Forward"
            in
+           assert (List.for_all
+                     (fun E.{content; _} -> Lists.is_sorted pat_weight content) tgs);
```
bclement-ocp: I meant for this assertion to be in the `matching` function, i.e. where we were sorting the list before (gated behind `Options.get_enable_assertions ()` to avoid undoing the performance improvements). Having it only here in `add_triggers` could cause us to miss some cases (I don't think there are any right now, but we could introduce some if we refactor things in the future).

Halbaroth: Done

Halbaroth: For the performance, it's not crucial to guard the assertion with `Options.get_enable_assertions` as assertions disappeared in releases and we test release artifacts on Marvin. I'm not sure `Option.get_enable_assertions` is really useful. For instance `dune runtest` runs the test with the `dev` profile by default, which means the assertions are present on CI and we have to use CLI option `--enable-assertions` only because of this flag option.

bclement-ocp: > assertions disappeared in releases

This should not be the case unless we are compiling with `-noassert`, which I don't think we are.

Halbaroth: oh really? Sad. I thought all the assertions were removed but `assert false` in the release profile. I'll check.

Halbaroth: The `release` profile of dune doesn't add `-noassert`. Now, I remembered we used this flag only for the SMT-COMP.

bclement-ocp (approved):
Let's wait for the benchmark result and merge

PR #1071, opened 2024-03-22
`ml_z_extract_small` is available in `zarith_stubs_js` v0.16.1

Halbaroth:
We don't need to define this primitive anymore.

### DIFF
--- a/.github/workflows/build_js.yml
+++ b/.github/workflows/build_js.yml
@@ -5,7 +5,7 @@ name: Build Javascript
 on: [push,pull_request]
 
 env:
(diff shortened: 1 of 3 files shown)


bclement-ocp:
Is the pin for the stubs required? Can we add a `>= v0.16.1` version constraint instead?

Halbaroth:
I also relaxed the constraint on `js_of_ocaml` and `zarith_stubs_js` requires `ocaml>=4.14.0` so I bumped the tested version of the CI for the JS worker.

bclement-ocp (approved):

PR #1072, opened 2024-03-22
Fix js-deps target

Halbaroth:
Using `opam install package>=version` creates a file `=version` in the pwd. It seems that `opam install 'package>=version'` works better.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -247,11 +247,11 @@ dev-switch:
 
 js-deps:
 	opam install \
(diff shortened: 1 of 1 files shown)


bclement-ocp (approved):
Shell scripting is such a wonderful world.

Halbaroth:
rofl.... the CI will be broken because of the minimal version required by `zarith_subs_js`. 
I shouldn't push the auto merge button :(

PR #1073, opened 2024-03-25
fix(BV): Properly set fully interpreted operators

bclement-ocp:
Alt-Ergo calls "fully interpreted" symbols that are fully internalized in semantic values and do not need to be taken into consideration by the CC(X) algorithm (see function `congruents` in `ccx.ml`).

(Note that `fully_interpreted` is only called for the Shostak theory associated with the symbol according to `is_min_symb`.)

The BV theory incorrectly states that all its symbols are fully interpreted, including `bv2nat`, `int2bv`, etc.; however, only `concat`, `extract` and `bv2nat` are fully interpreted. This means that Alt-Ergo never performs congruence closure for these functions, leading to missed reasoning opportunities.

### DIFF
--- a/src/lib/reasoners/bitv.ml
+++ b/src/lib/reasoners/bitv.ml
@@ -1464,7 +1464,10 @@ module Shostak(X : ALIEN) = struct
       let compare = compare_mine
     end)
 *)
-  let fully_interpreted _ = true
+  let fully_interpreted sb =
+    match sb with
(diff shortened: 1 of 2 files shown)


bclement-ocp:
(CI failures are unrelated and should be fixed by #1074)

Halbaroth (approved):

PR #1074, opened 2024-03-25
fix(CI): Use OCaml 4.14.2 for Makefile tests

bclement-ocp:


Halbaroth:
I hesitated to bump the minimal OCaml version as you did. The only reason for which we need `OCaml >= 4.14` is because the latest version of `js_zarith_stubs` as such a requirement but this package isn't a dependency of `Alt-Ergo` itself. I believe no ones use the AE js worker but us.

bclement-ocp:
This does not bump the minimal OCaml version, it only changes the version that is only used on the CI to test the "Make all" target. Version constraints are unchanged, and we are still otherwise testing OCaml 4.08 / 4.14 / 5.0.

Halbaroth:
By bumping the minimal version, I meant bumping the minimal version tested by the CI. 
If we still test `OCaml 4.08` on the CI for the binary and the library, I agree ;)

Halbaroth (approved):

bclement-ocp:
Yes — the change only concerns the "Make all" workflow (which simply checks that we can build using `make`); it does not apply to the other workflows (those that actually run the tests) as they use a different [matrix](https://github.com/OCamlPro/alt-ergo/blob/d0da55ff3040decbd5cb1c14f6f0107d4a44b327/.github/workflows/build.yml#L24-L31)!

PR #1075, opened 2024-03-25
Fix wrong documentation of `Tadt`

Halbaroth:
The documentation of the variant `Tadt` in the module `Ty` was wrong. `Tadt (a, params)` denotes the application of a polymorphic adt to the type parameters `params`.

### DIFF
--- a/src/lib/structures/ty.mli
+++ b/src/lib/structures/ty.mli
@@ -58,9 +58,9 @@ type t =
   (** Enumeration, with its name, and the list of its constructors. *)
 
   | Tadt of Hstring.t * t list
-  (** Algebraic types applied to arguments. [Tadt (s, args)] is
-      the application of the datatype constructor [s] to
-      arguments [args]. *)
+  (** Application of algebraic data types. [Tadt (a, params)] denotes
+      the application of the polymorphim datatype [a] to the types parameters
+      [params]. *)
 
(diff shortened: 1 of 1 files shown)


src/lib/structures/ty.mli
```diff
@@ -58,9 +58,9 @@ type t =
   (** Enumeration, with its name, and the list of its constructors. *)
 
   | Tadt of Hstring.t * t list
-  (** Algebraic types applied to arguments. [Tadt (s, args)] is
-      the application of the datatype constructor [s] to
-      arguments [args]. *)
+  (** Application of algebraic data types. [Tadt (a, params)] denotes
+      the application of the polymorphim datatype [a] to the types parameters
```
bclement-ocp:
```suggestion
      the application of the polymorphic datatype [a] to the types parameters
```

The existing documentation is not technically wrong if you read "datatype constructor" as "(data) *type constructor*" but it is confusing and the new version is more clear (even better would be to include `Tadt ("List", ["Int"])` to represent `(List Int)` as an example).

bclement-ocp (approved):

src/lib/structures/ty.mli
Halbaroth: The documentation was really confusing for me and I wrote it! I added an example. Thanks :)

PR #1076, opened 2024-03-26
Prevent from adding dependencies of OCaml 5 in the lock file

Halbaroth:
If you run `make lock` with an `opam` switch using `OCaml 5`, you got two new packages in the lock files: `base-domains` and `base-nnp`. Not surprisingly, the former requires `OCaml 5`. The latter is an optional dependency for `OCaml 4` so I prevent from adding it too in the lock file.


bclement-ocp (approved):
Every time we have do do this sort of manipulation my belief that we shouldn't be using OPAM lock files grows stronger :cry:

PR #1082, opened 2024-03-29
feat(BV): Display BV constants using hexadecimal

bclement-ocp:
We currently always print bit-vector constants using binary representation. This can lead to quite large values in models when using 32- or 64-bit values that are essentially unreadable. When the bit-vector width is a multiple of 4, this patch uses hexadecimal notation instead, as allowed by the SMT-LIB specification.

Note that this only applies to bit-vector terms; bit-vector semantic values are always printed in their binary representation for debugging purposes.

### DIFF
--- a/src/lib/structures/expr.ml
+++ b/src/lib/structures/expr.ml
@@ -462,7 +462,10 @@ module SmtPrinter = struct
       pp_rational ppf q
 
     | Sy.Bitv (n, s), [] ->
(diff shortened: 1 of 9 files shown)


bclement-ocp:
Looks like I forgot to update the tests for the new output syntax… derp.

Halbaroth (approved):

PR #1083, opened 2024-03-29
 feat(BV, CP): Propagators for addition and multiplication 

bclement-ocp:
This patch adds constraints to represent bit-vector addition and
multiplication, with propagators in the interval domain and a simple
propagator for multiplication that only considers the factors-of-two
multiples in the bit-vector domain (this information cannot be captured
by the interval domain due to imprecise handling of overflow).

No propagator for addition in the bit-vector domain is provided yet,
although the plan is to add a full-adder propagator to provide a limited
form of *local* bit-blasting, as suggested in

  S. Bardin, P. Herrmann, and F. Perroud.
  “An Alternative to SAT-Based Approaches for Bit-Vectors”.
  In: TACAS. 2010.

Note: this requires and currently includes #1058 ; the commit titled "feat(BV, CP): Propagators for addition and multiplication" is new.

### DIFF
--- a/src/lib/reasoners/bitlist.ml
+++ b/src/lib/reasoners/bitlist.ml
@@ -49,7 +49,7 @@ let explanation { ex; _ } = ex
 
 let exact width value ex =
   { width
-  ; bits_set = value
+  ; bits_set = Z.extract value 0 width
   ; bits_clr = Z.extract (Z.lognot value) 0 width
   ; ex }
 
@@ -264,3 +264,39 @@ let fold_domain f b acc =
         (ofs + 1) { b with bits_set = Z.logor b.bits_set mask } acc
   in
   fold_domain_aux 0 b acc
+
+(* simple propagator: only compute known low bits *)
+let mul a b =
+  let sz = width a in
+  assert (width b = sz);
+
+  let ex = Ex.union (explanation a) (explanation b) in
+
+  (* (a * 2^n) * (b * 2^m) = (a * b) * 2^(n + m) *)
+  let zeroes_a = Z.trailing_zeros @@ Z.lognot a.bits_clr in
+  let zeroes_b = Z.trailing_zeros @@ Z.lognot b.bits_clr in
+  if zeroes_a + zeroes_b >= sz then
+    exact sz Z.zero ex
+  else
+    let low_bits =
+      if zeroes_a + zeroes_b = 0 then empty
+      else exact (zeroes_a + zeroes_b) Z.zero ex
+    in
+    let a = extract a zeroes_a (zeroes_a + sz - width low_bits - 1) in
+    assert (width a + width low_bits = sz);
+    let b = extract b zeroes_b (zeroes_b + sz - width low_bits - 1) in
+    assert (width b + width low_bits = sz);
+    (* ((ah * 2^n) + al) * ((bh * 2^m) + bl) =
+        al * bl  (mod 2^(min n m)) *)
+    let low_a_known = Z.trailing_zeros @@ Z.lognot @@ bits_known a in
+    let low_b_known = Z.trailing_zeros @@ Z.lognot @@ bits_known b in
+    let low_known = min low_a_known low_b_known in
+    let mid_bits =
+      if low_known = 0 then empty
+      else exact
+          low_known
+          Z.(extract (value a) 0 low_known * extract (value b) 0 low_known)
+          ex
+    in
+    concat (unknown (sz - width mid_bits - width low_bits) Ex.empty) @@
+    concat mid_bits low_bits
--- a/src/lib/reasoners/bitlist.ml
+++ b/src/lib/reasoners/bitlist.ml
@@ -265,7 +265,10 @@ let fold_domain f b acc =
   in
   fold_domain_aux 0 b acc
 
-(* simple propagator: only compute known low bits *)
+(* simple propagator: only compute known low bits
+
+   Example: ???100 * ???000 = ?00000 (trailing zeroes accumulate)
+            ???111 * ????11 = ????01 (min of low bits known) *)
 let mul a b =
   let sz = width a in
   assert (width b = sz);
@@ -278,6 +281,7 @@ let mul a b =
   if zeroes_a + zeroes_b >= sz then
     exact sz Z.zero ex
   else
+    (* Factor out the low zeroes *)
     let low_bits =
       if zeroes_a + zeroes_b = 0 then empty
       else exact (zeroes_a + zeroes_b) Z.zero ex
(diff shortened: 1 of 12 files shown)


bclement-ocp:
Rebased with the new Intervals module from #1108

src/lib/reasoners/bitlist.mli
```diff
@@ -118,6 +118,9 @@ val logor : t -> t -> t
 val logxor : t -> t -> t
 (** Bitwise xor. *)
 
+val mul : t -> t -> t
+(** Multiplication. *)
```
Halbaroth: We may mention that `mul` is the multiplication of unsigned integers represented as bitvectors. It is obvious that it cannot be the bitwise multiplication because we have already an `and` operator.

Halbaroth:
Can you rebase this PR please? I will review it today :)

bclement-ocp:
I don't understand why GitHub was saying there were merge conflicts in a bunch of files there were none? I force pushed again and that seems to have solved it.

src/lib/reasoners/bitlist.ml
```diff
@@ -264,3 +264,39 @@ let fold_domain f b acc =
...
+  let ex = Ex.union (explanation a) (explanation b) in
+
+  (* (a * 2^n) * (b * 2^m) = (a * b) * 2^(n + m) *)
+  let zeroes_a = Z.trailing_zeros @@ Z.lognot a.bits_clr in
+  let zeroes_b = Z.trailing_zeros @@ Z.lognot b.bits_clr in
+  if zeroes_a + zeroes_b >= sz then
+    exact sz Z.zero ex
+  else
+    let low_bits =
+      if zeroes_a + zeroes_b = 0 then empty
```
Halbaroth, line 286: I am not sure to understand why we can omit the explanation `ex` here?

src/lib/reasoners/bitlist.ml
```diff
@@ -264,3 +264,39 @@ let fold_domain f b acc =
...
+      else exact (zeroes_a + zeroes_b) Z.zero ex
+    in
+    let a = extract a zeroes_a (zeroes_a + sz - width low_bits - 1) in
+    assert (width a + width low_bits = sz);
+    let b = extract b zeroes_b (zeroes_b + sz - width low_bits - 1) in
+    assert (width b + width low_bits = sz);
+    (* ((ah * 2^n) + al) * ((bh * 2^m) + bl) =
+        al * bl  (mod 2^(min n m)) *)
```
Halbaroth: What are `ah` and `bh` here?

src/lib/reasoners/bitlist.ml
```diff
@@ -264,3 +264,39 @@ let fold_domain f b acc =
         (ofs + 1) { b with bits_set = Z.logor b.bits_set mask } acc
   in
   fold_domain_aux 0 b acc
+
+(* simple propagator: only compute known low bits *)
```
Halbaroth: Can you add a simple example on two bitlists?

src/lib/reasoners/bitv_rel.ml
```diff
@@ -296,10 +319,28 @@ end = struct
...
+  let propagate_interval_binop ~ex sz dr op dx dy =
+    let open Interval_domains.Ephemeral in
+    let norm i = Intervals.Int.extract i ~ofs:0 ~len:sz in
+    match op with
+    | Badd ->
+      update ~ex dr @@ norm @@ Intervals.Int.add !!dx !!dy;
+      update ~ex dy @@ norm @@ Intervals.Int.sub !!dr !!dx;
+      update ~ex dx @@ norm @@ Intervals.Int.sub !!dr !!dy
+
+    | Bmul -> (* Only forward propagation for now *)
```
Halbaroth, line 338: What do you mean by `forward propagation`. I don't know this notion.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -355,20 +397,22 @@ end = struct
...
     let inf = if strict then Intervals.map_bound Z.succ inf else inf in
-    Intervals.Int.of_bounds ~ex inf  Unbounded
+    Intervals.Int.of_bounds ~ex:(Ex.union ex ex') inf  Unbounded
 
   let propagate_less_than ~ex ~strict dx dy =
     let open Interval_domains.Ephemeral in
-    update ~ex dx (less_than_sup ~strict !!dy);
-    update ~ex dy (greater_than_inf ~strict !!dx)
+    (* Do not use [update] to make sure that the justification is only stored on
+       the upper/lower bound. *)
```
Halbaroth: I guess you mean that you do not use the named attribute of `update`.

Halbaroth (commented):
Just few questions but looks already pretty good for me.

src/lib/reasoners/bitlist.ml
bclement-ocp: `low_bits` is a bit-vector of width `zeroes_a + zeroes_b`; a bit-vector of width `0` does not need justification.

(This code is simplified in #1144)

src/lib/reasoners/bitlist.ml
bclement-ocp: High and low bits of `a` and `b`, respectively -- `a = ah * 2^n + al` with `0 <= al < 2^n` and similar for `b`. Clarified the comment.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: We are propagating information forward (from arguments `x`, `y` to result `r = x * y`). Backward propagation means we would be updating the domain of one argument from the domain of the result and the other arguments.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Yes, "do not use `update` for justifications" — clarified the comment.

src/lib/reasoners/bitlist.ml
bclement-ocp: Sure.

src/lib/reasoners/bitlist.mli
bclement-ocp: Done

bclement-ocp (commented):
Thanks for the review — ready for round 2

src/lib/reasoners/bitlist.ml
Halbaroth: I am still a bit lost about the meaning of explanations of domains. I thought that a binding `r |--> (d, ex)` means `ex --> r in d` is a tautology (true in any model).

bclement-ocp: That is correct. Note that we then concatenate the `low_bits` with higher bits that do have a justification.

Here `low_bits` represent the extraction of the low bits up to bit number `zeroes_a + zeroes_b` exclusive. If `zeroes_a + zeroes_b` is `0`, then we represent a bit-vector of size `0`.

There is a single bit-vector of size `0` (the "empty" bit-vector), so we do not need a justification — any value of type "bit-vector of size 0" is equal to the same empty bit-vector. If `r` has width `0`, then the equality `r = empty` is true in any model.

(The type of bit-vectors of size 0 does not actually exist, and we can't build values of that type — it is simply a reasoning artifact here)

Halbaroth (approved):

PR #1084, opened 2024-03-29
feat(BV, CP): Add propagators for bvudiv and bvurem

bclement-ocp:
These are interval propagators only and respect the SMT-LIB semantics
for division by 0. This requires custom operators in the `Intervals`
module that update the bounds appropriately, sincethe existing `div`
operator is undefined on `0`.

**Note**: this requires and currently includes https://github.com/OCamlPro/alt-ergo/pull/1058 and #1083  ; the commit titled "feat(BV, CP): Add propagators for bvudiv and bvurem" is new.

### DIFF
--- a/src/lib/reasoners/intervals.ml
+++ b/src/lib/reasoners/intervals.ml
@@ -654,16 +654,19 @@ module Int = struct
                 (* x < y : bvurem is the identity *)
                 interval_set i1
               else if (
-                ZEuclideanType.equal i2.lb ZEuclideanType.zero ||
-                ZEuclideanType.compare i1.ub i2.ub < 0
+                ZEuclideanType.equal i2.lb ZEuclideanType.zero
               ) then
                 (* The range [0, i1.ub] is always valid; it is also the best we
-                   can do if [y] can be either [0] or bigger than [x]. *)
-                interval_set @@ { i1 with lb = ZEuclideanType.zero }
+                   can do if [y] can be [0]. *)
+                interval_set { i1 with lb = ZEuclideanType.zero }
               else
-                (* y is non-zero and smaller than x; use [0, i2.ub[ *)
-                interval_set @@
-                { lb = ZEuclideanType.zero ; ub = ZEuclideanType.pred i2.ub }
+                (* y is non-zero; we have both [x % y < y] and [x % y <= x] so
+                   take the min of these upper bounds. *)
+                let ub =
+                  if ZEuclideanType.compare i1.ub i2.ub < 0 then i1.ub
+                  else ZEuclideanType.pred i2.ub
+                in
+                interval_set { lb = ZEuclideanType.zero ; ub }
             ) u1
       ) u2
 end
--- a/src/lib/reasoners/intervals.ml
+++ b/src/lib/reasoners/intervals.ml
@@ -632,16 +632,22 @@ module Int = struct
     trace1 "lognot" u @@ map_strict_dec ZEuclideanType.lognot u
 
   let bvudiv ~size:sz u1 u2 =
-    (* [bvudiv] is euclidean division where division by zero is -1 *)
+    (* [bvudiv] is euclidean division where division by zero is -1 (as an
+       integer of width [sz], so 2^sz - 1) *)
     let mone = Z.extract Z.minus_one 0 sz in
     ediv ~div0:(Interval.of_bounds (Closed mone) (Closed mone)) u1 u2
 
   let bvurem u1 u2 =
+    (* In the following, [x] is the implicit variable associated with [u1] and
+       [y] the implicit variable associated with [u2]. *)
     of_set_nonempty @@
     map_to_set (fun i2 ->
         if ZEuclideanType.equal i2.ub ZEuclideanType.zero then
           (* [y] is always zero -> identity *)
           map_to_set interval_set u1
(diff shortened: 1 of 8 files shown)


bclement-ocp:
Rebased on #1108

src/lib/reasoners/intervals.ml
```diff
@@ -630,6 +630,36 @@ module Int = struct
 
   let lognot u =
     trace1 "lognot" u @@ map_strict_dec ZEuclideanType.lognot u
+
+  let bvudiv ~size:sz u1 u2 =
+    (* [bvudiv] is euclidean division where division by zero is -1 *)
```
Halbaroth: This comment is a bit ambiguous in this module. In the bitvector theory, `-1` is clearly represented as two's complement because there is no sign on bitvector. Zarith's numbers have a sign and `-1` can be:
- the negative number represented by `11111....`
- the positive number `2^sz-1`.

Here, the image when `u2 = 0` is the interval `[2^sz-1; 2^sz-1]`.

src/lib/reasoners/intervals.ml
```diff
@@ -630,6 +630,36 @@ module Int = struct
...
+    let mone = Z.extract Z.minus_one 0 sz in
+    ediv ~div0:(Interval.of_bounds (Closed mone) (Closed mone)) u1 u2
+
+  let bvurem u1 u2 =
+    of_set_nonempty @@
+    map_to_set (fun i2 ->
+        if ZEuclideanType.equal i2.ub ZEuclideanType.zero then
+          (* [y] is always zero -> identity *)
```
Halbaroth, line 646: There is no `y` here. I guess `y` denotes the variable whose the interval is `u2`.

src/lib/reasoners/intervals.ml
```diff
@@ -630,6 +630,36 @@ module Int = struct
...
+          map_to_set interval_set u1
+        else
+          map_to_set (fun i1 ->
+              if ZEuclideanType.compare i1.ub i2.lb < 0 then
+                (* x < y : bvurem is the identity *)
```
Halbaroth, line 654: Same here, there is no `x` and `y`.

src/lib/reasoners/intervals.ml
```diff
@@ -630,6 +630,36 @@ module Int = struct
...
+
+  let bvudiv ~size:sz u1 u2 =
+    (* [bvudiv] is euclidean division where division by zero is -1 *)
+    let mone = Z.extract Z.minus_one 0 sz in
+    ediv ~div0:(Interval.of_bounds (Closed mone) (Closed mone)) u1 u2
+
+  let bvurem u1 u2 =
+    of_set_nonempty @@
+    map_to_set (fun i2 ->
+        if ZEuclideanType.equal i2.ub ZEuclideanType.zero then
```
Halbaroth, line 645: I believe we assume that `y` cannot be negative, which makes sense for the modulo. We should add a test:
```ocaml 
if ZEuclideanType.compare i2.ub ZEuclideanType.zero < 0 then 
  invalid_arg ...
```

Halbaroth (changes requested):

src/lib/reasoners/intervals.ml
bclement-ocp: The result must be a `sz`-bit integer, so it seems fairly clear to me that `-1` means `2^sz - 1`. Clarified the comment.

src/lib/reasoners/intervals.ml
bclement-ocp: Indeed -- good catch, I will add such an assertion.

src/lib/reasoners/intervals.ml
bclement-ocp: Yes `x` represents the variable associated with `u1` and `y` with `u2`, I believe those were the name of the intervals in the first version of the code. Clarified.

src/lib/reasoners/intervals.ml
bclement-ocp: Ditto

bclement-ocp (commented):
Ready for round 2

src/lib/reasoners/intervals.ml
```diff
@@ -630,6 +630,36 @@ module Int = struct
...
+                interval_set i1
+              else if (
+                ZEuclideanType.equal i2.lb ZEuclideanType.zero ||
+                ZEuclideanType.compare i1.ub i2.ub < 0
+              ) then
+                (* The range [0, i1.ub] is always valid; it is also the best we
+                   can do if [y] can be either [0] or bigger than [x]. *)
+                interval_set @@ { i1 with lb = ZEuclideanType.zero }
+              else
+                (* y is non-zero and smaller than x; use [0, i2.ub[ *)
```
Halbaroth: I don't know what's happen... the UI is terrible and it seems the most important part of the review has been lost :/ 

So:
`y` is not necessary smaller than `x` here but still `[0, i2.ub[` is a valid interval because `x % y < y <=  i2.ub`.

Halbaroth (changes requested):

src/lib/reasoners/intervals.ml
bclement-ocp: Duh, this comment and the one above were bogus indeed — fixed

Halbaroth (approved):

PR #1085, opened 2024-03-29
 feat(BV, CP): Add propagators for bvshl and bvlshr

bclement-ocp:
This patch adds interval and bitlist propagators for the bvshl (left
shift) and bvlshr (logical right shift) in the intervals and bitlist
domains for bit-vectors.

The interval propagator for left shift needs to be written specially in
order to properly deal with overflow, but the propagator for bvlshr is
written using a generic propagator for (bi)-monotone functions.

The bitlist propagator for bvshl is required because it needs to
propagate information regarding low bits that are not tracked by
intervals. However, I am not sure that the bitlist propagator for bvlshr
is actually needed since it might be subsumed by the interval propagator
for bvlshr (and consistency constraints) entirely, and we might want to
remove it.

**Note**: this requires and currently includes https://github.com/OCamlPro/alt-ergo/pull/1058, https://github.com/OCamlPro/alt-ergo/pull/1083 and #1084 ; the commit titled " feat(BV, CP): Add propagators for bvshl and bvlshr" is new.

### DIFF
--- a/src/lib/reasoners/bitlist.ml
+++ b/src/lib/reasoners/bitlist.ml
@@ -307,3 +307,61 @@ let mul a b =
     in
     concat (unknown (sz - width mid_bits - width low_bits) Ex.empty) @@
     concat mid_bits low_bits
+
+let shl a b =
+  (* If the minimum value for [b] is larger than the bitwidth, the result is
+     zero.
+
+     Otherwise, any low zero bit in [a] is also a zero bit in the result, and
+     the minimum value for [b] also accounts for that many minimum zeros (e.g.
+     ?000 shifted by at least 2 has at least 5 low zeroes).
+
+     NB: [increase_lower_bound b Z.zero] is the smallest positive integer that
+     matches the bitlist pattern, and so is a lower bound. Ideally we would
+     like to use the lower bound from the interval domain for [b] instead. *)
+  match Z.to_int (increase_lower_bound b Z.zero) with
+  | n when n < width a ->
+    let low_zeros = Z.trailing_zeros @@ Z.lognot @@ a.bits_clr in
+    if low_zeros + n >= width a then
+      exact (width a) Z.zero (Ex.union (explanation a) (explanation b))
+    else if low_zeros + n > 0 then
+      concat (unknown (width a - low_zeros - n) Ex.empty) @@
+      exact (low_zeros + n) Z.zero (Ex.union (explanation a) (explanation b))
+    else
+      unknown (width a) Ex.empty
+  | _ | exception Z.Overflow ->
+    exact (width a) Z.zero (explanation b)
+
+let lshr a b =
+  (* If the minimum value for [b] is larger than the bitwidth, the result is
+     zero.
+
+     Otherwise, any high zero bit in [a] is also a zero bit in the result, and
+     the minimum value for [b] also accounts for that many minimum zeros (e.g.
+     000??? shifted by at least 2 is 00000?).
+
+     NB: [increase_lower_bound b Z.zero] is the smallest positive integer that
+     matches the bitlist pattern, and so is a lower bound. Ideally we would
+     like to use the lower bound from the interval domain for [b] instead. *)
+  match Z.to_int (increase_lower_bound b Z.zero) with
+  | n when n < width a ->
+    let sz = width a in
+    if Z.testbit a.bits_clr (sz - 1) then (* MSB is zero *)
+      let low_msb_zero = Z.numbits @@ Z.extract (Z.lognot a.bits_clr) 0 sz in
+      let nb_msb_zeros = sz - low_msb_zero in
+      assert (nb_msb_zeros > 0);
+      let nb_zeros = nb_msb_zeros + n in
+      if nb_zeros >= sz then
+        exact sz Z.zero (Ex.union (explanation a) (explanation b))
+      else
+        concat
+          (exact nb_zeros Z.zero (Ex.union (explanation a) (explanation b)))
+          (unknown (sz - nb_zeros) Ex.empty)
+    else if n > 0 then
+      concat
+        (exact n Z.zero (explanation b))
+        (unknown (sz - n) Ex.empty)
+    else
+      unknown sz Ex.empty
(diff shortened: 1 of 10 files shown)


bclement-ocp:
Rebased on #1108

src/lib/reasoners/bitlist.ml
```diff
@@ -300,3 +300,59 @@ let mul a b =
...
+  (* If the minimum value for [b] is larger than the bitwidth, the result is
+     zero.
+
+     Otherwise, any low zero bit in [a] is also a zero bit in the result, and
+     the minimum value for [b] also accounts for that many minimum zeros (e.g.
+     ?000 shifted by at least 2 has at least 5 low zeroes).
+
+     NB: we would like to use the lower bound from the interval domain for [b]
+     here. *)
+  match Z.to_int (increase_lower_bound b Z.zero) with
```
Halbaroth, line 322: It is a bit weird to use `increase_lower_bound` in order to extract the lower bound of `b`. I think we should write a specialization: 
```ocaml 
let extract_lower_bound b = increase_lower_bound Z.zero 
```

Alternatively, we can rename `increase_lower_bound` because its name is not very descriptive but I have no good suggestions for it right now.

src/lib/reasoners/bitlist.ml
```diff
@@ -300,3 +300,59 @@ let mul a b =
...
+  (* If the minimum value for [b] is larger than the bitwidth, the result is
+     zero.
+
+     Otherwise, any high zero bit in [a] is also a zero bit in the result, and
+     the minimum value for [b] also accounts for that many minimum zeros (e.g.
+     000??? shifted by at least 2 is 00000?).
+
+     NB: we would like to use the lower bound from the interval domain for [b]
+     here. *)
+  match Z.to_int (increase_lower_bound b Z.zero) with
```
Halbaroth, line 346: Same remark here.

src/lib/reasoners/bitlist.ml
```diff
@@ -300,3 +300,59 @@ let mul a b =
...
+    let low_zeros = Z.trailing_zeros @@ Z.lognot @@ a.bits_clr in
+    if low_zeros + n >= width a then
+      exact (width a) Z.zero (Ex.union (explanation a) (explanation b))
+    else if low_zeros + n > 0 then
+      concat (unknown (width a - low_zeros - n) Ex.empty) @@
+      exact (low_zeros + n) Z.zero (Ex.union (explanation a) (explanation b))
+    else
+      unknown (width a) Ex.empty
+  | _ | exception Z.Overflow ->
+    exact (width a) Z.zero (explanation b)
```
Halbaroth, line 333: Let's bind `width a` to a shorter name like `sz`.

src/lib/reasoners/bitlist.ml
```diff
@@ -300,3 +300,59 @@ let mul a b =
...
+     NB: we would like to use the lower bound from the interval domain for [b]
+     here. *)
+  match Z.to_int (increase_lower_bound b Z.zero) with
+  | n when n < width a ->
+    let low_zeros = Z.trailing_zeros @@ Z.lognot @@ a.bits_clr in
+    if low_zeros + n >= width a then
+      exact (width a) Z.zero (Ex.union (explanation a) (explanation b))
+    else if low_zeros + n > 0 then
+      concat (unknown (width a - low_zeros - n) Ex.empty) @@
+      exact (low_zeros + n) Z.zero (Ex.union (explanation a) (explanation b))
```
Halbaroth, line 329: I understand we cannot keep bits set to one because we only consider the lower bound of `b`, so we may clear all the bits if the value in `b` is too large. Should we consider the upper bound of `b` to keep bits set to one if possible?
For instance, if the bitlist of `a` is `1?11` and the width of `b`is `1`, we can return the bitlist `?110` instead of `?000`?

src/lib/reasoners/intervals.ml
```diff
@@ -311,6 +311,47 @@ module ZEuclideanType = struct
...
     | Finite n -> Finite (Z.lognot n)
+
+  (* Values larger than [max_int] are treated as +oo *)
+  let shift_left ?(max_int = max_int) x y =
+    match y with
+    | Neg_infinite ->
+      Fmt.invalid_arg "shl: must shift by nonnegative amount"
+    | Pos_infinite -> Pos_infinite
+    | Finite y when Z.sign y < 0 ->
+      Fmt.invalid_arg "shl: must shift by nonnegative amount"
```
Halbaroth: Nit: it is clearer to group all the invalid cases.

Halbaroth (changes requested):

src/lib/reasoners/bitlist.ml
bclement-ocp: As the comment above notes, we would want to use the bounds from the interval domain instead, so this should disappear at some point. I will clarify that `increase_lower_bound Z.zero` is the smallest positive valid lower bound.

src/lib/reasoners/bitlist.ml
bclement-ocp: If `a` is `1?11` and `b` is at most `1`, the possible values of the shift are `1?11` and `?110`, so we cannot do better than `??1?`.

We could consider trying to keep track of runs of `1` in the low bits whose length is at smaller than the maximum value of `b`, but this is fairly complex code to write — we could do it, but I'm not sure the complexity/benefits ratio is good enough.

src/lib/reasoners/bitlist.ml
bclement-ocp: Since this code is removed in #1144 I'd rather not — it would only add rebase conflicts.

src/lib/reasoners/bitlist.ml
bclement-ocp: Ditto

src/lib/reasoners/intervals.ml
bclement-ocp: Sure

Halbaroth (approved):
I trust you to do the appropriate changes and accept this PR now ;)

PR #1086, opened 2024-04-02
ADT destructors are delayed functions

Halbaroth:
Before this PR, the ADT destructors worked as follow:
- Destructor applictions from the user input were guarded: there are considered as aliens by the ADT theory;
- If the domain of an ADT semantic values became a singleton or if the input contains a tester application on this value, the ADT destructor application on this semantic value became interpreted.

The reason of this contruction is because all the functions are total in the SMT-LIB standard, even the destructors. So a destructor is partially interpreted:

If `v` is an ADT value and `d` is a destructor of this ADT associated with the constructor `c`, then:
- `d v` can be anything if the constructor of `v` isn't necessary `c`;
- `d v` is the field of `c` associated with `d` if the only possible constructor for `v` is `c`.

This PR uses the new mechanism `Rel_utils.Delayed` to delay the computation of `d v` until we know that `v` is of the form `(c ...)`.

This PR is rebased on #1078.


src/lib/reasoners/rel_utils.ml
```diff
@@ -87,7 +87,8 @@ module Delayed : sig
       by [dispatch].
 
       [dispatch] must be pure. *)
-  val create : (Symbols.operator -> delayed_fn option) -> t
+  val create :
+    is_ready:(X.r -> bool) -> (Symbols.operator -> delayed_fn option) -> t
```
bclement-ocp, line 94: Doc should be updated for the new parameter

src/lib/reasoners/adt_rel.ml
```diff
@@ -798,41 +619,72 @@ let assume env uf la =
...
+
+  | _ -> assert false
+
+exception Found of X.r * Hstring.t
+
+(* Pick a delayed destructor application in [env.delayed].
+
+   @raise Found (r, d) if there is a delayed application of the destructor
+          [d] on the semantic value [r]. *)
+let pick_delayed_destructor env =
```
bclement-ocp, line 652: Not a big fan of returning values through an exception, I personally wouldn't mind paying the cost of a `try .. with` to return an option here.

bclement-ocp (approved):
Getting rid of this of guarded/unguarded destructors is very nice! Doc needs to be updated and #1078 resolved first but the use of delayed functions here looks good to me :+1:

src/lib/reasoners/adt_rel.ml
Halbaroth: Actually I hesitated between the two versions. I will rollback to the previous one.

PR #1087, opened 2024-04-02
Refactoring `Adt_rel` using domains on class representatives only

Halbaroth:
This PR is rebased on #1078 and #1086.

### DIFF
--- a/src/lib/reasoners/adt_rel.ml
+++ b/src/lib/reasoners/adt_rel.ml
@@ -45,6 +45,9 @@ module MHs = Hs.Map
 let timer = Timers.M_Adt
 
 module Domain = struct
+  (* Set of possible constructors. The explanation justifies that any value
+     assigned to the semantic value has to use a constructor lying in the
+     domain. *)
   type t = {
     constrs : HSS.t;
     ex : Ex.t;
@@ -60,7 +63,7 @@ module Domain = struct
 
   let domain ~constrs ex =
     if HSS.is_empty constrs then
-      raise_notrace @@ Inconsistent ex
+      raise @@ Inconsistent ex
     else
       { constrs; ex }
 
@@ -242,9 +245,6 @@ type t = {
      inconsistency. *)
 
   domains : Domains.t;
-  (* Map of the uninterpreted semantic values to domains of their possible
-     constructors. The explanation justifies that any value assigned to
-     the semantic value has to use a constructor lying in the domain. *)
 
   delayed : Rel_utils.Delayed.t;
   (* Set of delayed destructor applications. The computation of a destructor
@@ -339,13 +339,6 @@ let assume_subst ~ex r1 r2 env =
   let domains = Domains.subst ~ex r1 r2 env.domains in
   { env with domains }
 
-let is_tightenable ty c =
-  match ty with
-  | Ty.Tadt (name, params) ->
-    let Adt cases = Ty.type_body name params in
-    Lists.is_empty @@ Ty.assoc_destrs c cases
-  | _ -> assert false
-
 (* Update the domains of the semantic values [r1] and [r2] according to the
    disequality [r1 <> r2].
 
@@ -361,25 +354,27 @@ let is_tightenable ty c =
 
    @raise Domain.Inconsistent if the disequality is inconsistent with
           the current environment [env]. *)
-let assume_distinct ~ex r1 r2 env =
-  let d1 = Domains.get r1 env.domains in
-  let d2 = Domains.get r2 env.domains in
-  let env =
-    match Domain.as_singleton d1 with
-    | Some (c, ex1) when is_tightenable (X.type_info r2) c ->
-      let ex = Ex.union ex ex1 in
-      let nd = Domain.remove ~ex c d2 in
-      tighten_domain r2 nd env
-    | Some _ | None ->
+let assume_distinct =
+  let aux ~ex r1 r2 env =
+    match Th.embed r1 with
+    | Constr { c_args; _ } when List.length c_args = 0 ->
+      begin
+        let d1 = Domains.get r1 env.domains in
+        let d2 = Domains.get r2 env.domains in
+        match Domain.as_singleton d1 with
+        | Some (c, ex') ->
+          let ex = Ex.union ex ex' in
+          let nd = Domain.remove ~ex c d2 in
+          tighten_domain r2 nd env
+        | None -> env
+      end
+    | _ ->
+      (* If `d1` is a singleton domain but its constructor has a payload,
+         we cannot tighten the domain of `r2` because they could have
+         distinct payloads. *)
       env
-  in
-  match Domain.as_singleton d2 with
-  | Some (c, ex2) when is_tightenable (X.type_info r1) c ->
-    let ex = Ex.union ex ex2 in
-    let nd = Domain.remove ~ex c d1 in
-    tighten_domain r1 nd env
-  | Some _ | None ->
-    env
+  in fun ~ex r1 r2 env ->
+    aux ~ex r1 r2 env |> aux ~ex r2 r1
 
 (* Assumes the tester `((_ is c) r)` where [r] can be a constructor
    application or a uninterpreted semantic value.
@@ -487,7 +482,15 @@ let build_constr_eq r c =
           try Ty.assoc_destrs c body with Not_found -> assert false
         in
         let xs = List.map (fun (_, ty) -> E.fresh_name ty) ds in
-        Some (r, E.mk_constr (Hstring.view c) xs ty)
+        let cons = E.mk_constr (Hstring.view c) xs ty in
+        let r', ctx = X.make cons in
+        (* In the current implementation of `X.make`, we produce
+           a nonempty context only for interpreted semantic values
+           of the `Arith` and `Records` theories. The semantic
+           values `cons` never involves such values. *)
+        assert (Lists.is_empty ctx);
+        let eq = Shostak.L.(view @@ mk_eq r r') in
+        Some (eq, E.mk_constr (Hstring.view c) xs ty)
 
       | _ -> assert false
     end
@@ -506,14 +509,7 @@ let propagate_domains env =
        match Domain.as_singleton d with
        | Some (c, ex) ->
          begin match build_constr_eq rr c with
-           | Some (r, cons) ->
-             let r', ctx = X.make cons in
-             (* In the current implementation of `X.make`, we produce
-                a nonempty context only for interpreted semantic values
-                of the `Arith` and `Records` theories. The semantic
-                values `cons` never involves such values. *)
-             assert (Lists.is_empty ctx);
-             let eq = Shostak.L.(view @@ mk_eq r r') in
+           | Some (eq, cons) ->
              let new_terms = SE.add cons new_terms in
              (Literal.LSem eq, ex, Th_util.Other) :: eqs, new_terms
            | None ->
@@ -521,7 +517,7 @@ let propagate_domains env =
          end
        | None ->
          eqs, new_terms
-    ) ([], SE.empty) env.domains
+    ) ([], env.new_terms) env.domains
 
 (* Remove duplicate literals in the list [la]. *)
 let remove_redundancies la =
@@ -547,6 +543,7 @@ let count_splits env la =
 
 let assume env uf la =
   Debug.pp_env "before assume" env;
+  (* should be done globally in CCX *)
   let la = remove_redundancies la in
   let delayed, result = Rel_utils.Delayed.assume env.delayed uf la in
   let env =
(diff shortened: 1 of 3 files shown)


Halbaroth:
As explained offline, we cannot use `X.extract_term` as we was done in `casesplit` in a previous version of this PR because there is no implementation of `term_extract` for AC symbols. For instance the following input file:
```
type t = A | B of { d : int }
logic e1, e2 : t
logic ac u : t, t -> t 

axiom a1 : e1 <> e2
axiom a2 : (u(e1, e2) ? B)
goal g : (u(e2, e1) ? B)
```
raises an exception with `Alt-Ergo v2.5.3` for the same reason because we used `X.term_extract` in `deduce_is_constr`. 

In the new implementation of `casesplit`, we produces a semantic literal instead of a syntactical one.

Halbaroth:
@bclement-ocp ping

src/lib/reasoners/adt_rel.ml
```diff
@@ -29,81 +29,181 @@
...
-         - When we add a guarded destructor and we have already assumed the
-           tester of its associated constructor, we add a new equation to the
-           queue between this destructor and its unguarded version.
-
-         This pending queue is flushed at the end of [assume] and its
-         content is propagated to the environment of CC(X). *)
+module Domain = struct
+  type t = {
+    constrs : HSS.t;
+    ex : Ex.t;
```
bclement-ocp, line 53: There was some nice documentation for the corresponding type in the old implementation that could be moved here.

src/lib/reasoners/adt_rel.ml
```diff
@@ -29,81 +29,181 @@
...
+
+  let[@inline always] as_singleton { constrs; ex } =
+    if HSS.cardinal constrs = 1 then
+      Some (HSS.choose constrs, ex)
+    else
+      None
+
+  let domain ~constrs ex =
+    if HSS.is_empty constrs then
+      raise_notrace @@ Inconsistent ex
```
bclement-ocp:
```suggestion
      raise @@ Inconsistent ex
```

I am uneasy about using `raise_notrace` here, because this is the only place where we `raise` the `Inconsistent` exception for all the possible cases in the module, and it will make debugging errors in this module difficult.

(In contrast, in the SatML module, `raise_notrace` is used in a function that has exactly one caller — although in retrospect I think that using it there was a mistake)

src/lib/reasoners/adt_rel.ml
```diff
@@ -134,18 +234,42 @@ let dispatch = function
...
+  classes : E.Set.t list;
+  (* State of the union-find represented by all its equivalence classes.
+     This state is kept for debugging purposes only. It is updated after
+     assuming literals of the theory and returned by queries in case of
+     inconsistency. *)
+
+  domains : Domains.t;
+  (* Map of the uninterpreted semantic values to domains of their possible
+     constructors. The explanation justifies that any value assigned to
+     the semantic value has to use a constructor lying in the domain. *)
```
bclement-ocp: This part should be moved to the doc of `Domain.t` type.

src/lib/reasoners/adt_rel.ml
```diff
@@ -218,334 +327,201 @@ let assume_th_elt t th_elt _ =
...
-        end
-      | _ ->
-        Printer.print_err "%a" X.print r;
-        assert false
-    end
-  | Constr _ | Select _ -> env, eqs
-
-(* Collect all the constructors of the ADT type [ty]. *)
-let values_of ty =
+let is_tightenable ty c =
```
bclement-ocp: This looks like it checks whether `c` is an enum constructor of `ty`, is that correct? If so `is_enum_constructor` would be a more clear name, or a short comment on `is_tightenable` to explain this.

src/lib/reasoners/adt_rel.ml
```diff
@@ -218,334 +327,201 @@ let assume_th_elt t th_elt _ =
...
+   as we did in the Enum theory.
+
+   @raise Domain.Inconsistent if the disequality is inconsistent with
+          the current environment [env]. *)
+let assume_distinct ~ex r1 r2 env =
+  let d1 = Domains.get r1 env.domains in
+  let d2 = Domains.get r2 env.domains in
+  let env =
+    match Domain.as_singleton d1 with
+    | Some (c, ex1) when is_tightenable (X.type_info r2) c ->
```
bclement-ocp: Nit: It might be simpler to keep a match on `extract r1` and `extract r2` here as previously and tighten when seeing a `Constr { c_args = []; _ }` values. It should have the same effect (propagation ensures that singleton domain become `Constr` semantic values already) while being more readable (no need for `is_tightenable`).

tests/adts/distinct_same_constructor.smt2
```diff
@@ -0,0 +1,9 @@
+(set-logic ALL)
+
+(declare-datatype list
+  ((cons (value Int) (tail list)) (nil)))
+
+(declare-fun l () list)
+(assert (distinct l (cons 0 nil) nil))
```
bclement-ocp, line 7: I am not sure I understand what this is trying to test, can you explain?

src/lib/reasoners/adt_rel.ml
```diff
@@ -218,334 +327,201 @@ let assume_th_elt t th_elt _ =
...
+       match Domain.as_singleton d with
+       | Some (c, ex) ->
+         begin match build_constr_eq rr c with
+           | Some (r, cons) ->
+             let r', ctx = X.make cons in
+             (* In the current implementation of `X.make`, we produce
+                a nonempty context only for interpreted semantic values
+                of the `Arith` and `Records` theories. The semantic
+                values `cons` never involves such values. *)
+             assert (Lists.is_empty ctx);
```
bclement-ocp: Ideally the call to `X.make` should be called at the same point that the term is created (otherwise we might change the code in `build_constr_eq` and break this property accidentally). Could `build_constr_eq` call `X.make` and return `Some (r, cons, r')`?

src/lib/reasoners/adt_rel.ml
```diff
@@ -558,75 +534,47 @@ let remove_redundancies la =
...
+  List.fold_left
+    (fun nb (_, _, _, i) ->
+       match i with
+       | Th_util.CS (Th_util.Th_adt, n) -> Numbers.Q.mult nb n
+       | _ -> nb
+    ) env.size_splits la
 
+let assume env uf la =
+  Debug.pp_env "before assume" env;
+  let la = remove_redundancies la in
```
bclement-ocp, line 547: Please keep the comment about doing this globally in CCX which is still relevant.

src/lib/reasoners/adt_rel.ml
```diff
@@ -558,75 +534,47 @@ let remove_redundancies la =
...
+  let delayed, result = Rel_utils.Delayed.assume env.delayed uf la in
+  let env =
+    try
+      assume_literals la uf env
+    with Domain.Inconsistent ex ->
+      raise_notrace (Ex.Inconsistent (ex, env.classes))
+  in
+  let (assume, new_terms), domains = propagate_domains env in
+  let assume = List.rev_append result.assume assume in
```
bclement-ocp: Is there a reason to add the existing assumptions before the new ones? It was done the other way before, and usually we can expect `assume` to be empty (any assumption not involving ADTs) even when `result.assume` is not.

src/lib/reasoners/adt_rel.ml
```diff
@@ -218,334 +327,201 @@ let assume_th_elt t th_elt _ =
...
+             assert (Lists.is_empty ctx);
+             let eq = Shostak.L.(view @@ mk_eq r r') in
+             let new_terms = SE.add cons new_terms in
+             (Literal.LSem eq, ex, Th_util.Other) :: eqs, new_terms
+           | None ->
+             eqs, new_terms
+         end
+       | None ->
+         eqs, new_terms
+    ) ([], SE.empty) env.domains
```
bclement-ocp:
```suggestion
    ) ([], env.new_terms) env.domains
```

src/lib/reasoners/adt_rel.ml
```diff
@@ -558,75 +534,47 @@ let remove_redundancies la =
...
+    with Domain.Inconsistent ex ->
+      raise_notrace (Ex.Inconsistent (ex, env.classes))
+  in
+  let (assume, new_terms), domains = propagate_domains env in
+  let assume = List.rev_append result.assume assume in
+  let env = {
+    delayed; domains;
+    classes = Uf.cl_extract uf;
+    size_splits = count_splits env la;
+    new_terms = SE.union env.new_terms new_terms;
```
bclement-ocp: It would be more efficient to have `propagate_domains` accumulate over the existing `env.new_terms` since it has access to the `env` already (and then the `new_terms` returned by `propagate_domains` already contains `env.new_terms`).

bclement-ocp (approved):
I have not done a very exhaustive side-by-side comparison with the previous implementation, but the new one looks correct. There are some nitpicks here and there, and I'd really like the `raise_notrace` removed for the reasons explained inline, but overall I am OK with this change.

src/lib/reasoners/adt_rel.ml
Halbaroth: I agree. Actually I hesitated to remove it because I had to replace it by `raise` in order to debug the implementation.

src/lib/reasoners/adt_rel.ml
Halbaroth: I believe it's matter of taste. If I rename it into `is_enum_constructor`, I have to explain why we use this filter in `assume_distinct`. I suggest to rename it but I'll add a comment in `assume_distinct` too.

Halbaroth: I suggest this slightly different implementation:
```ocaml 
let assume_distinct =
  let aux ~ex r r' env =
    let d = Domains.get r env.domains in
    match Domain.as_singleton d with
    | Some (c, ex') when is_enum_cons (X.type_info r) c ->
      let ex = Ex.union ex ex' in
      let nd = Domain.remove ~ex c d in
      tighten_domain r' nd env
    | Some _ | None ->
      (* If `d` is a singleton domain but its constructor has a payload,
         we cannot tighten the domain of `r'` because they could have
         distinct payloads. *)
      env
  in fun ~ex r1 r2 env ->
    aux ~ex r1 r2 env |> aux ~ex r2 r1
```

tests/adts/distinct_same_constructor.smt2
Halbaroth: I added this test to ensure we don't answer `unsat` by tightening to much the domain of `l` with `assume_distinct`.

src/lib/reasoners/adt_rel.ml
bclement-ocp: That looks OK to me and I agree, an explanatory comment is probably warranted whatever the name is.

I'd still suggest to do `match Shostak.Adt.embed r with Constr { c_args = []; _ } -> ...` in `aux` and get rid of `is_tightenable` / `is_enum_cons` entirely.

tests/adts/distinct_same_constructor.smt2
Halbaroth: Ok, this test is wrong. I want to write
```smt2
(set-logic ALL)

(declare-datatype list
  ((cons (value Int) (tail list)) (nil)))

(declare-fun l () list)
(assert (distinct l (cons 0 nil)))
(assert (distinct l nil))
(check-sat)
```

src/lib/reasoners/adt_rel.ml
Halbaroth: Actually this implementation was buggy :p  but thanks to the `distinct_same_constructor` I caught the bug. I removed the constructor from the singleton domain! 

```ocaml 
let assume_distinct =
  let aux ~ex r1 r2 env =
    let d1 = Domains.get r1 env.domains in
    let d2 = Domains.get r2 env.domains in
    match Domain.as_singleton d1 with
    | Some (c, ex') when is_enum_cons (X.type_info r1) c ->
      let ex = Ex.union ex ex' in
      let nd = Domain.remove ~ex c d2 in
      tighten_domain r2 nd env
    | Some _ | None ->
      (* If `d1` is a singleton domain but its constructor has a payload,
         we cannot tighten the domain of `r2` because they could have
         distinct payloads. *)
      env
  in fun ~ex r1 r2 env ->
    aux ~ex r1 r2 env |> aux ~ex r2 r1
```

tests/adts/distinct_same_constructor.smt2
Halbaroth: I'm confused. It was correct but this form is clearer.

src/lib/reasoners/adt_rel.ml
Halbaroth: I think I used `List.rev_append` because this function is tail-rec. Since `5.1` it doesn't matter as `List.append` is TRMC. Let's keep the older order (funny wording :p).

bclement-ocp: The older was: `List.rev_append assume result.assume`. I am OK with using `rev_append` here; my point is that we should recurse on `assume` — not on `result.assume`.

(But I was confused and forgot that `result.assume` just came from `Rel_utils.delayed` and so the order does not actually matter)

PR #1089, opened 2024-04-04
Fully disable heavy assertions

Halbaroth:
In some part of the codebase, we use heavy assertions as follow:
```
assert (not @@ Options.get_enable_assertions () || heavy_assertion)
```
This trick has two advantages:
- As the logical `or` in OCaml has short-circuit evaluation, if the option `enable_assertions` is not `true`, we don't evaluate the heavy assertion.
- If we use the compiler option `no-asserts`, we don't even test the option `Options.get_enable_assertions`.

This PR adds a new function in `Options` to reproduce this good design everywhere in the codebase.

Notice that I put the function `heavy_assert` in `Options` but it would be better to put it in `Util`. Alas, we got a cycle dependency :/

### DIFF
--- a/src/lib/reasoners/arith.ml
+++ b/src/lib/reasoners/arith.ml
@@ -711,7 +711,7 @@ module Shostak
       (*
         This assert is not TRUE because of AC and distributivity of '*'
         Options.heavy_assert
-          (lazy (X.equal (apply_subst a sbs) (apply_subst b sbs)));
+          (fun () -> X.equal (apply_subst a sbs) (apply_subst b sbs));
       *)
     List.iter
       (fun (p, _) ->
--- a/src/lib/reasoners/arith.ml
+++ b/src/lib/reasoners/arith.ml
@@ -710,8 +710,8 @@ module Shostak
     let sbs = List.filter (fun (p,_) -> SX.mem p lvs || is_non_lin p) sbs in
       (*
         This assert is not TRUE because of AC and distributivity of '*'
-        assert (not (get_enable_assertions ()) ||
-        X.equal (apply_subst a sbs) (apply_subst b sbs));
+        Options.heavy_assert
+          (lazy (X.equal (apply_subst a sbs) (apply_subst b sbs)));
       *)
     List.iter
       (fun (p, _) ->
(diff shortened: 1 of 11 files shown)


bclement-ocp:
I'm OK with the idea (although my personal opinion is that there isn't much value in using `no-assert`), but I don't understand what `Lazy.t` brings to the equation? We don't share `Lazy.t` objects so I fail to see what it brings compared to `unit -> 'a` functions (I believe `Lazy.force` requires a synchronization point in OCaml 5).

>     * If we use the compiler option `no-asserts`, we don't even test the option `Options.get_enable_assertions`.

I believe (but did not check) that even without flambda, `if Options.get_enable_assertions () then () else ()` should be eliminated (if `get_enable_assertions` is inlined) anyways.

Halbaroth:
> I believe (but did not check) that even without flambda, `if Options.get_enable_assertions () then () else ()` should be eliminated (if `get_enable_assertions` is inlined) anyways.

Uhm maybe. In this case `If Options.get_enable_assertions () then heavy_assert;` is more readable than `assert (not ... || ...)`.

I used a lazy value to reproduce the short circuit behavior of the logic operator `or`. If you write `heavy_assert (your_condition)`, we can imagine that `your_condition` will be evaluate before calling `heavy_assert`.

EDIT: ok, I just want a suspension and I don't care to memoize the value. I just forgot `lazy` is more than lazy evaluation but demand evaluation.

bclement-ocp:
> I used a lazy value to reproduce the short circuit behavior of the logic operator `or`. If you write `heavy_assert (your_condition)`, we can imagine that `your_condition` will be evaluate before calling `heavy_assert`.

I meant that we can write `heavy_assert (fun () -> your_condition)`

Halbaroth:
I asked some gurus about the simplification of `if cond then () else ()`. It seems that `flambda 1` doesn't simplify it, but `flambda 2` does. 

Besides, our condition is actually a call to an impure function but we only read a reference and `flambda 2` can simplify this case too.

bclement-ocp:
> I asked some gurus about the simplification of `if cond then () else ()`. It seems that `flambda 1` doesn't simplify it, but `flambda 2` does.

Seems I keep forgetting how much OCaml *doesn't* optimize stuff (:

bclement-ocp (approved):

PR #1093, opened 2024-04-10
feat(ADT): model generation

Halbaroth:
This PR implements the model generation for ADT. The model generation is
done by the casesplit mechanism in `Adt_rel`.

- If `--enable-adts-cs` is present, we do casesplits by choosing a
  delayed destructor as before after asserting.

- If there is no more delayed constructors, we look for a domain which
  contains a tightenable constructor (that is a constructor without
  payload) and we choose the domain as small as possible as it does in
  the `Enum` theory.

- If we are in model generation, we can also choose constructors with
  payload (but we still choose first constructors without payload).

- The termination of the model generation is a bit tricky in the case of
  mutually recursive ADT. Please see the tests added for some
  complicated examples. I hope that I caught all the corner cases.
  To ensure the termination, the basic idea is to sort ADT's
  constructors in the module `Ty` during the parsing and to use the fact
  that the SMT-LIB standard only accepts well-founded ADT.

  We choose constructors in domains with the following order:
  - Constructor with the less destructors using the same nest;
  - Constructor with the less destructors using another nest of the same
    mutually recursive declaration;

This PR is rebased on #1087.


Halbaroth:
**EDIT:** this comment is outdated. I completely rewrote this algorithm.

The algorithm I used to ensure the termination of the model generation isn't sufficient. Last Friday, I found very simple examples for which the generation loops infinitely. For instance:
```smt2
(set-logic ALL)
(declare-datatypes ((u 0) (v 0) (w 0)) (
  (
    (c1 (d1 v))
    (c2 (d2 u)))
  (
    (c3 (d3 u))
    (c4 (d4 w)))
  (
    (c5 (d5 Int)))
))
(declare-const a u)
(assert ((_ is c1) a))
(check-sat)
(get-model)
```
In fact we cannot find an appropriate order with the information we got in the module `Ty`. We need to pass all the nests at once during the parsing in order to perform a topological order on the nests:
```ocaml 
type case = string * (string * t) list
type adt = string * case list * t list

let find_map (type c) =
  let exception Found of c
  in fun (p : 'a -> 'b -> c option) (h : ('a, 'b) Hashtbl.t) ->
    try
      Hashtbl.iter
        (fun k v ->
           match p k v with Some c -> raise @@ Found c | None -> ()
        ) h;
      raise Not_found
    with Found c -> c

let count_in_set (set : (string, case list * t list) Hashtbl.t) c =
  Lists.sum
    (fun (_, ty) ->
       match ty with
       | Tadt (name, _) when Hashtbl.mem set (Hstring.view name) -> 1
       | _ -> 0
    ) (snd c)

let cons_weight (set : (string, case list * t list) Hashtbl.t) c1 c2 =
  let c = count_in_set set c1 - count_in_set set c2 in
  if c <> 0 then c
  else
    List.compare_lengths (snd c1) (snd c2)

let topological_sort (adts : adt list) =
  let seens : adt list ref = ref [] in
  let set : (string, case list * t list) Hashtbl.t = Hashtbl.create 17 in
  List.iter (fun (name, cases, params) -> Hashtbl.add set name (cases, params)) adts;
  while Hashtbl.length set > 0 do
    let adt, (cases, params) =
      find_map (fun adt (cases, params) ->
          let b =
            List.exists (fun (_ , case) ->
                List.for_all (fun (_, ty) ->
                    match ty with
                    | Tadt (n, _) ->
                      not @@ Hashtbl.mem set (Hstring.view n)
                    | _ -> true
                  ) case
              ) cases
          in
          if b then Some (adt, (cases, params)) else None
        ) set
    in
    let cases = List.stable_sort (cons_weight set) cases in
    Hashtbl.remove set adt;
    seens := (adt, cases, params) :: !seens
  done;
  List.rev !seens

let declare_adts adts =
  let adts = topological_sort adts in
  List.iter (fun (name, cases, params) ->
      let name = Hstring.make name in
      let cases =
        List.map (fun (c, destrs) ->
          let destrs =
            List.map (fun (d, dty) -> Hstring.make d, dty) destrs
          in
          { constr = Hstring.make c; destrs }
        ) cases
      in
      Decls.add name params (Adt cases)
    ) adts
```  

It seems to be okay. I have still some cleanup to perform before submitting the PR for reviews. 

I think that we also have to produce unique identifiers instead of `Hstring` for constructors. As I explained to @bclement-ocp last week, I believed we could trigger a bug with some push/pop and ADT's declarations sharing constructor names.
This bug isn't easy to trigger because after a `pop` statement, the `Hstrings` of the constructors of a dead ADT are dead values for OCaml's runtime and the GC may free them before the declaration of a new ADT using these constructor names.

src/lib/frontend/d_cnf.ml
```diff
@@ -729,7 +728,7 @@ let mk_mr_ty_decls (tdl: DE.ty_cst list) =
             ) cases ([], true)
           in
           let uid = Uid.of_dolmen ty_c in
-          if is_enum
+          if is_enum && not contains_adts
```
Halbaroth: Before this PR, we sent enum types in mutually recursive definitions to the `Sum` theory. We cannot do that anymore because we need to send wellf-founded nest to the ADT theory. I didn't write a comment here because this code will disappear after merging `Sum` theory into `ADT` theory.

src/lib/reasoners/adt_rel.ml
```diff
@@ -568,43 +609,112 @@ exception Found of X.r * Uid.t
...
            ()
       ) env.delayed;
     None
   with Found (r, d) -> Some (r, d)
 
+let pick_tightenable_domain ~for_model uf =
+  let ds = Uf.(GlobalDomains.find (module Domains) @@ domains uf) in
+  let r = Domains.fold_enums
+      (fun r c best ->
+         let rr, _ = Uf.find_r uf r in
```
Halbaroth, line 634: I'm not sure this line is still necessary after #1119.

src/lib/reasoners/adt_rel.ml
```diff
@@ -568,43 +609,112 @@ exception Found of X.r * Uid.t
...
+         | _ -> Some (cd, r, c)
+      ) ds None
+  in
+  match r with
+  | Some _ as r -> r
+  | None ->
+    if for_model then
+      Domains.fold
+        (fun r d best ->
+           let rr, _ = Uf.find_r uf r in
```
Halbaroth, line 649: Same here.

src/lib/reasoners/adt_rel.ml
```diff
@@ -568,43 +609,112 @@ exception Found of X.r * Uid.t
...
       (fun r sy _e ->
          match sy with
-         | Sy.Destruct d ->
-           raise_notrace @@ Found (r, d)
+         | Sy.Destruct destr ->
+           let d = Domains.get r ds in
+           if Domain.cardinal d > 1 then
+             raise_notrace @@ Found (r, destr)
+           else
+             ()
```
Halbaroth, line 623: I'm not sure it's necessary after #1119. I will check.

src/lib/reasoners/adt_rel.ml
```diff
@@ -566,45 +608,106 @@ let constr_of_destr ty d =
...
+    with Found (r, d) ->
+      let c = constr_of_destr (X.type_info r) d in
+      Some (LR.mkv_builtin false (Sy.IsConstr c) [r])
+
+(* Pick a enum constructor in a tracked domain with minimal cardinal.
+   Returns [None] if there is no such constructor. *)
+let pick_enum ds uf =
+  Domains.fold_enums
+    (fun r c best ->
+       let rr, _ = Uf.find_r uf r in
```
Halbaroth: I'm not sure this line is still necessary after https://github.com/OCamlPro/alt-ergo/pull/1119.

src/lib/reasoners/adt_rel.ml
```diff
@@ -566,45 +608,106 @@ let constr_of_destr ty d =
...
+       | _ -> Some (cd, r, c)
+    ) ds None
+
+(* Pick a constructor in a tracked domain with minimal cardinal.
+   Returns [None] if there is no such constructor. *)
+let pick_best ~for_model ds uf =
+  if for_model then
+    Domains.fold
+      (fun r d best ->
+         let rr, _ = Uf.find_r uf r in
```
Halbaroth: Same here

src/lib/reasoners/adt_rel.ml
```diff
@@ -566,45 +608,106 @@ let constr_of_destr ty d =
...
+    None
+  else
+    let ds = Uf.(GlobalDomains.find (module Domains) @@ domains uf) in
+    try
+      Rel_utils.Delayed.iter_delayed
+        (fun r sy _e ->
+           match sy with
+           | Sy.Destruct destr ->
+             let d = Domains.get r ds in
+             if Domain.cardinal d > 1 then
```
Halbaroth: I'm not sure this line is still necessary after https://github.com/OCamlPro/alt-ergo/pull/1119. I will check.

Halbaroth:
Finally, this PR is ready for a first review! I rewrote the algorithm for sorting constructors, see the new module `Nest`.

As I explained offline, it's not easy to avoid a global hash table to store the total order of constructors with the current state of the codebase. I plan to write a better code after removing the legacy frontend.

The PR have been tested on `ae-format`, `UFDT` and `QF_DT` and doesn't introduce regression or slow down the solver.

Halbaroth:
There is a bug in the module `Nest`. I will investigate later.

Halbaroth:
I fixed the bug.

src/lib/util/util.mli
```diff
@@ -131,3 +131,6 @@ val print_list_pp:
   Format.formatter -> 'a list -> unit
 
 val internal_error : ('a, Format.formatter, unit, 'b) format4 -> 'a
+
+val (<?>) : ('a -> 'b option) -> ('a -> 'b option) -> 'a -> 'b option
+(** [(f <?> g) x] returns [f x] if [f x] succeeded, [g x] otherwise. *)
```
bclement-ocp: I am very much against introducing non-standard """handy""" combinators such as these, because they make the code harder to read for everyone except their author.

(Things are different when this is done locally as a small DSL, because usage and definition are in close proximity and the local nature of the syntax is explicit.)

src/lib/util/nest.mli
```diff
@@ -0,0 +1,60 @@
...
+(** [add_nest defs] sorts and adds the nest [defs] in the current
+    internal order.
+
+    {b Note:} Assume that [defs] is a complete nest (in any order). *)
+
+val compare : Uid.t -> Uid.t -> int
+(** [compare id1 id2] compares the constructors [id1] and [id2] with
+    the order generated by [add_nest] if they are both Dolmen identifiers.
+
+    @raise Failure if the nests of the constructors [id1] and [id2] haven't
```
bclement-ocp:
```suggestion
    @raise Failure if the nests of the constructors [id1] and [id2] have not
```

(Rule of thumb — contracted forms are almost always wrong in formal or semi-formal writing such as documentation)

src/lib/util/nest.mli
```diff
@@ -0,0 +1,60 @@
...
+                   been added with [add_nest] before. *)
+
+val reinit : unit -> unit
+(** [reinit ()] resets the internal order. *)
```
bclement-ocp: Does this also remove the nests for the purpose of `compare` raising `Failure`?

src/lib/util/nest.ml
```diff
@@ -0,0 +1,205 @@
...
+(*                                                                        *)
+(*     More details can be found in the directory licenses/               *)
+(*                                                                        *)
+(**************************************************************************)
+
+module DStd = Dolmen.Std
+module DE = DStd.Expr
+module DT = DE.Ty
+module B = Dolmen.Std.Builtin
+open DE
```
bclement-ocp: Please avoid global `open`s when possible, it makes code harder to follow.

src/lib/util/nest.ml
```diff
@@ -0,0 +1,205 @@
...
+  (* find_weight *)
+  (fun cstr ->
+     try
+       H.find order cstr
+     with Not_found ->
+       Fmt.failwith "cannot find uid %a" Uid.pp cstr),
+  (* reinit *)
+  (fun () ->
+     ctr := 0;
+     H.clear order)
```
bclement-ocp:
```suggestion
let add_cstr, find_weight, reinit =
  let ctr = ref 0 in
  let order : int H.t = H.create 100 in
  let add_weight cstr =
     H.add order cstr !ctr;
     incr ctr
  and find_weight cstr =
     try
       H.find order cstr
     with Not_found ->
       Fmt.failwith "cannot find uid %a" Uid.pp cstr
  and reinit () =
     ctr := 0;
     H.clear order
  in add_cstr, find_weight, reinit
```

src/lib/util/nest.ml
```diff
@@ -0,0 +1,205 @@
...
+
+(* Sort the constructors of the nest using a sorting based on the
+   Kahn's algorithm. *)
```
bclement-ocp:
```suggestion
(* Sort the constructors of the nest using a sorting based on
   Kahn's algorithm. *)
```

"The Kahn's algorithm" translates to "l'algorithme du Kahn" (the possessive "Kahn's" already makes "algorithm" determined — "the" before a possessive applies to the first noun, e.g. "the King's algorithm").

src/lib/util/heap.ml
```diff
@@ -148,7 +122,7 @@ module Make(Rank : RankedType) = struct
   let[@inline] grow_to_by_double { heap } sz =
     Vec.grow_to_by_double heap sz
 
-  let remove_min ({ heap } as s) =
+  let pop_minimum ({ heap } as s) =
```
bclement-ocp: Why does this needs to be changed?

src/lib/reasoners/uf.ml
```diff
@@ -1179,9 +1183,9 @@ let compute_concrete_model_of_val cache =
   and get_abstract_for = Cache.get_abstract_for cache.abstracts
   in fun env t ((mdl, mrepr) as acc) ->
     let { E.f; xs; ty; _ } = E.term_view t in
-    if X.is_solvable_theory_symbol f ty
-    || Sy.is_internal f || E.is_internal_name t || E.is_internal_skolem t
-    || E.equal t E.vrai || E.equal t E.faux
+    if X.is_solvable_theory_symbol f ty || is_destructor f
```
bclement-ocp, line 1190: Why do we need this additional test? Shouldn't `Destructor` be a solvable theory symbol?

src/lib/reasoners/adt.ml
```diff
@@ -401,12 +401,17 @@ module Shostak (X : ALIEN) = struct
...
+    None
+
+  let to_model_term r =
+    match embed r with
+    | Constr { c_name; c_ty; c_args } ->
+      let args = Lists.try_map (fun (_, arg) -> X.to_model_term arg) c_args in
+      Option.bind args @@ fun args ->
+      Some (E.mk_constr c_name args c_ty)
+
+    | Select _ -> None
```
bclement-ocp, line 415: Is this actually called? I think we shouldn't have `Select` remaining once we are in model generation — if that is correct this should be either an error or at least a warning.

src/lib/frontend/d_cnf.ml
bclement-ocp: Not sure I understand — we can remove `enum` types from the nest and treat them as external types, no? enum types are always well-founded.

src/lib/reasoners/adt_rel.ml
```diff
@@ -566,45 +584,81 @@ let constr_of_destr ty d =
...
-           raise_notrace @@ Found (r, d)
-         | _ ->
-           ()
-      ) env.delayed;
-    None
-  with Found (r, d) -> Some (r, d)
+let (let*) = Option.bind
 
-(* Do a case-split by choosing a semantic value [r] and constructor [c]
+(* Do a casesplit by choosing a semantic value [r] and constructor [c]
```
bclement-ocp: "casesplit" is not a word, please revert this change.

src/lib/reasoners/adt_rel.ml
bclement-ocp: Is it still necessary?

src/lib/reasoners/adt_rel.ml
bclement-ocp: I don't think it is.

src/lib/reasoners/adt_rel.ml
```diff
@@ -566,45 +584,81 @@ let constr_of_destr ty d =
...
+    let ds = Uf.(GlobalDomains.find (module Domains) @@ domains uf) in
+    let* _, r, c = pick_best ds uf in
+    let _, cons = Option.get @@ build_constr_eq r c in
+    let nr, ctx = X.make cons in
+    assert (Lists.is_empty ctx);
+    Some (LR.mkv_eq r nr)
+
+let next_casesplit ~for_model env =
+  let open Util in
+  (split_delayed_destructor env) <?> (split_best_domain ~for_model)
```
bclement-ocp: This seems to be the only use of the `<?>` combinator; it can easily be expanded which will result in clearer code (if only because the `uf` argument will no longer be implicit).

src/lib/reasoners/adt_rel.ml
```diff
@@ -566,45 +584,81 @@ let constr_of_destr ty d =
...
+         let c = Domain.choose d in
+         Some (cd, r, c)
+    ) ds None
+
+let split_best_domain ~for_model uf =
+  if not for_model then
+    None
+  else
+    let ds = Uf.(GlobalDomains.find (module Domains) @@ domains uf) in
+    let* _, r, c = pick_best ds uf in
```
bclement-ocp: The first argument `cd` should be removed from the tuple, since it is always ignored.

src/lib/reasoners/adt_rel.ml
```diff
@@ -566,45 +584,81 @@ let constr_of_destr ty d =
...
+    let _, cons = Option.get @@ build_constr_eq r c in
+    let nr, ctx = X.make cons in
+    assert (Lists.is_empty ctx);
```
bclement-ocp, line 637: It is not immediately clear why this `assert` would be true, so it should have a short explanatory comment.

bclement-ocp (commented):
Before I dig into the `Nest` module: I had left the previous discussion on the topic with the impression that we were doing a topological order on *type constructors* (always taking a term constructor during model generation for which the maximum of their arguments is minimal), not on term constructors.

I think you might have explained why that is not sufficient when this work started, but I can't remember the reasoning — can you explain that again? It seems to me that sorting on type constructors works better wrt polymorphism. Or is that a design choice?

src/lib/util/nest.ml
Halbaroth: Oups, you taught me that again

src/lib/reasoners/adt.ml
Halbaroth: It shouldn't happen during the model generation. Notice that we raise an assertion in `Uf` here:
```ocaml
    match X.to_model_term rep with
    | Some v -> v, ME.add t v mrepr
    | None ->
      (* [X.to_model_term] cannot fail on constant semantic values. *)
      assert false
```
`X.to_model_term` is supposed to return `None` on nonconstant semantic values (for the predicate `X.is_constant`).

src/lib/reasoners/adt_rel.ml
Halbaroth: I copied the explanation for the same assert in `build_constr_eq`.

src/lib/util/heap.ml
Halbaroth: Just to make uniform the signature of functors `MakeOrdered` and `MakeRanked`. I can revert the modification or rename the `pop_minimum` operation of `MakeOrdered` to `remove_min`.

src/lib/util/nest.ml
Halbaroth: Indeed, it's better :)

src/lib/util/heap.ml
bclement-ocp: I think we should go for either `remove_min` or `pop_min` as those are the names that are likely to end up [in the Stdlib](https://github.com/ocaml/ocaml/pull/12871).

src/lib/reasoners/adt_rel.ml
Halbaroth: I'm not sure. I remembered that the casesplit mechanism can loop forever without this test after the cleanup of `Adt_rel`. As this fix isn't necessary for model generation, I removed it and I will test `--enable-adts-cs` later.

src/lib/util/heap.ml
Halbaroth: Ok, let's go for `pop_min`.

src/lib/util/nest.ml
```diff
@@ -0,0 +1,205 @@
...
+                id = cstr;
+                outgoing = Hashtbl.find map def;
+                in_degree = -1; (* dummy value *)
+                weight = Array.length dstrs
+              }
+              in
+              let in_degree =
+                Array.fold_left
+                  (fun acc2 dstr ->
+                     let d = let* dstr = dstr in def_of_dstr dstr in
```
bclement-ocp:
```suggestion
                     let d = Option.bind dstr def_of_dstr in
```

bclement-ocp (commented):
Waiting for an updated version to give formal approval; the Nest module looks OK. As discussed offline, if we can adapt the code to store priorities for types rather than constructors I think that would be better (especially w.r.t. polymorphism) but I would be OK with merging with the current algo.

src/lib/util/nest.ml
Halbaroth: At the origin it was a let-punning but the sugar syntax has been introduced in `4.13`. I will apply your suggestion.

Halbaroth:
> Waiting for an updated version to give formal approval; the Nest module looks OK. As discussed offline, if we can adapt the code to store priorities for types rather than constructors I think that would be better (especially w.r.t. polymorphism) but I would be OK with merging with the current algo.

I think I prefer to merge this early version and apply your suggestion in a separated PR. I will open an issue with example where the current algorithm fails to terminate.

src/lib/reasoners/uf.ml
Halbaroth: I'm not absolutely sure it's okay to declare `Destructor` as a solvable theory symbol. I'm preparing another PR in which I clean and review the current code of the module `Adt`, so I plan to check that during this PR.

src/lib/reasoners/adt.ml
Halbaroth: Is it ok for you? I can update the documentation of `X.to_model_term` if you want.

src/lib/reasoners/uf.ml
bclement-ocp: Makes sense; this would be worth having a TODO for.

src/lib/reasoners/adt.ml
bclement-ocp: OK, I did not have in mind that `to_model_term` should be callable on non-constant terms.

src/lib/reasoners/uf.ml
Halbaroth: Sure

Halbaroth:
@bclement-ocp is it good for you?

bclement-ocp (approved):
I was not sure if this was ready for a second pass or not. There is still a question on the `not contains_adt` case that I don't really understand, and I think the headers for the `Nest` module are wrong, but other than that LGTM.

src/lib/frontend/d_cnf.ml
Halbaroth: I think it's okay to remove it so I did it.

src/lib/util/nest.mli
Halbaroth: @bclement-ocp  I wrote this function for Alt-Ergo-Fuzz. The `Failure` in `compare` is there because we expect that the module Nest sorts all the constructors of a nest. If we use this function without declaring again the nest, it will raise the failure exception. I think we should remove this function completely.

bclement-ocp: I am fine with keeping this function — but the doc should clarify something along the line of "This clears all registered constructors." because it is not clear what "resets the internal order" means.

PR #1094, opened 2024-04-13
Merge `Enum` theory into `ADT` theory

Halbaroth:
After refactoring both `Enum` and `ADT` theories, they shared most of
their implementation.

This PR merges `Enum` theory into `ADT` ones.

This PR is rebased on #1093


Halbaroth:
This PR have been tested on our internal benchmarks.

src/lib/reasoners/adt_rel.ml
```diff
@@ -147,14 +150,20 @@ module Domains = struct
...
 
-  let empty = { domains = MX.empty; changed = SX.empty }
+  let empty = { domains = MX.empty; enums = SX.empty; changed = SX.empty }
 
   let filter_ty = is_adt_ty
 
+  let is_enum r =
+    match X.type_info r with
+    | Ty.Tadt (_, [], true) -> true
+    | _ -> false
```
bclement-ocp: It looks like this is the only place where we care about the new `is_enum` flag of the `Tadt` constructor. I understand it is probably here for performance reasons (seeing #1138 and previous offline discussions), but ideally we would like to remove it if possible (I think a time regression is OK for this, but not a number-of-solved-problems regression).

Did you try splitting only if *all* the current constructors have no payload rather than splitting on any constructor that has no payload?

src/lib/reasoners/adt_rel.ml
```diff
@@ -622,25 +639,58 @@ let pick_best ds =
...
+    let _, cons = Option.get @@ build_constr_eq r c in
+    let nr, ctx = X.make cons in
+    (* In the current implementation of `X.make`, we produce
+       a nonempty context only for interpreted semantic values
+       of the `Arith` and `Records` theories. The semantic
+       values `cons` never involves such values. *)
+    assert (Lists.is_empty ctx);
+    Some (LR.mkv_eq r nr)
+  else
+    None
```
bclement-ocp: I think there is some duplication with `split_best_domain` that can be eliminated by defining a `make_split n r c` function that makes the split, and a `pick` function that dispatches to either `pick_enum` or `pick_best` as appropriate.

src/lib/structures/ty.mli
```diff
@@ -48,20 +48,19 @@ type t =
...
+
   | Tfarray of t * t
   (** Functional arrays. [TFarray (src,dst)] maps values of type [src]
       to values of type [dst]. *)
-  | Tsum of Uid.t * Uid.t list
-  (** Enumeration, with its name, and the list of its constructors. *)
 
-  | Tadt of Uid.t * t list
-  (** Application of algebraic data types. [Tadt (a, params)] denotes
+  | Tadt of Uid.t * t list * bool
```
bclement-ocp: Unnamed booleans should generally be avoided; if we keep this flag, we should either `adt_kind = Adt | Enum` and use that instead of `bool`, use `[` Adt | `Enum ]`, or define `Tadt` with an inline record with an `is_enum : bool` field.

src/lib/frontend/typechecker.ml
```diff
@@ -171,6 +164,10 @@ module Types = struct
             from_labels =
               List.fold_left
                 (fun fl (l,_) -> MString.add l id fl) env.from_labels lbs }
+    | Enum l ->
+      let body = List.map (fun constr -> Uid.of_string constr, []) l in
+      let ty = Ty.t_adt ~enum:true ~body:(Some body) (Uid.of_string id) [] in
+      ty, { env with to_ty = MString.add id ty env.to_ty }
```
bclement-ocp: Not sure why this was moved; the check for `PolymorphicEnum` error should be kept (we do bad things when type variables are unused and this is a case where we can catch it).

bclement-ocp (commented):
Can you please change the name of the PR? "Merge enum" is very obscure; "Merge the enum and ADT theories" or "Merge the enum theory into the ADT theory" would be better.

> This PR have been tested on our internal benchmarks.

What are the results of those tests? Is it +0-0?

As far as the code goes, this looks mostly OK to me. Outside of some small suggestions, my main concern is the new boolean flag in the `Tadt` constructor. I know you did some things to try to get rid of it — can you say a bit more about what other approaches you tried and the issues with them? I want to make sure we don't miss an easy way of not needing this flag.

src/lib/frontend/typechecker.ml
Halbaroth: It's an artifact of the last rebase. I will fix that.

src/lib/structures/ty.mli
Halbaroth: Sure. I didn't make much effort here as we plan to get rid of the legacy types soon. I will create an extra ADT.

Halbaroth:
> What are the results of those tests? Is it +0-0?

It's +0-0 on `ae-format`, `QF_DT` and `UFDT`. In fact, the enum theory is slightly faster after this PR.

Halbaroth:
> As far as the code goes, this looks mostly OK to me. Outside of some small suggestions, my main concern is the new boolean flag in the `Tadt` constructor. I know you did some things to try to get rid of it — can you say a bit more about what other approaches you tried and the issues with them? I want to make sure we don't miss an easy way of not needing this flag.

I only tried one other approach, which is currently in the draft #1138. See the diff between these branches:
https://github.com/Halbaroth/alt-ergo/compare/merge-enum...Halbaroth:alt-ergo:casesplit-enum-cstr

Instead of splitting on enum types only after each assert, I split on domains that contains a constructor without payload. I got 4 regressions (with OCaml 5) and 0 improvements and the enum theory is slower (14m30s before and 15m28s). These tests have been run on `next` and PR #1138 with only the `UFDT` test suite.

src/lib/reasoners/adt_rel.ml
Halbaroth: It's not just a time regression, we lost some problems. As we decided offline, it's better to do a PR for merging `Enum` into `ADT` without any regression, then testing better approaches later. My only concern with this extra flag is the fact there is no equivalent flag in Dolmen's types, so we have to get rid of it before switching to these types.

A solution without flag consists in checking if the type is an enum in `Adt_rel.Domains`. The cost should be acceptable (we could cache the result in a weak cache, but there are too much weak caches in AE in my opinion).

> Did you try splitting only if _all_ the current constructors have no payload rather than splitting on any constructor that has no payload?

Not yet, I will try today.

bclement-ocp: If we want to cache it, I think that this can be easily added to the `Ty` module that already has a cache (for the definition of a type) — we can add it to `Ty.Adt` which is better than adding it to `Ty.t` because it is part of the type definition, not of the type name.

Halbaroth:
I will test your suggestion in another PR. I think we can merge this one. I removed the extra flag as you wish.

bclement-ocp (approved):
This looks cleaner. I'm OK with trying to change the split strategy separately.

bclement-ocp (changes requested):
Temporarily revoking approval to avoid accidental merging since it seems a regression snuck in the benchmarks.

Halbaroth:
There is no regression. I forgot to rebase the PR after merging the fix for #1099. I ran again benchmarks on #1138 with the fix and there is no regression anymore.

bclement-ocp (approved):

bclement-ocp:
> I forgot to rebase the PR after merging the fix for https://github.com/OCamlPro/alt-ergo/issues/1099.

Does this mean that it is that fix that introduced regressions?

Halbaroth:
> > I forgot to rebase the PR after merging the fix for #1099.
> 
> Does this mean that it is that fix that introduced regressions?

No. The fix solved regressions introduced with the Dolmen frontend in `2.5.0` and I didn't rebase this PR with it. So `next` solved new problems.

bclement-ocp:
Ah, I got things backwards — thanks for the clarification.

PR #1095, opened 2024-04-19
Remove the `Record` theory

### LINKED ISSUE #1008: Simple ADT test with different behavior with or without push
The following test have a different behavior given if the `push` instruction is here or not:

File `test_adt.smt2`
```
(set-logic ALL)
(declare-datatype t ((A) (B) (C (i Int))))

(declare-const e t)

(push 1)
(assert (not ((_ is A) e)))
(assert (not ((_ is B) e)))
(assert (not (exists ((n Int)) (= e (C n)))))
(check-sat)
```

Command: 
`alt-ergo test_adt.smt2 --enable-assertions --output smtlib2 --sat-solver CDCL --no-minimal-bj`

Without the push, alt-ergo returns `unsat` (as expected); with the push, it never returns. 

The culplrit may be the presence of the existential quantifier and `--no-minimal-bj`; remove the option and everything works fine. This test is a translation of goal `g_valid_5_1` of `tests/adts/simple_1.ae`.

Halbaroth:
This PR removes the Record theory and sends all the records to the ADT theory. 

This modification does not fix #1008 as expected because SatML does not send some terms generated by `Adt_rel` to the instantiation engine.


Halbaroth:
As I expected, we got regressions if we do not produce a context in the `make` function of `Adt` as we did in `Records`. Consider for instance this example (I got it from a real regression):
```smt2 
(set-logic ALL)
(declare-datatype unit ((void)))
(declare-datatype t ((box (unbox unit))))
(declare-fun f (t t) Bool)
(assert (forall ((u t) (v t)) (f u (box (unbox v)))))
(assert (not (f (box void) (box void))))
(check-sat)
```
We expect that `Alt-Ergo` outputs `unsat`. It works on `next` because `X.make` adds the extra equality `(unbox (box void)) = void`. The trigger of the first assertion is `(f u (box (unbox v)))`. First, we match `(box void)` with `(box (unbox v))`, then we try to match `(unbox v)` with `void`. Of course, we cannot match them if we only consider syntactical equality, but `Matching` allows to match terms modulo equality of the union-find. `(unbox (box void))` is in the class of `void` and can match `(unbox v)` with `v --> (box void)`. 

If you modify the input file as follows:
```smt2
(set-logic ALL)
(declare-datatype unit ((void)))
(declare-datatype t ((box (unbox unit))))
(declare-fun f (t) Bool)
(assert (forall ((v t)) (f (box (unbox v)))))
(assert (not (f (box void))))
(check-sat)
``` 
Alt-Ergo outputs `unsat` on `next` and this branch. It succeeds because after many matching rounds, it will try with variable triggers and there is one variable trigger `v` which covers all the variables of the axiom. In the previous case, it won't generate it because of `u`. So the instantiation engine produces a fresh instantiation with `v = void`.

Notice that there is specific code in `Matching` to manage another corner case of the Records theory. For instance, consider the input: 
```smt2
(set-logic ALL)
(declare-datatype t ((c (b Bool))))
(declare-fun f (t) Bool)
(declare-const x t)
(assert (forall ((y Bool)) (f (c y))))
(assert (not (f x)))
(check-sat)
```
Alt-Ergo outputs `unsat` on `next` and this branch. On next, Alt-Ergo need to match `x` with `(c y)` (it is the pattern of the axiom). Again, we need to match modulo equality. But it does not know that `x = (c (b x))`. The function `xs_modulo_records` in the module `Matching` will add the term `(c (b x))` to the list of terms equal to `x` and Alt-Ergo produces the right instance. 

I removed this code in the present branch and did not write anything to replace it but still the test passes. It works for a good reason! First of all, the ADT `x` is sent to `Adt_rel` thanks to the new function `init` in `Uf`.
Since `t` is a record, we immediately obtain a domain of size 1 for `x` and we produce a fresh boolean term `k` and the equality `x = (c k)`. Of course `(c k)` is the new representative element and we can match it against the pattern of the axiom.

Halbaroth:
In the last patch, I use the context of `Adt.make` to propagate the same equalities as we did in `Records.make`. Actually I produce these equalities as soon as the type of the term is an ADT with only one constructor, which is bit more general. 

I am running benchmarks to check if this patch removes most of the regressions. Even if it does, I would prefer a solution without using the context of `X.make`.

Halbaroth:
+55-12 on ae-format and the solver is slightly faster :)

Halbaroth:
Okay, I was not completely satisfied by the last commit. It is sad to still use the context of `X.make`. 

Another way to solve this issue consists in eliminating as earlier as possible terms of the form `(box ... (unbox u) ...)` where `box` is the unique constructor of a record type and `unbox` is one of its destructor (at the appropriate position).  Let us imagine that we have the axiom:
```
forall u : t. (P u (box (unbox u)))
```
where `P` is a undefined predicate and `t` is our record type. 

If we replace it by:
```
forall w : unit. u = box w --> (P u u)
```

Alt-Ergo should generate the trigger `(P u u)` for this axiom. If we have the term `(box unit)` in the union-find, we can match `(P (box unit) (box unit))` against `(P u u)` with `u --> (box unit)` and we can take `w --> unit`. We do not need to add the equality `(unbox (box unit)) = unit`.

Actually, I tried another transformation:
```
forall w : unit. 
let u = box w in 
(P u u)
```

It works if we disable some simplifications done in the smart constructor `Expr.mk_let` and during triggers generation but we lost other problems.

Halbaroth:
I think that this PR is now ready to review. I keep the current code with context in `X.make` even if I dislike this design because I didn't find a proper approach and I think we should merge this PR and look for a better design later.

src/lib/reasoners/adt.ml
```diff
@@ -105,21 +105,20 @@ module Shostak (X : ALIEN) = struct
...
+        Fmt.pf ppf "%a" DE.Term.Const.print c_name
+      else
+        Fmt.pf ppf "(%a %a)"
+          DE.Term.Const.print c_name
+          Fmt.(box @@ list ~sep:sp @@ pair nop X.print) c_args
 
     | Select d ->
-      Fmt.pf ppf "%a#!!%a" X.print d.d_arg DE.Term.Const.print d.d_name
-
+      Fmt.pf ppf "(%a %a)" X.print d.d_arg DE.Term.Const.print d.d_name
```
bclement-ocp: This looks like it is an AE-style printer rather than an SMT-lib style printer, not sure it should be changed.

src/lib/reasoners/adt.ml
```diff
@@ -168,21 +167,18 @@ module Shostak (X : ALIEN) = struct
...
   let make t =
     assert (not @@ Options.get_disable_adts ());
-    if Options.get_debug_adt () then
-      Printer.print_dbg
-        ~module_name:"Adt" ~function_name:"make"
-        "make %a" E.print t;
+    Log.debug (fun k -> k "make %a" E.print t);
     let { E.f; xs; ty; _ } = E.term_view t in
-    let sx, ctx =
+    let ssx, ctx =
```
bclement-ocp: We should call this `rev_rs` and `rs` below to denote that they are semantic value -- `ssx` and `xxs` are not particularly readable names!

src/lib/reasoners/adt_rel.ml
```diff
@@ -469,12 +472,7 @@ let build_constr_eq r c =
...
         let cons = E.mk_constr c xs ty in
-        let r', ctx = X.make cons in
-        (* In the current implementation of `X.make`, we produce
-           a nonempty context only for interpreted semantic values
-           of the `Arith` and `Records` theories. The semantic
-           values `cons` never involves such values. *)
-        assert (Compat.List.is_empty ctx);
+        let r', _ctx = X.make cons in
         let eq = Shostak.L.(view @@ mk_eq r r') in
         Some (eq, E.mk_constr c xs ty)
```
bclement-ocp:
```suggestion
        Some (eq, cons)
```

src/lib/reasoners/adt_rel.ml
```diff
@@ -469,12 +472,7 @@ let build_constr_eq r c =
         in
         let xs = List.map (fun (_, ty) -> E.fresh_name ty) ds in
         let cons = E.mk_constr c xs ty in
-        let r', ctx = X.make cons in
-        (* In the current implementation of `X.make`, we produce
-           a nonempty context only for interpreted semantic values
-           of the `Arith` and `Records` theories. The semantic
-           values `cons` never involves such values. *)
-        assert (Compat.List.is_empty ctx);
```
bclement-ocp, line 477: We can't keep the assertion because now we create side conditions for ADTs (when they are records), but we shold keep a comment to explain why it is OK to ignore these side conditions.

src/lib/reasoners/adt_rel.ml
```diff
@@ -632,12 +630,7 @@ let split_domain ~for_model env uf =
   let* cd, r, c = pick_domain ~for_model uf in
   if for_model || can_split env (Numbers.Q.from_int cd) then
     let _, cons = Option.get @@ build_constr_eq r c in
-    let nr, ctx = X.make cons in
-    (* In the current implementation of `X.make`, we produce
-       a nonempty context only for interpreted semantic values
-       of the `Arith` and `Records` theories. The semantic
-       values `cons` never involves such values. *)
-    assert (Compat.List.is_empty ctx);
```
bclement-ocp, line 640: Same as above -- we should keep a comment why it is OK to ignore the second return value of `X.make`.

src/lib/reasoners/matching.ml
```diff
@@ -411,15 +405,9 @@ module Make (X : Arg) : S with type theory = X.t = struct
...
               (fun t l ->
-                 let { E.f = f; xs = xs; ty = ty; _ } = E.term_view t in
-                 if Symbols.compare f_pat f = 0 then xs::l
-                 else
-                   begin
-                     match f_pat, ty with
-                     | Symbols.Op (Symbols.Record), Ty.Trecord record ->
-                       (xs_modulo_records t record) :: l
-                     | _ -> l
-                   end
```
bclement-ocp, line 422: Why is this not needed for records as ADTs?

src/lib/structures/ty.ml
```diff
@@ -41,15 +41,6 @@ type t =
...
   | Tfarray of t * t
   | Tadt of DE.ty_cst * t list
-  | Trecord of trecord
-
-and trecord = {
-  mutable args : t list;
-  name : DE.ty_cst;
-  mutable lbs :  (DE.term_cst * t) list;
-  record_constr : DE.term_cst;
-  (* for ADTs that become records. default is "{" *)
```
bclement-ocp, line 51: Can't say I am sad to see this one go.

tools/gentest.ml
```diff
@@ -264,6 +264,8 @@ end = struct
                 {acc with accepted_exit_codes = [142]}
               | "default" ->
                 {acc with filters = Some ["default"]}
+              | "tableaux" ->
+                {acc with filters = Some ["tableaux"]}
```
bclement-ocp, line 268: I seem to remember the difference between Tableaux and CDCL went away when adding the field equations to the context — is it not the case?

If not, a comment should be added to the test to explain why it is failing with CDCL.

src/lib/reasoners/adt.ml
```diff
@@ -192,20 +188,27 @@ module Shostak (X : ALIEN) = struct
...
+      let ctx =
+        match cases with
+        | [{ destrs; _ }] ->
+          List.fold_left2
+            (fun ctx x (d, d_ty) ->
+               let access = E.mk_term (Sy.destruct d) [t] d_ty in
+               E.mk_eq ~iff:false x access :: ctx
+            ) ctx xs destrs
+        | _ -> ctx
+      in
```
bclement-ocp: Can you add a comment explaining what we are doing here?

bclement-ocp (changes requested):
Looks good overall, but there are a few things that seem non-obvious to me that I'd like to see explained.

src/lib/reasoners/adt.ml
Halbaroth: It is a debug message, so we can print whatever we want. In my opinion, we should avoid using different formats for debug messaged based on the input format. We could use a similar printer than Dolmen for these debug messages. This modification contradicts with my own opinion (because infix notations are more readable) ;)

src/lib/reasoners/matching.ml
Halbaroth: I previously explained in this PR why this piece of code is no longer necessary. I forgot about it too and investigated it again! I'm copying my explanation here with additional details:

The specific code in Matching is there to handle a corner case of the Record theory. For instance, consider the input file:
```smt2
(set-logic ALL)
(declare-datatype t ((c (b Bool))))
(declare-fun f (t) Bool)
(declare-const x t)
(assert (forall ((y Bool)) (f (c y))))
(assert (not (f x)))
(check-sat)
```
Alt-Ergo outputs unsat on next and on this branch. On next, Alt-Ergo needs to match `x` with  `(c y)`, as it is the pattern of the first axiom. We need to match modulo equality. But it does not know that `x = (c (b x))`. This equation is only generated for record construction terms and `x` is a variable. The function `xs_modulo_records` adds the term `(c (b x))` into the class of `x` and Alt-Ergo produces the right instance.

I removed this code in the present branch and did not write anything to replace it but still the above test succeeds. It works for a good reason! 
First of all, the ADT `x` is sent to `Adt_rel` thanks to the new function `init` in `Uf`.
Since `t` is a record type, we immediately obtain a domain of size 1 for `x` and we produce a fresh boolean term k and the equality `x = (c k)`. Of course `(c k)` is the new representative element and we can match it against the pattern of the axiom.

tools/gentest.ml
Halbaroth: Good catch. It is a regression. The test: 
```smt2 
(set-logic ALL)the 
(declare-datatype t ((B (i Int))))
(declare-const e t)
(assert ((_ is B) e))
(assert (forall ((n Int)) (distinct e (B n))))
(check-sat)
```
succeeds on next but fails on this branch. Note that this input file is very closed to issue #1295 but `(assert ((_ is B) e)` is a ground constraint on `e`. Currently, this branch manages to discover the equality `e = (B .k)` for a fresh integer `.k` but the new term `(B .k)` never reaches the matching environment. More precisely, the new term is discarded here: 
https://github.com/OCamlPro/alt-ergo/blob/837293ef8d4c55eafdcc152089866a24db4fca91/src/lib/reasoners/satml.ml#L1084-L1099

In tableaux, all the new terms generated by relations are sent to the matching environment. 
Actually, I tried to fix this issue months ago. See https://github.com/OCamlPro/alt-ergo/commit/cca32808d8f8348728214b9d25b4ea22ff6ecaa9. The benchmark was pretty bad (+25-195 after correction).

Halbaroth: The test is still ignored on CDCL but I put a link to the issue #1262

src/lib/reasoners/adt_rel.ml
Halbaroth: I put a link to the issue #1296

src/lib/reasoners/matching.ml
```diff
@@ -229,6 +229,22 @@ module Make (X : Arg) : S with type theory = X.t = struct
...
+            List.map
+              (fun (d, ty) ->
+                  E.mk_term Sy.(Op (Destruct d)) [t] ty
+              ) destrs
+          in
+          let c = E.mk_constr constr xs ty in
+          add_term info c env
+        | _ -> add_term info t env)
+     | _ -> add_term info t env
+  *)
```
bclement-ocp: Should be removed?

bclement-ocp (approved):
I am really uneasy about losing such simple proofs as the 1008 example — we really should figure out a way to send those terms back to matching.

Halbaroth:
> I am really uneasy about losing such simple proofs as the 1008 example — we really should figure out a way to send those terms back to matching.

I can work on it this week, no rush to merge this PR.

Halbaroth:
Ok, this PR is quite old... I thought `1008/record.smt2` was a regression but it wasn't! Before this PR, we cannot prove the test with both `Tableaux` and `CDCL`. After this PR, we can prove it with `Tableaux`. The test have been introduced in this PR https://github.com/OCamlPro/alt-ergo/pull/1211 and you suggested to add it in the issue https://github.com/OCamlPro/alt-ergo/issues/1008. In #1008, you noticed that we are able to solve:
```smt2 
(set-logic ALL)
(declare-datatype t ((A) (B (i Int))))

(declare-const e t)

(assert ((_ is B) e))
(assert (forall ((n Int)) (distinct e (B n))))
(check-sat)
``` 
which means that the issue with `1008/record.smt2` is not related with #1262 as I thought. I remove the link to #1262 and merge this PR as it does not introduce regressions.

PR #1097, opened 2024-04-23
Unique identifiers

Halbaroth:
The current module `Id` uses `Hstring` to compare identifiers. This PR uses a hidden tag in the type `t` of `Id` to compare identifiers. This tag is unique by construction and the `Hstring` field is only used for printing.

In particular, calling twice `Sy.name` with the same input will produce two differents symbols. The full id (that is with the unique tag stored in it) can be print with `Id.pp ~full:true`.

This PR has to be tested a lot because we have to be sure we never store a symbol in a string in order to recreate it later for instance.


Halbaroth:
It seems this modification is too complicated until we get rid of the legacy parser.

PR #1098, opened 2024-04-24
Use the Dolmen identifiers for constructors and ADT names

Halbaroth:
We don't manage correctly incremental type declarations. For instance the following input is accepted by the SMT-LIB standard:
```smt2 
(set-logic ALL)
(push 1)
(declare-datatype t
  ((c1 (d1 Int)) (c2)))
(pop 1)
(declare-datatype t
  ((c2 (d1 Int)) (c1)))
(declare-const x t)
(declare-const y t)
(assert ((_ is c1) x))
(assert ((_ is c1) y))
(assert (distinct x y))
(check-sat)
```
(Notice that we switch the constructors `c1` and `c2` in our second declaration of the type `t`). AE answers `unknown` on `next` but it answers `unsat` if we remove the first declaration of `t`.

This PR fixes the issue for the new frontend only. So we got `unsat` with `--frontend dolmen` but we still got `unknown` with `--frontend legacy`.

The solution consists in using the unique identifiers produced by `Dolmen` for the constructors, destructors and names of ADTs. To reduce as much as possible the amount of modification done in the legacy typechecker, I wrapped Dolmen's identifiers in a new module `Uid`. This module agrees with `Hstring` if we use the constructors `of_string` or `of_hstring` and it uses Dolmen's identifiers if we use the constructor `of_dolmen`. Of course, we'll get rid of this module as soon as we remove the legacy frontend. 

I didn't use a unique identifier for the constructor of builtin enumeration types. This case will be fixed after removing the legacy frontend.

This PR is supposed to be merged before #1093


src/lib/structures/uid.ml
```diff
@@ -0,0 +1,98 @@
...
+(*     More details can be found in the directory licenses/               *)
+(*                                                                        *)
+(**************************************************************************)
+
+module DStd = Dolmen.Std
+module DE = DStd.Expr
+
+type t =
+  | Fake of Hstring.t
+  | Unique of { name : Hstring.t; index : int }
```
bclement-ocp: Why use a separate `Unique` constructor with our own tag rather than directly use the Dolmen identifier?

src/lib/structures/ty.ml
```diff
@@ -535,59 +536,48 @@ let t_adt ?(body=None) s ty_vars =
...
-let trecord ?(sort_fields = false) ~record_constr lv n lbs =
-  let lbs = List.map (fun (l,ty) -> Hstring.make l, ty) lbs in
+let trecord ?(sort_fields = false) ~record_constr lv name lbs =
   let lbs, record_constr =
     if sort_fields then
-      List.sort (fun (l1, _) (l2, _) -> Hstring.compare l1 l2) lbs,
-      Format.sprintf "%s___%s" record_constr n
+      List.sort (fun (l1, _) (l2, _) -> Uid.compare l1 l2) lbs, record_constr
+      (* TODO: FIIIIXXX *)
+      (*       Format.sprintf "%s___%s" record_constr n *)
```
bclement-ocp: I believe this is required for the legacy typechecker not to confuse constructors for distinct record types ; it should be moved to the legacy typechecker.ml rather than removed entirely.

src/lib/structures/ty.ml
```diff
@@ -514,19 +514,20 @@ let type_body name args = Decls.body name args
 
 
 (* smart constructors *)
-let tunit = Text ([],Hstring.make "unit")
+(* HACK: we should create a unique Dolmen identifier here. *)
```
bclement-ocp: We should instead use `DStd.Ty.Const.unit`.

src/lib/structures/fpa_rounding.ml
```diff
@@ -88,7 +88,8 @@ let hstring_ae_reprs =
 (* The rounding mode is the enum with the SMT values.
    The Alt-Ergo values are injected in this type. *)
 let fpa_rounding_mode =
-  Ty.Tsum (Hs.make "RoundingMode", hstring_smt_reprs)
+  let fake_reprs = List.map Uid.of_hstring hstring_smt_reprs in
+  Ty.Tsum (Uid.of_string "RoundingMode", fake_reprs)
```
bclement-ocp: Why use hstring rather than dolmen IDs here?

src/lib/structures/uid.ml
```diff
@@ -0,0 +1,98 @@
...
+(*                                                                        *)
+(*     More details can be found in the directory licenses/               *)
+(*                                                                        *)
+(**************************************************************************)
+
+module DStd = Dolmen.Std
+module DE = DStd.Expr
+
+type t =
+  | Fake of Hstring.t
```
bclement-ocp: Nit: calling them `Fake` is confusing; what's fake about them?

bclement-ocp (commented):
I did not look thoroughly through the minutiæ of all mechanical changes from `Hstring` → `Uid` but the change looks correct overall. I do have some minor questions, and something needs to be done for record constructors which I think are broken in the legacy typechecker by this PR.

src/lib/structures/ty.ml
Halbaroth: I don't know `Dolmen` already contains a unit constant. I'll change it.

src/lib/structures/uid.ml
Halbaroth: It's fake unique identifier. It's not unique but I can rename it into `Hstring`.

src/lib/frontend/d_cnf.ml
```diff
@@ -46,11 +46,15 @@ let unsupported msg =
     (fun str -> Errors.(run_error (Unsupported_feature str)))
     msg
 
-module HT = Hashtbl.Make(
-  struct
-    type t = int
-    let equal = (=)
-    let hash = Fun.id
+type id = Id : 'a DE.id -> id
```
bclement-ocp:
```suggestion
type id = Id : 'a DE.id -> id[@@unboxed]
```

src/lib/structures/uid.ml
Halbaroth: Of course I tried to use the `Dolmen identifier` directly but I got a regression in our non-regression test suite. I should investigate.

Halbaroth: Ok, I tried again with the original index produced by Dolmen. Now I got five regressions and it seems that Alt-Ergo ignores the timeout for them (`make runtest` never returns). Actually, if I run `dune exec -- alt-ergo tests/adts/match_in_terms.ae`, I cannot kill the processus with `CTRL+C`. I have to type `pkill alt-ergo`.

Halbaroth: I suspect an issue with `DE.Id.compare` or `DE.Id.equal`.

Halbaroth: Uhm they seems to be reasonable: 
```ocaml 
  let compare v v' = compare v.index v'.index

  let equal v v' = compare v v' = 0
```
very strange ...

Halbaroth: Thanks to Vincent, I got the bug with `gdb`. The issue comes from the fact that `Dolmen` identifiers can be infinite values and the polymorphic equality loops infinitely on them. I wrote another PR to remove as much as possible polymorphic comparison in lists.

src/lib/structures/fpa_rounding.ml
Halbaroth: When I wrote this PR  I didn't do it because I didn't want to have to manage unique identifiers in the legacy typechecker. Basically, if we use real unique identifiers and hstring identifiers, we have to ensure that the legacy typechecker won't recreate ADTs' constructors for builtin enums (and it does...). 

Finally I found another way to use `dolmen` identifiers here. I modified the comparison functions of `Uid`. Now if we compare a `Hstring` and a `Dolmen` identifiers, we transform the `Dolmen` identifier into string.

bclement-ocp: > the legacy typechecker won't recreate ADTs' constructors for builtin enums (and it does...).

Ah, I guess this is the issue of having different names for the AE and SMT enums, got it.

> Now if we compare a Hstring and a Dolmen identifiers, we transform the Dolmen identifier into string.

Is this not re-introducing the dependency on string values that this PR is trying to eliminate?

Halbaroth: Now the `Dolmen` frontend doesn't produce `Uid.t` of type `Hstring` anymore (but for abstract types, I'll change that). It means we always compare `Dolmen` identifiers here (for constructors, destructors and ADTs' names).

Halbaroth:
The PR is ready for a second pass. I think we should test the legacy frontend with `fpa` on Marvin before merging it.

src/lib/frontend/d_cnf.ml
```diff
@@ -97,42 +101,40 @@ module Cache = struct
   let store_var ind v =
     store_sy ind (Sy.var v)
 
-  let find_ty ind =
-    HT.find ae_ty_ht ind
+  let find_ty id =
+    HT.find ae_ty_ht (Id id)
 
-  let fresh_ty ?(is_var = true) ?(name = None) () =
+  let fresh_ty ?(is_var = true) ?(id = None) () =
```
bclement-ocp: I think you want `?id` here and `~id` when `fresh_ty` is called.

(Currently `fresh_ty` has a `?id:Uid.t option` optional argument, simply using `?id` would make it a `?id:Uid.t` argument)

src/lib/frontend/d_cnf.ml
```diff
@@ -200,11 +202,11 @@ let builtin_enum = function
   | Ty.Tsum (name, cstrs) as ty_ ->
     let ty_cst =
       DStd.Expr.Id.mk ~builtin:B.Base
-        (DStd.Path.global (Hstring.view name))
+        (DStd.Path.global (Uid.show name))
         DStd.Expr.{ arity = 0; alias = No_alias }
```
bclement-ocp: This looks a bit weird; aren't we building a second Dolmen identifier with the same name as an existing one here?

src/lib/structures/uid.ml
```diff
@@ -0,0 +1,92 @@
...
+let[@inline always] of_hstring hs = Hstring hs
+let[@inline always] of_string s = of_hstring @@ Hstring.make s
+
+let hash = function
+  | Hstring hs -> Hstring.hash hs
+  | Dolmen { index; _ } -> (index :> int)
+
+let pp ppf = function
+  | Hstring hs -> Hstring.print ppf hs
+  | Dolmen { path; _ } -> Fmt.string ppf @@ get_basename path
```
bclement-ocp: Why not `DE.Id.print`?

bclement-ocp (commented):
Yeah on paper the changes look correct but we definitely should check that there are no regressions. I am a bit lost by the many small changes now; I will take a deeper look after the weekend. I think there are many unneeded string/Hstring conversions now but we can clean that up later.

src/lib/structures/uid.ml
Gbury: Yeah, dolmen identifiers should never be compared using the polymorphic identity: besides being slow (compared to the comparison defined by dolmen), it may loop forever (because the `Type` identifiers loops on itself), or may fail at runtime (because ids contain tags that can contain arbitrary data such as functions).

src/lib/frontend/d_cnf.ml
Halbaroth: Indeed, I thought it was correct but we duplicate the identifiers here. I fixed the issue by creating the constructors and the Dolmen type in `Fpa_rounding`.

Halbaroth:
I rebase this PR on #1117 to remove some polymorphic comparisons again ;)

src/lib/structures/uid.ml
```diff
@@ -0,0 +1,75 @@
...
+  | Dolmen id -> DE.Id.print ppf id
+
+let show = Fmt.to_to_string pp
+
+let equal u1 u2 =
+  match u1, u2 with
+  | Hstring hs1, Hstring hs2 -> Hstring.equal hs1 hs2
+  | Dolmen id1, Dolmen id2 -> DE.Id.equal id1 id2
+  | _ ->
+    Hstring.equal (Hstring.make @@ show u1) (Hstring.make @@ show u2)
```
bclement-ocp: Isn't this a complicated way of writing `String.equal (show u1) (show u2)`? The complexity should be the same since `Hstring.make` requires a linear scan.

src/lib/structures/uid.ml
```diff
@@ -0,0 +1,75 @@
...
+
+let compare u1 u2 =
+  match u1, u2 with
+  | Hstring hs1, Hstring hs2 -> Hstring.compare hs1 hs2
+  | Dolmen id1, Dolmen id2 -> DE.Id.compare id1 id2
+  | _ -> Hstring.compare (Hstring.make @@ show u1) (Hstring.make @@ show u2)
```
bclement-ocp: Do we need the `Hstring` comparison here? Otherwise `String.compare (show u1) (show u2)` might be sufficient (complexity should be the same since `Hstring.make` requires a linear scan).

src/lib/structures/fpa_rounding.mli
```diff
@@ -40,13 +42,20 @@ type rounding_mode =
...
+[@@deriving ord]
 
 (** Equal to ["RoundingMode"], the SMT2 type of rounding modes. *)
 val fpa_rounding_mode_type_name : string
 
 (** Equal to ["fpa_rounding_mode"], the Alt-Ergo native rounding mode type. *)
 val fpa_rounding_mode_ae_type_name : string
 
+(** The Dolmen rounding mode type. *)
+val fpa_rounding_mode_dty : Dolmen.Std.Expr.Ty.t
```
bclement-ocp, line 54: :+1: 

Do we still need the various string conversion functions with this?

src/lib/structures/fpa_rounding.ml
```diff
@@ -87,8 +88,15 @@ let hstring_ae_reprs =
...
-  Ty.Tsum (Hs.make "RoundingMode", hstring_smt_reprs)
+let fpa_rounding_mode_dty, d_cstrs, fpa_rounding_mode =
+  let module DStd = Dolmen.Std in
+  let ty_cst = DE.Ty.Const.mk (DStd.Path.global "RoundingMode") 0 in
+  let cstrs =
+    List.map (fun c -> DStd.Path.global @@ to_smt_string c, []) cstrs
+  in
+  let _, cstrs = DE.Term.define_adt ty_cst [] cstrs in
+  DE.Ty.apply ty_cst [], cstrs,
+  Ty.Tsum (Uid.of_dolmen ty_cst, List.map (fun (c, _) -> Uid.of_dolmen c) cstrs)
```
bclement-ocp: Note: I think this could be replaced with `DStd.Expr.Ty.roundingMode`. I don't think it is entirely trivial to do so, but might be worth a note in a comment.

bclement-ocp (commented):
Still a couple questions but overall I think this is ready to merge once rebased & if the benches are green.

src/lib/structures/uid.ml
Halbaroth: I thought it was a little bit faster because we compare integers if these `Hstring` already exist but we have to compute the hash of the strings, which may be linear in the length of them.

src/lib/structures/fpa_rounding.mli
Halbaroth: I don't think so

src/lib/structures/fpa_rounding.ml
Halbaroth: I didn't the enum type is shipped with `Dolmen`. We can replace it easily. It's exactly the same code.

Halbaroth: Uhm actually `roundingMode` is the SMT-LIB roundingMode, so it's not so obvious. I add a note.

src/lib/structures/fpa_rounding.mli
Halbaroth: We still need them in both legacy and new frontend. In the new frontend, we use these functions in the typechecker to inject the AE identifiers into SMT-LIB identifiers. 

We also use them in `Expr` because we need to convert a `Hstring` identifier into `Fpa_rounding.rounding_mode`. 

I'm sure we can remove some of them after getting rid of the legacy frontend.

src/lib/structures/fpa_rounding.ml
bclement-ocp: The `RoundingMode` type from AE (SMT-LIB flavor) should be identical to the SMT-LIB RoundingMode, we made sure with @Stevendeo that the constructors are the same.

Halbaroth: I put a note and I will do it in another PR, maybe after the release.

Halbaroth:
No regression with the following combinations:
- frontend `legacy` on `ae-format`
- frontend `legacy` on `ae-format-fpa` with `--enable-theories fpa` 
- frontend `dolmen` on `ae-format`
- frontend `dolmen` on `ae-format-fpa` with `--enable-theories fpa` 

So this PR is ready for I hope a last check ;)

bclement-ocp (approved):
:tada:

PR #1100, opened 2024-04-29
Normalize ADT in `assume`

### LINKED ISSUE #1096: ADT: Normalize representatives in assume
Following #1087 I believe the `assume_literals` function is incorrect: it should call `Uf.find_r` on all semantic values that appear in the literals.

Semantic values that appear in the literals given to `assume` are not guaranteed to be in normal form, even though they usually are: during CC(X) resolution I believe it is possible to generate a literal such as `x = y` then later pick a new representative for `y` or `x`. I have seen this happen with bit-vectors; presumably it can also happen with ADTs.

(Sorry, I missed it during review!)

Halbaroth:
Fix issue #1096

### DIFF
--- a/src/lib/reasoners/adt_rel.ml
+++ b/src/lib/reasoners/adt_rel.ml
@@ -431,6 +431,8 @@ let assume_literals la uf env =
        match lit with
        | Eq (r1, r2) as l, _, ex, Th_util.Subst when is_adt r1 ->
          Debug.assume l;
+         let r1, _ = Uf.find_r uf r1 in
+         let r2, _ = Uf.find_r uf r2 in
          (* Needed for models generation because fresh terms are not added with
             the function add. *)
          let env = add_rec r1 uf env in
@@ -439,6 +441,8 @@ let assume_literals la uf env =
 
        | Distinct (false, [r1; r2]) as l, _, ex, _ when is_adt r2 ->
          Debug.assume l;
+         let r1, _ = Uf.find_r uf r1 in
+         let r2, _ = Uf.find_r uf r2 in
          (* Needed for models generation because fresh terms are not added with
             the function add. *)
          let env = add_rec r1 uf env in
@@ -447,10 +451,12 @@ let assume_literals la uf env =
 
        | Builtin (true, Sy.IsConstr c, [r]) as l, _, ex, _ ->
          Debug.assume l;
+         let r, _ = Uf.find_r uf r in
          assume_is_constr ~ex r c env
 
        | Builtin (false, Sy.IsConstr c, [r]) as l, _, ex, _ ->
          Debug.assume l;
+         let r, _ = Uf.find_r uf r in
          assume_not_is_constr ~ex r c env
 
        | _ ->
--- a/src/lib/reasoners/adt_rel.ml
+++ b/src/lib/reasoners/adt_rel.ml
@@ -431,8 +431,6 @@ let assume_literals la uf env =
        match lit with
        | Eq (r1, r2) as l, _, ex, Th_util.Subst when is_adt r1 ->
          Debug.assume l;
-         let r1, _ = Uf.find_r uf r1 in
-         let r2, _ = Uf.find_r uf r2 in
          (* Needed for models generation because fresh terms are not added with
             the function add. *)
          let env = add_rec r1 uf env in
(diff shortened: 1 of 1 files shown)


src/lib/reasoners/adt_rel.ml
```diff
@@ -431,6 +431,8 @@ let assume_literals la uf env =
        match lit with
        | Eq (r1, r2) as l, _, ex, Th_util.Subst when is_adt r1 ->
          Debug.assume l;
+         let r1, _ = Uf.find_r uf r1 in
```
bclement-ocp: This one looks wrong?

src/lib/reasoners/adt_rel.ml
```diff
@@ -439,6 +441,8 @@ let assume_literals la uf env =
 
        | Distinct (false, [r1; r2]) as l, _, ex, _ when is_adt r2 ->
          Debug.assume l;
+         let r1, _ = Uf.find_r uf r1 in
+         let r2, _ = Uf.find_r uf r2 in
```
bclement-ocp: The explanations should be added to `ex`

bclement-ocp (changes requested):

src/lib/reasoners/adt_rel.ml
Halbaroth: Uhm, I wasn't sure it was necessary because the explanation may be added by a `subst` literal? I agree it's safer to add it.

src/lib/reasoners/adt_rel.ml
Halbaroth: Ok, I'm not sure to understand what's going on here. I didn't find a failure-induced input. I understood that CC(X) can send literals with new class representatives to the ADT theory before emitting a subst literal? Is it the problem?

bclement-ocp: There can only be an `Eq (r1, r2)` with `Subst` origin if `r2` was the representative for `r1` at some point, so `Uf.find_r uf r1` and `Uf.find_r uf r2` should always be equal here, which makes `assume_subst` not very useful.

We need to call `Uf.find_r uf r2` to ensure the most up-to-date representative, although after checking in ccx.ml  and contrary to what I said at the meeting for `Subst` equalities that should not be needed either (they are added in `add_touched` immediately before calling `assume`, so they are guaranteed to be up-to-date).

src/lib/reasoners/adt_rel.ml
bclement-ocp: It is necessary. If `ex` justifies that `r1 <> r2`, `ex` on its own is not enough to justify `rr1 <> rr2`.

(For instance if we have `x <> y` and `x = x'` ; `y = y'`, `x <> y` alone is not sufficient to justify `x' <> y'` ; we need all 3 facts (`x <> y`, `x = x'` and `y = y'`) to derive `x' <> y'`)

src/lib/reasoners/adt_rel.ml
bclement-ocp: To clarify; nothing needs to be done *for `Subst` equalities specifically*.

src/lib/reasoners/adt_rel.ml
bclement-ocp: Note that there is no guarantee about the order in which we treat the literals (they are shuffled in `make_unique`).

src/lib/reasoners/adt_rel.ml
```diff
@@ -439,18 +439,26 @@ let assume_literals la uf env =
...
-         let env = add_rec r1 uf env in
-         let env = add_rec r2 uf env in
-         assume_distinct ~ex r1 r2 env
+         let env = add_rec rr1 uf env in
+         let env = add_rec rr2 uf env in
+         assume_distinct ~ex rr1 rr2 env
 
        | Builtin (true, Sy.IsConstr c, [r]) as l, _, ex, _ ->
          Debug.assume l;
+         let r, _ = Uf.find_r uf r in
```
Halbaroth: I guess I have to add the explanation of `find_r` to `ex`?

bclement-ocp: I think so (here and in the `false` case below)

Halbaroth:
I run a benchmark to be sure.

bclement-ocp (approved):

PR #1106, opened 2024-05-03
Documentation of quantified type in `Expr`

Halbaroth:
While fixing the issue #1099, I wrote a bit of documentation for `Expr`.

### DIFF
--- a/src/lib/structures/expr.ml
+++ b/src/lib/structures/expr.ml
@@ -70,11 +70,9 @@ and quantified = {
   toplevel : bool;
   user_trs : trigger list;
   binders : binders;
-  (* These fields should be (ordered) lists ! important for skolemization *)
   sko_v : t list;
   sko_vty : Ty.t list;
-  loc : Loc.t; (* location of the "GLOBAL" axiom containing this quantified
-                  formula. It forms with name a unique id *)
+  loc : Loc.t;
   kind : decl_kind;
 }
 
@@ -1438,10 +1436,27 @@ and mk_forall_bis (q : quantified) =
         let q = {q with binders; user_trs = trs; sko_v; main = f } in
         mk_forall_bis q
 
+(* Look for a substitution [sbt] of term variables such that:
+   - the domain of [sbt] is a subset of [binders];
+   - the application of [sbt] on the formula [f] produces a formula [g]
+     such that:
+     -- the set of free term variables of [g] in [binders] is strictly smaller;
+     -- the universal closures of [f] and [g] are equivalent.
+
+   For instance, let's [f] be the formula:
+    x <> 1 or f(y) = x + 1
+   and the set of binders is `{ x, y }`. This function will produce the
+   substitution `{ x |-> 1 }`. Indeed, the ground formulas
+    forall x, y : int. x <> 1 or f(y) < x + 1
+   and
+    forall y : int. 1 <> 1 or f(y) < 1 + 1
+   are equivalent.
+
+   *Note*: this heuristic is disabled if the user has defined filters.
+
+   @return [None] if the heuristic failed to produce such a substitution. *)
 and find_particular_subst =
   let exception Found of Var.t * t in
-  (* ex: in "forall x, y : int. x <> 1 or f(y) = x+1 or P(x,y)",
-     x can be replaced with 1 *)
   let rec find_subst v tv f =
     match form_view f with
     | Unit _ | Lemma _ | Skolem _ | Let _ | Iff _ | Xor _ -> ()
@@ -1464,6 +1479,7 @@ and find_particular_subst =
       | _ -> ()
   in
   fun binders trs f ->
+    (* TODO: move the test for `trs` outside. *)
     if not (Ty.Svty.is_empty f.vty) || has_hypotheses trs ||
        has_semantic_triggers trs
     then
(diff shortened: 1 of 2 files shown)


src/lib/structures/expr.ml
```diff
@@ -2584,7 +2610,7 @@ module Purification = struct
         mk_term (Sy.Var fresh_var) [] t.ty , add_let fresh_var t lets
 
       | _ -> (* detect ITEs *)
-        match t.xs with
+        match tm.xs with
```
bclement-ocp: Is this change intentional?

src/lib/structures/expr.ml
```diff
@@ -1438,10 +1436,27 @@ and mk_forall_bis (q : quantified) =
...
         mk_forall_bis q
 
+(* Look for a substitution [sbt] of term variables such that:
+   - the domain of [sbt] is a subset of [binders];
+   - the application of [sbt] on the formula [f] produces a formula [g]
+     such that:
+     -- the set of free term variables of [g] in [binders] is strictly smaller;
+     -- the universal closures of [f] and [g] are equivalent.
+
+   For instance, let's [f] be the formula:
```
```suggestion
   For instance, let [f] be the formula:
```

src/lib/structures/expr.ml
```diff
@@ -70,11 +70,9 @@ and quantified = {
   toplevel : bool;
   user_trs : trigger list;
   binders : binders;
-  (* These fields should be (ordered) lists ! important for skolemization *)
```
bclement-ocp, line 73: I think it is helpful to document this invariant in the .ml file as well.

src/lib/structures/expr.ml
bclement-ocp: Also [f] appears to be both the formula and a function within, is that right?

src/lib/structures/expr.ml
```diff
@@ -1438,10 +1436,27 @@ and mk_forall_bis (q : quantified) =
         let q = {q with binders; user_trs = trs; sko_v; main = f } in
         mk_forall_bis q
 
+(* Look for a substitution [sbt] of term variables such that:
+   - the domain of [sbt] is a subset of [binders];
+   - the application of [sbt] on the formula [f] produces a formula [g]
+     such that:
+     -- the set of free term variables of [g] in [binders] is strictly smaller;
+     -- the universal closures of [f] and [g] are equivalent.
```
bclement-ocp: My understanding was that it we are only trying to transform implications of the form `x = e -> P(x)` into `P(e)`; is it actually more general?

src/lib/structures/expr.ml
bclement-ocp: Given that the CI fails on this I'll guess it was not :rofl:

Halbaroth: Nope, there is no code change in a documentation PR ;)

src/lib/structures/expr.ml
Halbaroth: It's a bit more general because it transforms the formula `∀ x : bool, x  -> Q(x)` into `Q(true)`.

I wrote this paragraph to explain why this transformation is valid, but I can write an exhaustive list of the transformations. I didn't think in terms of implications, which is simpler to understand indeed.

bclement-ocp: Right but if `x` is a boolean, `x` is actually `x = true` (: 

Having the general context where the transformation takes place is useful but I'd say explaining what the function actually does is more valuable.

src/lib/structures/expr.ml
```diff
@@ -70,11 +70,9 @@ and quantified = {
   toplevel : bool;
   user_trs : trigger list;
   binders : binders;
-  (* These fields should be (ordered) lists ! important for skolemization *)
-  sko_v : t list;
+  sko_v : t list; (* This list has to be ordered for the skolemization. *)
```
bclement-ocp, line 73: The original comment applies to *both* `sko_v` and `sko_vty` but the new comment only mentions `sko_v`.

src/lib/structures/expr.ml
```diff
@@ -1438,10 +1441,14 @@ and mk_forall_bis (q : quantified) =
         let q = {q with binders; user_trs = trs; sko_v; main = f } in
         mk_forall_bis q
 
+(* If [f] is a formula of the form [x = a -> P(x)] where [a] doesn't content
+   [x], this function produces the substitution { x |-> a }.
+
+   *Note*: this heuristic isn't used if the user has defined filters.
+
+   @return [None] if the formula hasn't the right form. *)
```
bclement-ocp: I think keeping the example as in your original rewriting was useful (or at least mention that we work with CNF so that `x = a -> P(x)` is written `x <> a \/ P(x)`).

src/lib/structures/expr.ml
```diff
@@ -1438,10 +1441,14 @@ and mk_forall_bis (q : quantified) =
         let q = {q with binders; user_trs = trs; sko_v; main = f } in
         mk_forall_bis q
 
+(* If [f] is a formula of the form [x = a -> P(x)] where [a] doesn't content
+   [x], this function produces the substitution { x |-> a }.
+
+   *Note*: this heuristic isn't used if the user has defined filters.
```
bclement-ocp: Nit: the way to write bold text in odoc format is: `{b Note}`, I don't think `*Note*` works. (Doesn't matter here because this is not a doc comment, but odoc syntax is used below with `@return` so might as well be consistent)

bclement-ocp (approved):

PR #1108, opened 2024-05-06
feat: Rewrite the Intervals module entirely

bclement-ocp:
The existing Intervals module suffers from multiple drawbacks. It is undocumented, uses a questionable implementation where explanations associated with each internal bounds have unclear semantics (did I mention it is not documented), and it has been the source of many soundness bug in the past due to the way it uses exceptions to indicate emptiness. This makes it hard to extend the module; for instance #1058 was delayed because we were not quite sure whether the implementation of functions related to bit-vectors were correct.

This patch is a reimplementation of the Intervals module from scratch. The new implementation keeps the representation used by the `intersect` futnction in the existing implementation (and in a way can be thought of as resolving the TODO in the existing implementation suggesting to generalize that type). It is thoroughly documented, and is split between a "core" module that provides safe functions to deal with explanations, and specialized implementations for common functions (addition, multiplication, etc.) using the "core" interface such that reasoning about the implementation of addition etc. does not require thinking about explanations at all. This makes it easier to extend the module with new specialized functions.

The implementation is done through a (small) family of functors, allowing separate implementations for real and integer intervals that prevent accidentally mixing them up.

For the time being, the old interface is re-implemented on top of the new interface (except where implementation details leaked too much) so as to keep the changes mostly confined to the `Intervals` module. Migrating to the new interface (and especially abandoning the use of exceptions) will be done in a second step.

The patch is relatively big, but can't realistically be split into smaller parts without ending up in intermediate states full of dead code. I suggest reviewers first take a look at the documentation of the `OrderedType`, `Interval` and `Union` signatures in the `Intervals_intf` module (note that this includes some LaTeX, and might be easier to read using `odoc` -- I tried to make sure the `odoc`-generated documentation was usable). This should provide a good understanding of the "core" functionality of the new implementation. The rest of the review can be split into parts that should be mostly independent:

 - Implementation of the `OrderedType` interface for `Z.t` and `Q.t` in `ZEuclideanType` and `QAlgebraicType`;

 - Implementation of the concrete functions for addition, multiplication, etc. in the `Intervals` module;

 - Re-implementation of the old API in the `Intervals.Legacy` module (and minor related changes in other modules, notably `IntervalCalculus`);

 - Implementation of the "core" functionality in the `Intervals_core` module.


bclement-ocp:
Benchmarks are +9-3 on benchtop-tests and +24-6 on ae-format (from a cursory look at the actual problems, this seems to come from the more precise tracking of explanations performed by the new implementation).

Overall things seems to be a tiny bit slower (1% or so), which I  would say is fine given improvements in readability & maintainability. I know a few places where the implementation is less than optimal and could be improved upon if necessary:

 - `add` is a translation, not an arbitrary monotone function; this was used by the previous implementation to perform less comparisons than the new implementation and we could also have a specialized implementation for `add` at the cost of extra complexity;
 - Same for `partial_map_inc` and `partial_map_dec`, as mentioned in the documentation;
 - In `map_mon_to_set` we are currently calling the user-provided function `f` on the initial/final segment when it is forbidden; this means we are often doing additions of `[-oo; x]` with `[-oo; y1]` then `[y2; y3]`, etc. which will all ultimately be subsumed by `[-oo; x + yn]`. We can do better and avoid doing this, at the cost of additional complexity again.

src/lib/reasoners/intervals_intf.ml
```diff
@@ -0,0 +1,735 @@
...
+  module Union(OT : OrderedType) : Union
+    with type 'a union = 'a union
+     and type value := OT.finite
+     and type bnd = OT.t
+     and type explanation := explanation
+end
+
+module type RingType = sig
+  (** An ordered ring is an ordered type with addition and multiplication that
+      follows ring laws. *)
```
Halbaroth, line 610: I guess you mean `ordered ring laws` here? The ring structure has to respect the order.

bclement-ocp:
@Gbury given that you have made a bunch of fixes to the intervals module in the past & some conversations we had wrt API for this module, I'd appreciate your input on the interface file `intervals_intf.ml` if you find the time to take a look.

src/lib/reasoners/intervals_intf.ml
```diff
@@ -0,0 +1,735 @@
...
+  val add : t -> t -> t
+  (** [add] will only be called with values that are compatible with its
+      monotonicity: its argument can never be two bounds of different kinds
+      (upper or lower). *)
+
+  val zero : t
+  (** Neutral element for addition. *)
+
+  val neg : t -> t
+  (** Inverse for addition: for all [u], [add u (neg u) = zero]. *)
```
Halbaroth, line 623: As we use the new math notation of `odoc` in this PR, we can use it here too.

src/lib/reasoners/intervals_intf.ml
bclement-ocp: I initially meant that addition and multiplication must follow ring laws (e.g. distributivity etc.) but you are right, this should follow ordered ring laws (including compatibility with order).

src/lib/reasoners/intervals_intf.ml
bclement-ocp: I (loosely) tried to use the actual OCaml functions (using inline code notation) for short documentation to maximize clarity and use mathematical notation where there was more involved stuff / longer explanations which are easier to follow in maths.

src/lib/reasoners/intervals_intf.ml
```diff
@@ -0,0 +1,735 @@
...
+  val add_explanation : ex:explanation -> t -> t
+  (** [add_explanation ~ex u] adds the explanation [ex] to the union [u].
+
+      More precisely: [add_explanation ~ex u] is equivalent to [u] in models
+      where [ex] holds, and is [Interval.full] otherwise. *)
+
+  val intersect : t -> t -> (t, explanation) kind
+  (** [intersect u1 u2] computes the intersection of [u1] and [u2].
+
+      The resulting intersection is valid in all models. *)
```
Halbaroth, line 417: I'm not sure to understand what you mean by `valid in all models`?

bclement-ocp: I mean that: in any model `M`, `intersect u1 u2` evaluates to the intersection of `M(u1)` and `M(u2)` (I guess it should rather be in all contexts). IOW; `intersect` is correct w.r.t explanations; I will clarify.

bclement-ocp: > in any model M, `intersect u1 u2` evaluates to the intersection of `M(u1)` and `M(u2)`

Sorry, it should be `intersect u1 u2` evaluates to *a superset of* the intersection of `M(u1)` and `M(u2)` — in other contexts we might lose precision (but in the current context, it evaluates to the intersection exactly).

src/lib/reasoners/intervals.mli
```diff
@@ -1,7 +1,7 @@
 (**************************************************************************)
 (*                                                                        *)
 (*     Alt-Ergo: The SMT Solver For Software Verification                 *)
-(*     Copyright (C) 2013-2023 --- OCamlPro SAS                           *)
+(*     Copyright (C) 2024 --- OCamlPro SAS                                *)
```
Halbaroth, line 4: Please don't remove `2013` here. Alt-Ergo wasn't the property of OCamlPro before this date.

src/lib/reasoners/intervals.mli
```diff
@@ -19,140 +19,137 @@
 (*                                                                        *)
 (*     CNRS - INRIA - Universite Paris Sud                                *)
 (*                                                                        *)
-(*     Until 2013, some parts of this code were released under            *)
-(*     the Apache Software License version 2.0.                           *)
-(*                                                                        *)
```
Halbaroth: It's far simpler to use the same header for all the files. So please don't change the header even if no part of the current file existed before 2013. I'll use `headache` to update them before the next release.

src/lib/reasoners/intervals.mli
```diff
@@ -19,140 +19,137 @@
...
+(** This module provides implementations of union-of-intervals over reals and
+    integers. *)
 
-val is_undefined : t -> bool
+type 'a union
+(** Polymorphic union type. This allows writing conversion functions between
+    integer and real unions. *)
 
-val point : Numbers.Q.t -> Ty.t -> Explanation.t -> t
+module Real : AlgebraicField
```
Halbaroth, line 40: My heart of arithmetician cries when I see this.

src/lib/reasoners/intervals.mli
bclement-ocp: This change was not intended to be part of the PR, I will revert it.

src/lib/reasoners/intervals.mli
```diff
@@ -19,140 +19,137 @@
...
-val merge : t -> t -> t
+  val add : t -> t -> t
 
-val abs : t -> t
+  val scale : Numbers.Q.t -> t -> t
 
-val pretty_print : Format.formatter -> t -> unit
+  val affine_scale : const:Numbers.Q.t -> coef:Numbers.Q.t -> t -> t
+  (** Perform an affine transformation on the given bounds.
+      Suposing input bounds (b1, b2), this will return
```
Halbaroth:
```suggestion
      Supposing input bounds (b1, b2), this will return
```

src/lib/reasoners/intervals.mli
```diff
@@ -19,140 +19,137 @@
...
+      (const + coef * b1, const + coef * b2).
+      This function is useful to avoid the incorrect roundings that
+      can take place when scaling down an integer range. *)
 
-val print : Format.formatter -> t -> unit
+  val pretty_print : Format.formatter -> t -> unit
 
-val finite_size : t -> Numbers.Q.t option
+  val print : Format.formatter -> t -> unit
```
Halbaroth:
```suggestion
  val print : t Fmt.t
```

src/lib/reasoners/intervals.mli
bclement-ocp: I am not happy with this name either but couldn't find anything better, even if not strictly correct mathematically (the only other good alternative I had was `FieldWithNthRoot`, which is a mouthful). I happily take suggestions, maybe `RadicalField`?

src/lib/reasoners/intervals.mli
```diff
@@ -19,140 +19,137 @@
...
-val pretty_print : Format.formatter -> t -> unit
+  val affine_scale : const:Numbers.Q.t -> coef:Numbers.Q.t -> t -> t
+  (** Perform an affine transformation on the given bounds.
+      Suposing input bounds (b1, b2), this will return
+      (const + coef * b1, const + coef * b2).
+      This function is useful to avoid the incorrect roundings that
+      can take place when scaling down an integer range. *)
 
-val print : Format.formatter -> t -> unit
+  val pretty_print : Format.formatter -> t -> unit
```
Halbaroth:
```suggestion
  val pretty_print : t Fmt.t
```

src/lib/reasoners/intervals.mli
```diff
@@ -19,140 +19,137 @@
...
 
-val contains : t -> Numbers.Q.t -> bool
+  val add_explanation : t -> Explanation.t -> t
 
-val add_explanation : t -> Explanation.t -> t
+  val equal : t -> t -> bool
 
-val equal : t -> t -> bool
+  val pick : is_max:bool -> t -> Numbers.Q.t
+  (** [pick ~is_max t] returns an elements of the set of intervals [t]. If
```
Halbaroth:
```suggestion
  (** [pick ~is_max t] returns an elements of the union of intervals [t]. If
```

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,807 @@
...
+(**************************************************************************)
+
+open Intervals_intf
+
+(** Pretty-printer for a bound when used as a lower bound. *)
+let pp_lower_bound pp_a ppf lb =
+  match lb with
+  | Unbounded -> Format.fprintf ppf "]-oo"
+  | Open x -> Format.fprintf ppf "]%a" pp_a x
+  | Closed x -> Format.fprintf ppf "[%a" pp_a x
```
Halbaroth: Nit: use `Fmt.pf`.

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,807 @@
...
+
+(** Pretty-printer for a bound when used as an upper bound. *)
+let pp_upper_bound pp_a ppf ub =
+  match ub with
+  | Unbounded -> Format.fprintf ppf "+oo["
+  | Open x -> Format.fprintf ppf "%a[" pp_a x
+  | Closed x -> Format.fprintf ppf "%a]" pp_a x
```
Halbaroth: Nit: use `Fmt.pf`

src/lib/reasoners/intervals.mli
```diff
@@ -19,137 +19,140 @@
 (*                                                                        *)
 (*     CNRS - INRIA - Universite Paris Sud                                *)
 (*                                                                        *)
+(*     Until 2013, some parts of this code were released under            *)
+(*     the Apache Software License version 2.0.                           *)
+(*                                                                        *)
```
Halbaroth, line 24: Has to be removed ;)

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+(*     The Alt-Ergo theorem prover                                        *)
+(*                                                                        *)
+(*     Sylvain Conchon, Evelyne Contejean, Francois Bobot                 *)
+(*     Mohamed Iguernelala, Stephane Lescuyer, Alain Mebsout              *)
+(*                                                                        *)
+(*     CNRS - INRIA - Universite Paris Sud                                *)
+(*                                                                        *)
+(*     Until 2013, some parts of this code were released under            *)
+(*     the Apache Software License version 2.0.                           *)
+(*                                                                        *)
```
Halbaroth, line 24: Same here

src/lib/reasoners/intervals_intf.ml
```diff
@@ -0,0 +1,738 @@
...
+(*     The Alt-Ergo theorem prover                                        *)
+(*                                                                        *)
+(*     Sylvain Conchon, Evelyne Contejean, Francois Bobot                 *)
+(*     Mohamed Iguernelala, Stephane Lescuyer, Alain Mebsout              *)
+(*                                                                        *)
+(*     CNRS - INRIA - Universite Paris Sud                                *)
+(*                                                                        *)
+(*     Until 2013, some parts of this code were released under            *)
+(*     the Apache Software License version 2.0.                           *)
+(*                                                                        *)
```
Halbaroth, line 24: Same here

src/lib/reasoners/intervals.mli
Halbaroth: `RadicalField` sounds a little bit better because there exists a notion of `Radical extension` in `Galois Theory` and our `Real` field is actually a tower of radical extensions of the rational field.

Let's keep the old name. As we have a comment to explain this field is closed by root extractions, it's acceptable.

Halbaroth:
I'm doing a second pass on this PR. Sorry for the partial review.

src/lib/reasoners/intervals.mli
Halbaroth: We decided to use separated headers so this comment is irrelevant now.

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+
+  type explanation = Ex.t
+
+  type 'a union = 'a * ('a interval, Ex.t) kind list * 'a
+  (* We represent intervals as a list of maximal disjoint allowed intervals with
+     precise bounds and explanations justifying that the (nonempty) gaps between
+     each consecutive allowed interval is forbidden.
+
+     This induces a loss of
+     precision: for instance, if [0, 4] and [10, 20] are allowed and [6, 7] is
```
Halbaroth:
```suggestion
     This induces a loss of precision: 
     for instance, if [0, 4] and [10, 20] are allowed and [6, 7] is
```

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+     forbidden with justification [ex], this will "widen" the forbidden interval
+     to [4+eps, 10-eps] instead; if we find that the associated value must be
+     equal to [5], we will include [ex] in the conflict even though it is not
+     strictly necessary. On the other hand, there are diminishing returns: we
+     don't want to keep track too precisely of the forbidden intervals if they
+     are not going to occur in a conflict.
```
Halbaroth: Can you explain why we don't want to keep them?

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+
+     Note that this is always correct, because it is equivalent to adding
```
Halbaroth: By `correct`, you mean `sound`?

src/lib/reasoners/intervals_intf.ml
```diff
@@ -0,0 +1,738 @@
...
+
+  (** {1 Creation}
+
+      The following functions are used to create unions. *)
+
+  val of_interval :
+    ?ex:explanation -> bnd interval -> t
+  (** Build an union from an interval.
+
+      [of_interval ?ex:None i] evaluates to [i] in all contexts.
```
Halbaroth, line 326:
```suggestion
      [of_interval ~ex:None i] evaluates to [i] in all contexts.
```

src/lib/reasoners/intervals_intf.ml
```diff
@@ -0,0 +1,738 @@
...
+  type explanation
+  (** The type of explanations. *)
+
+  type value
+  (** The type of values. *)
+
+  type bnd
+  (** The type of bounds. *)
+
+  type 'a union
```
Halbaroth, line 305: Is it true that a `NonEmpty`interval is always followed by an `Empty` interval (or nothing) and 
an `Empty` interval is always followed by an `NonEmpty` interval (of nothing)? In other words, this list alternates the two constructors of `kind`?

We can enforce this invariant as follows:
```ocaml
type ('a, 'b) nonempty = [
  | `NonEmpty of 'a * ('a, 'b) empty
  | `Nil
]

and ('a, 'b) empty = [
  | `Empty of 'b * ('a, 'b) nonempty
  | `Nil
]

type ('a, 'b) union = [('a, 'b) nonempty | ('a, 'b) empty]
``` 
or with a GADT, but I'm ok with your code too. It's easy to preserve this invariant in the module. We should at least mention this invariant at the type definition.

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+
+    let value_opt (_, u, _) =
+      let rec loop ex v = function
+        | [] -> Option.map (fun v -> (v, ex)) v
+        | Empty ex' :: u -> loop (Ex.union ex ex') v u
+        | NonEmpty i :: u ->
+          match v with
+          | None when i.lb = i.ub -> loop ex (OT.value_opt i.lb) u
+          | _ -> None
+      in loop Ex.empty None u
```
Halbaroth: If an union of intervals always alternate empty and nonempty intervals, we can write this function as follows? 
```ocaml 
    let value_opt (_, u, _) =
      let v, ex =
        match u with
        | Empty ex :: NonEmpty i :: ([Empty _] | []) when i.lb = i.ub ->
          OT.value_opt i.lb, ex
        | NonEmpty i :: ([Empty _] | []) when i.lb = i.ub ->
          OT.value_opt i.lb, Ex.empty
        | _ -> None, Ex.empty
      in
      Option.map (fun v -> (v, ex)) v
```

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+       they do not deal with explanations. There are the exception to the rule
+       that only functions related to explanations are implemented in [Union],
+       because it just makes sense to have them here. *)
+
+    let rec subset_seq ~strict i1 s1 i2 s2 =
+      if i2.ub < i1.lb then
+        (* i1 is entirely after i2; skip i2 *)
+        match Stdcompat.Seq.uncons s2 with
+        | None -> false
+        | Some (i2', s2') -> subset_seq ~strict:false i1 s1 i2' s2'
```
Halbaroth, line 310: We should mention that we turn this flag into `false` because in this case we know that `i1 U s1` is included strictly into `i2 U s2` if and only if `i1 U s1` is included into `i2' U s2'`.

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+    let nonempty = function
+      | NonEmpty u -> u
+      | Empty _ -> invalid_arg "nonempty"
+
+    let checked_opt = function
+      | None -> Empty Ex.empty
+      | Some u -> checked u
+
+    let add_explanation ~ex ((_, u, _) as u') =
+      if Ex.is_empty ex then u' else
```
Halbaroth:
```suggestion
      if Ex.is_empty ex then u' 
      else
```

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+        let[@tail_mod_cons] rec loop ex' = function
+          | [] -> [ Empty ex' ]
+          | [ NonEmpty i ] as u when i.ub = OT.pinfty -> Empty ex' :: u
+          | NonEmpty i :: u -> Empty ex' :: NonEmpty i :: loop ex u
+          | Empty ex'' :: u -> loop (Ex.union ex' ex'') u
+        in
```
Halbaroth: I am not sure this code is correct. If our list ends with an empty interval, it will create an extra empty interval, which isn't necessary. It's probably fine but it would be better to preserve the alternation of the two constructors.

There are three explanations `ex, ex', ex''` and we have to be very careful to avoid a soudness bug here. I think this piece of code would be easier to read if we avoid to use the accumulator as follows:
```ocaml 
        let[@tail_mod_cons] rec loop = function
          | [] -> [ Empty ex ]
          | [ NonEmpty i ] as u when i.ub = OT.pinfty -> Empty ex :: u
          | [ Empty ex' ] -> [ Empty (Ex.union ex ex') ]
          | Empty ex' :: NonEmpty i :: u ->
            Empty (Ex.union ex ex') :: NonEmpty i :: loop u
          | NonEmpty i :: u -> NonEmpty i :: loop u
          | _ -> assert false
        in
```

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+        !(union ex1 ex2) U [2; 3]
+
+       We need [ex1] to forbid ]1; 2[ and [ex2] to forbid [0; 1] and [2; 4].
+
+       In addition, the current implementation respects the known global
+       lower/upper bounds, which adds a low amount of complexity (everything
+       before/after the global lower/upper bound can be removed without
+       justification). An alternative implementation could ignore the global
+       lower and upper bounds and always return [-oo, +oo] as an enclosing
+       interval; which might improve efficiency of further computations. *)
```
Halbaroth, line 403: Did you run benchmarks with this alternative implementation?

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+         greater than or equal to [ub], but there may be allowed intervals that
+         end inside [lb; ub] (or at [pred lb]).
+
+         The list [fs] collects a set of forbidden half-intervals that must be
+         added before [lb; ub] (unless they get subsumed by merging). *)
+      let rec loop gub lb ub u fs = function
+        | [] ->
+          prepend_forbid gub lb (NonEmpty { lb ; ub } :: u) fs
+        | Forbid (i, _) :: s when lb <= i.lb ->
+          (* [i] is subsumed by [lb; ub] *)
```
Halbaroth:
```suggestion
          (*  [i] is subsumed by [lb; ub]:
              --------------------------------------------------
                    |ooooo|øøøøøøø|oooooooooo|
              --------------------------------------------------
                    lb    i.lb    i.ub       ub
          *)
```

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+          loop gub lb ub u fs s
+        | Forbid (i, ex) :: s ->
+          (* [i] starts before [lb]; need to keep it *)
+          loop gub lb ub u ((i.lb, ex) :: fs) s
+        | Allow i :: s when lb <= i.ub || OT.succ i.ub = lb ->
+          (* No gap: merge intervals ([i.ub] is smaller by sorting)*)
+          loop gub (min i.lb lb) ub u fs s
+        | Allow i :: s ->
+          (* Gap with another allowed interval *)
```
Halbaroth:
```suggestion
          (*  Gap with another allowed interval:
              --------------------------------------------------
                    |/////|       |oooooooooo|
              --------------------------------------------------
                    i.lb  i.ub    lb         ub
          *)
```

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+         The list [fs] collects a set of forbidden half-intervals that must be
+         added before [lb; ub] (unless they get subsumed by merging). *)
+      let rec loop gub lb ub u fs = function
+        | [] ->
+          prepend_forbid gub lb (NonEmpty { lb ; ub } :: u) fs
+        | Forbid (i, _) :: s when lb <= i.lb ->
+          (* [i] is subsumed by [lb; ub] *)
+          loop gub lb ub u fs s
+        | Forbid (i, ex) :: s ->
+          (* [i] starts before [lb]; need to keep it *)
```
Halbaroth:
```suggestion
      (*  [i] starts before [lb]; need to keep it:
          --------------------------------------------------
                |/////|øøøøøøø|oooooooooo|
          --------------------------------------------------
                i.lb  lb      i.ub       ub

        OR

          --------------------------------------------------
                |/////|       |oooooooooo|
          --------------------------------------------------
                i.lb  i.ub    lb         ub
      *)
```

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+          loop gub lb ub u ((i.lb, ex) :: fs) s
+        | Allow i :: s when lb <= i.ub || OT.succ i.ub = lb ->
+          (* No gap: merge intervals ([i.ub] is smaller by sorting)*)
```
Halbaroth:
```suggestion
      (*  No gap: merge intervals ([i.ub] is smaller by sorting):
          --------------------------------------------------
                |/////|øøøøøøø|oooooooooo|
          --------------------------------------------------
                i.lb  lb      i.ub       ub

        OR

          --------------------------------------------------
                |ooooo|øøøøøøø|oooooooooo|
          --------------------------------------------------
                lb    i.lb    i.ub       ub
      *)
```

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -1875,36 +1860,37 @@ let fm_simplex_unbounded_integers_encoding env uf =
...
+           Some Sim.Core.{ bvalue = R2.of_r lb ; explanation = ex_lb }
+         | None -> None
+       in
+       let max =
+         match ub with
+         | Some (ub, ex_ub) ->
+           let ub = Q.from_z ub in
+           Some Sim.Core.{ bvalue = R2.of_r ub ; explanation = ex_ub }
+         | None -> None
+       in
```
Halbaroth: We can factorize the `min` and `max` cases as it was done in `bnd_to_simplex_bound` before.

src/lib/reasoners/intervals.mli
```diff
@@ -19,137 +19,140 @@
...
-val abs : t -> t
+  val scale : Numbers.Q.t -> t -> t
 
-val pretty_print : Format.formatter -> t -> unit
+  val affine_scale : const:Numbers.Q.t -> coef:Numbers.Q.t -> t -> t
+  (** Perform an affine transformation on the given bounds.
+      Supposing input bounds (b1, b2), this will return
+      (const + coef * b1, const + coef * b2).
+      This function is useful to avoid the incorrect roundings that
+      can take place when scaling down an integer range. *)
```
Halbaroth:
```suggestion
  (** Perform an affine transformation on the given bounds.
      Supposing input bounds (b1, b2), this will return
      (const + coef * b1, const + coef * b2).
      This function is useful to avoid the incorrect roundings that
      can take place when scaling down an integer range. 
      
      @raise Invalid_arg if [coef] is zero. *)
```

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+          let f_i = { lb = f i.ub ; ub = f i.lb } in
+          loop f_i.ub [NonEmpty f_i] u'
+        | u -> loop (f glb) [] u
+      in init u
+
+    (** {2 Image by arbitrary functions} *)
+
+    type 'a basic =
+      | Allow of OT.t interval
+      | Forbid of OT.t interval * Ex.t
```
Halbaroth: I guess you want to write:
```ocaml
type 'a basic = 
  | Allow of 'a interval 
  | Forbid of 'a interval * Ex.t
```

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+    let map_to_set f (glb, u, gub) acc =
+      let rec loop acc lb ex = function
+        | [] -> forbid_set ~ex (f { lb ; ub = gub }) acc
+        | Empty ex' :: u -> loop acc lb (Ex.union ex ex') u
+        | NonEmpty i :: u ->
+          let acc =
+            if Ex.is_empty ex then acc
+            else forbid_set ~ex (f { i with lb }) acc
+          in
+          loop (f i acc) i.lb Ex.empty  u
```
Halbaroth: Same remark as before. I think it is easier to read this kind of function if we traverse the list two per two elements.

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+    let of_set_checked s = checked_opt (of_set s)
+
+    let of_set_nonempty s = nonempty @@ checked_opt @@ of_set s
+
+    let forbid_set ~ex s acc =
+      if Ex.is_empty ex then acc
+      else
+        List.rev_append
+          (List.rev_map
+             (function Allow i | Forbid (i, _) -> Forbid (i, ex)) (s []))
```
Halbaroth, line 746: Can you add a comment to explain why we can ignore the explanation stored in the `Forbid` intervals here?

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+      | (Allow i | Forbid (i, _)) :: s ->
+        Some (
+          List.fold_left (fun { lb = glb ; ub  = gub } b ->
+              let { lb ; ub } = match b with Allow i | Forbid (i, _) -> i in
+              { lb = min lb glb ; ub = max ub gub }
+            ) i s
+        )
+
+    (* Similar to [map_to_set] except that we are keeping track of the
+       global hull of each call to [f i] where [i] is allowed.
```
Halbaroth:
```suggestion
       convex hull of each call to [f i] where [i] is allowed.
```
The function `hull_set` computes the convex hull of a set. `Global hull` is not a common terminology.

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+
+       Then we know that if [x] is between two consecutive intervals
+       [i] and [i'] (with [i.ub < i'.lb]), [f x] is included in the
+       hull of [union (f i) (f i')].
```
Halbaroth, line 784:
```suggestion
       convex hull of [union (f i) (f i')].
```

src/lib/reasoners/intervals_core.ml
Halbaroth: Nit: it's more readable to align `if` and `else`.

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+            if Ex.is_empty ex then acc
+            else forbid_set ~ex (f { i with lb }) acc
+          in
+          loop (f i acc) i.lb Ex.empty  u
+      in loop acc glb Ex.empty u
+
+    let union_hull i1 i2 =
+      { lb = min i1.lb i2.lb ; ub = max i1.ub i2.ub }
+
+    let hull_set = function
```
Halbaroth: I suggest to rename this function `convex_hull_set`.

src/lib/reasoners/intervals_intf.ml
```diff
@@ -0,0 +1,738 @@
...
+      the current context. *)
+
+  val map_to_set : ('a interval -> set) -> 'a union -> set
+  (** [map_to_set f u] computes the image of [u] by [f].
+
+      There are no restrictions on [f] (except that it be isotone, i.e. it must
+      be compatible with inclusion), which means that [map_to_set] needs
+      to call [f] on currently allowed intervals, but also on some intervals
+      that are currently impossible but would be possible in other contexts
+      (depending on explanations).
```
Halbaroth: As far as I know, `isotone` is not a common terminology. I think we should inline its definition here:
``` 
a ⊂ b  => f(a) ⊂ f(b) 
```

src/lib/reasoners/intervals_intf.ml
```diff
@@ -0,0 +1,738 @@
...
+
+  val of_set_nonempty : set -> t
+  (** Converts a [set] to an union, assuming that it is nonempty in the current
+      context.
+
+      @raise Invalid_argument if the [set] represents an union that is empty in
+      the current context. *)
+
+  val map_to_set : ('a interval -> set) -> 'a union -> set
+  (** [map_to_set f u] computes the image of [u] by [f].
```
Halbaroth, line 512: What's the definition of `image`? I expect that the image of a map of type `X -> Y` is of type `subset of Y`. Here, we have `Y = set` and `type (Im f) = set` too. I guess we compute `U f(i)` where `i` is any interval of `u`.

I understand that we see an element of type union as:
- a subset of the real line 
- a set of disjoint intervals

If we see `u` as a subset of the real line and `f` has a map of this subset to the real line, there is no restriction on `f`. The isotone restriction is just because our internal representation of unions leaks in this API. Is there any way to avoid that?

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+        | u -> loop (f glb) [] u
+      in init u
+
+    (** {2 Image by arbitrary functions} *)
+
+    type 'a basic =
+      | Allow of OT.t interval
+      | Forbid of OT.t interval * Ex.t
+
+    (* The ['a set] type represents a loose union of [basic] intervals. There
```
Halbaroth:
```suggestion
    (* The [set] type represents a loose union of [basic] intervals. There
```

src/lib/reasoners/intervals_core.ml
Halbaroth: Does it mean that for a `set` S, an element `x` is in the S if and only if there is an allow interval of S that contains `x`? 

If it is the case, we can see a `set` S as an union of implication `ex => I \subset S` where `ex` is an explanation. The `Allow` case is essentially the case with `ex = Explanation.empty` ?

src/lib/reasoners/intervals_core.ml
```diff
@@ -0,0 +1,811 @@
...
+       In addition, the current implementation respects the known global
+       lower/upper bounds, which adds a low amount of complexity (everything
+       before/after the global lower/upper bound can be removed without
+       justification). An alternative implementation could ignore the global
+       lower and upper bounds and always return [-oo, +oo] as an enclosing
+       interval; which might improve efficiency of further computations. *)
+    let intersect =
+      let[@tail_mod_cons] rec loop ex1 u1 ex2 u2 gub =
+        match u1, u2 with
+        | [], [] -> (empty [@tailcall false]) (min_ex ex1 ex2) []
```
Halbaroth, line 407: Why do you use this annotation here? Is it crucial to ensure that the tail call is optimized?

Halbaroth (changes requested):

src/lib/reasoners/intervalCalculus.ml
bclement-ocp: Sure.

src/lib/reasoners/intervals_core.ml
bclement-ocp: Keeping track of forbidden intervals is only useful if we find a conflict, so if we don't find a conflict involving these intervals, we have spent resources computing values that are not used.

src/lib/reasoners/intervals_core.ml
bclement-ocp: Sure.

src/lib/reasoners/intervals_core.ml
bclement-ocp: No, we need both explanations in the `[ Empty ex ; NonEmpty i ; Empty ex' ]` case.

src/lib/reasoners/intervals_core.ml
bclement-ocp: That is what the comment "i1 is entirely after i2; skip i2" intended to convey; I will clarify the comment.

src/lib/reasoners/intervals_core.ml
bclement-ocp: I somewhat disagree; I like the one liner to get rid of a trivial case. I can change it.

src/lib/reasoners/intervals_core.ml
bclement-ocp: > I am not sure this code is correct. If our list ends with an empty interval, it will create an extra empty interval,

I don't think that is true? `loop ex' [ Empty ex'' ]` becomes `loop (Ex.union ex' ex'') []` which becomes `[ Empty (Ex.union ex' ex'') ]`.

On the other hand, your code is wrong for `[NonEmpty i]` when `i.ub <> OT.pinfty` (and `i.lb <> OT.minfty`) because it returns `[ NonEmpty i ; Empty ex ]` where it should return `[ Empty ex ; NonEmpty i ; Empty ex ]`, and it is also wrong for `[ NonEmpty i1 ; NonEmpty i2 ]` because it does not add an `Empty` constructor in between. `Empty` entries with an empty explanation are skipped (see e.g. `of_interval`; I think I did not make this information explicit in the documentation and I will update it).

In general, while writing the code, I found that trying to exploit this "strict alternation" invariant in places where it was not strictly necessary resulted in writing more code with more cases to analyze and was generally more susceptible to making a mistake causing a soundness bug. This is why I settled on the general pattern of accumulating the explanation in a `loop` function — it requires to keep a bit more stuff in mind, but the pattern is always the same and once I got used to it it resulted in less mistakes in general.

I can add a comment to explain this.

src/lib/reasoners/intervals_core.ml
bclement-ocp: Note that this is `[@tailcall false]`, not `[@tailcall true]`. Without this annotation, the `tail_mod_cons` transformation complains (because `empty` is not mutually recursive, this tail call no longer is a tail call after the transformation).

src/lib/reasoners/intervals_core.ml
bclement-ocp: No. I don't think this will have an impact that is much higher than noise and unless we identify performance issues with the interval module, this implementation is interesting enough to keep because it allows to keep track of global bounds (e.g. `bv2nat(x)` is between `[0, 2^n]`).

src/lib/reasoners/intervals_core.ml
bclement-ocp: I think I actually want `type basic = Allow of OT.t interval | Forbid of OT.t interval * Ex.t` here, I will fix it.

src/lib/reasoners/intervals_core.ml
bclement-ocp: I am not sure I understand what this graphic is trying to say. I can rewrite to: "[i] is forbidden but included in [lb; ub] that is allowed.", is that more clear?

bclement-ocp: Yes and no — explanations are true in the current context where the forbidden intervals are, well, forbidden, so there is some negation trickery. I thought I explained this somewhere but I can't find it right now, I wlil add some documentation explaining.

An element `x` is in `S` *in context `C`* if and only if:

 - There is an allowed interval `I \subset S` with `x in I`, or
 - There is a forbidden interval `ex \/ I \subset S` with `x in I` and `ex` an explanation that is *false* in context `C`

So it is an union (disjunction) of implications $\neg \mathrm{ex} \rightarrow I \subseteq S$ where $\mathrm{ex}$ is an implication and the `Allow` case is essentially the case with `ex = Explanation.full` (where `Explanation.full` is false in all contexts, whereas `Explanation.empty` is true in all contexts).

(You can flip things around and say that `Explanation.empty` is false in all contexts etc. but it remains that a forbidden interval with `Ex.empty` explanation provides no information and can be removed, which is the opposite of `Allow`)

src/lib/reasoners/intervals_core.ml
bclement-ocp: OK, I get the drawings now, I will add some (but without unicode symbols).

src/lib/reasoners/intervals_core.ml
bclement-ocp: If an interval is forbidden with explanation `ex` it means: "if `ex` is `false`, then this interval is possible".

Here we add a second explanation `ex2` to a forbidden interval with explanation `ex1`. So: "if `ex2 is false, then if `ex1` is false, then this interval is possible" — that is "if `ex2` is false and `ex1` is false, then this interval is possible".

It is always sound to weaken an explanation, so we can go from "if `ex2` is false and `ex1` is false, then this interval is possible" to just "if `ex1` is false, then this interval is possible" or just "if `ex2` is false, then this interval is possible. In the limit we simply get: "this interval is possible" (which correspond to the `Allow` case) — and indeed it is always sound to over-approximate the set of possible values.

Note that `Ex.union ex1 ex2` corresponds to "if `ex1` is false *or* `ex2` is false, then this interval is possible"; in which case eliminating either `ex1` or `ex2` would be unsound.

src/lib/reasoners/intervals_core.ml
bclement-ocp: This is not a "global hull", it is a "(convex) hull that is global" — i.e. we are computing a hull that is valid in all contexts. I will clarify the terminology.

src/lib/reasoners/intervals_core.ml
bclement-ocp: Same answer as before: this was my initial intuition as well when writing this code (that traversing the list by chunks would be better), but then I repeatedly got lost in too many corner cases, causing soundness issues and uncertainty about correctness, so I switched to this style.

In practice I find that once you get used to thinking "we accumulate an explanation for the gap between two consecutive allowed intervals" it is much easier to work with than the other style because there really are only two cases: `[]` and `NonEmpty i :: u`, the `Empty` case is always just stashing away the explanation into the accumulated one. On the other hand, the "two per two elements" style is full of corner cases.

(We can start with an `Empty` *or* a `NonEmpty`, and we can end with an `Empty` *or* a `NonEmpty`, and sometimes we don't have an `Empty` between two consecutive `NonEmpty`s because the associated explanation is `Ex.empty`)

src/lib/reasoners/intervals_core.ml
bclement-ocp: Sure.

src/lib/reasoners/intervals_intf.ml
bclement-ocp: This is not entirely true because we allow skipping the `Empty` between two consecutive `NonEmpty` if it has an associated `Ex.empty` explanation.

I do not want to use polymorphic variants here due to memory usage concerns (non-constant polymorphic variants take up one additional word per entry) ; I did try writing a GADT with these invariants but it resulted in awkward and hard-to-read code so I gave up on the idea.

src/lib/reasoners/intervals_intf.ml
bclement-ocp: The syntax is correct; `[of_interval ~ex:None i]` would not be well typed.

src/lib/reasoners/intervals_intf.ml
bclement-ocp: It is common enough to be mentioned on Wikipedia. I used isotone for "monotone with respect to inclusion" (in general it means "monotone with respect to a partial order"); I will clarify this.

src/lib/reasoners/intervals_intf.ml
bclement-ocp: In fact the intent is that we compute $\bigcup_{x \in U} f(\\{ x \\})$, but we exploit the interval-monotonicity of $f$ to compute `f` on a "judiciously chosen and sufficient" (from the point of view of the user) set of intervals. It turns out that the intervals we use are disjoint, but it does not change anything (actually I am not sure they are — we may be extending the forbidden intervals to also include the allowed intervals they touch, I don't remember).

(Obviously there is no way for us to compute $\\{ f(x) \mid x \in U \\}$ directly when $U$ is a non-singleton rational or real interval anyways, so we have to represent `f` by a function that takes and returns some sort of set representation)

I will try to clarify the doc.

bclement-ocp (commented):
Many replies inline, I tried to add a bunch of documentation. I also moved all the `loop ... (Ex.union ex ex')` lines to always be the first of the pattern-matching; hopefully that helps showing that this is not the important case.

Should be ready for round 2.

src/lib/reasoners/intervals_core.ml
Halbaroth: It took me a lot of time to understand this piece of code but now I understood it, it is clear. Let's keep the comment ;)

src/lib/reasoners/intervals_core.ml
Halbaroth: Doesn't matter much for me. Let's keep it. At some point, we will use a formatter like ocamlformat.

src/lib/reasoners/intervals_core.ml
Halbaroth: This graphic represents the relative position of the intervals `[lb; ub]` and `i`. The `o` letter means the element is in `[lb; ub]`, `/` means the element is in `I` and `ø` means the element is in both intervals.

src/lib/reasoners/intervals_core.ml
Halbaroth: oh! It didn't understand that at all! Thanks for the clarification.

src/lib/reasoners/intervals_core.ml
Halbaroth: > Does it mean that for a `set` S, an element `x` is in the S if and only if there is an allow interval of S that contains `x`?
> 
> If it is the case, we can see a `set` S as an union of implication `ex => I \subset S` where `ex` is an explanation. The `Allow` case is essentially the case with `ex = Explanation.empty` ?

For me it's far easier to understand these domains like that. Thanks ;)

src/lib/reasoners/intervals.mli
Halbaroth: Did you skip this comment? @bclement-ocp

bclement-ocp: I was confused for a moment because I did make that change. VScode was acting up and opening files in the wrong directory with about 50% chance yesterday, and it looks like I made this specific change in the wrong worktree, thanks for catching it!

Halbaroth (approved):
Thanks for the clarification :)

PR #1110, opened 2024-05-07
fix(UF): Use proper explanation for neqs when initializing leaf

bclement-ocp:
When initializing a leaf, the equality `p = rp` between `p` and its normal form `rp` w.r.t the current environment is justified by `ex_rp`, which is usually but not necessarily `Ex.empty` and needs to be added to a conflict found while adding the equality `p = rp`.


bclement-ocp:
Actually, this case looks like it is covered by `explain_repr_of_distinct`.

PR #1112, opened 2024-05-10
fix(CP): Remove `ex` parameter from `Domain.intersect`

bclement-ocp:
Adding an explicit `ex` parameter to the `Domain.intersect` signature was intended to make it behave more like an update and avoid issues such as #479. The other intent was to be a bit more efficient by avoiding additional `add_explanation` calls.

Given that I made this design yet have now introduced the same soundness issue twice because of it, it is now clear that this was not a good idea.

This patch reverts that design decision and the `ex` parameter in the `Domain.intersect` API, replacing it with explicit calls to `add_explanation` instead.

### DIFF
--- a/src/lib/reasoners/bitlist.ml
+++ b/src/lib/reasoners/bitlist.ml
@@ -80,12 +80,12 @@ let value b = b.bits_set
 let is_fully_known b =
   Z.(equal (shift_right (bits_known b + ~$1) b.width) ~$1)
 
-let intersect ~ex b1 b2 =
+let intersect b1 b2 =
   let width = b1.width in
   let bits_set = Z.logor b1.bits_set b2.bits_set in
   let bits_clr = Z.logor b1.bits_clr b2.bits_clr in
   bitlist ~width ~bits_set ~bits_clr
-    (Ex.union b1.ex (Ex.union ex b2.ex))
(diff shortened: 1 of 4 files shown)


bclement-ocp:
> Given that I made this design yet have now introduced the same soundness issue twice because of it, it is now clear that this was not a good idea.

To clarify: [this](https://github.com/bclement-ocp/alt-ergo/commit/71258eb2e333ccdc37be8026aeac9c553d354b02#diff-9c506f9751f739f250c812250496dc8865030750c6c64eb579d3680e28f3dfdfR97-R101) is what I am talking about.

Halbaroth:
Is ADT theory exposed to same risk? Because we use the previous API for their domains.

Halbaroth (approved):

bclement-ocp:
We can make the same change in the enum/ADT theories yes.

(Note that I don't think there are soundness issues on `next` related to this currently; the soundness issue I mention is related to integration of the Intervals API that does not add explanations on intersection. But arguably the API here is weird: the explanation does not justify that "the intersection is valid" -- whatever that means -- ; rather it justifies that one of the domain applies to a specific variable and that should be reflected by a call to `add_explanation`).

PR #1113, opened 2024-05-10
fix(FPA): Separate semantic triggers upon trigger construction

### LINKED ISSUE #1111: Possible unsoundness with FPA
[repr.ae.txt](https://github.com/OCamlPro/alt-ergo/files/15269432/repr.ae.txt)

On the attached file, I get this result:
```
$ alt-ergo repr.ae --enable-theories fpa
File "repr.ae", line 36, characters 3-111: Valid (0.0121) (107 steps) (goal defqtvc)
```
But I don't see any reason for the file to be valid.

This was originally reproduced with alt-ergo 2.4.0 and the options `--use-fpa --prelude fpa-theory-2019-10-08-19h00.ae`.

bclement-ocp:
Currently, the theories are responsible for populating the `semantic` field of triggers when a lemma with semantic triggers is added to the theory (call to `assume_th_elt`).

This means that such a lemma lives for some time *without* semantic triggers: in particular, it is initially created using `mk_forall` without semantic triggers. During that call, the guard in `find_particular_subst` preventing the "particular substitution" optimization from triggering for lemmas with semantic trigger is ignored; which means that the "particular substitution" optimization is actually applied to lemmas with semantic triggers.

In particular, this optimization makes the lemma
`float_of_pos_pow_of_two` in the FPA theory unsound, because it removes the check that `x` is actually a power of two, and causes #1111.

This patch makes the `Expr` module responsible for separating syntaxic and semantic triggers, rather than the theory. This ensures that a lemma with semantic triggers never appears as having no semantic triggers. In order to make sure this invariant is properly maintained, the `trigger` type is again made private to the `Expr` module. This required moving from sorting code from the `Matching` module to the `Expr` module, also ensuring that triggers are properly sorted for matching purposes at all times.

Fixes #1111

### DIFF
--- a/src/lib/frontend/d_cnf.ml
+++ b/src/lib/frontend/d_cnf.ml
@@ -1655,14 +1655,11 @@ and make_trigger ?(loc = Loc.dummy) ~name_base ~decl_kind
     mk_expr ~loc ~name_base ~decl_kind
   in
   let content = List.map mk_expr e in
-  let t_depth =
-    List.fold_left (fun z t -> max z (E.depth t)) 0 content in
   (* clean trigger:
      remove useless terms in multi-triggers after inlining of lets*)
-  let trigger =
-    { E.content; t_depth; semantic = []; (* will be set by theories *)
-      hyp; from_user; }
-  in
+  let trigger = E.mk_trigger ~user:from_user ~hyp content in
+  if not in_theory && not (Lists.is_empty trigger.semantic) then
+    Errors.typing_error ThSemTriggerError loc;
   E.clean_trigger ~in_theory name trigger
 
 (** Preprocesses the body of a goal by:
--- a/src/lib/reasoners/matching.ml
+++ b/src/lib/reasoners/matching.ml
@@ -482,19 +482,10 @@ module Make (X : Arg) : S with type theory = X.t = struct
     Debug.match_pats_modulo pat lsubsts;
     List.fold_left (match_one_pat mconf env tbox pat) [] lsubsts
 
-  let pat_weight s t =
-    let sf = (Expr.term_view s).f in
-    let tf = (Expr.term_view t).f in
-    match sf, tf with
-    | Symbols.Name _, Symbols.Op _ -> -1
-    | Symbols.Op _, Symbols.Name _ -> 1
-    | _ -> Expr.(depth t - depth s)
-
   let matching mconf env tbox pat_info =
     let open Matching_types in
     let pats = pat_info.trigger in
     let pats_list = pats.E.content in
-    Options.heavy_assert (fun () -> Lists.is_sorted pat_weight pats_list);
     Debug.matching pats;
     if List.length pats_list > Options.get_max_multi_triggers_size () then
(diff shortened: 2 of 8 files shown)


src/lib/frontend/d_cnf.ml
```diff
@@ -1655,14 +1655,11 @@ and make_trigger ?(loc = Loc.dummy) ~name_base ~decl_kind
...
-    List.fold_left (fun z t -> max z (E.depth t)) 0 content in
   (* clean trigger:
      remove useless terms in multi-triggers after inlining of lets*)
-  let trigger =
-    { E.content; t_depth; semantic = []; (* will be set by theories *)
-      hyp; from_user; }
-  in
+  let trigger = E.mk_trigger ~user:from_user ~hyp content in
+  if not in_theory && not (Lists.is_empty trigger.semantic) then
+    Errors.typing_error ThSemTriggerError loc;
```
bclement-ocp, line 1662: This is added to `d_cnf` (and not `cnf`) because this check was made by the legacy typechecker but is not made by Dolmen.

tests/issues/1111.expected
Halbaroth, line 1: This file is empty? Did you forget to add the expected file?

src/lib/structures/expr.mli
```diff
@@ -91,10 +91,8 @@ and semantic_trigger =
   | IsTheoryConst of t
   | LinearDependency of t * t
 
-and trigger = (*private*) {
+and trigger = private {
```
Halbaroth, line 94: Actually I planned to do this modification only to make this struct private :)

src/lib/reasoners/matching.ml
```diff
@@ -482,19 +482,10 @@ module Make (X : Arg) : S with type theory = X.t = struct
...
-    match sf, tf with
-    | Symbols.Name _, Symbols.Op _ -> -1
-    | Symbols.Op _, Symbols.Name _ -> 1
-    | _ -> Expr.(depth t - depth s)
-
   let matching mconf env tbox pat_info =
     let open Matching_types in
     let pats = pat_info.trigger in
     let pats_list = pats.E.content in
-    Options.heavy_assert (fun () -> Lists.is_sorted pat_weight pats_list);
```
Halbaroth, line 497: Why did you remove this assert? In the previous code, we have to sort triggers in different locations, so the design was a bit fragile but now we have to ensure that `Expr` always produce sorted pattern list. We can replace this assertion by documentation.

Halbaroth (changes requested):

src/lib/reasoners/matching.ml
bclement-ocp: Because `pat_weight` is now private to the `Expr` module and I am not sure it is worth exposing just for this assert. I did forget to add documentation to `trigger.content`, will do.

tests/issues/1111.expected
bclement-ocp: Yep, should be `unknown`

Halbaroth (approved):

PR #1114, opened 2024-05-10
Remove polymorphic comparison with lists

Halbaroth:
Some functions from stdlib's list contains polymorphic comparison. We cannot use them safely on `Dolmen` identifiers.

This PR replaces these functions by new functions in `Lists` that use monomorphic comparisons.

### DIFF
--- a/src/lib/reasoners/bitv.ml
+++ b/src/lib/reasoners/bitv.ml
@@ -552,7 +552,7 @@ module Shostak(X : ALIEN) = struct
 
     exception Valid
 
-    let add elt l = if List.mem elt l then l else elt::l
+    let add elt l = if Lists.mem (equal_signed X.equal) elt l then l else elt::l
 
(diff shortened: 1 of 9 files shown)


src/lib/reasoners/bitv.ml
```diff
@@ -587,11 +587,19 @@ module Shostak(X : ALIEN) = struct
        Ensures: there are no duplicates in the result (in particular, [x = y]
                 and [y = x] cannot be both present)
        Ensures: there are no trivial equalities [x = x] in the result. *)
+
+    let equal_pair_simple_term ((a1, b1) as p1) ((a2, b2) as p2) =
+      if p1 == p2 then true
```
bclement-ocp: This is never going to be true given the way we are using this function, no need for a shortcut.

bclement-ocp (approved):

PR #1117, opened 2024-05-13
Get rid of Stdlib.compare and Stdlib.(=)

Halbaroth:
The PR #1114 isn't sufficient. There were still some polymorphic comparison in the codebase of the form `Stdlib.compare` or `Stdlib.(=)`. This PR tries to remove all of them.

### DIFF
--- a/src/lib/reasoners/bitv_rel.ml
+++ b/src/lib/reasoners/bitv_rel.ml
@@ -164,7 +164,10 @@ end = struct
     | Bor -> Fmt.pf ppf "bvor"
     | Bxor -> Fmt.pf ppf "bvxor"
 
-  let equal_binop : binop -> binop -> bool = Stdlib.(=)
+  let equal_binop op1 op2 =
+    match op1, op2 with
+    | Band, Band | Bor, Bor | Bxor, Bxor -> true
+    | _ -> false
 
   let hash_binop : binop -> int = Hashtbl.hash
 
--- a/src/lib/reasoners/bitv_rel.ml
+++ b/src/lib/reasoners/bitv_rel.ml
@@ -167,7 +167,8 @@ end = struct
   let equal_binop op1 op2 =
     match op1, op2 with
     | Band, Band | Bor, Bor | Bxor, Bxor -> true
-    | _ -> false
+    | Band, _ | _, Band -> false
+    | Bor, Bxor | Bxor, Bor -> false
 
   let hash_binop : binop -> int = Hashtbl.hash
 
--- a/src/lib/structures/expr.ml
+++ b/src/lib/structures/expr.ml
@@ -1955,9 +1955,9 @@ module Triggers = struct
     | { f = (Name _) as s1; xs=tl1; _ }, { f = (Name _) as s2; xs=tl2; _ } ->
       let l1 = List.map score_term tl1 in
       let l2 = List.map score_term tl2 in
-      let l1 = List.fast_sort Stdlib.compare l1 in
-      let l2 = List.fast_sort Stdlib.compare l2 in
-      let c  = Util.cmp_lists l1 l2 Stdlib.compare in
+      let l1 = List.fast_sort Int.compare l1 in
+      let l2 = List.fast_sort Int.compare l2 in
+      let c  = Util.cmp_lists l1 l2 Int.compare in
       if c <> 0 then c
       else
         let c = Sy.compare s1 s2 in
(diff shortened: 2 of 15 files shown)


src/lib/reasoners/bitv_rel.ml
```diff
@@ -164,7 +164,10 @@ end = struct
     | Bor -> Fmt.pf ppf "bvor"
     | Bxor -> Fmt.pf ppf "bvxor"
 
-  let equal_binop : binop -> binop -> bool = Stdlib.(=)
+  let equal_binop op1 op2 =
+    match op1, op2 with
+    | Band, Band | Bor, Bor | Bxor, Bxor -> true
+    | _ -> false
```
bclement-ocp: This is not a good way to write such a comparison function as it will silently fail if we add new constructors to the `binop` type. Polymorphic comparison is OK on enumerated types, which `binop` is (and is intended to stay). I'd rather add a comment mentioning this on `binop`.

src/lib/structures/expr.ml
```diff
@@ -1990,15 +1990,15 @@ module Triggers = struct
 
     | { f = Op (Access a1) ; xs=[t1]; _ },
       { f = Op (Access a2) ; xs=[t2]; _ } ->
-      let c = Stdlib.compare a1 a2 in (* should be Hstring.compare *)
+      let c = Hstring.compare a1 a2 in (* should be Hstring.compare *)
```
bclement-ocp, line 1993:
```suggestion
      let c = Hstring.compare a1 a2 in
```

src/lib/structures/expr.ml
```diff
@@ -1990,15 +1990,15 @@ module Triggers = struct
...
       if c<>0 then c else cmp_trig_term t1 t2
 
     | { f = Op (Access _); _ }, _ -> -1
     | _, { f = Op (Access _); _ } -> 1
 
     | { f = Op (Destruct a1) ; xs = [t1]; _ },
       { f = Op (Destruct a2) ; xs = [t2]; _ } ->
-      let c = Stdlib.compare a1 a2 in (* should be Hstring.compare *)
+      let c = Hstring.compare a1 a2 in (* should be Hstring.compare *)
```
bclement-ocp, line 2001:
```suggestion
      let c = Hstring.compare a1 a2 in
```

src/lib/structures/xliteral.ml
```diff
@@ -143,7 +143,7 @@ module Make (X : OrderedType) : S with type elt = X.t = struct
 
   type t = { at : atom; neg : bool; tpos : int; tneg : int }
 
-  let compare a1 a2 = Stdlib.compare a1.tpos a2.tpos
+  let compare a1 a2 = compare a1.tpos a2.tpos
```
bclement-ocp:
```suggestion
  let compare a1 a2 = Int.compare a1.tpos a2.tpos
```

bclement-ocp (approved):
LGTM; some minor comments but I think that the change for the `binop` type should be reverted (or at least the implementation of `equal_binop` fixed to be future-proof).

src/lib/structures/expr.ml
Halbaroth: I removed it another PR ;)

src/lib/structures/expr.ml
Halbaroth: Same here

src/lib/reasoners/bitv_rel.ml
Halbaroth: I know that polymorphic comparison is ok on enum types. I removed it because if someone adds a new constructor to it, maybe this constructor will contain a payload with a Dolmen identifier.

bclement-ocp: I understand. What I am saying is that this should not happen: this particular type is designed to be an enumerated type, and it should never contain Dolmen identifiers. Hence, documenting that fact is a better fix

(Furthermore, I don't think the change is currently "Dolmen identifier proof" because I don't think the `hash` function below will work properly with Dolmen identifiers either)

Halbaroth: I think removing the polymorphic comparison is a better fix. 

I know the hash isn't correct.

src/lib/structures/expr.ml
bclement-ocp: I may be missing something but given that this is the PR that fixes the issue identified by the comment, it should also be the one removing the comment.

Halbaroth: I know but in fact this PR is extracted from #1098 . I can do it twice if you want...

bclement-ocp: It would have been better to avoid the inconsistent state one way or another. That said, this being replaced by `Uid.compare` (I suppose?) in that PR makes sense; I have to review PRs as they are and can't really infer what "another PR" refers to — thanks for the clarification.

PR #1119, opened 2024-05-13
feat(UF): Store domains inside the union-find

bclement-ocp:
This patch adds the ability to store domains (as used in the bit-vector, enum and ADT theories) directly in the union-find structure, which is responsible for updating them automatically as class representatives are updated.

The advantages of doing this are twofold: we no longer need to manually keep track of substitutions (avoiding the issue that some of them are not propagated from the union-find to the relations), and having the domains stored in a central place rather than inside each relations module opens the path to more communication between the domains (e.g. accessing integer domains from the bit-vector module).

Note that constraints (notably for the bit-vector theory) are *not* integrated into the union-find and still need to be kept track of manually.

### DIFF
--- a/src/lib/reasoners/adt_rel.ml
+++ b/src/lib/reasoners/adt_rel.ml
@@ -167,9 +167,9 @@ module Domains = struct
       internal_update r nd t
     | Alien _ | Constr _ | Select _ -> t
 
-  (** [tighten r d t] replaces the domain of [r] in [t] by a domain [d] contains
-      in the current domain of [r]. The representative [r] is marked [changed]
-      after this call if the domain [d] is strictly smaller. *)
+  (** [tighten r d t] replaces the domain of [r] in [t] by a domain [d]
+      contained in the current domain of [r]. The representative [r] is marked
+      [changed] after this call if the domain [d] is strictly smaller. *)
   let tighten r d t =
     let od = get r t in
     (* For sake of completeness, the domain [d] has to be a subset of the old
@@ -268,7 +268,7 @@ let empty uf = {
   delayed = Rel_utils.Delayed.create ~is_ready dispatch;
   size_splits = Numbers.Q.one;
   new_terms = SE.empty;
-}, Uf.Domains.add (module Domains) Domains.empty (Uf.domains uf)
+}, Uf.GlobalDomains.add (module Domains) Domains.empty (Uf.domains uf)
 
 (* ################################################################ *)
 (*BISECT-IGNORE-BEGIN*)
@@ -520,7 +520,7 @@ let count_splits env la =
 
 let assume env uf la =
   let ds = Uf.domains uf in
-  let domains = Uf.Domains.find (module Domains) ds in
+  let domains = Uf.GlobalDomains.find (module Domains) ds in
   Debug.pp_domains "before assume" domains;
   (* should be done globally in CCX *)
   let la = remove_redundancies la in
@@ -541,7 +541,7 @@ let assume env uf la =
   in
   Debug.pp_domains "after assume" domains;
   env,
-  Uf.Domains.add (module Domains) domains ds,
+  Uf.GlobalDomains.add (module Domains) domains ds,
   Sig_rel.{ assume; remove = [] }
 
 (* ################################################################ *)
@@ -598,7 +598,7 @@ let case_split env uf ~for_model =
     assert (not for_model);
     if Options.get_debug_adt () then
       Debug.pp_domains "before cs"
-        (Uf.Domains.find (module Domains) (Uf.domains uf));
+        (Uf.GlobalDomains.find (module Domains) (Uf.domains uf));
     match pick_delayed_destructor env with
     | Some (r, d) ->
       if Options.get_debug_adt () then
@@ -619,7 +619,7 @@ let optimizing_objective _env _uf _o = None
 let query _env uf (ra, _, ex, _) =
   if Options.get_disable_adts () then None
   else
-    let domains = Uf.Domains.find (module Domains) (Uf.domains uf) in
+    let domains = Uf.GlobalDomains.find (module Domains) (Uf.domains uf) in
     try
       match ra with
       | Xliteral.Builtin(true, Sy.IsConstr c, [r]) ->
--- a/src/lib/reasoners/adt_rel.ml
+++ b/src/lib/reasoners/adt_rel.ml
@@ -108,6 +108,12 @@ module Domain = struct
     domain ~constrs ex
 end
 
+let is_adt_ty = function
+  | Ty.Tadt _ -> true
+  | _ -> false
+
+let is_adt r = is_adt_ty (X.type_info r)
+
 module Domains = struct
   (** The type of simple domain maps. A domain map maps each representative
       (semantic value, of type [X.r]) to its associated domain. *)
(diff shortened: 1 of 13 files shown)


bclement-ocp:
After the usual 5.0 cross-check, this PR is +0-2 on ae-format (but the -2 seems spurious/weak-table-related; at least they don't reproduce on my machine) and +0-17 on the QF_BV subset. The regressions on the QF_BV subset seem to be related to changes in the way the splits are performed; they disappear when increasing the `--max-split` limit or when using `--enable-sat-cs` (with #1118 applied). I think we can treat these as a particular case of #1120, as also discussed in #1027.

src/lib/reasoners/uf.ml
```diff
@@ -78,6 +78,87 @@ end
...
+
+type 'a domains = (module Domains with type t = 'a)
+
+module Domains = struct
+  type (_, _) eq = Equal : ('a, 'a) eq
+
+  let[@inline] uid (type a) ((module Doms) : a domains) =
+    Obj.Extension_constructor.id (Obj.Extension_constructor.of_val Doms.Id)
+
+  module MapI = Map.Make(Int)
```
Halbaroth: Let's use `Util.MI`;)

src/lib/reasoners/uf.mli
```diff
@@ -34,6 +34,77 @@ type t
...
+
+      The explanation [ex] justifies the equality [rr = nrr].
+
+      @raise Inconsistent if the domains for [rr] and [nrr] are incompatible. *)
+end
+
+type 'a domains = (module Domains with type t = 'a)
+(** The type for global domain modules with storage type ['a]. *)
+
+module Domains : sig
```
Halbaroth: It's a bit confusing to use the same name for the signature `Domains` and the module `Domains`.
The signature `Domains` is a signature for sets of domains with a common unique identifier.
The module `Domains` is actually a set of domains with different unique identifiers that we represent as a map of sets (with common unique identifier) for efficiency purposes. We should mention that in the documentation.

src/lib/reasoners/sig_rel.mli
```diff
@@ -54,10 +54,11 @@ module type RELATION = sig
 
   val timer : Timers.ty_module
 
-  val empty : Expr.Set.t list -> t
+  val empty : Uf.t -> t * Uf.Domains.t
```
Halbaroth:
```suggestion
  val empty : Uf.t -> t * Uf.Domains.t
  (** [empty uf] creates a new environment for this relation theory and registers domains' identifiers in [uf]. The second value returned is used as an accumulator in [Relation.empty]. *)
```

src/lib/reasoners/adt_rel.ml
```diff
@@ -148,16 +158,14 @@ module Domains = struct
...
+    | Alien _ when not (MX.mem r t.domains) ->
+      (* We have to add a default domain if the key `r` isn't in map in order
+          to be sure that the case-split mechanism will attempt to choose a
+          value for it. *)
+      let nd = Domain.unknown (X.type_info r) in
+      internal_update r nd t
+    | Alien _ | Constr _ | Select _ -> t
 
   (** [tighten r d t] replaces the domain of [r] in [t] by a domain [d] contains
       in the current domain of [r]. The representative [r] is marked [changed]
```
Halbaroth:
```suggestion
  (** [tighten r d t] replaces the domain of [r] in [t] by a domain [d] contained
      in the current domain of [r]. The representative [r] is marked [changed]
```

src/lib/reasoners/sig_rel.mli
```diff
@@ -54,10 +54,11 @@ module type RELATION = sig
...
 
   val assume : t ->
-    Uf.t -> (Shostak.Combine.r input) list -> t * Shostak.Combine.r result
+    Uf.t -> (Shostak.Combine.r input) list ->
+    t * Uf.Domains.t * Shostak.Combine.r result
```
Halbaroth:
```suggestion
    Uf.t -> (Shostak.Combine.r input) list ->
    t * Uf.Domains.t * Shostak.Combine.r result
    (** [assume env uf la] assumes the annotated literals [la] in the environment [env] and propagated these constraints in the domains of [uf]. The second value returned 
    is an accumulator used in [Rel.assume]. *)
```

src/lib/reasoners/uf.mli
```diff
@@ -34,6 +34,77 @@ type t
...
+      no longer be tracked and the intersection of domains should be associated
+      with [nrr].
+
+      The explanation [ex] justifies the equality [rr = nrr].
+
+      @raise Inconsistent if the domains for [rr] and [nrr] are incompatible. *)
+end
+
+type 'a domains = (module Domains with type t = 'a)
+(** The type for global domain modules with storage type ['a]. *)
```
Halbaroth: `global domain` is confusing. This is a set of domains actually, not a domain. We may say `set of domains of common identifier` or `common kind` if we rename `Id` to `Kind`.

src/lib/reasoners/rel_utils.ml
```diff
@@ -388,16 +370,19 @@ struct
     (** Map from leaves to the *tracked* representatives that contains them *)
   }
 
+  type _ Uf.id += Id : t Uf.id
```
Halbaroth, line 373: I think `Kind` is a better name here.

Halbaroth (approved):
Looks very good for me. I've very few remarks.

src/lib/reasoners/uf.mli
bclement-ocp: The `s` means different things in these cases — in the module signature it means that there are domains for multiple variables, and in the module it means that there are domains of different kinds (for multiple variables, but `Domainss` does not really work). I will rename the `Domains` signature and `domains` type into `GlobalDomain` and `global_domain`, and the `Domains` module into `GlobalDomains` — I think this makes sense and leaves some space to introduce `LocalDomain` / `local_domain` / `LocalDomains` later if needed.

src/lib/reasoners/uf.mli
bclement-ocp: It is a domain (i.e. set of possible values) for the tuple comprising all of the currently defined variables (hence "global"). It so happens that all the implementations we currently have are represented as a product of per-variable ("local") domains, but I would rather not enshrine that fact in the documentation since one of the motivations is to use this API for the "delta" domain of [Sharpening Constraint Programming approaches for Bit-Vector Theory](https://cea.hal.science/cea-01795779/document).

I can clarify the documentation to mention that a global domain encodes a set of (correlated) values for multiples variables.

src/lib/reasoners/rel_utils.ml
bclement-ocp: I am not convinced — `kind` makes it seem like there can be multiple domain modules of the same kind but the goal is that each domain has a single unique ID distinct from that of all other domain modules.

What I think of as a "domain's kind" is rather the module itself, and the `Id` constructor is just an identifier for the module (i.e. an identifier for the kind), so it would be weird to have a kind for the kind — if that makes sense.

src/lib/reasoners/uf.mli
Halbaroth: > I can clarify the documentation to mention that a global domain encodes a set of (correlated) values for multiples variables.

Yes, please ;)

Issue #1123, opened 2024-05-14
long running time with low steps


kanigsson:
Hello, 

[repr.ae.txt](https://github.com/OCamlPro/alt-ergo/files/15304733/repr.ae.txt)

On the attached reproducer, Alt-Ergo runs for a long time, despite specifying a low steps number:
```
$ alt-ergo repr.ae --steps-bound=200
... runs for several minutes ...
```

Probably some code doesn't increment the steps counter?

Halbaroth:
Thanks for reporting this issue. 

It seems there are at least two issues here (and maybe three?):
1. We don't increment steps in the reasoner of the `Arrays` theory.
2. The `Arrays` theory slows down Alt-Ergo in presence of very long terms, even if the input file doesn't involve terms of this theory, as it's the case for your example.
3. Alt-Ergo produces huge terms for this input file.

- In #1126, I wrote a quick fix for the second point and now Alt-Ergo solves your example immediately. The fix consists in caching the function responsible of the bottleneck in `Arrays`. The purpose of this function is to register relevant terms for the instantiation engine.. The current implementation explores all the terms of the problem to find `select` or `store` calls, which is very slow in presence of long terms.
- To solve the first point, I plan to add a counter for the `Arrays` theory but I don't plan to increment them in the function `update_gets_sets`.
- I didn't investigate the last point.

bclement-ocp:
Fixed by #1126.

PR #1126, opened 2024-05-15
Cache the slow function `update_gets_sets`

Halbaroth:
The function `update_gets_sets` of `Arrays_rel` explores recursively all the terms of the problem appearing in literals of `Rel.assumes`. This function becomes a bottle neck if we have very large terms as in the issue #1123. For instance, the input file of this issue is solved on `next` in several minutes but instantly after this patch.

### DIFF
--- a/src/lib/reasoners/arrays_rel.ml
+++ b/src/lib/reasoners/arrays_rel.ml
@@ -80,6 +80,8 @@ end
 
 let timer = Timers.M_Arrays
 
+module H = Ephemeron.K1.Make (Expr)
+
 type t =
   {gets  : G.t;               (* l'ensemble des "get" croises*)
    tbset : S.t TBS.t ;        (* map t |-> set(t,-,-) *)
@@ -88,6 +90,7 @@ type t =
    seen  : E.Set.t Tmap.t;    (* combinaisons (get,set) deja splitees *)
    new_terms : E.Set.t;
    size_splits : Numbers.Q.t;
+   cached_relevant_terms : (G.t * S.t TBS.t) H.t;
   }
 
 
@@ -99,6 +102,7 @@ let empty uf =
    seen  = Tmap.empty;
    new_terms = E.Set.empty;
    size_splits = Numbers.Q.one;
+   cached_relevant_terms = H.create 1024;
   }, Uf.domains uf
 
 (*BISECT-IGNORE-BEGIN*)
@@ -184,12 +188,12 @@ let merge_revelant_terms (gets, tbset) (g, t) =
    presence of very large terms in the problem.
 
    See issue https://github.com/OCamlPro/alt-ergo/issues/1123 *)
-let rec relevant_terms t =
+let rec relevant_terms env t =
   let { E.f; xs; ty; _ } = E.term_view t in
   let gets, tbset =
     List.fold_left
       (fun acc x ->
-         merge_revelant_terms acc (cached_relevant_terms x)
+         merge_revelant_terms acc (cached_relevant_terms env x)
       ) (G.empty, TBS.empty) xs
   in
   match Sy.is_get f, Sy.is_set f, xs with
@@ -199,16 +203,13 @@ let rec relevant_terms t =
   | false, false, _ -> (gets,tbset)
   | _  -> assert false
 
-and cached_relevant_terms =
-  let module H = Ephemeron.K1.Make (Expr) in
-  let cache : (G.t * S.t TBS.t) H.t = H.create 1024
-  in fun t ->
-    match H.find cache t with
-    | r -> r
-    | exception Not_found ->
-      let r = relevant_terms t in
-      H.add cache t r;
-      r
+and cached_relevant_terms env t =
+  match H.find env.cached_relevant_terms t with
+  | r -> r
+  | exception Not_found ->
+    let r = relevant_terms env t in
+    H.add env.cached_relevant_terms t r;
+    r
 
 (* met a jour les composantes gets et tbset de env avec les termes
    contenus dans les atomes de la *)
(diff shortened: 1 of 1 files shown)


Halbaroth:
The previous patch contained a bug. I prevented from caching the result of `update_gets_sets` with its accumulator at the toplevel of the recursive function but I haven't been careful at some recursive calls. This mistake was detected by only one test in our test suite! 

I believe the new fix is correct but we should check it a last time on Marvin.

Halbaroth:
+0-0 on both `ae-format` on `QF_AX` :)

Halbaroth:
@bclement-ocp ping?

src/lib/reasoners/arrays_rel.ml
```diff
@@ -166,25 +166,59 @@ module Debug = struct
...
   match Sy.is_get f, Sy.is_set f, xs with
   | true , false, [a;i]   -> G.add {g=t; gt=a; gi=i; gty=ty} gets, tbset
   | false, true , [a;i;v] ->
     gets, TBS.add a {s=t; st=a; si=i; sv=v; sty=ty} tbset
   | false, false, _ -> (gets,tbset)
   | _  -> assert false
 
+and cached_relevant_terms =
+  let module H = Ephemeron.K1.Make (Expr) in
+  let cache : (G.t * S.t TBS.t) H.t = H.create 1024
```
bclement-ocp: I think this breaks Alt-Ergo-Fuzz and makes @hra687261 sad

bclement-ocp (commented):
Oof I missed when skimming that code that we were doing the recursion on *terms* and not semantic values.

If I am not mistaken, we don't need the cache to be global and could store the table in the environment here? Since it is just a cache it is OK if it is mutable state inside the environment, and this would avoid breaking Alt-Ergo-Fuzz (in addition to avoiding more global state).

src/lib/reasoners/arrays_rel.ml
hra687261: Thanks for keeping it in mind!
I suppose it can be kept in the environment or it can expose a cache resetting function which should be called in the right places.
But the first solution seems cleaner. I would assume that performance differences would be negligible.

Halbaroth: @hra687261 Out of sheer curiosity, why is it important for a fuzzing tool?

Halbaroth:
> Oof I missed when skimming that code that we were doing the recursion on _terms_ and not semantic values.
> 
> If I am not mistaken, we don't need the cache to be global and could store the table in the environment here? Since it is just a cache it is OK if it is mutable state inside the environment, and this would avoid breaking Alt-Ergo-Fuzz (in addition to avoiding more global state).

Using a cache per environment seems a bit weird for me. I agree that a global cache is ugly but at least the environment of theories contains only information about terms of the current problem. The purpose of this cache is to store old relevant terms discovered while exploring some branches with the SAT solver and these terms aren't necessary relevant in the current branch.

We currently stores global caches in `IntervalCalculus` and we have `reinit_ctx` in SAT solvers to reinitialize them. If we use local caches, it's not obvious where and when we should reinitialize these caches.

src/lib/reasoners/arrays_rel.ml
hra687261: It uses Alt-Ergo as a library, and to make sure that two independent runs have the same behavior, it is necessary to be able to reset and the resolution context (hash-tables, refs, environment). Otherwise the fuzzing doesn't work. (cf. https://github.com/OCamlPro/alt-ergo/pull/490)

hra687261:
(IMO: adding a way to reinitialize the cache (And calling from the SAT solvers) should also work)

bclement-ocp:
> Using a cache per environment seems a bit weird for me. I agree that a global cache is ugly but at least the environment of theories contains only information about terms of the current problem. The purpose of this cache is to store old relevant terms discovered while exploring some branches with the SAT solver and these terms aren't necessary relevant in the current branch.

As long as we do not ask "is this key in the cache" (I think we can't do this with the ephemeron API anyways), this is not a problem. As an aside, the fact that "the current problem" is a thing that exists as a global state is something we are trying to move away (see #377).

The ideal solution would be to only cache the values for the current problem — this would require removing entries from the table as we backtrack, and we don't currently have a mechanism for that.

> 
> We currently stores global caches in `IntervalCalculus` and we have `reinit_ctx` in SAT solvers to reinitialize them. If we use local caches, it's not obvious where and when we should reinitialize these caches.

The point of using local caches is that we do not have to reinitialize them. We need to reinitialize the global caches because they are global state and hence a previous run can affect a subsequent run — but since each run has its own environment, using local caches will not have this issue.

(FTR, I am fine with using context reinitialization hooks as this is what we currently do elsewhere; storing the info on the local environment is easier to do & simpler to migrate to a backtracking table when we get the ability to do so hence why I suggested it)

Halbaroth:
> As long as we do not ask "is this key in the cache" (I think we can't do this with the ephemeron API anyways), this is not a problem. As an aside, the fact that "the current problem" is a thing that exists as a global state is something we are trying to move away (see #377).

I know using a global state isn't the best design. I didn't know I don't need to flush the cache with your solution.

> The ideal solution would be to only cache the values for the current problem — this would require removing entries from the table as we backtrack, and we don't currently have a mechanism for that.

Not sure to understand your point here. By `current problem`, I meant the current state of the SAT. Let's image we have a formula `A \/ B` and we have explored `A` and now we explore `B` after backtracking. Without cache or with cache that follows the backtracking mechanism of the SAT, we have to explore completely all terms of `B`even if we have explored some of them in `A`. Actually the `new_terms` field of `Arrays_rel` already contains all the relevant terms of the current branch and we can use it as a cache (we replace it by a functional map). It may be sufficient for the submitted input file. I will test.

I thought a bit about this problem last week and I think that an acceptable design consists in attaching relevant terms to the semantic values (or terms?) directly. So the cache we use for `X.make` would store relevant terms too. I didn't implement this solution because I'm not sure there is no better ways and I prefer a minimal patch before the next release.

Halbaroth:
I replaced the global cache by a local cache. Still working with the input file of #1123 but I can run again benchmarks?

Halbaroth:
+0-0 on `ae-format` again :)

bclement-ocp (approved):

hra687261 (approved):

PR #1129, opened 2024-05-17
Fix locales for Nix shell

Halbaroth:
The locales aren't correctly supported by Nix on non-NixOs system. See https://nixos.wiki/wiki/Locales

I cannot compile with Nix the Sphinx documentation without this fix...

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -5,6 +5,7 @@ let
 in
 
 pkgs.mkShell {
+  LOCALE_ARCHIVE = "${pkgs.glibcLocales}/lib/locale/locale-archive";
   nativeBuildInputs = with ocamlPackages; [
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -5,6 +5,7 @@ let
 in
 
 pkgs.mkShell {
+  LOCALE_ARCHIVE = "${pkgs.glibcLocales}/lib/locale/locale-archive";
```
bclement-ocp, line 10: This looks magical; can you add a comment explaining why this is needed (e.g. a link to https://nixos.wiki/wiki/Locales)?

Halbaroth: Sure

bclement-ocp (approved):

Issue #933, opened 2023-11-15
Missing value for option `--input` for polymorphic SMT syntax


claudemarche:
With the Dolmen frontend of Alt-Ergo, it is possible to give an input file in the SMT syntax. It even supports the upcoming 3.0 version of the syntax which allows one to add prenex type quantification on hypotheses. Yet, Alt-Ergo will accepts the polymorphic syntax only if the given file has the `.psmt2` extension.

It should be possible though to use this frontend parser even of the input file has another extension, or if input is taken from standard input, say, by passing the option `--input`. Unfortunately there does not seem to have a possible value for this option corresponding to `psmt2`. Please add it!

bclement-ocp:
Thanks for the report! 

With the legacy frontend, Alt-Ergo *only* supported psmt2 (so with the legacy frontend we actually parse all smt2 files as psmt2). Hence, historically, the `--input smtlib2` for Alt-Ergo actually meant "psmt2".

When writing the Dolmen frontend we interpreted `--input smtlib2` as requesting "plain" smt2, not "psmt2". But given that "psmt2" is an extension of smt2, all valid smt2 files should also be valid psmt2 files. I would prefer interpreting "smtlib2" as requesting psmt2 format, like was done in the legacy frontend: I am not sure there are benefits of having a strict adherence to the standard here, and it would streamline things a bit.

@claudemarche would that solution be acceptable for you, or is having separate `--input` options for "polymorphic smt2" and "non-polymorphic smt2" useful?

claudemarche:
In fact I don't understand anymore. At first, it seemed to me that Alt-Ergo was behaving differently on `f.smt2` and `f.psmt2` with the same contents for these files. Now I cannot see the difference anymore.
Is it the case that it may behave differently?
If it makes no difference at all, then we should just forget about all that and just get rid of this extension 'psmt2' which is meaningless to me

bclement-ocp:
Alt-Ergo should behave identically on `.smt2` and `.psmt2` files _that do not use polymorphism_; if it doesn't, that is a bug (either on our end or in Dolmen) and we will fix it.

If the file uses polymorphism, it currently depends on the frontend:
 - With the (deprecated) legacy frontend, it should behave identically
 - With the Dolmen frontend, polymorphic syntax in a `.smt2` file should be rejected

[...]

PR #1132, opened 2024-05-27
fix(CLI): Do not use psmt2 format for .smt2 file

bclement-ocp:
Following discussions at the Alt-Ergo Club meeting last week, it was agreed that it was better to support SMT-LIB2 format by default for ".smt2" files and to only enable the polymorphic extensions for ".psmt2" files.

This patch performs that change, partially reverting #948 / 42e9ef10dca19490b502db3b5f323aeb0417100e and uses the same polymorphic variant type as Dolmen for the same purpose.

In order to accomodate #933, introduce a new "psmt2" input mode.

### DIFF
--- a/src/lib/util/options.mli
+++ b/src/lib/util/options.mli
@@ -55,14 +55,27 @@ type interpretation =
   | ILast        (** Compute only the last interpretation just before
                      returning SAT/Unknown *)
 
+type smtlib2_version =
+  [ `Latest
+  (** Latest version of the SMT-LIB standard. *)
+  | `V2_6
+  (** The SMT-LIB standard: Version 2.6
+      https://smt-lib.org/papers/smt-lib-reference-v2.6-r2021-05-12.pdf *)
+  | `Poly
+    (** Polymorphic extension of the SMT-LIB standard.
+
+        See https://inria.hal.science/hal-01960203/document *)
+  ]
+(** Version of the SMT-LIB standard used. *)
+
 (** Type used to describe the type of input wanted by
     {!val:set_input_format} *)
 type input_format =
   | Native                     (** Native Alt-Ergo format  *)
-  | Smtlib2
-  (** {{:
-      http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf}
-      Smtlib} default format *)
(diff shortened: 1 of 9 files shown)


src/lib/util/options.mli
```diff
@@ -55,14 +55,27 @@ type interpretation =
   | ILast        (** Compute only the last interpretation just before
                      returning SAT/Unknown *)
 
+type smtlib2_version =
+  [ `Latest
+  (** Latest version of the SMT-LIB standard. *)
+  | `V2_6
+  (** The SMT-LIB standard: Version 2.6
+      https://smt-lib.org/papers/smt-lib-reference-v2.6-r2021-05-12.pdf *)
```
Halbaroth: We should use the link syntax of `odoc` here: https://ocaml.github.io/odoc/odoc_for_authors.html#links-and-references

src/lib/util/options.mli
```diff
@@ -55,14 +55,27 @@ type interpretation =
...
+  | `Poly
+    (** Polymorphic extension of the SMT-LIB standard.
+
+        See https://inria.hal.science/hal-01960203/document *)
```
Halbaroth: Same here

src/lib/util/options.mli
```diff
@@ -55,14 +55,27 @@ type interpretation =
...
 type input_format =
   | Native                     (** Native Alt-Ergo format  *)
-  | Smtlib2
-  (** {{:
-      http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf}
-      Smtlib} default format *)
+  | Smtlib2 of smtlib2_version
+  (** SMT-LIB default format.
+
+      See https://smt-lib.org/language.shtml. *)
```
Halbaroth: Same here

Halbaroth (approved):

src/lib/util/options.mli
bclement-ocp: This was written as is because I intended it to render as the full URL. I can make it a text link.

Halbaroth: I thought it wasn't clickable without this syntax. I'm ok with both solution ;)

PR #1133, opened 2024-05-27
fix(CLI): Remove `--optimize` option in favor of `--strict`

bclement-ocp:
Following discussions at the Club's meeting last week, it was decided to enable optimization statements (`maximize`, `minimize`, and `get-objectives`) by default and to forbid their usage through a different option to enable strict mode with SMT-LIB spec compliance.

Currently, the `--strict` option only disables optimization. It should also disable extensions such as `ae.round` and `bv2nat` / `int2bv`.

This patch implements that change. Note that the `--optimize` option did not go into any release yet.


tools/gentest.ml
```diff
@@ -398,8 +398,7 @@ let () =
      ; "--no-minimal-bj" ])
   ; ("runtest", "optimize",
      [ "--output=smtlib2"
-     ; "--frontend dolmen"
-     ; "--optimize" ])]
+     ; "--frontend dolmen" ])]
```
Halbaroth: We should remove the `optimize` tag in our test suite.

src/bin/common/solving_loop.ml
```diff
@@ -534,33 +535,15 @@ let main () =
...
 
-  let set_optimize optim st =
-    let sat = DO.SatSolver.get st in
-    match sat with
-    | Util.Tableaux | Tableaux_CDCL when optim ->
-      warning
-        "Sat-solver %a is incompatible with optimization: ignoring command."
-        Util.pp_sat_solver
-        sat;
-      st
```
Halbaroth, line 549: Just a remainder. As we don't explore different SAT branches in `FunSAT` and we don't force the SAT solver to be `SatML` in presence of the MaxSMT extension, it means the feature is broken in presence of `--sat-solver tableaux` for instance. We should mention it at least in the documentation of AE for the next release.

Halbaroth (changes requested):
Looks good to me.

src/bin/common/solving_loop.ml
bclement-ocp: The Tableaux solver errors out if it receives an optimization command (see implementation of `Fun_sat_frontend.optimize`).

bclement-ocp:
> Looks good to me.

You marked the PR as "changes requested" so I am a bit confused by this comment.

tools/gentest.ml
bclement-ocp: Done.

src/bin/common/solving_loop.ml
Halbaroth: Uhm I'm not sure we should crash with an error here. We don't produce wrong model but we don't produce the best one (according to our definition of best one).

Halbaroth:
> > Looks good to me.
> 
> You marked the PR as "changes requested" so I am a bit confused by this comment.


:D my request changes weren't essential.

src/bin/common/solving_loop.ml
bclement-ocp: I agree that this is a bad error mechanism, but I'd argue that returning an error here is appropriate — the user is requesting an optimized model, but we can't satisfy that request. I'll change the mechanism when we encounter an `optimize` in the Tableaux solver.

bclement-ocp:
> :D my request changes weren't essential.

In that case it is better to mark as "Approved" — marking as "Changes requested" prevents the PR from being merged until you (or someone else) later approve(s).

bclement-ocp:
@Halbaroth Added a test to return an `(error)` reply when trying to use `(maximize)`, `(minimize)`, or `(get-objectives)` with a Tableaux-based solver. Whether or not we stop or ignore the corresponding statement depends on the `--continue-on-error` flag, as usual for errors per the SMT-LIB standard.

Halbaroth (approved):

PR #1138, opened 2024-06-03
Case splits on enum domains

Halbaroth:
This PR changes the way we perform split on ADT domains.
Before this PR, after asserting formulas, we can split on ADT semantic
values of enum type, that is all its constructors have no payload.
After this patch, we can split on an ADT domain if all its constructors
are enum constructors, that is they have no payload.

### DIFF
--- a/src/lib/reasoners/adt_rel.ml
+++ b/src/lib/reasoners/adt_rel.ml
@@ -139,7 +139,7 @@ module Domains = struct
     (** Set of tracked representatives whose the domain only contains
         enum constructors, that is constructors without payload.
 
-        We can split on these values after asserting new formulas. *)
+        This field is used by the case split mechanism, see [pick_enum]. *)
 
     changed : SX.t;
     (** Representatives whose domain has changed since the last flush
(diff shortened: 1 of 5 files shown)


Halbaroth:
As expected, this modification introduces regressions but improve a bit the total wall time. I'll probably close this PR soon.

src/lib/reasoners/adt_rel.ml
```diff
@@ -134,7 +136,10 @@ module Domains = struct
         We don't store domains for constructors and selectors. *)
 
     enums: SX.t;
-    (** Set of tracked representatives of enum type. *)
+    (** Set of tracked representatives whose the domain only contains
+        enum constructors, that is constructors without payload.
+
+        We can split on these values after asserting new formulas. *)
```
bclement-ocp: Let's simply mention case splits or the `case_split` function here. The specific condition under which we are performing case splits is an implementation detail of the SAT solver that the theories know nothing about, and "after asserting new formulas" only correspond to the behavior for a specific value of the `--case-split-policy` option.

bclement-ocp (approved):
LGTM but we should not talk about "[splitting] after asserting new formulas" because the specific condition changes depending on options.

PR #1140, opened 2024-06-05
Fix documentation of `make_triggers`

Halbaroth:
While trying to understand the issue 1139, I realized that my documentation of `make_triggers` in `Expr` was completely wrong. I misunderstood the meaning of the flag `escaped_vars` in the module `Triggers`. This PR fixes and clarifies this documentation.

### DIFF
--- a/src/lib/structures/expr.ml
+++ b/src/lib/structures/expr.ml
@@ -1869,6 +1869,7 @@ module Triggers = struct
 
   module Svty = Ty.Svty
 
+  (* Set of patterns with their sets of free term and type variables. *)
   module STRS =
     Set.Make(
     struct
@@ -2115,15 +2116,12 @@ module Triggers = struct
         if c <> 0 then c else Var.Set.compare y1 y2
     end)
 
-  (* let's [e] be a subterm of a quantified formula [f] and [bv] the set
-     of quantified variables of [f]. Then [underscore bv e] returns the term
-     obtained by substituting all the free term variables in [e] that aren't
-     in [bv] by the underscore term.
+  (* [underscore bv e] replaces all the free term variables in [e] that
+     aren't in [bv] by the underscore term.
 
-     For instance, if [f] is the formula:
-       forall x:int, forall y:int, g(x, y, z) = 0.
-     where the free variable of [g(x, y, z)] are exactly [{x, y, z}],
-     and [e = g(x, y, z)], then this function returns the term [g(x, y, _)]. *)
+     For instance with [bv = {x, y}] and [e = g(x, y, z)] where [{x, y, z}]
+     is the set of free term variables of [e], this functions returns the term
+     [g(x, y, _)]. *)
   (* TODO: rename this function. *)
   let underscore =
     let aux t s =
@@ -2145,6 +2143,12 @@ module Triggers = struct
         let vt = Var.Set.add Var.underscore (Var.Set.inter vt bv) in
         t,vt,vty
 
+  (* [parties mconf bv vty l escaped_vars] generates all the multi-triggers such
+     that their patterns lie in [l] and they cover all the free variables [bv]
+     and [vty].
+
+     If [escaped_vars] is [true], replace the free term variables of these
+     patterns that do not lie in [bv]. *)
   let parties mconf bv vty l escaped_vars =
     let l =
       if mconf.Util.triggers_var then l
@@ -2165,6 +2169,8 @@ module Triggers = struct
                  let vty3 = Svty.union vty2 vty1 in
                  let e = t::l, bv3, vty3 in
                  if Var.Set.subset bv bv3 && Svty.subset vty vty3 then
+                   (* The multi-trigger [e] cover all the free variables [bv]
+                      and [vty]. *)
                    llt, SLLT.add e llt_ok
                  else
                    SLLT.add e llt, llt_ok
@@ -2178,28 +2184,39 @@ module Triggers = struct
     let l = STRS.elements s in (* remove redundancies in old l *)
     SLLT.elements (parties_rec (SLLT.empty, SLLT.empty) l)
 
+  (* Simplify the multi-trigger [l] by removing a pattern [p] in [l] if
+     its set of free (term and type) variables:
+     - isn't maximal (for the inclusion) among the set of free variables of
+         patterns in [l];
+     - contains the union of [bv_a] and [vty_a];
+     - is disjoint with the union of [bv_a] and [vty_a]. *)
   let simplification =
-    let strict_subset bv vty =
+    (* Check if there is a pattern in [l] whose the set of free variables
+       contains strictly the union of [bv] and [vty]. *)
+    let strict_subset bv vty l =
       List.exists
         (fun (_, bv',vty') ->
            (Var.Set.subset bv bv' && not(Var.Set.equal bv bv')
             && Svty.subset vty vty')
            || (Svty.subset vty vty' && not(Svty.equal vty vty')
-               && Var.Set.subset bv bv') )
-    in
-    let rec simpl_rec bv_a vty_a acc = function
-      | [] -> acc
-      | ((_, bv, vty) as e)::l ->
-        if strict_subset bv vty l || strict_subset bv vty acc ||
-           (Var.Set.subset bv_a bv && Svty.subset vty_a vty) ||
-           (Var.Set.equal (Var.Set.inter bv_a bv) Var.Set.empty &&
-            Svty.equal (Svty.inter vty_a vty) Svty.empty)
-        then simpl_rec bv_a vty_a acc l
-        else  simpl_rec bv_a vty_a (e::acc) l
+               && Var.Set.subset bv bv') ) l
     in fun bv_a vty_a l ->
-      simpl_rec bv_a vty_a [] l
+      let rec simpl_rec acc = function
+        | [] -> acc
+        | ((_, bv, vty) as e)::l ->
+          if strict_subset bv vty l || strict_subset bv vty acc ||
+             (Var.Set.subset bv_a bv && Svty.subset vty_a vty) ||
+             (Var.Set.equal (Var.Set.inter bv_a bv) Var.Set.empty &&
+              Svty.equal (Svty.inter vty_a vty) Svty.empty)
+          then simpl_rec acc l
+          else simpl_rec (e::acc) l
+      in
+      simpl_rec [] l
 
   let multi_triggers menv bv vty trs escaped_vars =
+    (* The simplification removed all the patterns of the multi-trigger [trs]
+       that cover all the free variables [bv] and [vty]. Indeed, such patterns
+       have already been generated as mono-trigger before. *)
     let terms = simplification bv vty trs in
     let l_parties = parties menv bv vty terms escaped_vars in
     let lm = List.map (fun (lt, _, _) -> lt) l_parties in
@@ -2212,6 +2229,8 @@ module Triggers = struct
     at_most menv.Util.nb_triggers m
 
   let mono_triggers menv vterm vtype trs =
+    (* We only keep mono-trigger that cover all the free variables [vterm]
+       and [vtype]. *)
     let mono = List.filter
         (fun (_, bv_t, vty_t) ->
            Var.Set.subset vterm bv_t && Svty.subset vtype vty_t) trs
@@ -2296,6 +2315,9 @@ module Triggers = struct
   let free_vars_as_set e =
     Var.Map.fold (fun v _ s -> Var.Set.add v s) e.vars Var.Set.empty
 
+  (* Collect all the subterms of the expression [e] that
+     are pure and contain at least one free (term or type) variables in
+     [vars]. *)
   let potential_triggers =
     let has_bvar bv_lf bv =
       Var.Map.exists (fun e _ -> Var.Set.mem e bv) bv_lf
@@ -2310,15 +2332,15 @@ module Triggers = struct
         Var.Map.add let_v x lets, [let_e; in_e]
       | _ -> lets, e.xs
     in
-    let rec aux ((vterm, vtype) as vars) ((strs, lets) as acc) e =
-      let strs, lets =
+    let rec aux ((vterm, vtype) as vars) (strs, lets) e =
+      let strs =
         if e.pure && (has_bvar e.vars vterm || has_tyvar e.vty vtype) &&
            not (is_prefix e.f)
         then
           let vrs = free_vars_as_set e in
-          STRS.add (e, vrs, e.vty) strs, lets
+          STRS.add (e, vrs, e.vty) strs
         else
-          acc
+          strs
       in
       let lets, args = args_of e lets in
       List.fold_left (aux vars) (strs, lets) args
@@ -2335,6 +2357,8 @@ module Triggers = struct
     e.ty == Ty.Tbool &&
     match e.f with Sy.Form _ -> false | _ -> true
 
+  (* Keep only patterns in [full_trs] whose all the free term
+     variables are free in [f] with the same type. *)
   let trs_in_scope full_trs f =
     STRS.filter
       (fun (e, _, _) ->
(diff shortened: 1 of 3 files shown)


src/lib/structures/expr.ml
```diff
@@ -2115,6 +2115,16 @@ module Triggers = struct
         if c <> 0 then c else Var.Set.compare y1 y2
     end)
 
+  (* let's [e] be a subterm of a quantified formula [f] and [bv] the set
```
```suggestion
  (* let [e] be a subterm of a quantified formula [f] and [bv] the set
```

bclement-ocp: I think that mentioning the formula `f` is confusing here since it is not an argument of the function (and in full generality might not exist). I'd say something along the lines of:

> [underscore bv e] substitutes all free (term) variables in [e] that are not in [bv] by the special `underscore` variable.

and give the example of `g(x, y, z)` with `bv = {x, y}` without mentioning a formula at all.

src/lib/structures/expr.ml
```diff
@@ -2124,6 +2134,7 @@ module Triggers = struct
              else Var.Map.add v (mk_term (Sy.var Var.underscore) [] ty) sbt
           )t.vars Var.Map.empty
       in
+      (* TODO: remove the useless returned boolean value. *)
```
bclement-ocp: Note: doing this is OK to do in a doc PR in my book, since it is both obviously correct and extremely localized. If the "TODO" is longer than the change itself, might as well do the change.

src/lib/structures/expr.mli
```diff
@@ -309,17 +309,26 @@ val make_triggers:
...
-    1. In greedy mode:
-    - Mono-triggers without variables;
-    - Mono-triggers with variables;
-    - Multi-triggers without variables;
-    - Multi-triggers with variables.
-      2. Otherwise:
-    - Mono and multi-triggers without variables;
-    - Mono and multi-triggers with variables.
+    The following documentation describes the strategies used by SatML
+    to generate multi-triggers.
```
bclement-ocp: If this is SatML-specific should the documentation be here?

src/lib/structures/expr.mli
```diff
@@ -309,17 +309,26 @@ val make_triggers:
...
+
+    Assuming that [Options.get_triggers_var ()] is [true] (the default),
+    we do not allow trigger variables in frugal, normal and greedy mode
+    but we allow them in greedier mode.
+
+    While generating multi-triggers (with at least two patterns), we can
+    escape free variables of their patterns that aren't bound in the quantified
+    formula.
```
bclement-ocp: I cannot parse this sentence. Maybe it would be more clear with a definition of "escaping variable" and an example?

src/lib/structures/expr.mli
```diff
@@ -309,17 +309,26 @@ val make_triggers:
...
+
+    The strategies of multi-trigger generations are:
+      1. In frugal and normal modes, we try in order:
+    - Mono-triggers;
+    - Multi-triggers without escaping variables;
+    - Multi-triggers with escaping variables.
+      2. In greedy mode, we try in order:
+    - Mono and multi-triggers without escaping variables.
+    - Mono and multi-triggers with escaping variables;
+      3. In greedier mode: same as greedy mode.
```
bclement-ocp:
```suggestion
      3. In greedier mode: same as greedy mode, except that variables are allowed as valid triggers.
```

bclement-ocp (changes requested):

src/lib/structures/expr.ml
Halbaroth: Sure, I wasn't happy with this formulation too.

src/lib/structures/expr.mli
Halbaroth: `FunSAT` doesn't define the same strategies (`frugal`, `normal`, ...). Instead, it uses the global option `Options.get_greedy ()` to choose between `greedy` or not and there is no `greedier mode` too. I can write a documentation for `FunSAT` if you want.

Halbaroth: I will try to factorize the documentation for both solvers.

bclement-ocp: If this function is only called by SatML it is fine to document it that way — if the instantiation strategies are different I was thinking it might make more sense to document in each solver what they do.

bclement-ocp: Or rather, to put the documentation for SatML in satml_frontend

Halbaroth: Finally, I don't talk about `frugal`, `normal`, ... modes because they have different names or meaning in `FunSAT`.

src/lib/structures/expr.mli
Halbaroth: I tried to reformulate and add an example. 
Honestly, I don't understand the purpose of this feature. Consider the example:
```
logic P: int -> prop
logic Q: int, int -> prop
logic a: int

axiom a: forall z: int. forall x:int. P(x) -> forall y:int. Q(y, z)
goal g1: P(1) -> Q(a, 1)
```
`Alt-Ergo` will generate the multi-trigger `{P(x), Q(_,z)}`.

As I tried to understand the purpose of this feature, I documented more functions in `Triggers`, sorry :/

bclement-ocp: `y` is not a quantified variable of the toplevel quantifier. As I understand it, the multi-trigger `{P(x), Q(_, z)}` will instantiate the axiom when there is a term of the form `P(x)` (binding `x`) and of the form `Q(_, z)` (binding `z`) in the context to generate `P(x) -> forall y:int. Q(y, z)`. Later we might instanciate the inner `forall` quantifier but only if `P(x)` holds.

src/lib/structures/expr.ml
```diff
@@ -2115,6 +2116,13 @@ module Triggers = struct
         if c <> 0 then c else Var.Set.compare y1 y2
     end)
 
+  (* [underscore bv e] replaces all the free term variables in [e] that
+     aren't in [bv] by the underscore term.
```
```suggestion
     are not in [bv] by the underscore term.
```

src/lib/structures/expr.ml
```diff
@@ -2124,18 +2132,23 @@ module Triggers = struct
...
+        let t = aux t s in
         let vt = Var.Set.add Var.underscore (Var.Set.inter vt bv) in
         t,vt,vty
 
+  (* [parties mconf bv vty l escaped_vars] generates all the multi-triggers such
+     that their patterns lie in [l] and they cover all the free variables [bv]
+     and [vty].
+
+     If [escaped_vars] is [true], replace the free term variables of these
+     patterns that do not lie in [bv]. *)
```
bclement-ocp: Replace the free term variables of these patterns with what?

src/lib/structures/expr.ml
```diff
@@ -2169,39 +2184,53 @@ module Triggers = struct
     let l = STRS.elements s in (* remove redundancies in old l *)
     SLLT.elements (parties_rec (SLLT.empty, SLLT.empty) l)
 
+  (* Simplify the multi-trigger [l] by removing a pattern [p] in [l] if
+     its set of free (term and type) variables:
+     - isn't maximal (for the inclusion) among the set of free variables of
```
```suggestion
     - is not maximal (for the inclusion) among the set of free variables of
```

src/lib/structures/expr.ml
```diff
@@ -2169,39 +2184,53 @@ module Triggers = struct
...
+         patterns in [l];
+     - contains the union of [bv_a] and [vty_a];
+     - is disjoint with the union of [bv_a] and [vty_a]. *)
```
bclement-ocp: I don't understand how these two conditions can hold at the same time, and I thought this was an AND-list. Is it an OR-list?

src/lib/structures/expr.ml
```diff
@@ -2169,39 +2184,53 @@ module Triggers = struct
...
+             (Var.Set.subset bv_a bv && Svty.subset vty_a vty) ||
+             (Var.Set.equal (Var.Set.inter bv_a bv) Var.Set.empty &&
+              Svty.equal (Svty.inter vty_a vty) Svty.empty)
+          then simpl_rec acc l
+          else simpl_rec (e::acc) l
+      in
+      simpl_rec [] l
 
   let multi_triggers menv bv vty trs escaped_vars =
+    (* The simplification removed all the patterns of the multi-trigger [trs]
```
```suggestion
    (* [simplification] removes all the patterns of the multi-trigger [trs]
```

src/lib/structures/expr.ml
```diff
@@ -2322,6 +2357,8 @@ module Triggers = struct
     e.ty == Ty.Tbool &&
     match e.f with Sy.Form _ -> false | _ -> true
 
+  (* Keep only patterns in [full_trs] whose all the free term
```
```suggestion
  (* Keep only patterns in [full_trs] for which all the free term
```

src/lib/structures/expr.ml
```diff
@@ -2322,6 +2357,8 @@ module Triggers = struct
...
+     variables are free in [f] with the same type. *)
```
bclement-ocp, line 2360: What does "free with the same type" means?

src/lib/structures/expr.mli
```diff
@@ -309,19 +309,25 @@ val make_triggers:
...
-    - Mono-triggers with variables;
-    - Multi-triggers without variables;
-    - Multi-triggers with variables.
-      2. Otherwise:
-    - Mono and multi-triggers without variables;
-    - Mono and multi-triggers with variables.
-
-    Mono-trigger with `Sy.Plus` or `Sy.Minus` top symbols are ignored
+    We can escape unbounded variables in patterns of multi-triggers, that
+    is replaced them by the underscore term. For instance, if [binders] is
```
bclement-ocp:
```suggestion
    is replace them by the underscore term. For instance, if [binders] is
```

src/lib/structures/expr.mli
```diff
@@ -309,19 +309,25 @@ val make_triggers:
...
+    the set [{x, y}] and [g(x, y, z)] is a pattern where [{(x, y, z)}] are
+    free term variables, we can replace [z] by [_].
+
+    Our strategy for multi-trigger generations depends on the matching
+    environment [menv] as follows:
+
+    If [menv.greedy] is [false], we try to generate in order:
+    - Mono-triggers;
+    - Multi-triggers with escaping variables.
```
bclement-ocp: Should that be `without`? Does "with escaping variables" means "we allow underscore terms" and "without escaping variables" means "we do not use the underscore term"?

bclement-ocp (commented):
Some more questions :)

src/lib/structures/expr.ml
Halbaroth: with the `underscore term`.

src/lib/structures/expr.ml
Halbaroth: It's a OR-list.

src/lib/structures/expr.mli
Halbaroth: `with escaping variables` means we have replaced variables not bound at the top level of this formula by the underscore variable. Here, `escaping` has to be understand as in string context: in the string "aaa\\"aaa", the inner " is escaped.

The name is really confusing because we escape variables that escape the top level binders...

It's `with` and not `without`. Actually we compute multi-triggers without escaping variables but these triggers are threw away:
```ocaml
        let mono, multi =
          make_triggers mconf vterm vtype trs ~escaped_vars:false
        in
        triggers_of_list @@
        match mono with
        | _::_ -> mono (* try to take nb_triggers ? *)
        | [] ->
          let mono', multi' =
            make_triggers mconf vterm vtype f_trs ~escaped_vars:true
          in
          if mono' != [] then mono'
          else if multi != [] then multi
          else multi'
```
If `mono` isn't empty, we ignore the multi-trigger without escaping variables.
If `mono` is empty, we regenerate everything with escaping variables.

In the greedy mode, `mono` and `multi` lists are merged in `make_triggers`.... (terrible design)

bclement-ocp: > with escaping variables means we have replaced variables not bound at the top level of this formula by the underscore variable.

The part I do not understand is if this is exclusive -- are "multi triggers without escaping variables" a subset of "multi triggers with escaping variables" or are they two distinct sets of multi-triggers.

> Here, escaping has to be understand as in string context: in the string "aaa\"aaa", the inner " is escaped.

This confuses me even more. I thought that "escaping variable" referred to a variable escaping its scope (e.g. in `forall x, P(x) -> forall y, Q(x, y)` we must move `Q(x, y)` to the scope of `forall x`, which causes `y` to escape its `forall y` scope). I don't understand the link with character escaping in a string context.

src/lib/structures/expr.ml
Halbaroth: This comment describe the code as it is. I know that comments are supposed to describe why we do something, not how we do it but here I have no idea why we check the type of variables. Variables of the module `Var` have unique identifiers, so `Var.equal x y` should imply that `x` and `y` have the same type in any expressions of `Expr`. I believe we can replace this portion of code by:
```ocaml
  let trs_in_scope full_trs f =
    STRS.filter
      (fun (e, _, _) ->
         Var.Map.for_all
           (fun v _ -> Var.Map.mem v)
           )
           e.vars
      )full_trs
```

src/lib/structures/expr.mli
Halbaroth: > The part I do not understand is if this is exclusive -- are "multi triggers without escaping variables" a subset of "multi triggers with escaping variables" or are they two distinct sets of multi-triggers.

They are distinct.

Halbaroth: Is it better with the last commit?

src/lib/structures/expr.mli
Halbaroth: Sure, I understood that but I don't understand why in `greedy` mode we first try multi-triggers with escaped variables. I expected the reverse order.

src/lib/structures/expr.ml
bclement-ocp: I was not asking about why we do it but I only read the documentation and did not understand what it meant :)

Possibly this is due to polymorphism? For instance [f] might have variable [x] of type [int list] but the trigger has [x] of type ['a list] (or the opposite)? I am not sure if we rename variables on type instantiation.

src/lib/structures/expr.mli
bclement-ocp: What I don't understand is why we try both — from what I understand, the only difference is whether we bind `z` or not, so doing only the version with escaped variables looks sufficient.

src/lib/structures/expr.mli
bclement-ocp: Yes that looks better!

(Are you sure they are distinct? The only difference seems to be this line `if escaped_vars  then List.rev_map (underscore bv) l else l in` which I think have the same multi-triggers with and without the `escaped_vars` flag for quantifiers with no escapes?

bclement-ocp (approved):

src/lib/structures/expr.ml
Halbaroth: I put an assert in a separated PR and I run `Marvin` on it.

src/lib/structures/expr.mli
Halbaroth: I will check if we can remove the other one in the separated PR.

src/lib/structures/expr.mli
Halbaroth: It's the only difference but `l` is the list of multi-triggers and we apply substitutions on them. We don't keep the original multi-triggers. I guess I don't understand your question.

bclement-ocp: What I was asking is: are the triggers returned by `make_triggers ~escaped_vars:true` and `make_triggers ~escaped_vars:false` distinct (i.e. no trigger can be in both lists) or not. I understand this is not the case, because triggers with no escaped variables (i.e. only toplevel variables) appear in both lists.

Halbaroth: Ok... you mean: are these sets *disjoint*? no but these sets are distinct in general.

bclement-ocp: Oh my bad! Yes I meant "disjoint", not "distinct", in spite of making the same mistake twice (:

PR #1142, opened 2024-06-07
Use separated headers

Halbaroth:
Use separated headers to distinguish our contributions.


Halbaroth:
Is it better @bclement-ocp ?

bclement-ocp:
I meant that the year should ideally depend on the file. If we can't do that, keeping the start of the project date (2013) everywhere is better.

Halbaroth:
So you want a copyright of the form `2013-xxxx` where `xxxx` is the last year we have modified the file?

bclement-ocp:
I don't really see the point of this format: the main point here is to not claim copyright before the code was actually written.

So ideally we would have `XXXX - YYYY` where `XXXX` is the year the file was created and `YYYY` is the year it was last modified, but `XXXX - 2024` where `XXXX` is the year the file was created works too.

If that is technically challenging with the tools we have, `2013 - 2024` is the next best (and simplest) option. My understanding is that the copyright years in the notice are informational only anyways.

Halbaroth:
I removed years in all the headers as we decided during the last dev meeting.

bclement-ocp:
I think this is correct wrt the files changed. I understood we chose the option of keeping "2013 - 2024" everywhere, but dropping the years is fine by me  should be fine (this is what [curl](https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/) did recently, and code from big companies such as VScode do not have copyright years) and should simplify our life. Would prefer an ack from @lefessan or @Ninjapouet before merging.

Halbaroth:
I think we can merge. It is really easy to rollback it with `ack` ;)

Halbaroth:
Fabrice agreed so we can merge @bclement-ocp

bclement-ocp (approved):

PR #1143, opened 2024-06-10
Use a custom attribute to name lemma

Halbaroth:
This PR adds a new attribute ":lemma" to name lemmas instead of using the builin attribute ":named". So the new syntax is:
```smt
(assert (! (forall ...) :lemma my_lemma))

```
Notice that we produce a custom error `Expect_simple_sexpr` if the user gives an inappropriate s-expressions. Currently, we cannot define our own printer for the typechecker but Guillaume will implement this soon.

### DIFF
--- a/src/lib/frontend/d_cnf.ml
+++ b/src/lib/frontend/d_cnf.ml
@@ -1847,7 +1847,7 @@ let make dloc_file acc stmt =
     | { id = DStd.Id.{name = Simple name; _}; contents = `Hyp t; loc; attrs;
         implicit=_ } ->
       let name =
-        match DStd.Tag.get t.term_tags lemma_attr with
+        match DStd.Tag.get t.term_tags lemma_name_attr with
         | Some n -> n
         | None ->
           match DStd.Tag.get t.term_tags DE.Tags.named with
--- a/src/lib/frontend/d_cnf.ml
+++ b/src/lib/frontend/d_cnf.ml
@@ -444,10 +444,11 @@ let smt_fpa_builtins =
       end
     | _ -> `Not_found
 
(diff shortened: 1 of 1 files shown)


src/lib/frontend/d_cnf.ml
```diff
@@ -444,11 +444,38 @@ let smt_fpa_builtins =
       end
     | _ -> `Not_found
 
+let lemma_attr : string DStd.Tag.t = DStd.Tag.create ()
+
+type _ Dl.Typer.T.err +=
+  | Expect_simple_sexpr
```
bclement-ocp: Nit: This is not very explicit about where/why we expect a simple sexpr — `Invalid_lemma_name` or `Non_simple_symbol_lemma_name` would be more explicit.

src/lib/frontend/d_cnf.ml
```diff
@@ -1817,7 +1844,7 @@ let make dloc_file acc stmt =
     | { id = DStd.Id.{name = Simple name; _}; contents = `Hyp t; loc; attrs;
         implicit=_ } ->
       let name =
-        match DStd.Tag.get t.term_tags DE.Tags.named with
+        match DStd.Tag.get t.term_tags lemma_attr with
```
bclement-ocp: Do we want to fall back on `:named` if it is provided but `:lemma` is not provided?

bclement-ocp (approved):

src/lib/frontend/d_cnf.ml
Halbaroth: Naming is so hard. I wasn't a big fan of this one too.

PR #1144, opened 2024-06-11
feat(BV): Do not store width in Bitlist

bclement-ocp:
This patch rewrites the Bitlist module to represent infinite-width
bit-vectors rather than fixed-width bit-vectors, making it able to
represent unbounded integers. This improves the symmetry with the
interval domain (which also applies to unbounded integers) and is
intended to simplify the implementation of BV-to-int and int-to-BV
conversions.

In order to avoid having to use negative numbers in bitlists, the
internal representation is changed from a pair of (bits equal to [1],
bits equal to [0]) masks to a pair (bits equal to [1], unknown bits)
masks. This change should also be good for memory consumption, as we no
longer keep two copies of the value around when all bits are known.

Note: depends on #1108, #1058, #1083, #1084, and #1085. Only the latest commit with title "feat(BV): Do not store width in Bitlist" is included in this PR.

### DIFF
--- a/src/lib/reasoners/bitlist.ml
+++ b/src/lib/reasoners/bitlist.ml
@@ -29,139 +29,154 @@ module Ex = Explanation
 
 exception Inconsistent of Ex.t
 
-(** A bitlist representing the known bits of a bit-vector of width [width].
+(** A bitlist representing the known bits of an infinite-width bit-vector.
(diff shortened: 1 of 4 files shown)


bclement-ocp:
This is +0-0 on QF_BV and slightly faster.

Halbaroth (approved):
LGTM

PR #1130, opened 2024-05-21
Simplify `is_mine_symb`

Halbaroth:
The argument `ty` of `is_mine_symbol` was only used by
`Enum` and `Adt` theories because they shared the `Construct` symbol.
As we merged these theories, we can remove this argument now.

This PR is rebased on #1094

PR #1146, opened 2024-06-17
Simple ADTs with a single constructor are records

Halbaroth:
This PR fixes a bug in `D_cnf` that have been exposed by #1130. Before merging `Enum` into `ADT` the situation was as follows:
- simple ADT with several constructors and with at least one payload was sent to the `Adt` theory
- simple ADT with several constructors but no payload was sent to `Enum` theory
- simple ADT with a single constructor was sent to the `Record` theory.

The last corner case includes a corner corner case: an ADT with a single constructor without payload (so basically the unit type with an extra step...)

The `D_cnf` included a bug. It produced single constructors without payload of type ADT instead of type record but the check done in `is_mine_symb` prevented from sending it to `Adt`. As I removed this check, we got an assert on the following input:
```smt2
(set-logic ALL)
(declare-datatype t ((Record)))
(define-fun a () t Record)
(check-sat)
```

After this patch, the situation is as follows:
- simple ADT with several constructors is sent to the `Adt` theory
- simple ADT with only one constructor is sent to the `Record` theory
- mutually recursive ADT whose all the types are record is sent to the `Record` theory

### DIFF
--- a/src/lib/frontend/d_cnf.ml
+++ b/src/lib/frontend/d_cnf.ml
@@ -970,8 +970,14 @@ let rec mk_expr
             E.mk_term sy [] ty
 
           | B.Constructor _ ->
-            let ty = dty_to_ty term_ty in
-            E.mk_constr (Uid.of_dolmen tcst) [] ty
+            begin match dty_to_ty term_ty with
+              | Trecord _ as ty ->
+                E.mk_record [] ty
+              | Tadt _ as ty ->
+                E.mk_constr (Uid.of_dolmen tcst) [] ty
+              | _ ->
+                assert false
+            end
 
           | _ -> unsupported "Constant term %a" DE.Term.print term
         end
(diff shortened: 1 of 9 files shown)


hra687261 (approved):

src/lib/frontend/d_cnf.ml
```diff
@@ -975,8 +975,10 @@ let rec mk_expr
                 E.mk_record [] ty
               | Tadt _ as ty ->
                 E.mk_constr (Uid.of_dolmen tcst) [] ty
-              | _ ->
-                assert false
+              | Tunit ->
+                E.void
+              | ty ->
+                Fmt.failwith "unexpected type %a@." Ty.print ty
```
hra687261, line 981: Is there a test that shows that Tunit is used? And that we crash if we don't match it here?

Halbaroth: There are 2000 regressions on `ae-format` without this case, so yes, we have tests :)
Actually the builtin `unit` type of `Dolmen` is an ADT but a distinct builtin type `Tunit` in Alt-Ergo. It's a corner corner corner case...

I tested this PR on `ae-format` and now there is no regression anymore ;)

Halbaroth: I will add a unit test to catch unit (sic) case

Halbaroth: Finally, I removed the `Ty.Tunit` type of AE in #1148, so this corner case disappears.

PR #1152, opened 2024-06-20
feat(BV): Only store domains on variable parts

bclement-ocp:
This patch changes the way that bit-vector domains are stored in order to share domains between multiple bit-vectors with the same variable part, including across different bit-width. Currently, if we had terms [x], [00 @ x], and [10 @ x], we would store domains for each of these. With this patch, we only store a domain for [x] and rebuild the domains for [00 @ x] and [10 @ x] dynamically; we do this by computing normal forms composed of a variable part ([x] here) and a constant offset (either [0] or [10]).

This reduces the number of variables handled by the system (which is important for global domains such as global difference) and reduces the number of trivial propagations.

Note that this patch is not reverting #1044. In particular, we are considering [x @ y] as a unique (composite) variable built up from the atomic variables [x] and [y] (in the same way that the polynomial [x + y] is built up from the monomial [x] and [y] in the IntervalCalculus module; in fact, the implementation is intentionally abstracted in order to be useable with polynomials). This is necessary because, as noted in #1044, the interval domain for [x @ y] cannot be built reconstructed precisely from the interval domains for [x] and [y].

The patch includes a bit of refactoring of the way we handle constraints; notably, instead of aggressively substituting constraints, we now store the original constraint and call [Uf.find] at propagation time. This avoids having to define a substitution operation on normal forms and allows storing the constraint dependencies directly inside the domain (in the previous design, we could not do it easily without repeating the expensive substitution of constraint arguments).

Note: depends on https://github.com/OCamlPro/alt-ergo/pull/1144. Only the latest commit with title "[feat(BV): Interval domains for bit-vectors](https://github.com/OCamlPro/alt-ergo/commit/b5cbc3cf20af2ea46159ade40db8b5e15e281a9c)" is included in this PR.

**Benchmarks**: This PR provides a small reasoning boost (+171-48 on the usual QF_BV subset) and a significant performance boost for bit-vector reasoning (-25% on the same QF_BV subset) compared to #1144.


src/lib/reasoners/bitv_rel.ml
```diff
@@ -111,128 +115,131 @@ module Interval_domain = struct
...
+
+      We decompose non-constant bit-vector compositions as a variable part,
+      where all constant bits are set to [0] and all high constant bits are
+      chopped off, and an offset with all the constant bits. We consider the
+      variable part atomic if it is a single non-negated variable.
+
+      Assuming [x] and [y] are bit-vectors of width 2:
+      - [101 @ x] is [x + 10100] ;
+      - [10 @ x @ 01] is [(x @ 00) + 100001] ;
+      - [10 @ y<0, 0> @ y<1, 1>] is [(y<0, 0> @ y<1>1) + 1000] ;
```
Halbaroth: I do not understand the notation `x<.., ...>`. Is it an extraction? I guess there is a typo with `y<1>1` instead of `y<1, 1>`?

bclement-ocp:
Rebased on `next`. @Halbaroth ping

bclement-ocp:
Note: this fixes rare crashes with bitvector fields in ADTs/records (previous implementation incorrectly assumed that `X.leaves` of a composition returned the bit-vectors in the composition, but that is not true if the composition contains ADT/record destructors). I tried to make a test case for it but couldn't figure out the exact conditions to trigger this (it depends on the way propagation is done so would be brittle anyways).

src/lib/reasoners/rel_utils.ml
```diff
@@ -250,634 +386,863 @@ module type Domain = sig
...
+        update the associated domain imperatively. A single (physical) entry is
+        associated with a given key. *)
+
+    val domain : t -> domain
+    (** Return the domain associated with this entry. *)
+
+    val set_domain : t -> domain -> unit
+    (** Intersect the domain associated with this entry and the provided
+        [domain]. The explanation [ex] justifies that the [domain] applies to
+        the entry's key.
```
Halbaroth: This documentation is outdated. There is no `ex` argument.

src/lib/reasoners/rel_utils.ml
```diff
@@ -250,634 +386,863 @@ module type Domain = sig
...
+
+    let entry t x =
+      { Entry.entry = EX.entry t.domains x
+      ; key = x
+      ; notify = t.notify }
+  end
 
-    let (!!) handle = handle.domain
+  let edit ~notify ~default t =
+    SX.iter notify t .changed;
```
```suggestion
    SX.iter notify t.changed;
```

src/lib/reasoners/rel_utils.ml
```diff
@@ -250,634 +386,863 @@ module type Domain = sig
...
+        the entry's key.
+
+        @raise Domain.Inconsistent if the intersection is empty. *)
+  end
+
+  val entry : t -> key -> Entry.t
+  (** [entry t k] returns the [handle] associated with [k].
+
+      There is a unique entry associated with each key [k] that is created
+      on-the-fly when [handle t k] is called for the first time.
```
Halbaroth: There is no `handle` function in this module. I guess you meant `entry t k`?

src/lib/reasoners/rel_utils.ml
```diff
@@ -250,634 +386,863 @@ module type Domain = sig
...
+
+      The domain associated with the entry is initialized from the underlying
+      persistent domain the first time it is accessed, and updated with
+      [update]. *)
```
Halbaroth: What is `update` here? Cannot find it.

src/lib/reasoners/rel_utils.ml
```diff
@@ -250,634 +386,863 @@ module type Domain = sig
...
-    (** Perform structural propagation for the given representative.
+    type domain = D.t
+    (** The type of per-variable domains. *)
 
-        More precisely, if [r] is a leaf, the domain of [r] is propagated to any
-        semantic value that contains [r] as a leaf according to the structure of
-        that semantic value (using [Domain.map_leaves]); if [r] is not a leaf,
-        its domain is propagated to any of the leaves it contains.
+    val find : key -> t -> domain
+    (** Find the domain associatd with the given key.
```
```suggestion
    (** Find the domain associated with the given key.
```

src/lib/reasoners/rel_utils.ml
```diff
@@ -250,634 +386,863 @@ module type Domain = sig
...
-    let size = 17 in
-    let ephemeral =
-      { Ephemeral.persistent = { t with changed = SX.empty }
-      ; handles = HX.create size
-      ; dirty_cache = HX.create size
-      ; changed_set = HX.create size }
+  val add : X.t -> W.t -> t -> t
+  (** [add x w r] adds the tuple [(x, w)] to the relation. *)
+
+  val add_many : X.t -> W.Set.t -> t -> t
```
Halbaroth: Please add documentation

src/lib/reasoners/rel_utils.ml
```diff
@@ -250,634 +386,863 @@ module type Domain = sig
...
+
+  val range : X.t -> t -> W.Set.t
```
Halbaroth: Same here

src/lib/reasoners/rel_utils.ml
```diff
@@ -250,634 +386,863 @@ module type Domain = sig
...
+    ; composite_watches = CW.remove_range w t.composite_watches
+    ; parents = t.parents
+    ; triggers = t.triggers }
+
+  let needs_propagation t =
+    DMA.needs_propagation t.atoms ||
+    DMC.needs_propagation t.composites ||
+    not (W.Set.is_empty t.triggers)
+
+  let variables { variables ; _ } = variables
```
Halbaroth: For this kind of function, I am used to add `[@inline always]`, but I am not sure it is necessary?

src/lib/reasoners/rel_utils.ml
```diff
@@ -250,634 +386,863 @@ module type Domain = sig
...
+
+  let parents { parents ; _ } = parents
```
Halbaroth: Same question for this one

Halbaroth:
I still believe the code would be easier to read with less functors and signatures. I understand your argument. It is probably easier to write it like this for you but reading the file `Rel_utils` gives me the same feeling that reading Bourbaki, a bunch of abstract axioms before reaching the real stuff.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Yes, this is the syntax that is used elsewhere in documentation comments for extraction (and indeed there is a typo, it should be `y<1, 1>`). I will clarify.

src/lib/reasoners/rel_utils.ml
bclement-ocp: True — the justification is in the domain.

src/lib/reasoners/rel_utils.ml
bclement-ocp: Yes, `handle` was renamed to `entry` at some point.

src/lib/reasoners/rel_utils.ml
bclement-ocp: It should be `set_domain`, will fix.

src/lib/reasoners/rel_utils.ml
bclement-ocp: It's probably going to be inlined anyways. I can add the annotation.

bclement-ocp (commented):
> I still believe the code would be easier to read with less functors and signatures.

I have:

 - Removed the `EphemeralMap` functor that turned out not to be useful 
 - Merged the OffsetDomain and CompositeDomain signature into the Domain signature (the goal was to be able to extend the code to domains such as ADT constructors, which don't have an offset and are never composite, but that is not really an useful generalization at the moment)
 - Merged AtomicType and CompositeType into the NormalForm signature (I intended to do this one prior to submitting the PR but forgot to actually do it)
 - Removed the UfHandle functor in favor of a `Canon` sub-module in `Domains_make.Ephemeral`
 - Inlined the `HandleNotations` functor into the `EphemeralDomainMap` signature (I was not really happy with this functor in the first place; the goal was to avoid code duplication for the definition of `update` but all solutions I found, including `HandleNotations`, were too clunky and verbose, and the code for `update` is small, so I think it's OK to duplicate here)
 - Renamed the `BinRel` functor to `WatchMap` to make it less abstract, and removed some functions from it that were unused.

I think the remaining abstractions (except for `WatchMap`) are necessary to be able to share the implementation between interval/bitlist domains and (in the future) between bit-vector/arithmetic normal forms. I am not too happy with the remaining `WatchMap` functor but I don't have a good solution to avoid it without significant changes. 

> It is probably easier to write it like this for you

It is not that it is easier to write but that I find it easier to understand code with small encapsulated abstractions. I agree I went a little overboard here though, hopefully it is better now.

> reading the file Rel_utils gives me the same feeling that reading Bourbaki, a bunch of abstract axioms before reaching the real stuff.

That is a common and unfortunate side-effect of OCaml processing definitions in order, and the solution is to read from the bottom up — OCaml code reads like a research paper, out of order. I have moved the signature definitions to a new `Domains_intf` module, and the implementation itself to a new `Domains` module.

bclement-ocp:
> Benchmarks: This PR provides a small reasoning boost (+171-48 on the usual QF_BV subset) and a significant performance boost for bit-vector reasoning (-25% on the same QF_BV subset) compared to https://github.com/OCamlPro/alt-ergo/pull/1144.

In the current state of both the PR and the `next` branch, this is now a smaller reasoning boost (+19-14, essentially a wash) and performance boost (10%, still nothing to sneer at) on the QF_BV dataset. On the J3 dataset it gives a small reasoning boost (+51-6) and larger performance boost (15%).

bclement-ocp:
> I am not too happy with the remaining WatchMap functor but I don't have a good solution to avoid it without significant changes.

Actually, I can just fold `WatchMap` to `DomainsMap`. Pushed a last commit that does just that. This should be ready for round 2 now @Halbaroth .

Halbaroth:
Thanks for your effort, I will do a second pass now and we can merge :)

Halbaroth (approved):

PR #1153, opened 2024-06-20
Use Dolmen attribute to store well founded order

Halbaroth:
This PR used the well designed Dolmen to store custom attributes on constructors of ADT and ADT themself. This design has two advantages:
1. We don't need to store an extra global state for the total order generated in the module `Nest`.
2. We have a total order per ADT instead of a global order for all the ADT types.

To be more specific, we generate a bijection between ADT constructors and the interval `[0, n[` where `n` is the number of constructors.

This PR have been tested and we have no regression at all. I'm running benchmarks for the legacy frontend too.


src/lib/frontend/d_cnf.ml
```diff
@@ -607,12 +607,12 @@ let mk_ty_decl (ty_c: DE.ty_cst) =
...
     let record_constr = Uid.of_dolmen cstr in
-    let ty = Ty.trecord ~record_constr tyvl (Uid.of_dolmen ty_c) lbs in
+    let ty = Ty.trecord ~record_constr tyvl uid lbs in
     Cache.store_ty ty_c ty
 
   | Some (Adt { cases; _ } as adt) ->
-    Nest.add_nest [adt];
-    let uid = Uid.of_dolmen ty_c in
+    let uid = Uid.of_ty_cst ty_c in
+    List.iter (fun (i, h) -> Uid.set_hash i h) @@ Nest.generate [adt];
```
bclement-ocp: Is there a reason to call `Uid.set_hash` here? Why not do it inside `Nest.generate` which would become an effectful function (returns `unit`)?

src/lib/frontend/d_cnf.ml
```diff
@@ -729,14 +729,14 @@ let mk_mr_ty_decls (tdl: DE.ty_cst list) =
           assert false
     ) ([], false) tdl
   in
-  Nest.add_nest rev_tdefs;
+  List.iter (fun (i, h) -> Uid.set_hash i h) @@ Nest.generate rev_tdefs;
```
bclement-ocp: Same remark as above.

src/lib/reasoners/adt_rel.ml
```diff
@@ -36,48 +36,68 @@ module LR = Uf.LX
...
+module DT = Dolmen.Std.Expr.Ty
+module B = Dolmen.Std.Builtin
+module TSet = Set.Make (Int)
 
 let timer = Timers.M_Adt
 
+(* HACK: this wrapper around [Uid.get_hash] is a temporary hack
+   used by the legacy frontend only. For the legacy frontend, we generate a
+   trivial perfect hash for the type [ty], that is a hash whose the associated
+   order does not ensure the termination of the model generation. *)
```
bclement-ocp: As I have already stated: it would have been better to wait for the release to do this, so that we would avoid churn with this sort of hacks.

"Trivial perfect hash for the type [ty]" is not really clear (I had to read the definition of the `hash` type elsewhere to understand). "a perfect hash for the constructors of type [ty]" would be clearer (since we hash the constructors, not the type itself)

Also I suppose that it should be a hash whose associated order does ensure termination (rather than does *not* ensure termination)?

src/lib/reasoners/adt_rel.ml
```diff
@@ -36,48 +36,68 @@ module LR = Uf.LX
...
-      type t = Uid.t
-
-      (* We use a dedicated total order on the constructors to ensure
-         the termination of the model generation. *)
-      let compare = Nest.compare
-    end)
+module DE = Dolmen.Std.Expr
+module DT = Dolmen.Std.Expr.Ty
+module B = Dolmen.Std.Builtin
+module TSet = Set.Make (Int)
```
bclement-ocp: This is no longer a set of types — `Hset`, or even `SI`, would be better.

Nit: there is `Util.SI` :upside_down_face:

src/lib/reasoners/adt_rel.ml
```diff
@@ -86,38 +106,43 @@ module Domain = struct
     | Ty.Tadt (name, params) ->
       (* Return the list of all the constructors of the type of [r]. *)
       let cases = Ty.type_body name params in
+      let hash = get_hash ty in
```
bclement-ocp: Does not look like `get_hash` is only used by the legacy frontend after all?

src/lib/reasoners/adt_rel.ml
bclement-ocp: Can we store this in the existing database in the `Ty` module, since it is now keyed on types (and not constructors)?

src/lib/structures/fpa_rounding.ml
```diff
@@ -93,8 +93,16 @@ let fpa_rounding_mode_dty, d_cstrs, fpa_rounding_mode =
     List.map (fun c -> DStd.Path.global @@ to_smt_string c, []) cstrs
   in
   let _, d_cstrs = DE.Term.define_adt ty_cst [] cstrs in
+  let () =
+    match DStd.Expr.Ty.definition ty_cst with
+    | Some def ->
+      let name, hash = Nest.generate [def] |> List.hd in
```
bclement-ocp: This should not use `List.hd` (which fails if the list is empty but might silently introduce errors if it has more than one element) but rather an explicit pattern matching.

src/lib/structures/ty.ml
```diff
@@ -536,15 +536,17 @@ let t_adt ?(body=None) s ty_vars =
 module DE = Dolmen.Std.Expr
 
 let tunit =
-  let name = Uid.of_dolmen DE.Ty.Const.unit in
-  let void = Uid.of_dolmen DE.Term.Cstr.void in
+  let name =
+    match Dolmen.Std.Expr.Ty.definition DE.Ty.Const.unit with
+    | Some def ->
+      let name, hash = Nest.generate [def] |> List.hd in
```
bclement-ocp: Same as above

src/lib/structures/uid.ml
```diff
@@ -26,38 +26,81 @@
...
+      and [n] where [n] is the number of constructors of the ADT. *)
+
+  of_int : int -> t;
+  (** Return the associated constructor to the integer.
+
+      @raises Invalid_argument if the integer does not correspond to
+       a constructor of this ADT. *)
+}
 
 let[@inline always] of_dolmen id = Dolmen id
```
bclement-ocp: Do we still need this case?

src/lib/structures/uid.ml
```diff
@@ -26,38 +26,81 @@
...
+  | Ty_cst id1, Ty_cst id2 -> DE.Id.compare id1 id2
   | _ -> String.compare (show u1) (show u2)
 
+let hash_tag : hash DStd.Tag.t = DStd.Tag.create ()
+
+let set_hash id hash =
+  match id with
+  | Ty_cst ty_c ->
+    DE.Ty.Const.set_tag ty_c hash_tag hash
+  | _ -> Fmt.invalid_arg "set_hash %a" pp id
```
bclement-ocp: To avoid potential misuses of this API, would it work to have `of_ty_cst` fail if no hash is provided, and have a separate function `create_ty_cst` that takes the type constructor and the hash?

bclement-ocp (commented):
Can you clarify why we store the `hash` on the type?

Since the (term) constructors for distinct types are distinct, couldn't we just store the hash associated with each term constructor on the term itself? It seems like it would avoid a bunch of ceremony to get the hashing function from the type.

If the issue is that this is not compatible with the legacy frontend (which it might very well not be): do we need this in 2.6 or can we postpone to 2.7 so that we don't have to deal with the legacy frontend? It looks like it would simplify this PR a great deal.

src/lib/frontend/d_cnf.ml
Halbaroth: There are no reason but concentrating effects in `D_cnf`. I can change it and rename `Nest.generate` accordingly.

src/lib/reasoners/adt_rel.ml
Halbaroth: Sure

src/lib/reasoners/adt_rel.ml
Halbaroth: I meant that the purpose of this wrapper is to preserve the support for the legacy frontend but we have to call it to retrieve the hash function computed in `Nest` too.

src/lib/frontend/d_cnf.ml
Halbaroth: Same answer

Halbaroth:
> Can you clarify why we store the `hash` on the type?
> 
> Since the (term) constructors for distinct types are distinct, couldn't we just store the hash associated with each term constructor on the term itself? It seems like it would avoid a bunch of ceremony to get the hashing function from the type.
> 

Sure, it's simpler. This PR have been really complicated to write actually, I have redone it from scratch several times.

src/lib/structures/fpa_rounding.ml
Halbaroth: I removed this code.

src/lib/reasoners/adt_rel.ml
Halbaroth: I removed this code.

src/lib/structures/uid.ml
Halbaroth: I removed this code.

src/lib/structures/uid.ml
Halbaroth: We store three kind of Dolmen's identifiers: `ty id`, `type_fun id` and `type_ id`. I can replace this case by `Ty_var`?

src/lib/frontend/d_cnf.ml
```diff
@@ -585,11 +585,11 @@ and handle_ty_app ?(update = false) ty_c l =
 let mk_ty_decl (ty_c: DE.ty_cst) =
   match DT.definition ty_c with
   | Some (
-      Adt { cases = [| { cstr = { id_ty; _ } as cstr; dstrs; _ } |]; _ } as adt
+      Adt { cases = [| { cstr = { id_ty; _ } as cstr; dstrs; _ } |]; _ }
     ) ->
     (* Records and adts that only have one case are treated in the same way,
        and considered as records. *)
-    Nest.add_nest [adt];
+    let uid = Uid.of_ty_cst ty_c in
```
bclement-ocp, line 588: Why no `Nest.attach_orders` here?

src/lib/structures/uid.ml
bclement-ocp: If there is a single other kind of identifier we store, it would be better as it would avoid an existential in the type.

src/lib/util/lists.mli
```diff
@@ -67,3 +67,9 @@ val try_map : ('a -> 'b option) -> 'a list -> 'b list option
 val is_sorted : ('a -> 'a -> int) -> 'a list -> bool
 (** [is_sorted cmp l] checks that [l] is sorted for the comparison function
     [cmp]. *)
+
+val find_index : ('a -> bool) -> 'a list -> int option
+(** [find_index f l] returns [Some i], where [i] is the index of the first
+    element of the list [l] that satisfies [f x], if there is such an element.
+
+    It returns [None] if there is no such element. *)
```
bclement-ocp: Looks like this is no longer used?

src/lib/structures/uid.ml
```diff
@@ -26,38 +26,65 @@
...
+module DE = Dolmen.Std.Expr
 
 type t =
   | Hstring : Hstring.t -> t
   | Dolmen : 'a DE.id -> t
+  | Term_cst : DE.term_cst -> t
+  | Ty_cst : DE.ty_cst -> t
 
 let[@inline always] of_dolmen id = Dolmen id
+let[@inline always] of_term_cst id = Term_cst id
```
bclement-ocp: Is it correct to say that `of_term_cst` should only be called on term constructors that have had an order attached? Could we had an assertion for this here?

src/lib/structures/uid.mli
```diff
@@ -30,10 +30,22 @@ module DE = Dolmen.Std.Expr
...
 val of_string : string -> t
 val of_hstring : Hstring.t -> t
+
+val to_term_cst : t -> DE.term_cst
+
+val order_tag : int Dolmen.Std.Tag.t
+(** Tag used to attach the order of constructor. Used to
+    retrieve efficiency the order of the constructor in [to_int]. *)
+
+val perfect_hash : t -> int
```
bclement-ocp: This should be documented, and in particular it should be documented that it must only be called on uids to which an order has been attached by the `Nest` module, and that it can only be called on term constructors.

src/lib/structures/uid.mli
```diff
@@ -30,10 +30,22 @@ module DE = Dolmen.Std.Expr
 type t = private
   | Hstring : Hstring.t -> t
   | Dolmen : 'a DE.id -> t
+  | Term_cst : DE.term_cst -> t
+  | Ty_cst : DE.ty_cst -> t
```
bclement-ocp: This is orthogonal to the PR but if we define

```ocaml
type _ t =
  | Hstring : Hstring.t -> 'a t
  | Term_cst : DE.term_cst -> DE.ty t
  | Ty_cst : DE.ty_cst -> DE.type_fun t
  | Ty_var : DE.ty_var -> DE.type_ t

type term_cst = DE.ty t

type ty_var = DE.type_ t

type ty_cst = DE.type_fun t
```

then we can expose the "kind" of uid in the type and gain type safety (in particular, `perfect_hash` can be of type `Uid.term_cst -> int`).

src/lib/structures/uid.mli
```diff
@@ -30,10 +30,22 @@ module DE = Dolmen.Std.Expr
...
+val of_term_cst : DE.term_cst -> t
+val of_ty_cst : DE.ty_cst -> t
 val of_string : string -> t
 val of_hstring : Hstring.t -> t
+
+val to_term_cst : t -> DE.term_cst
+
+val order_tag : int Dolmen.Std.Tag.t
+(** Tag used to attach the order of constructor. Used to
+    retrieve efficiency the order of the constructor in [to_int]. *)
```
bclement-ocp: Nit: in the current design, I believe moving `order_tag` and `perfect_hash` to `Nest` would allow hiding `order_tag` from the interface.

src/lib/frontend/d_cnf.ml
```diff
@@ -607,12 +607,12 @@ let mk_ty_decl (ty_c: DE.ty_cst) =
...
     let record_constr = Uid.of_dolmen cstr in
-    let ty = Ty.trecord ~record_constr tyvl (Uid.of_dolmen ty_c) lbs in
+    let ty = Ty.trecord ~record_constr tyvl uid lbs in
     Cache.store_ty ty_c ty
 
   | Some (Adt { cases; _ } as adt) ->
-    Nest.add_nest [adt];
-    let uid = Uid.of_dolmen ty_c in
+    let uid = Uid.of_ty_cst ty_c in
+    Nest.attach_orders [adt];
```
bclement-ocp: Note: I think that keeping the `let uid = Uid.of_ty_cst ty_c in` after the call to `Nest.attach_orders` allows to have an assertion in `Uid.of_ty_cst` that the type has had order tags attached.

bclement-ocp (changes requested):
I am a bit concerned by the amount of implicit dependencies currently and have suggested places where (I think) we can add runtime checks ensuring that we catch errors early if we try to create uids without an order attached, which should help with maintainability in the future.

I also made some other comments regarding dead code and missing documentation that should be fixed.

src/lib/frontend/d_cnf.ml
Halbaroth: Because we never send record values to the ADT theory so we never use the total order.

src/lib/structures/uid.ml
Halbaroth: It is not true because we use it also for unique constructor of record but I can generate an order for these constructors too and add an assertion here.

src/lib/structures/uid.mli
Halbaroth: I applied the change in this PR. It is definitively a better approach. Thanks :)

src/lib/structures/uid.ml
Halbaroth: Actually, we cannot put an assertion here as destructors are also of type `term_cst`.

src/lib/frontend/d_cnf.ml
```diff
@@ -101,20 +101,20 @@ module Cache = struct
...
-  let update_ty_store ?(is_var = true) id =
-    let ty = fresh_ty ~is_var ~id () in
+  let update_ty_store ?(is_var = true) ?name id =
+    let ty = fresh_ty ~is_var ~name () in
     store_ty id ty
 
-  let update_ty_store_ret ?(is_var = true) id =
-    let ty = fresh_ty ~is_var ~id () in
+  let update_ty_store_ret ?(is_var = true) ?name id =
+    let ty = fresh_ty ~is_var ~name () in
```
Halbaroth: Using a polymorphic GADT for `Uid.t` almost works every but I ran into an annoying issue here. 

It makes sense to use `ty_cst` for the constructor of applications of abstract types in `Ty.t` but in `elim_toplevel_forall` in `D_cnf`, we remove the top level binders for both type and term variables. Basically, we replace a type variable `ty_var` by an uninterpreted type using `Text`. So we need to convert a `ty_var` into a `ty_cst`, which is not permit by `Dolmen`. 

I rollback my modifications done in #1098.

src/lib/frontend/d_cnf.ml
```diff
@@ -101,20 +101,20 @@ module Cache = struct
...
 
-  let fresh_ty ?(is_var = true) ?id () =
+  let fresh_ty ?(is_var = true) ?(name = None) () =
     if is_var
     then Ty.fresh_tvar ()
     else
-      match id with
-      | Some id -> Ty.text [] (Uid.of_dolmen id)
+      match name with
+      | Some n -> Ty.text [] (Uid.of_string n)
```
bclement-ocp: I think this might re-introduce some of the potential name clashes that #1098 was intended to remove.

When converting a variable → type for the purpose of quantifier elimination, we should fall back to the `Ty.fresh_empty_text ()` here — see my comment in `store_tyv` below.

src/lib/frontend/d_cnf.ml
```diff
@@ -125,7 +125,8 @@ module Cache = struct
       update_ty_store_ret ~is_var id
 
   let store_tyv ?(is_var = true) t_v =
-    update_ty_store ~is_var t_v
+    let name = get_basename t_v.DE.path in
+    update_ty_store ~is_var ~name t_v
```
bclement-ocp: `update_ty_store` is a simple wrapper here, we could call `fresh_ty` directly and keep the existing code, which would avoid using `Uid.of_string`:

```suggestion
    let ty = fresh_ty ~is_var () in
    store_ty t_v ty
```

Alternatively, we can create an identifier from the name of the variable:

```suggestion
    let id = DE.Ty.Const.mk t_v.DE.path 0 in
    let ty = fresh_ty ~is_var ~id in
    store_ty t_v ty
```

src/lib/frontend/d_cnf.ml
bclement-ocp: Makes sense — the order is also trivial. I think it still makes sense to have an order attached to all constructors for consistency.

src/lib/structures/fpa_rounding.ml
```diff
@@ -93,10 +93,15 @@ let fpa_rounding_mode_dty, d_cstrs, fpa_rounding_mode =
     List.map (fun c -> DStd.Path.global @@ to_smt_string c, []) cstrs
   in
   let _, d_cstrs = DE.Term.define_adt ty_cst [] cstrs in
+  let () =
+    match DStd.Expr.Ty.definition ty_cst with
+    | Some def -> Nest.attach_orders [def]
+    | None -> assert false
+  in
```
bclement-ocp:
```suggestion
  let def, d_cstrs = DE.Term.define_adt ty_cst [] cstrs in
  Nest.attach_orders [def];
```

src/lib/structures/uid.ml
```diff
@@ -26,46 +26,83 @@
...
+  | Ty_var : DE.ty_var -> DE.ty_var t
 
-let[@inline always] of_dolmen id = Dolmen id
+type term_cst = DE.term_cst t
+type ty_cst = DE.ty_cst t
+type ty_var = DE.ty_var t
+
+let order_tag : int DStd.Tag.t = DStd.Tag.create ()
+
+let has_order id =
```
bclement-ocp: It is somewhat confusing that `has_order` can return `true` for terms that have no order attached. Something like `has_order_if_constructor` would be clearer.

src/lib/structures/uid.ml
```diff
@@ -26,46 +26,83 @@
...
+  let has_attached_order id =
+    DE.Term.Const.get_tag id order_tag |> Option.is_some
+  in
+  (not (is_cstr id)) || has_attached_order id
+
+let[@inline always] of_term_cst id =
+  (* This assertion ensures that the API of the [Nest] module have been
+     correctly used, that is [Nest.attach_orders] have been called on
+     the nest of [id] if [id] is a constructor of ADT. *)
+  if not @@ has_order id then Fmt.failwith "not order on %a" DE.Id.print id;
```
bclement-ocp:
```suggestion
  if not @@ has_order id then Fmt.invalid_arg "no order on %a" DE.Id.print id;
```

bclement-ocp (changes requested):
Looks good overall — two small changes regarding Dolmen API use and we should be ready to merge :+1:

src/lib/frontend/d_cnf.ml
Halbaroth: I did this modification. We have an order on all constructors because we check it in `Uid.of_ty_cst` now ;)

bclement-ocp (approved):
LGTM

Nit: I did not notice earlier, but I am not a fan of the "cstr" abbreviation and would very much prefer "constr", if we need to abbreviate at all — we are not writing C and I don't think we need to import their hatred for vowels. Plus my brain reads "cstr" as "C string".

Halbaroth:
> LGTM
> 
> Nit: I did not notice earlier, but I am not a fan of the "cstr" abbreviation and would very much prefer "constr", if we need to abbreviate at all — we are not writing C and I don't think we need to import their hatred for vowels. Plus my brain reads "cstr" as "C string".

Sadly, I prefer this option but I use it mainly because it is the convention in Dolmen too :). I agree to uniform all of them to `constr`.

PR #1154, opened 2024-07-01
feat(BV): Add support for bv2nat/int2bv normal forms 

### LINKED ISSUE #824: BV/Int propagations
When we have (dis|in)equalities between `int2bv` / `bv2nat` terms we should propagate the relevant information to the other theory. For instance, if we have `bv2nat(x) = bv2nat(y)`, we should be able to learn `x = y`. This is the relational version of #801 (which is about computing `int2bv` /  `bv2nat` values when their argument is known).

See also #749 for an incomplete implementation of this.

bclement-ocp:
Currently, Alt-Ergo knows the value of [bv2nat] and [int2bv] on
constants. We also perform some simplifications during semantic value
creation; however, these simplifications are not integrated into the
rest of the solver. For instance, we know that [bv2nat(x @ y)] is equal
to [bv2nat(x) * 2^n + bv2nat(y)] (where [n] is the bit-width of [y]),
but we don't know that [z] is equal to the same under the hypothesis
that [z = x @ y].

This patch moves these simplifications to the [Bitv_rel] module, making
them an integral part of the solving process. Since we can't compute
[bv2nat] (or [int2bv]) of an arbitrary semantic value, we generate fresh
variables to represent them when needed.

Note: this PR depends on #1144 and #1152. Only the last commit titled "feat(BV): Add support for bv2nat/int2bv normal forms" is included in this PR.


bclement-ocp:
Fixes #824 

(I don't know why it does not appear in the "Development" dropdown...)

Halbaroth:
I'm ready to review this PR as soon as you rebase it ;)

bclement-ocp:
@Halbaroth done!

tests/bitv/testfile-bv2nat-007.smt2
```diff
@@ -0,0 +1,6 @@
+(set-logic ALL)
+(declare-const x Int)
+(declare-const y Int)
+(assert (= ((_ int2bv 1024) x) ((_ int2bv 1024) y)))
+(assert (distinct (mod (- x y) 179769313486231590772930519078902473361797697894230657273430081157732675805500963132708477322407536021120113879871393357658789768814416622492847430639474124377767893424865485276302219601246094119453082952085005768838150682342462881473913110540827237163350510684586298239947245938479716304835356329624224137216) 0))
```
Halbaroth, line 5: We may use the pow function here? Or you want to respect the SMT-LIB2 standard?

src/lib/reasoners/bitv_rel.ml
```diff
@@ -27,51 +27,558 @@ module L = Xliteral
...
+      let ( - ) = P.sub
+
+      let ( +$$ ) p n = P.add_const (Q.of_bigint n) p
+
+      let ( *$$ ) p n = P.mult_const (Q.of_bigint n) p
+
+      let mkv_eq p p' = Uf.LX.mkv_eq (to_repr p) (to_repr p')
+
+      let mkv_le p p' = Uf.LX.mkv_builtin true LE [to_repr p; to_repr p']
+    end
```
Halbaroth: We may move these helper functions to `Polynomial` in a notation module?

src/lib/reasoners/bitv_rel.ml
```diff
@@ -27,51 +27,557 @@ module L = Xliteral
...
+  type t =
+    { bv2nat : (P.t * Ex.t) Extraction.Map.t MX.t
+    (** Map from bit-vector extractions to their integer representation.
+
+        If [(bv, ext)] is associated with [(p, ex)] in this map, it means
+        that [p = bv2nat(bv<ext>)] is justified by [ex].
+
+        We maintain the invariant that if [(bv, ext)] is associated with
+        [(p, ex)] in this map, then [(bv, ext, ex)] is associated with [p]
+        in [nat2bv]. *)
```
Halbaroth:
```suggestion
        We maintain the invariant that if [(bv, ext)] is associated with
        [(p, ex)] in this map, then [p] is associated with [(bv, ext, ex)]
        in [nat2bv]. *)
```

src/lib/reasoners/bitv_rel.ml
```diff
@@ -27,51 +27,557 @@ module L = Xliteral
...
+      MX.update bv (function
+          | None -> None
+          | Some m ->
+            let m = Extraction.Map.remove ext m in
+            if Extraction.Map.is_empty m then None else Some m
+        ) t.bv2nat
+    in
+    { use ; nat2bv ; bv2nat ; eqs = t.eqs }
+
+  (* Returns the polynomial associated with [bv2nat(bv) asr ofs], or raises
```
Halbaroth:
```suggestion
  (* Returns the polynomial associated with [bv2nat(bv asr ofs)], or raises
```

src/lib/reasoners/bitv_rel.ml
```diff
@@ -27,51 +27,557 @@ module L = Xliteral
...
+     [Not_found] if there is none. *)
+  let find_asr bv ofs t =
+    find_ext bv (Extraction.shift_right ~size:(bitwidth bv) ofs) t.bv2nat
+
+  (* Returns the polynomial associated with [bv2nat(bv) asr ofs], creating a
```
Halbaroth:
```suggestion
  (* Returns the polynomial associated with [bv2nat(bv asr ofs)], creating a
```

src/lib/reasoners/bitv_rel.ml
Halbaroth: Note that it is not very important because we can define the usual operator `lasr` on integers as: `x lasr ofs = bv2nat(nat2bv(x) asr ofs)` but it is better to be consistent with our notations and we use `asr` on bitvectors in other comments.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -27,51 +27,557 @@ module L = Xliteral
...
+  let init_asr bv ofs t =
+    let (p, ex), t = find_or_create_asr bv 0 t in
+    if ofs = 0 then (p, ex), t
+    else
+      let (p_asr, ex_asr), t = find_or_create_asr bv ofs t in
+      let delta = T.Ints.(p - p_asr *$$ Z.(one lsl ofs)) in
+      let eqs = add_width_ineqs ~ex:(Ex.union ex ex_asr) ofs delta t.eqs in
+      (p_asr, ex_asr), { t with eqs }
+
+  (* Returns the polynomial associated with [bv2nat(bv) asr ofs].
```
Halbaroth:
```suggestion
  (* Returns the polynomial associated with [bv2nat(bv asr ofs)].
```

src/lib/reasoners/bitv_rel.ml
```diff
@@ -27,51 +27,557 @@ module L = Xliteral
...
+        remove_aux bv' ext' p { t with eqs }
+    in
+    let is_new, t =
+      match find_ext bv ext t.bv2nat with
+      | exception Not_found ->
+        true, t
+      | (p', ex') ->
+        let eqs = (T.Ints.mkv_eq p p', Ex.union ex ex') :: t.eqs in
+        false, remove_aux bv ext p' { t with eqs }
+    in
```
Halbaroth: I think it is a bit clearer to factorize these two exception handlers as follows:
```ocaml 
try 
   let bv', ext', ex' = T.Ints.Map.find p t.nat2bv in 
   ...
   let p', ex' = find_ext bv ext t.bv2nat in 
   ...
   false, t
with Not_found -> true, t
```

src/lib/reasoners/bitv_rel.ml
```diff
@@ -27,51 +27,557 @@ module L = Xliteral
...
+  (* Add the equality [bv_r = int2bv(int_r)].
+
+     We do not have a table mapping polynomials to their (truncated) bit-vector
+     representation; instead, we simply add the equation
+
+        bv2nat(int2bv(int_r)) = int_r + 2^|bv_r| * k
+
+     for a fresh variable [k].
+
+     If we know that [int_r] is an exact bit-vector representation (i.e. [int_r]
```
Halbaroth:
```suggestion
     If we know that [int_r] has an exact bit-vector representation (i.e. [int_r]
```

src/lib/reasoners/bitv_rel.ml
```diff
@@ -27,51 +27,557 @@ module L = Xliteral
...
+        entries in the map). *)
+    ; use : T.Ints.Set.t MX.t
+    (** Map from integer variables to the polynomials that use them.
+
+        This is used for integer substitutions because of our encoding of
+        extractions in [bv2nat]: we encode [bv2nat(bv<i, j>)] in
+        terms of [bv2nat(bv<0, i>)] and [bv2nat(bv<0, j>)] which we create
+        dynamically. The polynomials associated with these expressions are
+        thus not tracked by the UnionFind module, and are not given to
+        [subst]. *)
```
Halbaroth: Can you justify this design please?

Halbaroth (changes requested):

tests/bitv/testfile-bv2nat-007.smt2
bclement-ocp: I'd prefer keeping the SMT2 tests as compliant as possible. I don't think it is an issue to have a huge constant in a test.

src/lib/reasoners/bitv_rel.ml

src/lib/reasoners/bitv_rel.ml
bclement-ocp: I am always very wary when seeing calls to `find` functions that are not immediately within an exception handler because it is very easy to introduce subtle bugs this way.

In fact, the code you suggest has one such subtle bug: when adding the relation between `p` and `ext`, both `p` and `ext` might already have a recorded integer/bit-vector counterpart independently. If `T.Ints.Map.find p t.nat2bv` raises `Not_found` but `find_ext bv ext t.bv2nat` does not, we must return `false` and remove the existing mapping for `ext`, but your suggested code returns `true` and an unmodified `t`.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Done

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Done

src/lib/reasoners/bitv_rel.ml
bclement-ocp: It actually matters because `asr` on bit-vectors is truncated whereas on integers it is not!

src/lib/reasoners/bitv_rel.ml
bclement-ocp: It creates less variables. Instead of having $O(n^2)$ possible variables for each extraction, we only have $O(n)$ possible variables for each shift.

It also means the arithmetic theory knows equalities such as `x<2, 5> - x<2, 3> = 4 * x<4, 5>` by normalization, which is nice, but really the motivation is to use less variables.

I updated the comment to explain this.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Fixed, and also clarified what "associated with" means here to clarify which is the key and which is the value.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Not the `mkv_eq` and `mkv_le` helpers, but otherwise yes. Done (although I did not use a notation module, following the usual convention e.g. in Zarith that operators are defined directly in the corresponding module, unlike syntax extensions such as `let*`).

Halbaroth (approved):

PR #1159, opened 2024-07-11
Rename `cstr` into `constr`

Halbaroth:
We have no convention to abbrevate `constructor` or `destructor`. This commit chooses the convention:
constructor -> constr -> c
destructor -> destr -> d

Notice that `Dolmen` uses the `cstr` and `dstr`, so there are still some places, mainly in `D_cnf`, where these abbreviations appear.


bclement-ocp (approved):
>  destructor -> destr -> d

I don't see any change regarding this; I suppose we were already using this convention?

Halbaroth:
> > destructor -> destr -> d
> 
> I don't see any change regarding this; I suppose we were already using this convention?

I thought I would need to change some of them, but I did not. I kept the message to make clear we will stick this convention.

PR #1160, opened 2024-07-11
Use `Array.fold_right` in `D_cnf`

Halbaroth:
There are several places in `D_cnf` where we use the iterator
`Array.fold_left` to build lists, then we call `List.rev` on it to
preserve the order from the AST.

We can use `Array.fold_right` here because:
- This iterator is tail-rec too;
- There is no side effects which require to be perform in the order of
  the AST;
- We do not need to reverse the output list.

This PR is rebase on #1159.


bclement-ocp (approved):
My only concern is that this might result in slightly different behavior compared to the legacy typechecker, but I don't think the order in which we process these should have any observable side-effect.

Halbaroth:
> My only concern is that this might result in slightly different behavior compared to the legacy typechecker, but I don't think the order in which we process these should have any observable side-effect.

Let's run benchmark before merging.

PR #1164, opened 2024-07-16
Do not erase type constraints in `Ty.fresh`

### LINKED ISSUE #1161: Discrepancy between the legacy frontend and the Dolmen frontend
Consider the following input in native language:
```
type 'b sy = { V : 'b }

function rename(u: 'b sy, s: ('b,'c) farray) : 'c sy = { V = s[u.V] }

logic f : ('b,'b sy) farray

goal g:
  forall t:int.
  forall s1:(int,int) farray.
  forall s2:(int,int) farray.
  s1[t] = s2[t] ->
  rename({ V = t }, s1) = rename({ V = t }, s2)
```
Alt-Ergo answers `Valid` with Dolmen frontend in both `v2.5.4` and `next` but `I don't know` with the legacy frontend. 
If we inline the definition of `rename` in the goal as follows:
```
goal g:
  forall t:int.
  forall s1:(int,int) farray.
  forall s2:(int,int) farray.
  s1[t] = s2[t] ->
  s1[t] = s2[t]
```
 Alt-Ergo answers `Valid` in all the cases.

Halbaroth:
The purpose of the function `Ty.fresh` is to replace all the type variables occurring in its input type by fresh type variables. This function is only used by the legacy typechecker in order to simulate local typechecking contexts. We have to perform this substitution before unifying types because the unification function [Ty.unify] used by the legacy frontend is destructive.

The variant `Tvar` of `Ty.t` is a node used in a kind of union-find that store type constraints. It is important to preserve these constraints in `Ty.fresh`. So we have to generate a fresh type variable only if `Tvar` is the representative of its own class in the implicit union-find, that is if the `value` is field is `None`. Otherwise, we have to call recursively `Ty.fresh` on `value`.

Notice that this patch proceed a little bit differently. We use the function `Ty.shorten` that performs two tasks:
1. Path compressing for the implicit union-find. This task is performed in place.
2. Return a type whose `Tvar` subterms that are not roots in the union-find have been replaced by a concrete type or a free type variable.

So in particular `Ty.shorten t` does not contain subterms of the form `Ty.Tvar { value = Some _; _ }` (but `t` can still contain such subterms after calling the function).

~~This PR modifies also the way we print type variables because the previous printer `Ty.pp_smtlib` does not print differently type variables with constraints and without.~~

This PR fixes #1161.


Halbaroth:
I tested the patch with Marvin and I got `+200-1` and a slight slow down. I will run benchmarks to compare it with `Dolmen` now.

src/lib/structures/ty.ml
```diff
@@ -60,7 +60,7 @@ module Smtlib = struct
     | Trecord { args; name; _ } | Tadt (name, args) ->
       Fmt.(pf ppf "(@[%a %a@])" Uid.pp name (list ~sep:sp pp) args)
     | Tvar { v; value = None; _ } -> Fmt.pf ppf "A%d" v
-    | Tvar { value = Some t; _ } -> pp ppf t
+    | Tvar { value = Some t; _ } -> Fmt.pf ppf "[%a]" pp t
```
bclement-ocp: I think this should be reverted. Types that include variables that are no longer canonical should behave identically to types where the corresponding substitution has been performed: the `value` field here is just an artifact of an efficient implementation of substitutions using mutability; it should not appear in user-facing output (notwithstanding the fact that this syntax is not SMT-LIB compliant).

src/lib/structures/ty.ml
```diff
@@ -684,9 +689,10 @@ let rec monomorphize ty =
   | Tadt(name, params) ->
     Tadt(name, List.map monomorphize params)
 
-let print_subst fmt sbt =
-  M.iter (fun n ty -> Format.fprintf fmt "%d -> %a" n print ty) sbt;
-  Format.fprintf fmt "@?"
+let print_subst =
+  let sep ppf () = Fmt.pf ppf " -> " in
+  Fmt.(box @@ braces
+       @@ iter_bindings ~sep:comma M.iter (pair ~sep int pp_smtlib))
```
bclement-ocp: Before, we were using `print` and now we use `pp_smtlib`; why the change?

src/lib/structures/ty.ml
```diff
@@ -370,7 +370,12 @@ let apply_subst =
   fun s ty -> if M.is_empty s then ty else apply_subst s ty
 
 let rec fresh ty subst =
-  match ty with
+  match shorten ty with
+  | Tvar { value = Some _; _ } ->
+    (* This case is eliminated by the normalization performed
+       in [shorten]. *)
+    assert false
+
```
bclement-ocp: `fresh` can call itself recursively on sub-terms, and `shorten` itself recurses into the term, so this introduces quadratic complexity. `shorten` should only be called once on the top-level call of `fresh` but not in the recursive calls. This can be done by shadowing `fresh`:

```ocaml
let fresh ty subst = fresh (shorten ty) subst
```

bclement-ocp (changes requested):

src/lib/structures/ty.ml
Halbaroth: no reason

src/lib/structures/ty.ml
Halbaroth: Sure. Actually we should create a printer for debugging only instead of using the SMT-LIB or AE ones depending on the input format. For types, Dolmen's one should be sufficient.

bclement-ocp (approved):

PR #1165, opened 2024-07-16
feat(BV, Opt): Support bit-vector optimization

### LINKED ISSUE #884: Support optimization for the BV theory

bclement-ocp:
Since we track intervals for bit-vector variables, adding support for optimization in the bit-vector theory is fairly inconsequential.

Fixes #884


tests/models/bitv/optim-4.models.smt2
```diff
@@ -0,0 +1,17 @@
...
+(define-fun bvabs32 ((x (_ BitVec 32))) (_ BitVec 33)
+  (ite
+    (= ((_ extract 31 31) x) #b1)
+    (concat #b0 x)
+    (concat #b1 x)))
+(maximize (bvabs32 x))
+(assert (bvslt x (_ bv0 32)))
+; https://github.com/OCamlPro/alt-ergo/issues/1163
+(declare-fun P ((_ BitVec 33)) Bool)
+(assert (P (bvabs32 x)))
```
Halbaroth: We can remove these lines after merging #1166.

src/lib/structures/expr.mli
```diff
@@ -522,6 +522,8 @@ end
     https://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml
     https://smtlib.cs.uiowa.edu/logics-all.shtml#QF_BV *)
 module BV : sig
+  val of_Z : size:int -> Z.t -> t
```
Halbaroth, line 525: We expose this function that raises an assertion if `size <= 0`. We should raise `invalid_arg` instead.

Halbaroth (changes requested):

Halbaroth (approved):
Great :) 

You just need to push new expected files.

bclement-ocp:
I did, but looks like I messed up a newline by editing the file directly. Fixed.

PR #1169, opened 2024-07-17
Deploy several documentations

Halbaroth:
This PR adds the ability to deploy several documentations of Alt-Ergo. The new documentation workflows work as follows:
- the workflow `build_doc` builds both sphinx and odoc documentations and merges them in a single artifact. This workflow is trigger after pushing on PR;
- while pushing on `next`, the new `gh-pages` workflow retrieves the previous artifact and deploy it in the directory `/dev` of the `gh-pages` branch;
- if we want to add a new release documentation, we have to do by hand using the target `doc` of our makefile and update the `index.mld` file of `gh-pages`.

This PR fixes #693 

I also removed the support for `gtk-lang` of the native input because we removed the GUI.

PR #1170, opened 2024-07-19
fix(BV): Do not lose explanations in bvmul

bclement-ocp:
The implementation of bvmul from #1144 introduced a soundness bug: when we do not know anything about the result, the explanation is dropped.

This is because the implementation was performing mixing bitlist computation and creation of raw bitlist values. This patch fixes the implementation by performing all computations in [Z] and only adding the explanation at the end.

### DIFF
--- a/src/lib/reasoners/bitlist.ml
+++ b/src/lib/reasoners/bitlist.ml
@@ -321,20 +321,19 @@ let mul a b =
     let low_a_known = Z.trailing_zeros @@ a.bits_unk in
     let low_b_known = Z.trailing_zeros @@ b.bits_unk in
     let low_known = min low_a_known low_b_known in
-    let mid_bits = exact Z.(value a * value b) ex in
+    let mid_bits = Z.(value a * value b) in
     let mid_bits =
       if low_known = max_int then mid_bits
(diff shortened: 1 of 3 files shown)


Halbaroth (approved):
Looks good to me. Please don't merge before #1169

Halbaroth:
Can you add a test that caught the soundness bug in the previous code?

bclement-ocp:
Yes and no — I don't have a reproducer on next. I have one on #1152 where it was found, but the propagators run in a different order on `next` and #1152 

I can add the reproducer that breaks on #1152 even though i don't think it is very robust to changes in propagation strategy. Now that I understand the issue I will see if I can make a more robust test

Halbaroth:
Sure, if the test is not robust we shouldn't add it.

bclement-ocp:
Thinking about it more, I think the test is actually fairly robust but very hard to trigger on `next` due to a combination of us being over-conservative in adding explanations to constraints (whenever a substitution is performed on an argument of the constraint, the corresponding explanation is added to the whole constraint) *and* transforming bit-level information into Shostak equations. Without this specific combination the test would trigger.

I have added the test.

PR #1171, opened 2024-07-22
fix(CI): Ensure binary releases have proper version number

### LINKED ISSUE #1131: version number for binary releases
There seems to be an issue with how binary releases are produced, in that the `--version` they return is the hash of the corresponding git commit, instead of a version number.
This is problematic because : 
- it is not consistent with binaries produced when installing alt-ergo through opam
- it makes why3 fail to detect the alt-ergo version (since why3 is expecting `--version` to behave as it does for opam-built releases)

bclement-ocp:
The issue boils down to GitHub not downloading tag information when running the workflow, which means that `dune-build-info` and `dune-subst` see a lightweight tag rather than an annotated tag and ignore it.

In order to facilitate deployment, update the workflow to automatically upload the artifact to the release if applicable.

Fixes #1131

### DIFF
--- a/.github/workflows/build_static.yml
+++ b/.github/workflows/build_static.yml
@@ -4,16 +4,18 @@ on:
   workflow_dispatch:
   pull_request:
   push:
+    tags:
+      - "v*.*.*"
     branches:
       - next
       - main
 
 jobs:
-  macos-static:
+  build-macos-static:
     name: Build statically linked macOS binaries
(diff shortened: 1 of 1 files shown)


Halbaroth (approved):

bclement-ocp:
Do I need to wait for #1169 before merging?

Halbaroth:
Alas, yes... I am struggling with github permissions :/

PR #1173, opened 2024-07-22
fix(AC, Arith): Use a separate namespace for AC abstractions

### LINKED ISSUE #1172: Confusion between fresh AC and Shostak abstraction procedures
Related to: #989

We use the same symbol name space `Fresh` for fresh names generated for the AC and Shostak generation procedures. However, the arithmetic module tries to detect when abstracted constants have been created in polynomials and selectively applies distributivity based on this information.

I do not understand exactly what is going on there, but I do know that fresh names that do not come from the AC abstraction procedure (e.g. come from Shostak abstraction of record fields) should not impact whatever condition is used here for distributivity. Here is an example where this causes us to reply `unknown` where we should be able to reply `unsat`:

```smt-lib
(set-logic ALL)
(declare-datatype record ((mkrecord (field Int))))
(declare-const boxed record)
(define-fun add ((x Int) (y Int)) Int (+ x y))
(assert
 (distinct
  (* (field boxed) (add (field boxed) 1))
  (add (* (field boxed) (field boxed)) (field boxed))))
(check-sat)
```

Here, `boxed` becomes abstracted into `{ field = .k0 }`, which causes `(field boxed)` to canonicalize to `.k0`, which is a fresh name, and so we do not apply distributivity. Adding the following lines before the `assert`:

```smt-lib
(declare-const x Int)
(assert (= (field boxed) x))
```

makes us correctly reply `unsat`.

This is also a problem for the variables introduced by #743 that are fresh names but should not prevent distributive simplifications from occurring.

bclement-ocp:
We are using the same name space for all the constants introduced by any abstraction procedure. However, we also have code in the arithmetic module that cares specifically about abstractions introduced by AC(X) in order to determine whether distributivity should be applied or not.

This patch uses different namespaces for symbols introduced for AC(X) purposes and symbols introduced by other abstraction procedures (such as Shostak's record abstraction). After reviewing the different places where we use `Expr.fresh_name`, I believe only the two locations changed to `Expr.fresh_ac_name` in this commit are related to AC(X), and other uses of `Expr.fresh_name` are for separate, independent abstraction procedures.

Fixes #1172


src/lib/reasoners/uf.ml
```diff
@@ -1188,8 +1188,7 @@ let compute_concrete_model_of_val cache =
        We should check if these symbols can be defined as solvable to
        remove this particular case here. *)
     if X.is_solvable_theory_symbol f || is_destructor f
-       || Sy.is_internal f || E.is_internal_name t || E.is_internal_skolem t
-       || E.equal t E.vrai || E.equal t E.faux
+       || Sy.is_internal f || E.equal t E.vrai || E.equal t E.faux
```
bclement-ocp, line 1191: `Sy.is_internal` is true for all names with a non-`User` namespace, so checking again for `Fresh` and `Skolem` namespaces with `E.is_internal_name` and `E.is_internal_skolem` is not necessary.

src/lib/structures/expr.mli
```diff
@@ -256,6 +254,11 @@ val real : string -> t
 val bitv : string -> Ty.t -> t
 val fresh_name : Ty.t -> t
 
+(* Special names used for AC(X) abstraction.
+   These corresponds to the K sort in the AC(X) paper. *)
```
Halbaroth: We use `odoc` comment in `mli`.

Halbaroth (approved):

bclement-ocp:
(Waiting on benchmarks to merge)

bclement-ocp:
This is +11-0 and slight performance improvement on internal dataset — merging

PR #1175, opened 2024-07-23
chore: Support bytecode-only environments

### LINKED ISSUE #1066: Check that we are compatible with bytecode-only environments
I think #1049 fixed the issue, but we should double-check.

See https://github.com/ocaml/opam-repository/pull/25294

bclement-ocp:
We use dune-site for the inequalities plugin, but not for the AB Why3 plugin. Since this plugin is deprecated, simply mark it as incompatible with bytecode-only environments.

Fixes #1066


Halbaroth:
It is incompatible with bytecode compilation because Dynlink cannot work in bytecode?

bclement-ocp:
No, it is incompatible with bytecode compilation because the `install` rule lists the `.cmxs` file which is only produced by native compilers. I am sure we could find a way to conditionally list this file, but I didn't find such a way from a quick look at the `dune` doc. Given that the plugin is deprecated anyways, I don't think it is worth spending time on it — especially since this is to please the OPAM CI only, the alt-ergo-plugin-ab-why3 package already does not build in bytecode-only environments.

Halbaroth (approved):

PR #1176, opened 2024-07-23
feat(doc): Versioned documentation pages

### LINKED ISSUE #693: Make sure that the documentation on the website is that of the last release

bclement-ocp:
This patch overhauls documentation generation to support separate documentation pages for each Alt-ergo version on the GitHub pages website.

We generate one version for the `next` and `main` branch, and one version for each tag of the form `v*.*.*`. The version selector from Sphinx is used to allow users to change the version (it appears in the lower-right corner of the page).

We also keep a `latest` version that points to the latest tag that was built (using a symbolic link). The default page redirects to the `latest` version.

As part of this overhaul of the documentation workflow, the OCaml documentation built by `odoc` is now also versioned and added to the `API` sub-directory of the Sphinx documentation (with a placeholder to prevent Sphinx from complaining).

Fixes #693

This PR will require some cleanup of the gh-pages branch that I plan to do after merging. It should work however with the current state of the branch.

See this PR in action [here](https://bclement-ocp.github.io/alt-ergo/next/). I have also back-ported the documentation for version v2.5.4 (but it does not have the version selector on the bottom) [here](https://bclement-ocp.github.io/alt-ergo/v2.5.4/), which is also the place the previous documentation URL would go.

Note: the website should be updated to point to `https://ocamlpro.github.io/alt-ergo/latest` once this PR is merged.


.github/workflows/documentation.yml
```diff
@@ -44,62 +51,58 @@ jobs:
...
+          echo "target=${GITHUB_REF#*/*/}" >> $GITHUB_OUTPUT
+
       # Build the sphinx documentation
       # and automatically print any error or warning
       - name: Build sphinx documentation
         uses: ammaraskar/sphinx-action@master
         with:
           docs-folder: "docs/sphinx_docs/"
-          build-command: "sphinx-build -b html . _build"
+          build-command: |
```
Halbaroth, line 104: We may add the flag `-W` to fail on warnings?

bclement-ocp: Unfortunately not yet because we have warnings. We can do it once they are fixed.

bclement-ocp: (The warning existed before this PR, it is related to licenses — I did not investigate)

Halbaroth: Ok let's merge and we will fix the warnings later.

Halbaroth (approved):

PR #1178, opened 2024-07-24
Fix optimization with incremental mode

### LINKED ISSUE #993: Optimization and `push` and `pop`

Halbaroth:
This PR fixes the issue 993. Before this PR we could not use
optimization in incremental mode.

Now we store all the objectives in a vector of lists in the environment
of SatML. We push/pop this vector in `Satml.push/Satml.pop` and we add
new objective in the last list in `Satml.optimize`. All the objectives
to one incremental guard are given to the theory just after decide the
guard. The toplevel is a corner case: we do not store objectives at the
toplevel. Instead we give them directly to the theory in `Satml.optimize`.
Notice that `Satml.optimize` has to be used at the decision level 0.
This invariant was not ensure by an assertion.

This commit also fixes another tiny issue. If you consider the following
input file:
```smt2
(set-logic ALL)
(declare-const x Int)
(assert (<= x 10))
(maximize x)
(maximize x)
(check-sat)
(get-objectives)
```
we expect that AE will output twice `(x 10)` but before this commit it will
output it only once because we do not use a unique identifier per `maximize`
statement. I add a unique identifier for them in `Solving_loop` and use
it to to order them in the objective model. This design is more robust and
we can add the objectives in any order to the theory in a same incremental level.

As I explained during the last dev meeting, I also have to ensure that
SatML decides all the incremental guards before answering `sat`.
Otherwise, it could answer `sat` before giving all the objective
functions to the theory. This case happens if we use push/pop without
`assert` at this incremental level:
```smt2
(set-logic ALL)
(declare-const x Int)
(assert (<= x 10))
(push 1)
(maximize x)
(check-sat)
(get-model)
(get-objectives)
(check-sat)
(pop 1)
```
On `next`, AE answers `x = 0` here.

I also rename the field `next_objective` to `next_optimized_split`
because it contains the last split produced by the optimization
procedure of a theory.

Fix issue #993


src/lib/util/vec.ml
```diff
@@ -25,6 +25,8 @@
 (*                                                                        *)
 (**************************************************************************)
 
+module A = Array
```
bclement-ocp: Not sure what this change has to do with the PR, please revert it (especially since `A` is already used as an abbreviation for `Xliteral` elsewhere).

src/lib/util/vec.mli
```diff
@@ -83,13 +83,19 @@ val push : 'a t -> 'a -> unit
...
+    @raise Assert_failure if the index is not valid. *)
 
 val set : 'a t -> int -> 'a -> unit
 (** Set the element at the given index, either already set or the first
     free slot if [not (is_full vec)], or
     @raise Invalid_argument if the index is not valid. *)
 
+val replace : (' a -> 'a) -> 'a t -> int -> unit
+(** [replace f vec n] replaces the n-th element of [vec]
+    by [f (get vec n)].
```
bclement-ocp:
```suggestion
(** [replace f vec n] is equivalent to
     [set vec n (f (get vec n))], but with a single bound check.
```

src/lib/reasoners/satml.ml
```diff
@@ -2260,7 +2280,12 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
       assert (env.next_dec_guard <= Vec.size env.increm_guards);
     enqueue env g.neg 0 None
 
-  let optimize env ~is_max obj =
-    let fn = Objective.Function.mk ~is_max obj in
-    env.tenv <- Th.add_objective env.tenv fn Objective.Value.Unknown
+  let optimize env fn =
+    if decision_level env > 0 then Fmt.invalid_arg "Satml.optimize";
+    assert (decision_level env = 0);
```
bclement-ocp: No need for an `assert` after the test above.

src/lib/reasoners/satml.ml
```diff
@@ -1840,13 +1849,22 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
         let a = Vec.get env.increm_guards env.next_dec_guard in
         (assert (not (a.neg.is_guard || a.neg.is_true)));
         env.next_dec_guard <- env.next_dec_guard + 1;
-        make_decision env a
+        make_decision env a;
+        let objs = Vec.get env.objectives (env.next_dec_guard-1) in
+        let tenv =
+          List.fold_left
```
bclement-ocp: It would make more sense to use `List.fold_right` here to ensure objectives are added in the same order as in the input problem. It should not change anything with the current implementation (except if API users create multiple objective functions with the same index), but it would be at least helpful for debugging and should not impact performance noticeably.

bclement-ocp (changes requested):
> I add a unique identifier for them in Solving_loop and use
> it to to order them in the objective model. This design is more robust and
> we can add the objectives in any order to the theory in a same incremental level.

It feels a bit icky to give control of the unique identifiers to users of the `Objective` module. I'd be OK with merging this implementation (w/ my other remarks addressed), however keeping the existing order-based API with unique identifiers generated in `Objective.Function.mk` would seem to provide less coupling.

src/lib/util/vec.ml
Halbaroth: rofl

src/lib/reasoners/satml.ml
Halbaroth: oups I forgot to remove it

Halbaroth:
> > I add a unique identifier for them in Solving_loop and use
> > it to to order them in the objective model. This design is more robust and
> > we can add the objectives in any order to the theory in a same incremental level.
> 
> It feels a bit icky to give control of the unique identifiers to users of the `Objective` module. I'd be OK with merging this implementation (w/ my other remarks addressed), however keeping the existing order-based API with unique identifiers generated in `Objective.Function.mk` would seem to provide less coupling.

This was my previous version. I did all this thing to avoid adding an extra global state hidden in a module.

bclement-ocp (approved):
LGTM once the tests are green

PR #1181, opened 2024-07-26
fix(BV): Implement abstract_selectors

### LINKED ISSUE #1180: BV does not properly implement `abstract_selectors`
For some reason, the bit-vector theory does not implement `abstract_selectors`.

```console
$ cat abstract_selectors.smt2
(set-logic ALL)
(declare-datatype boxed ((box (unbox (_ BitVec 16)))))
(declare-const b boxed)
(assert (= (concat (unbox b) #xbeef) #xdeadbeef))
(check-sat)

$ alt-ergo abstract_selectors.smt2
[Error]b.unbox                         
Fatal error: exception File "src/lib/reasoners/shostak.ml", line 489, characters 12-18: Assertion failed
```

bclement-ocp:
Fixes #1180

### DIFF
--- a/src/lib/reasoners/bitv.ml
+++ b/src/lib/reasoners/bitv.ml
@@ -1482,7 +1482,20 @@ module Shostak(X : ALIEN) = struct
 
   let term_extract _ = None, false
 
-  let abstract_selectors v acc = is_mine v, acc
+  let abstract_selectors v acc =
+    let acc, v =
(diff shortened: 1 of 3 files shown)


bclement-ocp:
Updated with the test output files that were missing.

bclement-ocp:
@Halbaroth this is a simple patch that fixes a common bug on the J3 dataset and should be quick to review.

Halbaroth (approved):

PR #1184, opened 2024-07-29
Use Unix eof for shell scripts

Halbaroth:
This modification is mostly motivated by the fact that we cannot test `opam` files on `Windows` if we try to install a git repository with `opam pin https://...`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,6 @@
 # Declare files will always have LF line endings on checkout.
-configure eol=lf
+configure text eol=lf
+*.sh text eol=lf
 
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,5 +1,6 @@
 # Declare files will always have LF line endings on checkout.
-configure eol=lf
+configure text eol=lf
```
bclement-ocp:
```suggestion
```

We no longer have a configure file

bclement-ocp (approved):

.gitattributes
Halbaroth: Indeed :)

PR #1185, opened 2024-07-29
chore(BV, CP): Refactor propagation mechanism

bclement-ocp:
This patch simplifies the propagation mechanism (currently used in the
bit-vector relations only) in order to accomodate different types of
propagators more easily.

In particular, there is now a single generic (and configurable) loop
that runs the propagators instead of a spaghetti of different loops for
each kind of propagators.

*Note:* This PR depends on and includes #1152 and #1154. Only the commit titled "chore(BV, CP): Refactor propagation mechanism" (and later commits) are part of this PR.


src/lib/util/type.mli
```diff
@@ -0,0 +1,53 @@
...
+
+module Id : sig
+  type 'a t
+
+  val make : unit -> 'a t
+
+  val uid : 'a t -> int
+
+  val provably_equal : 'a t -> 'b t -> ('a, 'b) eq option
+end
```
Halbaroth: I guess it is an artefact ;)

src/lib/reasoners/bitv_rel.ml
```diff
@@ -1877,94 +1951,144 @@ let simplify_all uf eqs touched (dom, idom) =
...
+    (* We run all propagations over a single domain to completion, then run
+       the consistency propagators to perform cross-domain propagations. *)
+    ignore (
+      run_schedule state @@ Repeat (
+        Round_robin
+          [| (Repeat (Single queues.propagation_queue), 1)
+           ; (Repeat (Single queues.consistency_queue), 1)
+          |]
+      )
+    );
```
Halbaroth: Nice code! It is far simpler to understand :D

src/lib/reasoners/bitv_rel.ml
```diff
@@ -1877,94 +1951,144 @@ let simplify_all uf eqs touched (dom, idom) =
...
+  ; bitlist_changed = HX.create 17 }
+
+(* Schedule specifications *)
+type schedule =
+  | Single of Any_propagator.Queue.t
+  (** Propagate a single constraint from the given queue. *)
+  | Round_robin of (schedule * int) array
+  (** Performs [n] propagations from the provided schedules in round-robin
+      fashion the coefficient adjusts how many propagations are performed
+      before moving on to the next schedule. *)
```
Halbaroth: It seems that you only use coefficient 1. Do you plan to use different coefficients later?

Halbaroth (changes requested):

src/lib/util/type.mli
bclement-ocp: Indeed — this is in `Compat` now. Removed.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: Not yet. I think I initially wanted to experiment with it then didn't. Also using "round-robin" here is somewhat incorrect since we don't cycle. I renamed and simplified the constructors.

Halbaroth:
Please rebase and we can merge :)

Halbaroth (approved):

PR #1187, opened 2024-07-29
Fix MacOS static build

Halbaroth:
We do not install `nums` anymore so we need to update the static build workflow.


bclement-ocp:
Can you run the workflow on your fork and check that it actually works? We don't build static macOS binaries on PRs.

bclement-ocp (approved):
Ah actually this is broken since #1175 so this should be OK

PR #1188, opened 2024-07-29
Fix static build macos v2.5.x

Halbaroth:
Backport of #1187


bclement-ocp:
I am not sure we can backport this — IIRC we need it because the version of `ocplib-simplex` that we use for 2.5 brings in `num` as a dependency, so this needs to be tested before merging.

Given that the build works even if there is an extra dependency I think it is simpler to just not backport it.

Halbaroth:
Good point, it does not work:
https://github.com/Halbaroth/alt-ergo/actions/runs/10147125715/job/28056778107

I thought we get rid of `nums` in #600 but I forgot we have this dependency through `ocplib-simplex`.

PR #1189, opened 2024-07-30
Rewrite `gen_link_flags` in OCaml

Halbaroth:
As we noticed yesterday, Dune does not call an appropriate shell with the stanza `(run ...)` on 
Windows. I tried `(run bash ...)` but I ran into issues with absolute and relative paths. Another solution consists in using 
the stanza `(enabled_if ...)` to prevent Dune from calling the script on Windows. This solution is ugly because the mini-language for actions in Dune is really limited.

Finally, I rewrite the simple script in OCaml. This solution works fine! 

I tested the workflow here: https://github.com/Halbaroth/alt-ergo/actions/runs/10162591930


src/bin/text/dune
```diff
@@ -1,9 +1,18 @@
 (documentation
   (package alt-ergo))
 
+(executable
+  (name gen_link_flags)
+  (libraries unix fmt stdcompat cmdliner)
+  (modules gen_link_flags)
+  (promote (until-clean)))
```
bclement-ocp: There should be no need to promote this executable, let's not clutter the source directory.

src/bin/text/dune
```diff
@@ -1,9 +1,18 @@
...
+
 (rule
- (with-stdout-to link_flags.dune
-  (run ./gen-link-flags.sh %{env:LINK_MODE=dynamic} %{ocaml-config:system})))
+ (target link_flags.dune)
+ (deps (:gen gen_link_flags.exe))
```
bclement-ocp: I believe this is not needed and you can just use `./gen_link_flags.exe` (or `%{bin:gen_link_flags}`) in the `run` command below

src/bin/text/dune
```diff
@@ -1,9 +1,18 @@
 (documentation
   (package alt-ergo))
 
+(executable
+  (name gen_link_flags)
+  (libraries unix fmt stdcompat cmdliner)
```
bclement-ocp: It doesn't look like we use `cmdliner`, it should not be a dependency.

bclement-ocp (commented):
This is not a request to use it (I think the script is simple enough this way) but did you look at https://dune.readthedocs.io/en/stable/dune-libs.html#configurator ? I believe it is supposed to provide facilities to do this sort of things.

src/bin/text/dune
Halbaroth: I wrote a previous version with `cmdliner` but it was a bit over-engineering. I forgot to remove the dependency.

src/bin/text/dune
Halbaroth: My first attempt was:
```
(executable
  (name gen_link_flags)
  (libraries unix fmt stdcompat)
  (modules gen_link_flags))

(rule
 (with-stdout-to link_flags.dune
  (run %{bin:gen_link_flags} %{env:LINK_MODE=dynamic} %{ocaml-config:system})))
``` 
But it does not work. It works if I add `public_name` and `package` in the `executable` stanza. I do not want to distribute this script. This is the only workaround I found to call `gen_link_flags` without shipping it in an opam package. I also add `until-clean` for this purpose.

bclement-ocp: Right, `%{bin}` scope is for installed executable — the following should work with no promotion tho:

```
(executable
  (name gen_link_flags)
  (libraries unix fmt stdcompat)
  (modules gen_link_flags))

(rule
 (with-stdout-to link_flags.dune
  (run ./gen_link_flags.exe %{env:LINK_MODE=dynamic} %{ocaml-config:system})))
```

bclement-ocp (approved):

PR #1190, opened 2024-07-30
Rewrite `gen_link_flags` in OCaml

Halbaroth:
Backport of #1189 

The workflow have been tested here: 
https://github.com/Halbaroth/alt-ergo/actions/runs/10162501317


bclement-ocp:
I am not sure we need to backport all these build fixes to the 2.5 ; I don't think we plan on making further releases from that branch?

Halbaroth:
Sure, I planned to publish a version `v2.5.5` with Windows support but we add this support only from `v2.6.x`.

PR #1191, opened 2024-07-31
Get rid of `stdcompat`

Halbaroth:
We do not need to use `stdcompat` because we only need to support few new functions from `stdlib`. This PR removes the `stdcompat` dependency and create a new module `compat` which contains all the functions we need to compile on OCaml 4.08.

I also rename `Lists` to `my_list` and `myUnix` to `my_unix` and now `my_list` only contains extra functions (which are not part of any stdlib versions).


src/bin/common/parse_command.ml
```diff
@@ -462,13 +462,13 @@ let mk_theory_opt () no_contracongru
     no_fm no_nla no_tcp no_theory restricted tighten_vars
     _use_fpa (theories)
   =
-  set_no_ac (not (Lists.mem Theories.equal Theories.AC theories));
+  set_no_ac (not (My_list.mem Theories.equal Theories.AC theories));
```
bclement-ocp: Nit: I think we could get rid of `My_list.mem` everywhere with `List.exists (Theories.equal Theories.AC) theories`. Not sure we want to.

src/lib/reasoners/records.ml
```diff
@@ -322,7 +322,7 @@ module Shostak (X : ALIEN) = struct
      | Record (lbs, ty) ->
       Record (List.map (fun (n,e') -> n, subst_access x s e') lbs, ty)
      | Access (lb, e', _) when compare_mine x e' = 0 ->
-      Lists.assoc Uid.equal lb s
+      Compat.List.assoc Uid.equal lb s
```
bclement-ocp:
```suggestion
      My_list.assoc Uid.equal lb s
```

src/lib/util/my_list.ml
```diff
@@ -12,23 +12,10 @@
...
 (*                                                                        *)
-(*     The Alt-Ergo theorem prover                                        *)
-(*                                                                        *)
-(*     Sylvain Conchon, Evelyne Contejean, Francois Bobot                 *)
-(*     Mohamed Iguernelala, Stephane Lescuyer, Alain Mebsout              *)
-(*                                                                        *)
-(*     CNRS - INRIA - Universite Paris Sud                                *)
-(*                                                                        *)
-(*     ---------------------------------------------------------------    *)
-(*                                                                        *)
```
bclement-ocp: This needs to stay since we simply rename the file while keeping the existing code.

src/lib/util/my_list.mli
```diff
@@ -12,29 +12,10 @@
...
-(*     The Alt-Ergo theorem prover                                        *)
-(*                                                                        *)
-(*     Sylvain Conchon, Evelyne Contejean, Francois Bobot                 *)
-(*     Mohamed Iguernelala, Stephane Lescuyer, Alain Mebsout              *)
-(*                                                                        *)
-(*     CNRS - INRIA - Universite Paris Sud                                *)
-(*                                                                        *)
-(*     ---------------------------------------------------------------    *)
-(*                                                                        *)
 (*     More details can be found in the directory licenses/               *)
```
bclement-ocp: Same, needs to stay.

src/lib/util/my_list.mli
```diff
@@ -12,29 +12,10 @@
...
 (*                                                                        *)
 (**************************************************************************)
 
-(** Lists utilies
-
-    This modules defines some helper functions on lists
-*)
-
-(** {3 Misc functions} *)
```
bclement-ocp: Why remove this?

src/lib/util/compat.ml
```diff
@@ -0,0 +1,131 @@
...
+    let rec aux accu l_accu = function
+      | [] -> accu, rev l_accu
+      | x :: l ->
+        let accu, x = f accu x in
+        aux accu (x :: l_accu) l in
+    aux accu [] l
+end
+
+module Bytes = struct
+  include Bytes
```
bclement-ocp: I would argue it would be better to do `open Bytes` at the beginning of the module, and `include Bytes` at the end (with `[@@@ocaml.warning "-32-33"]` at the top of the file). This way, we would use the stdlib version when it is available and only fall back to our version if its not.

(Same for the `List`, `Array`, etc. in this module)

bclement-ocp (changes requested):
You asked for review but kept the PR as draft, is this intended?

Looks fine overall.

src/bin/common/parse_command.ml
Halbaroth: Sure

src/lib/util/compat.ml
Halbaroth: I was looking for such trick in fact, thanks :)

bclement-ocp (approved):

PR #1192, opened 2024-08-01
fix(BV): Internalize sign_extend and repeat

bclement-ocp:
Currently, we convert `sign_extend` and `repeat` to the appropriate concatenations and extractions at the term level. This has the advantage that it is simple, but the inconvenient that we create a potentially large amount of terms that do not appear in the original problem (proportional to the size of the sign extension or repetition).

In addition to preventing obvious simplifications with extractions, the creation of many terms clutters the union-find which drastically slows down the solver in presence of sign extensions to large bit-widths.

Note: this impacts some of the Décysif datasets.


bclement-ocp:
> Note: this impacts some of the Décysif datasets.

More precisely, this patch is on average 6x faster and solves 10% more problems than current `next` on the J3 dataset.

tests/bitv/testfile-sign-extend-003.smt2
```diff
@@ -0,0 +1,4 @@
+(set-logic ALL)
+(declare-const x (_ BitVec 4))
+(assert (distinct ((_ extract 3 2) ((_ sign_extend 4) x)) ((_ extract 3 2) x)))
+(check-sat)
```
Halbaroth: This test is identical with `testfile-sign-extend-002.smt2`.

src/lib/structures/expr.ml
```diff
@@ -3097,17 +3097,17 @@ module BV = struct
       (Sy.Op (Sy.Extract (j, i))) [s] (Ty.Tbitv (i - j + 1))
 
   (* Other operations *)
-  let rec repeat i t =
+  let repeat i t =
     assert (i >= 1);
```
Halbaroth: We may use `invalid_arg` here.

Halbaroth (approved):
LGTM 
Feel free to apply my suggestions.

tests/bitv/testfile-sign-extend-003.smt2
bclement-ocp: Oops, this one was intended to test extraction of the high bits. Fixed.

src/lib/structures/expr.ml
bclement-ocp: Done

PR #1195, opened 2024-08-05
Do not use lock file for all OCaml versions

Halbaroth:
In the current version of our workflow, we use an opam lock file to ensure that we can reproduce builds. Unfortunately, there are different conflicts depending on the plateform and OCaml version. We use a hack to prevent opam from saving the version of OCaml's toolchain (see the target `lock` in our Makfile). This hack is not sufficient if we want to use the current workflow `build` on Windows.

This PR modifies the workflow `build` and the Makefile as follows:
1. We check the reproducibility only on OCaml 4.14.1. I choose this version because we both develop on this version with Nix.
2. We checked the Makefile on OCaml 4.14.2, which is strange because we check the installation with opam on OCaml 4.14.1. I replace 4.14.1 by 4.14.2.
3. I remove the flag `-y` in the target `dev-switch`, `deps` and `test-deps` because we use `OCAMLYES` in our workflows and Makefile's users want to check dependencies before running a slow task as `opam switch create` or `opam install`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
 permissions: {}
 
 env:
-  OCAML_DEFAULT_VERSION: 4.14.1
+  OCAML_DEFAULT_VERSION: 4.14.2
   # Add OPAMYES=true to the environment, this is usefill to replace `-y` option
   # in any opam call
   OPAMYES: true
@@ -27,7 +27,7 @@ jobs:
           - ubuntu-latest
         ocaml-compiler:
           - 4.08.1
-          - 4.14.1
+          - 4.14.2
           - 5.0.0
         exclude:
           # No longer seems to be available
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
 permissions: {}
 
 env:
-  OCAML_DEFAULT_VERSION: 4.14.2
+  OCAML_DEFAULT_VERSION: 4.14.1
   # Add OPAMYES=true to the environment, this is usefill to replace `-y` option
   # in any opam call
   OPAMYES: true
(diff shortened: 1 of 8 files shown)


bclement-ocp:
I don't see a reason to test using 4.14.1 now that 4.14.2 is out. We can update the pinned nixpkgs if we want consistency with the nix version.

Couple questions:
 - Shouldn't this PR be updating the lockfile?
 - How is this going to work with the CI also testing 4.08 and 5.0? Don't we risk seeing more frequent build failures there? 
 - Can you clarify the "different conflicts depending on the plateform and OCaml version" that motivate this change? Is this related to Windows support?

Halbaroth:
In the current version of this PR, the CI on OCaml 4.08 and 5.0 uses only one constraint: the version of OCaml. In my opinion, the main purpose of the lock file is to automatize bisections and most of the time, it is sufficient to do it with our dev environment. Checking our code works with older version of libraries is made by the CI of opam-repository. We can reproduce this behaviour here. Basically the CI of opam-repository checks if our package builds with a smallest solution of the constraints on its dependencies. In my opinion, it makes more sense than checking arbitrary version of libraries.

All the current conflicts can be solved by choosing precise versions of libraries. Windows' support adds an extra layer of complexity when we have to choose the library.

bclement-ocp:
> In the current version of this PR, the CI on OCaml 4.08 and 5.0 uses only one constraint: the version of OCaml.

But they are still using the lockfile (at least the workflow sets `OPAMLOCKED`), which now (I suppose — you did not include the new lockfile in the PR, which I assume is an omission) explicitly states the version of OCaml to use. I admit I am not 100% up to date on how lockfiles work but I thought that would not work.

Halbaroth:
@bclement-ocp It seems I cannot ping you for this review, uhm strange. Anyway, this PR is ready for a second pass.

bclement-ocp (approved):

PR #1198, opened 2024-08-07
feat(shostak): Transparent abstracted constants

bclement-ocp:
This patch introduces a new type of leaf in the Shostak module, called *abstract leaves*. Abstract leaves behave like constant terms, and are intended to replace the `X.term_embed (E.fresh_name ty)` pattern where possible.

An abstract leaf is created by calling `X.abstract` on an existing semantic value `r`. The abstract leaf `X.abstract r` is automatically defined to be equal to `r` when processed by the `Uf` and `Ccx` modules.

The benefits of abstract leaves over fresh term constants are twofold:

 - With fresh constants, when the same semantic value is abstracted multiple times, a new constant is created for each abstraction. Semantically, this is correct; however, this causes the introduction of unnecessary constants in the union-find that will just end up in the same equivalence class. With abstract leaves, abstracting the same semantic value multiple times always returns the same constant.

 - With fresh constants, an additional equation between the new constant and the original term to abstract must be kept and processed separately. Since abstract leaves embed their definition, the additional equations can be introduced automatically by the `Uf` and `Ccx` modules and do not need to be carried separately.

Abstract leaves are currently only used by the AC theory. In the future, it is intended to be used as a basis for implementing application of interpreted-but-not-solvable functions to semantic values (so that for instance we can denote `bv2nat(r)` where `r` is a semantic value). While it would be straightforward to adapt abstract leaves to be able to create terminal leaves for `bv2nat(r)` (without definitional equations), we want congruence closure and computation (so that `bv2nat(x @ y)` automatically becomes `bv2nat(x) * 2^n + bv2nat(y)`) to work for such leaves. This require additional adaptations to CC(X), and is expected to need abstract leaves to enforce an AC(X)-compatible ordering.

### DIFF
--- a/src/lib/reasoners/ac.ml
+++ b/src/lib/reasoners/ac.ml
@@ -199,7 +199,7 @@ module Make (X : Sig.X) = struct
   let rec abstract2 sy r =
     match List.find (is_other_ac_symbol sy) (X.leaves r) with
     | ac_lv ->
-      (* Abstraction in depth: [f(x, y) + 1] -> [@ac(f(x, y) + 1]
+      (* Abstraction in depth: [f(x, y) + 1] -> [@ac(f(x, y)) + 1]
          and not [@ac(f(x, y) + 1)]. *)
       abstract2 sy (X.subst ac_lv (X.abstract ~kind:Ac ac_lv) r)
     | exception Not_found -> r
--- a/src/lib/reasoners/ac.ml
+++ b/src/lib/reasoners/ac.ml
@@ -183,10 +183,10 @@ module Make (X : Sig.X) = struct
      presentation in the paper to accomodate those differences, globally,
      between the implementation and theoretical description of AC(X).
 
-     More precisely, `abstract2` will abstract terms that *contain* AC leaves
-     when they appear as argument of an AC symbol. This ensures that AC terms
-     satisfy the T_AC definition from page 22 of the paper, although
-     correctness of the corresponding abstraction process has not been proven.
+     More precisely, `abstract2` will abstract all AC leaves that appear in the
+     arguments of an AC symbol. This ensures that AC terms satisfy the T_AC
+     definition from page 22 of the paper, although correctness of the
+     corresponding abstraction process has not been formally proven.
      See also https://github.com/OCamlPro/alt-ergo/issues/989
 
      [1]: Canonized Rewriting and Ground AC Completion Modulo Shostak Theories:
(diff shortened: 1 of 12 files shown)


src/lib/reasoners/shostak.ml
```diff
@@ -51,6 +67,7 @@ struct
 
   type rview =
```
Halbaroth, line 68: It is more readable to make these type mutually recursives without using the fact that `CX` is a recursive module itself.
```ocaml 
  type rview =
    | Term of Expr.t
    | Abstract of Sig.abs_kind * r
    | Ac of AC.t
    | Arith of ARITH.t
    | Records of RECORDS.t
    | Bitv of BITV.t
    | Adt of ADT.t

and r = {v : rview ; id : int}

File: src/lib/reasoners/shostak.ml [hunk 2]
```diff
@@ -72,6 +89,7 @@ struct
         | Bitv t -> fprintf fmt "%a" BITV.print t
         | Adt t -> fprintf fmt "%a" ADT.print t
         | Term t -> fprintf fmt "%a" Expr.print t
+        | Abstract (k, t) -> fprintf fmt "@@%a(%a)" pp_abs_kind k CX.print t
```
Halbaroth, line 92: This notion of abstract semantic value has no relation with `abstract value` in the SMT-LIB standard, we may use a different prefix for them? This printer is only used in debug message so we do not need to print SMT-LIB terms.

src/lib/reasoners/shostak.ml
```diff
@@ -214,6 +238,16 @@ struct
 
   let term_embed t = hcons {v = Term t; id = -1000 (* dummy *)}
 
+  let abstract ~kind arg =
+    (* No need to nest abstractions
+       XXX: would it be OK to abstract terms into themselves? *)
+    let arg =
+      match arg.v with
+      | Abstract (_, arg) -> arg
```
Halbaroth, line 246: Is this case is currently trigger by some tests?

src/lib/reasoners/ac.ml
```diff
@@ -196,42 +196,24 @@ module Make (X : Sig.X) = struct
...
-        let k = Expr.fresh_ac_name ty in
-        let eq = Expr.mk_eq ~iff:false k t in
-        X.term_embed k, eq::acc
-
-    else
-      r, acc
+  let rec abstract2 sy r =
+    match List.find (is_other_ac_symbol sy) (X.leaves r) with
+    | ac_lv ->
+      (* Abstraction in depth: [f(x, y) + 1] -> [@ac(f(x, y) + 1]
```
Halbaroth:
```suggestion
      (* Abstraction in depth: [f(x, y) + 1] -> [@ac(f(x, y)) + 1]
```

src/lib/reasoners/ac.ml
```diff
@@ -196,42 +196,24 @@ module Make (X : Sig.X) = struct
             Volume 8, Issue 3.
           doi:10.2168/LMCS-8(3:16)2012
           https://arxiv.org/pdf/1207.3262.pdf *)
-  let abstract2 sy t r acc =
-    if List.exists (is_other_ac_symbol sy) (X.leaves r) then
-      match X.ac_extract r, Expr.term_view t with
-      | Some ac, { f = Name { hs; kind = Ac; _ } ; xs; ty; _ } ->
-        (* It should have been abstracted when building [r] *)
-        assert (not (Sy.equal sy ac.h));
```
Halbaroth, line 204: This assertion is not relevant anymore?

bclement-ocp:
I expected this to have little to no impact but there seems to be some regressions on the internal dataset. I don't think we need it for 2.6.0 as the follow-up won't be complete by then, so I'm marking the PR as draft for now and will revisit in the next release cycle.

Halbaroth:
Do you want my review? I was reviewing it yesterday.

bclement-ocp:
Sure — I think it will be mostly identical once it is posted again, but I want to make sure I understand the regressions better. If needed we will keep `fresh_ac_name` for AC abstractions and use the new mechanism for new abstractions.

src/lib/reasoners/ac.ml
bclement-ocp: We are not looking at the toplevel symbol for `r` anymore, so `ac_lv` is guaranteed to have a different head symbol already (since it is returned by `List.find (is_other_ac_symbol sy)`).

src/lib/reasoners/shostak.ml
bclement-ocp: This is somewhat true. I wrote it, and would like to keep it, this way for two reasons:
 - It avoids nested recursion (type-level recursion inside a module-level recursion; the `rview` type already depends recursively on `CX.r` through the different Shostak modules)
 - Uniform with the use at the term level where we call e.g. `CX.print`; otherwise we would have to make all functions in this module recursive themselves.

src/lib/reasoners/shostak.ml
bclement-ocp: It does have a loose relationship with it since in both cases they are abstract names for constants, so I think that reusing `@` makes sense. Note that the abstracted semantic values are written `@kind(arg)` whereas the SMT-LIB abstract constants are written `@name` with no arguments.

I think this disambiguates them enough to make using `@` acceptable here, but I am OK with switching to another prefix (`%`?).

src/lib/reasoners/shostak.ml
bclement-ocp: I don't think we can reach this case currently; it is more of a future-proofing thing.

Note that since we don't look inside abstracted terms, except to enforce definitional equality, it is OK to replace the inner value with any other representation provided that the two representations are definitionally equal (and `Abstract (_, arg)` is definitionally equal to `arg`).

src/lib/reasoners/shostak.ml
Halbaroth: In fact this printer is only used for debugging. We use the printer of `Expr` to print models and currently, this is the only printer we really need to keep SMT-LIB compliant. Let us keep `@` with parentheses.

src/lib/reasoners/shostak.ml
Halbaroth: Indeed, let us keep your code so.

PR #1202, opened 2024-08-09
Remove warnings

Halbaroth:
After upgrading Dune, `ocamlc` outputs new warnings (I guess the default warning flags of Dune changed). This PR removes them.

Notice that there is still one warning:
```
record field acts_add_lit_view is never read
```
I guess this field will be used soon, so I do not touch it.


bclement-ocp (approved):
> Notice that there is still one warning:
> 
> ```
> record field acts_add_lit_view is never read
> ```
> 
> I guess this field will be used soon, so I do not touch it.

It is not used because we use it through the convenience `acts_add_eq`. It is OK to keep the function in `simplify_all` and remove the record field.

Halbaroth:
I prefer avoiding annoying rebases for you ;)

Issue #785, opened 2023-08-21
Refactoring the output channels


Halbaroth:
The current situation of the output management in AE is a bit of a mess.

Basically, the Alt-Ergo library should never print anything by itself. Such behaviour isn't expected from a library....
In my opinion, the library can only produce `Logs` sources to report debugging information on the diagnostic channel.

The Alt-Ergo binary prints results (answer, models and unsat core) on the regular output.

- [x] Removing useless output channels to fit with the SMT standard using only two channels (`regular` and `diagnostic`). See #782 
- [x] The Ocplib-simplex library shouldn't use the `App` level or set any level on its sources.
- [x] The Alt-Ergo library doesn't print any result by itself. In particular, we have to refactor the frontend of the SMT solver.
- [ ] Using Logs and sources everywhere in the Alt-Ergo library.
- [ ] Move the `module Output` from `options.ml` to the Alt-Ergo binary.

Halbaroth:
As we decided during the last dev meeting, I tried to rewrite #1068 (which adds logs library) in order to implement the current `Printer` module. The idea was to limit as much as possible the modifications to this module. In particular, I tried to preserve as much as possible the current API of `Printer`. Unfortunately, it seems there is no way to reimplement `print_std`, `print_err`, ... without breaking their API. The issue comes from the fact these functions are variadic. For instance, if we want to implement `print_dbg` with `Logs.debug (fun k -> k ...)`, we need to feed the continuation `k` with the variadic arguments of `print_dbg`. I and Vincent did not find a hack to work around this issue.

bclement-ocp:
Ah duh. Yeah I don't think we can preserve the current API then :cry:

PR #1206, opened 2024-08-12
Minimal `Logs` integration

Halbaroth:
This PR adds:
- `Logs` as a new dependency;
- Logs' sources for each debug flag;
- A reporter that prints on the appropriate output (regular or
  diagnostic) according to the source.

As explained in #785, we cannot do really better without breaking the API of `Printer`

This PR is rebased on #1204 

See #1207 for an example using sources introduced here.

Notice that I create two sources for theories with a Shostak and Relation modules. An alternative implementation consists in prefixing all the messages in `*_rel.ml` by `[Rel]` as it does in the legacy code. I have no strong opinion on this choice.

I did not create a bijection between debug flags and sources because it is more flexible to use two different lists. 
For instance, it could be useful to create flags that set different sources (and actually, `--dd adt` for instance set `adt` and `adt_rel` sources).


src/bin/js/worker_interface.ml
```diff
@@ -185,7 +185,7 @@ type options = {
   debug_cc : bool option;
   debug_combine : bool option;
   debug_constr : bool option;
-  debug_explanations : bool option;
+  debug_explanation : bool option;
```
bclement-ocp: The correct name for this feature is the plural "explanations"; please revert this change.

src/bin/common/parse_command.ml
```diff
@@ -213,50 +213,107 @@ module Debug = struct
     | Commands -> "commands"
     | Optimize -> "optimize"
 
+  let set_level src = Logs.Src.set_level src (Some Debug)
+
   let mk ~verbosity flags =
+    let module S = Sources in
```
bclement-ocp: I am not a big fan of using a global `Sources` module with all sources stuffed in. Since we are setting up sources in the `Parse_command` module which has everything has a dependency, would it be possible to have each module expose its own `val src : Logs.src` instead (like in the `Logs` examples in the documentation)?

Halbaroth: Sure, this was my first design in fact...

Halbaroth: Just, notice that some of them do not correspond with a module.

src/bin/js/worker_interface.ml
Halbaroth: Please justify `correct`. The name of module is `explanation`. I did this modification because the current name of the debug flag is `explanation` too.

src/bin/common/parse_command.ml
bclement-ocp: I'd say that the few that do not correspond with a module can either be:
 - Stuffed into `options.ml` for now, or
 - Moved to a relevant module (e.g. we could have `fm_src` in `intervalCalculus.ml`), or
 - Removed entirely as they are not used (we will figure out what to do with them at a later time)

src/bin/js/worker_interface.ml
bclement-ocp: > the current name of the debug flag is `explanation` too

I'd argue that this is incorrect as well (although I never noticed since I always write `-d ex`) ;)

What I mean is — there are multiple explanations used by Alt-Ergo that we want to debug; similar to `debug_commands`, `debug_warnings`, `debug_triggers`, `debug_types`, `debug_intervals`, etc. On the other hand the singular flags such as `debug_bitv`, `debug_interpretation`, `debug_matching`, `debug_uf`, etc. refer to a single data-structure / theory / step to debug (ignoring the fact that we duplicate the environment for case-split purposes).

"The correct name" was a bit strong, and I apologize — I don't think it matters too much either way. However I do believe that if it does not matter, it is best to not change whatever is there. Now that the change has been made let's keep it.

Halbaroth: I don't think it's matter too. When I see a discrepancy in our naming conventions and I am modifying this part of the code, I try to reduce it a bit. I agree to rename `explanation` to `explanations` everywhere (but for the module name of course).

Halbaroth: In fact, I removed the source from `Explanation` because `--d explanations` has no relation with this module. It increases the verbosity of explanations. We may add flags like `--verbosity explanations, ...`.

src/bin/common/parse_command.ml
```diff
@@ -154,7 +154,8 @@ module Debug = struct
     | Cc
     | Combine
     | Constr
-    | Explanation
+    | Explanation (* deprecated *)
+    | Explanations
```
bclement-ocp: We can simply use "explanations". Cmdliner uses prefix completion so that `-d explanation` will still work the same (on the other hand having both "explanation" and "explanations" breaks `-d ex`).

src/lib/reasoners/relation.ml
```diff
@@ -41,6 +41,9 @@ module Rel5 : Sig_rel.RELATION = Adt_rel
 
 module Rel6 : Sig_rel.RELATION = Ite_rel
 
+(* This source is unused. *)
+let src = Logs.Src.create ~doc:__MODULE__ ("AltErgoLib." ^ __MODULE__)
```
bclement-ocp: Maybe we don't need it then? I think all the other sources are also unused, why single this one out?

Halbaroth: It is an artefact. I added it because I tried to add the source in the signature of `Sig_rel`, which was not a good idea. We do not want to build all the modules in `Parse_commands` just to setup up sources.

src/bin/common/parse_command.ml
Halbaroth: Oh I do not know we can do that :) thanks

src/lib/reasoners/relation.ml
Halbaroth: In fact, it was not an artefact. I can remove `src` from `Sig_rel` but I have to introduce an extra signature in all the modules `*_rel.mli`.

bclement-ocp (approved):

PR #1207, opened 2024-08-12
Use logs in `Adt_rel`

Halbaroth:
This PR is rebased on #1204 and #1206

### DIFF
--- a/src/lib/reasoners/adt_rel.ml
+++ b/src/lib/reasoners/adt_rel.ml
@@ -323,7 +323,7 @@ let empty uf = {
 (*BISECT-IGNORE-BEGIN*)
 module Debug = struct
   let assume l =
-    Logs.debug ~src:Sources.adt_rel
+    Log.debug
       (fun k -> k "assume the literal:@ %a" (Xliteral.print_view X.print) l)
 end
 (*BISECT-IGNORE-END*)
@@ -403,7 +403,7 @@ let assume_not_is_constr ~ex r c domains =
 let add r uf domains =
   match X.type_info r with
   | Ty.Tadt _ ->
-    Logs.debug ~src:Sources.adt_rel (fun k -> k "add %a" X.print r);
+    Log.debug (fun k -> k "add %a" X.print r);
     let rr, _ = Uf.find_r uf r in
     Domains.init rr domains
   | _ ->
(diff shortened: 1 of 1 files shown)


src/lib/reasoners/adt_rel.ml
```diff
@@ -319,45 +319,9 @@ let empty uf = {
...
-
-  let add r =
-    if Options.get_debug_adt () then
-      print_dbg
-        ~module_name:"Adt_rel"
-        ~function_name:"add"
-        "%a" X.print r
-
+  let assume l =
+    Logs.debug ~src:Sources.adt_rel
```
bclement-ocp: Please don't use `Logs.debug ~src:...`. Instead, define `module Log = (val Logs.src_log Sources.adt_rel)` at the top of the file, and use `Log.debug` directly. See the [example](https://erratique.ch/software/logs/doc/Logs/index.html#setupreporter) from `Logs`'s documentation.

src/lib/reasoners/adt_rel.ml
```diff
@@ -319,45 +319,9 @@ let empty uf = {
...
-    if Options.get_debug_adt () then
-      print_dbg
-        ~module_name:"Adt_rel"
-        ~function_name:"assume"
-        "assume %a" LR.print (LR.make a)
-
-  let pp_domains loc domains =
-    if Options.get_debug_adt () then begin
-      print_dbg ~flushed:false ~module_name:"Adt_rel"
-        "@[<v 2>--ADT env %s ---------------------------------@ " loc;
```
bclement-ocp, line 337: Looks like this PR removes the the `---` separators that can make the logs more readable; is that intentional?

bclement-ocp (changes requested):

src/lib/reasoners/adt_rel.ml
Halbaroth: It is intentional because `Logs` uses correctly indentations. So we will get something like that:
```
[DEBUG][Adt_rel] environment before assume:
                 the environment here 
                 .... 
[DEBUG][Adt_rel] ... 
``` 
The indentation makes clear in which messages we are.

bclement-ocp (approved):
LGTM. Please merge #1206 first and rebase before merging.

Halbaroth:
Actually I already rebased it and I expected that github will skip empty commit. I rebased it again to be sure but now the CI is dead :D (it is not our fault).

PR #1208, opened 2024-08-14
Remove gtk-lang file

Halbaroth:
We do not need to keep this file as we remove the legacy GUI.


bclement-ocp (approved):
This file provides syntax coloring for the alt-ergo language to GTK widgets. Besides the old GUI, I think if it gets installed it might be picked up by GTK editors (e.g. gedit) to provide syntax coloring for alt-ergo files. Probably not worth worrying about.

Halbaroth:
The file is not installed by our package, which means it is very unlikely that someone uses it or even knows it exists.

PR #1209, opened 2024-08-14
Fix issue 929

### LINKED ISSUE #929: Wrong model for arrays
The current model produced by Alt-Ergo for the input `555.smt2` is wrong:
```
(set-logic ALL)
(declare-const x Int)
(declare-const y Int)
(declare-const a1 (Array Int Int))
(declare-const a2 (Array Int Int))
(assert (= (select a1 x) x))
(assert (= (store a1 x y) a1))
(assert (= (select a2 x) x))
(assert (= (store a2 x y) a1))
(check-sat)
(get-model)
``` 
The model is
```
(
  (define-fun x () Int 0)
  (define-fun y () Int 0)
  (define-fun a1 () (Array Int Int) (store (as @a1 (Array Int Int)) 0 0))
  (define-fun a2 () (Array Int Int) (store (as @a0 (Array Int Int)) 0 0))
)
```
But it should be 
```
(
  (define-fun x () Int 0)
  (define-fun y () Int 0)
  (define-fun a1 () (Array Int Int) (store (as @a0 (Array Int Int)) 0 0))
  (define-fun a2 () (Array Int Int) (store (as @a0 (Array Int Int)) 0 0))
)
```
My first thought was that Alt-Ergo didn't produce an appropriate case-split in `Arrays_rel` but it seems to be the very purpose of the field `splits` in its environment.
After a quick check, it seems Alt-Ergo discovers the equality `a0 = a1` but the equality never reaches the union-find.

Halbaroth:
This PR fixes the issue #929 about model arrays.

Our implementation of `ArrayEx` is complete but we do not split on literals `a = b` where `a` and `b` are any arrays of the problem. Doing these splits during the search proof would be a bad idea because we do not need them to be complete. In the example of the issue #929, we obtain an equality of the form `(store a1 x y) = (store a2 x y)` and we know that `(select a1 x) = (select a2 x)`. We never try to decide `a1 = a2`.

Notice that I update the set of arrays by tracking literals `Eq` of origin `Subst`. We know that this design is fragile because it could happen that substitutions are not sent to theories. The worst that can happen is to produced wrong models for arrays in very tricky cases. A better implementation consists in using our new global domains system to track `selects` in `Adt_rel` and use the keys of this map instead of `arrays`. I did not opt to this implementation to keep this PR simple and atomic.

### DIFF
--- a/src/lib/reasoners/arrays_rel.ml
+++ b/src/lib/reasoners/arrays_rel.ml
@@ -31,6 +31,7 @@ module A  = Xliteral
 module L  = List
 
 module X = Shostak.Combine
+module SX = Shostak.SXH
 module Ex = Explanation
 
 module LR = Uf.LX
@@ -91,6 +92,8 @@ type t =
    new_terms : E.Set.t;
    size_splits : Numbers.Q.t;
    cached_relevant_terms : (G.t * S.t TBS.t) H.t;
+   arrays : SX.t;
+   (* Set of all class representatives of arrays. *)
   }
 
 
@@ -103,6 +106,7 @@ let empty uf =
    new_terms = E.Set.empty;
    size_splits = Numbers.Q.one;
    cached_relevant_terms = H.create 1024;
+   arrays = SX.empty
   }, Uf.domains uf
 
 (*BISECT-IGNORE-BEGIN*)
@@ -415,17 +419,40 @@ let new_equalities env eqs la class_of =
   implied_consequences env eqs la
 
 (* choisir une egalite sur laquelle on fait un case-split *)
-let two = Numbers.Q.from_int 2
+let two = Numbers.Q.from_int 3
 
-let case_split env _uf ~for_model:_ =
+let split_arrays ~for_model env =
+  if for_model then
+    match SX.choose env.arrays with
+    | exception Not_found -> None
+    | a1 ->
+      match SX.find_first (fun a -> not @@ X.equal a a1) env.arrays with
+      | exception Not_found -> None
+      | a2 ->
+        Some (LR.neg @@ LR.mk_eq a1 a2)
+  else
+    None
+
+let (let*) = Option.bind
+
+let split_indices env =
+  let* cs = LR.Set.choose_opt env.split in
+  Some (LR.neg cs)
+
+let next_split ~for_model env =
+  match split_indices env with
+  | None -> split_arrays ~for_model env
+  | cs -> cs
+
+let case_split env _uf ~for_model =
   (*if Numbers.Q.compare
     (Numbers.Q.mult two env.size_splits) (max_split ()) <= 0  ||
     Numbers.Q.sign  (max_split ()) < 0 then*)
-  try
-    let a = LR.neg (LRset.choose env.split) in
-    Debug.case_split a;
-    [LR.view a, true, Th_util.CS (Th_util.Th_arrays, two)]
-  with Not_found ->
+  match next_split ~for_model env with
+  | Some cs ->
+    Debug.case_split cs;
+    [LR.view cs, true, Th_util.CS (Th_util.Th_arrays, two)]
+  | None ->
     Debug.case_split_none ();
     []
 
@@ -442,11 +469,27 @@ let count_splits env la =
   in
   {env with size_splits = nb}
 
+let is_array r =
+  match X.type_info r with
+  | Ty.Tfarray _ -> true
+  | _ -> false
+
 let assume env uf la =
   let are_eq = Uf.are_equal uf ~added_terms:true in
   let are_neq = Uf.are_distinct uf in
   let class_of = Uf.class_of uf in
   let env = count_splits env la in
+
+  (* Update the set of representatives of arrays. *)
+  let env =
+    List.fold_left
+      (fun env l ->
+         match l with
+         | Xliteral.Eq (r1, r2), _, _, Th_util.Subst when is_array r2 ->
+           { env with arrays = SX.remove r1 env.arrays |> SX.add r2 }
+         | _ -> env
+      ) env la
+  in
 
   (* instantiation des axiomes des tableaux *)
   Debug.assume la;
@@ -462,7 +505,15 @@ let assume env uf la =
   env, Uf.domains uf, { Sig_rel.assume = l; remove = [] }
 
 let query _ _ _ = None
-let add env uf _ _ = env, Uf.domains uf, []
+
+let add env uf r _ =
+  match X.type_info r with
+  | Ty.Tfarray _ ->
+    let rr, _ = Uf.find_r uf r in
+    let arrays = SX.add rr env.arrays in
+    { env with arrays }, Uf.domains uf, []
+  | _ ->
+    env, Uf.domains uf, []
 
 let new_terms env = env.new_terms
 let instantiate ~do_syntactic_matching:_ _ env _ _ = env, []
--- a/src/lib/reasoners/uf.ml
+++ b/src/lib/reasoners/uf.ml
@@ -1144,7 +1144,8 @@ module Cache = struct
 
     abstracts: (r, Expr.t) Hashtbl.t;
     (** Contains all the abstract values generated. This cache is necessary
-        to ensure we don't generate twice an abstract value for a given symbol. *)
+        to ensure we don't generate twice an abstract value for a given
+        symbol. *)
   }
 
   let mk () = {
--- a/src/lib/reasoners/uf.ml
+++ b/src/lib/reasoners/uf.ml
@@ -1134,34 +1134,56 @@ let terms env =
 
 (* Helper functions used by the caches during the computation of the model. *)
 module Cache = struct
-  let store_array_get arrays_cache (t : Expr.t) (i : Expr.t) (v : Expr.t) =
-    match Hashtbl.find_opt arrays_cache t with
-    | Some values ->
-      Hashtbl.replace values i v
-    | None ->
+  type t = {
+    arrays : (Expr.t, unit) Hashtbl.t;
+    (** Set of arrays for which we need to generate a model value.
+        Notice that this set also contains embedded arrays. *)
+
+    selects: (X.r, (Expr.t, Expr.t) Hashtbl.t) Hashtbl.t;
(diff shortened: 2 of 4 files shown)


src/lib/reasoners/arrays_rel.ml
```diff
@@ -415,17 +419,40 @@ let new_equalities env eqs la class_of =
   implied_consequences env eqs la
 
 (* choisir une egalite sur laquelle on fait un case-split *)
-let two = Numbers.Q.from_int 2
-
-let case_split env _uf ~for_model:_ =
+let two = Numbers.Q.from_int 3
```
bclement-ocp, line 422: Please don't write `let two = 3`; I would like to keep what remains of my sanity intact.

src/lib/reasoners/arrays_rel.ml
```diff
@@ -415,17 +419,40 @@ let new_equalities env eqs la class_of =
...
+let (let*) = Option.bind
+
+let split_indices env =
+  let* cs = LR.Set.choose_opt env.split in
+  Some (LR.neg cs)
+
+let next_split ~for_model env =
+  match split_indices env with
+  | None -> split_arrays ~for_model env
+  | cs -> cs
```
bclement-ocp, line 445: This is a somewhat verbose way of writing:

```ocaml
let next_split ~for_model env =
  match LR.Set.choose_opt env.split with
  | Some cs -> Some (LR.neg cs)
  | None -> split_arrays ~for_model env
```

src/lib/reasoners/uf.ml
```diff
@@ -1134,35 +1134,58 @@ let terms env =
...
+    selects: (X.r, (Expr.t, Expr.t) Hashtbl.t) Hashtbl.t;
+    (** Contains all the accesses to arrays. *)
+
+    abstracts: (r, Expr.t) Hashtbl.t;
+    (** Contains all the abstract values generated. This cache is necessary
+        to ensure we don't generate twice an abstract value for a given
+        symbol. *)
+  }
+
+  let mk () = {
```
bclement-ocp, line 1151: Let's call this `create`; it is the de-facto standard name in OCaml for this type of function.

src/lib/reasoners/uf.ml
```diff
@@ -1134,35 +1134,58 @@ let terms env =
...
 module Cache = struct
-  let store_array_get arrays_cache (t : Expr.t) (i : Expr.t) (v : Expr.t) =
-    match Hashtbl.find_opt arrays_cache t with
-    | Some values ->
-      Hashtbl.replace values i v
-    | None ->
+  type t = {
+    arrays : (Expr.t, unit) Hashtbl.t;
+    (** Set of arrays for which we need to generate a model value.
+        Notice that this set also contains embedded arrays. *)
```
bclement-ocp, line 1140: Can you clarify what you mean by an "embedded array"?

src/lib/reasoners/uf.ml
```diff
@@ -1134,35 +1134,58 @@ let terms env =
...
+
+    selects: (X.r, (Expr.t, Expr.t) Hashtbl.t) Hashtbl.t;
+    (** Contains all the accesses to arrays. *)
```
bclement-ocp, line 1143: Nit: This comment should explain what an entry in this table looks like. I know this comment is moved from an existing location, but it does not help understand the meaning of the table.

src/lib/reasoners/uf.ml
```diff
@@ -1134,35 +1134,58 @@ let terms env =
...
+    | exception Not_found ->
       let values = Hashtbl.create 17 in
       Hashtbl.add values i v;
-      Hashtbl.add arrays_cache t values
+      Hashtbl.add cache.selects r values
+    | values ->
+      Hashtbl.replace values i v
 
-  let get_abstract_for abstracts_cache env (t : Expr.t) =
+  let retrieve_selects cache env (t : Expr.t) =
```
bclement-ocp, line 1175: It is not clear from its name that this function can raise `Not_found`. If this is intentional, this function should be called `retrieve_selects_exn` instead (I'd even call it `find_selects_exn`; `retrieve` usually has an implication that work needs to be performed to construct the data, e.g. fetch it from the network).

src/lib/reasoners/uf.ml
```diff
@@ -1134,35 +1134,58 @@ let terms env =
...
+    selects = Hashtbl.create 17;
+    abstracts = Hashtbl.create 17;
+  }
+
+  let add_array cache env (t : Expr.t) =
+    Hashtbl.add cache.arrays t ();
+    let r, _ = find env t in
+    match Hashtbl.find cache.selects r with
+    | exception Not_found ->
+      Hashtbl.add cache.selects r (Hashtbl.create 17)
```
bclement-ocp, line 1162: I think the comment from the previous moved code is valuable and could be kept here.

src/lib/reasoners/uf.ml
bclement-ocp: Also, using `Expr.t` as keys to an `Hashtbl.t` is wrong. The hashing function from `Expr` should be used with `Hashtbl.Make(Expr)`.

src/lib/reasoners/uf.ml
bclement-ocp: Also, using `X.r` as a key to a `Hashtbl.t` is wrong. The `Shostak.HX` module should be used instead.

src/lib/reasoners/uf.ml
```diff
@@ -1134,35 +1134,58 @@ let terms env =
...
-    | None ->
+  type t = {
+    arrays : (Expr.t, unit) Hashtbl.t;
+    (** Set of arrays for which we need to generate a model value.
+        Notice that this set also contains embedded arrays. *)
+
+    selects: (X.r, (Expr.t, Expr.t) Hashtbl.t) Hashtbl.t;
+    (** Contains all the accesses to arrays. *)
+
+    abstracts: (r, Expr.t) Hashtbl.t;
```
bclement-ocp, line 1145: Ditto; this should use `Shostak.HX` instead of `Hashtbl.t`.

src/lib/reasoners/uf.ml
bclement-ocp: I also don't understand why this new table is useful. For any key `t` here I believe that the representative of `t` is always guaranteed to be in the `selects` table; can't we just use the `selects` table directly?

src/lib/reasoners/arrays_rel.ml
```diff
@@ -442,12 +469,28 @@ let count_splits env la =
...
   let are_neq = Uf.are_distinct uf in
   let class_of = Uf.class_of uf in
   let env = count_splits env la in
 
+  (* Update the set of representatives of arrays. *)
+  let env =
+    List.fold_left
+      (fun env l ->
+         match l with
+         | Xliteral.Eq (r1, r2), _, _, Th_util.Subst when is_array r2 ->
```
bclement-ocp, line 488: Note: ideally we would ignore `Sy.Set` arrays here but they can be UF representatives so I don't think we can :(

src/lib/reasoners/arrays_rel.ml
```diff
@@ -415,17 +419,40 @@ let new_equalities env eqs la class_of =
...
-
-let case_split env _uf ~for_model:_ =
+let two = Numbers.Q.from_int 3
+
+let split_arrays ~for_model env =
+  if for_model then
+    match SX.choose env.arrays with
+    | exception Not_found -> None
+    | a1 ->
+      match SX.find_first (fun a -> not @@ X.equal a a1) env.arrays with
```
bclement-ocp, line 429: This needs to take into consideration the types of arrays. We must not create an equality literal between arrays of different types. Probably the `arrays` table needs to be segregated by both index and value types.

bclement-ocp: Also, the current implementation loops trying to assert the same disequality. See for instance:

```smt-lib
(set-logic ALL)
(declare-const b (Array Int Int))
(declare-const c (Array Int Int))
(declare-const i Int)
(assert (= (select a i) (select b i)))
(assert (= (select a i) (select c i)))
(check-sat)
(get-model)
```

With `-d split` we can see that we loop repeatedly selecting `b <> a` as a split during model generation. `Uf.already_distinct` would need to be checked.

bclement-ocp (changes requested):
I understand the changes to the `arrays_rel` module to allow splitting on array equality. I am not sure I understand the changes to the `uf` module; at least I don't understand what the expected behavior change is.

src/lib/reasoners/arrays_rel.ml
Halbaroth: What happened??? I have never touch this piece of code!

src/lib/reasoners/uf.ml
Halbaroth: Honestly, both conventions are used in this codebase and other ones. I do not care about these kind of details.

Halbaroth: Uhm I guess you meant this is the convention in the stdlib for mutable data structures, indeed.

bclement-ocp: > Uhm I guess you meant this is the convention in the stdlib for mutable data structures, indeed.

Exactly this

src/lib/reasoners/uf.ml
Halbaroth: By `embedded arrays` I mean an array used in a payload of an ADT. Currently, we have a test `embedded-array.models.smt2` to ensure we produce correct models for these kind of expressions.

src/lib/reasoners/arrays_rel.ml
bclement-ocp: I'd wager on a misplaced C-a keystroke :wink:

Halbaroth: That's my guess too. Actually I do not use this feature, may be I should disable these keystrokes.

src/lib/reasoners/uf.ml
Halbaroth: I guess we can use `selects` table directly because we always add an array, even if there is no `select` term using it in our problem.

PR #1219, opened 2024-08-22
Remove some polymorphic hashtables

Halbaroth:
We should not use polymorphic equality on expressions because they may contain recursive terms producing by Dolmen. This PR removes these dangerous hashtables.

### DIFF
--- a/src/lib/reasoners/uf.ml
+++ b/src/lib/reasoners/uf.ml
@@ -1137,31 +1137,30 @@ let terms env =
 
 (* Helper functions used by the caches during the computation of the model. *)
 module Cache = struct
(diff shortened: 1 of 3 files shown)


bclement-ocp (approved):
> because they may contain recursive terms producing by Dolmen

This was already true before symbols could embed Dolmen terms, because `Expr.hash` is constant-time whereas `Hashtbl.hash` depends on the size of the expression.

Halbaroth:
The purpose of this PR is to ensure that hashing terms will always terminate. I did not try to optimize the model generation. I believe that we hadn't self recursive terms before using Dolmen's terms in Symbols but I might be wrong.

PR #1221, opened 2024-08-28
fix(BV, CP): Run cross-propagators to completion

bclement-ocp:
Since #1185, when substitutions change the domain of a variable, we no longer trigger propagations, which means that we can end up in a state where the bitlist and interval domains are not consistent (i.e. running [constrain_bitlist_from_interval] or [constrain_interval_from_bitlist] would shrink some domains).

This was initially part of #1185 but was accidentally removed as part of a simplification pass during review. Add it back.

### DIFF
--- a/src/lib/reasoners/domains.ml
+++ b/src/lib/reasoners/domains.ml
@@ -100,6 +100,11 @@ sig
 
   val watches : key -> t -> W.Set.t
   (** Returns the set of watches associated with object [x]. *)
+
+  val needs_propagation : t -> bool
+  (** Returns [true] if the domain needs propagation. This means that the domain
+      associated with some variables has been shrunk, and cross-domain
+      propagation might be needed. *)
 end
 =
 struct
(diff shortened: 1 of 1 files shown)


bclement-ocp:
CI failure seems unrelated:

```
#=== ERROR while installing mingw-w64-shims.0.2.0 =============================#
Cannot copy D:\a\alt-ergo\alt-ergo\_opam\.opam-switch\build\mingw-w64-shims.0.2.0\shim.exe to D:\a\alt-ergo\alt-ergo\_opam\bin\x86_64-w64-mingw32-gcc.exe (C:\hostedtoolcache\windows\opam\2.2.1\x86_64\opam.exe: "open" failed on D:\a\alt-ergo\alt-ergo\_opam\bin\x86_64-w64-mingw32-gcc.exe: Permission denied).
```

Halbaroth (approved):
Good catch.

Issue #1157, opened 2024-07-11
Native support for `distinct`


bclement-ocp:
After #890 we explode n-ary `distinct` constraints into a clique of pairwise disequalities. As mentioned in #889, we should implement n-ary `distinct` correctly at some point.

Currently this is tricky because while an n-ary `distinct` is a literal, its negation is not, and we need to figure out the best way to deal with this.

PR #1222, opened 2024-08-28
feat(BV): Support binary distinct on arbitrary bit-widths

bclement-ocp:
This used to be impossible to do in the general case when we have only bitlist domains, but is possible since we also have interval domains.

This implementation only supports binary distinct operators, and will need to be revisited as part of #1157.

Note: this currently raises an `Internal_error` with `--enable-assertions` due to cross-propagation not running to completion. #1221 fixes this error.

### DIFF
--- a/src/lib/reasoners/bitv_rel.ml
+++ b/src/lib/reasoners/bitv_rel.ml
@@ -34,6 +34,8 @@ let is_bv_ty = function
   | Ty.Tbitv _ -> true
   | _ -> false
 
+let is_bv_r r = is_bv_ty (X.type_info r)
+
 let bitwidth r =
   match X.type_info r with Tbitv n -> n | _ -> assert false
 
@@ -678,7 +680,7 @@ module Constraint : sig
   type fun_t =
     | Fbinop of binop * X.r * X.r
 
-  type binrel = Rule | Rugt
+  type binrel = Rule | Rugt | Rdistinct
 
   type rel_t =
     | Rbinrel of binrel * X.r * X.r
@@ -742,8 +744,13 @@ module Constraint : sig
   (** [bvshl r x y] is the constraint [r = x >> y] *)
 
   val bvule : X.r -> X.r -> t
+  (** [bvule x y] is the constraint [(bvule x y)]. *)
 
   val bvugt : X.r -> X.r -> t
+  (** [bvugt x y] is the constraint [(bvugt x y)]. *)
+
+  val distinct : X.r -> X.r -> t
+  (** [distinct x y] is the constraint [(distinct x y)]. *)
 end = struct
   type binop =
     (* Bitwise operations *)
@@ -818,11 +825,12 @@ end = struct
       Fbinop (op, y, x)
     | Fbinop _ as e -> e
 
-  type binrel = Rule | Rugt
+  type binrel = Rule | Rugt | Rdistinct
 
   let pp_binrel ppf = function
     | Rule -> Fmt.pf ppf "bvule"
     | Rugt -> Fmt.pf ppf "bvugt"
+    | Rdistinct -> Fmt.pf ppf "distinct"
 
   let equal_binrel : binrel -> binrel -> bool = Stdlib.(=)
 
@@ -923,6 +931,7 @@ end = struct
 
   let bvule = cbinrel Rule
   let bvugt = cbinrel Rugt
+  let distinct = cbinrel Rdistinct
 
   let equal c1 c2 = c1.tag = c2.tag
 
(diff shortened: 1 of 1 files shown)


bclement-ocp:
Rebased after the merge of #1221 CI should succeed now.

src/lib/reasoners/bitv_rel.ml
```diff
@@ -1489,6 +1522,13 @@ end = struct
       acts.acts_add_eq ~ex X.top X.bot;
       true
     | Rule | Rugt -> false
+    | Rdistinct when X.equal x y ->
+      acts.acts_add_eq ~ex X.top X.bot;
+      true
+    | Rdistinct ->
+      (* TODO(bclement): Is it is OK to simplify here if either side is
+         constant? *)
```
Halbaroth, line 1528: What simplifications did you have in your mind?

src/lib/reasoners/bitv_rel.ml
```diff
@@ -2229,13 +2269,33 @@ let case_split env uf ~for_model =
...
+        let bit_interval = Intervals.Int.extract int ~ofs:bitidx ~len:1 in
+        match Intervals.Int.value_opt bit_interval with
+        | Some (v, _) ->
+          if Options.get_enable_assertions () then
+            Util.internal_error
+              "High bit allowed by bitlists, but not intervals."
+          else
+            const 1 v, 1
+        | None -> const 1 Z.zero, 2
+      in
```
Halbaroth: I don't understand why you choose this test. I think that you check if the integer interval is tight enough to constraint the most significant unknown bit of `rr`. Clearly, it is a bug if our cross propagation is complete and you correct your split in this case, right?

Halbaroth (changes requested):

Halbaroth:
Did you run benchmarks to ensure there is slow down? If our implementation of cross-propagation is correct, we shouldn't raise an error but we could run this PR with `--enable-assertions` to check if we can catch such a mistake with this test.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: I mean remove the constraint entirely if we already know that the value is impossible.

src/lib/reasoners/bitv_rel.ml
bclement-ocp: If the most significant bit unknown bit in the bitlist domain is forced in the interval domain, it should have been forced in the bitlist domain by the cross-propagator (`constrain_bitlist_from_interval`).

If it has somehow *not* been propagated to the proper value from intervals to bitlist, and we pick the wrong value, we wlil loop in the case split. Let's say that the high bit is forced to `1` by the interval domain.

 - We split on `h = 0`; the theory propagates back the equality `h = 0`
 - We find a contradiction
 - Back-tracking: the theory propagates the disequality `h <> 0`
 - This is a disequality (distinct), so it is only added to the interval domain
 - Since the interval domain is not properly cross-propagated to the bitlist domain, we don't see that information and split on `h = 0` again.

Halbaroth: Ok, I think I can answer my own question. We cannot do better with lower unknown bit.

bclement-ocp (commented):
> Did you run benchmarks to ensure there is slow down?

Do you mean "to ensure there is no slow down"? Not on this version, but I believe I did with an earlier implementation of this PR. I don't think there is much risk tho.

> If our implementation of cross-propagation is correct, we shouldn't raise an error but we could run this PR with --enable-assertions to check if we can catch such a mistake with this test.

I believe that our implementation of cross-propagation is complete *currently*, but there might be corner cases where it is not, or it might no longer be in the future. If that happens, we want to know about it (so we want to have a way to detect it), but it's also fine and we don't want to bother our users with it: we are not doing anything incorrect, we are just emulating the little bit of cross-propagation that we need directly in the case split function in order to avoid infinite loops.

I'd be fine with a debug message instead of an error in `--enable-assertions` mode.

src/lib/reasoners/bitv_rel.ml
Halbaroth: And you think it could be unsound? I agree to postpone this improvement, I just want to understand what you wanted to do here.

Halbaroth (dismissed):

src/lib/reasoners/bitv_rel.ml
bclement-ocp: No, there is no soundness risk (there might be incompleteness though).

Currently, we don't have access to the domains during simplification (only to the semantic values), so we can't directly check whether the value is possible or not. However, simplification should only run on constraints that were already propagated, so if one of the argument is a constant, we should be guaranteed that it is fine to drop the constraint in this case. I am not 100% certain it is a guarantee (there could be a bug or an unexpected interaction somewhere), so I don't want to depend on this before thoroughly checking that it is OK to do so.

bclement-ocp:
Interesting, this seems to break `optim-4.models.smt2`? Looking into it

bclement-ocp:
> Interesting, this seems to break `optim-4.models.smt2`? Looking into it

This looks like a bug in the optimization procedure that was revealed by this change.

Halbaroth (approved):

PR #1223, opened 2024-08-28
Wrap the library Alt_ergo_prelude

Halbaroth:
Unwrapping this library was a big mistake, mea maxima culpa. When I added this prelude, I did know how to use `-open` with a wrapped library.

I couldn't compile smtml with `alt-ergo-lib` because the library of smtml depends on a module Prelude from another library.


src/lib/dune
```diff
@@ -7,7 +7,7 @@
 (library
   (name AltErgoLib)
   (public_name alt-ergo-lib)
-  (ocamlopt_flags (:standard -open Prelude))
+  (flags (:standard -open Alt_ergo_prelude__Prelude))
```
bclement-ocp: I think it would be better to keep `(wrapped false)`, rename `prelude.ml` to `alt_ergo_prelude.ml`, and use `-open Alt_ergo_prelude` here; this would avoid depending on the mangling performed by `(wrapped true)`.

Halbaroth: I have a better option: I kept `(wrapped true)` but I rename `prelude.ml` to `alt_ergo_prelude.ml`. In this case, I don't need to mangle the module in `(flags ...)`.

bclement-ocp: Does that work? I thought `(wrapped false)` was introduced because that was not working?

Halbaroth: It works. I introduced `wrapped false` because I didn't know how Dune mangles modules and I didn't know we can define a top level module with the same name of the library.

bclement-ocp: OK, then that's the best solution indeed.

bclement-ocp (approved):

PR #1225, opened 2024-08-29
fix(CP): Make sure domains do not overflow the default domain

bclement-ocp:
When reading domains through a non-canonical representative, we are intersecting it with the default domain of the representative (i.e. the range of the bit-vector type) in order to ensure that the resulting domain is known to be within the range of the type.

This is useful for interval domains because we keep track of global bounds, which we rely on in functions such as [bvshl], but are forgotten by the call to [add_explanation].

We also need to perform the same intersection when modifying a domain through a non-canonical representative, otherwise we might store a domain that overflows the bounds of the type.

(It is unfortunate that we have to do this dance instead of storing type information on the interval themselves, but that would be a bigger change).

This was found by fuzzing.


Halbaroth:
I was writing the issue :)

bclement-ocp:
I thought I sent a DM saying I was looking into it since it is my code, but apparently I did not — apologies.

src/lib/reasoners/domains.ml
```diff
@@ -530,8 +530,12 @@ struct
             D.intersect (D.unknown (X.type_info repr)) @@
             D.add_explanation ~ex (Entry.domain entry)
 
-        let set_domain { entry ; explanation = ex ; _ } d =
-          Entry.set_domain entry (D.add_explanation ~ex d)
+        let set_domain { repr ; entry ; explanation = ex } d =
+          if Explanation.is_empty ex then Entry.set_domain entry d
```
Halbaroth, line 534: Why you don't need to intersect the domain `d` if the explanation is empty?

bclement-ocp: We suppose that the domain is within the type bounds originally. The issue we are trying to fix is that `D.add_explanation` discards the global bounds (this is necessary for soundness, because the global bounds are not necessarily type bounds). If the explanation is empty, we don't need to call `add_explanation` (`D.add_explanation` is a no-op in this case), so we don't need to enforce the type bounds.

bclement-ocp: > We suppose that the domain is within the type bounds originally

Now that I think about it, I don't think this is documented, but it should.

bclement-ocp: Clarified the doc about this.

Halbaroth (approved):

PR #1234, opened 2024-09-07
fix: Make sure model generation is complete for more operators

bclement-ocp:
Some RIA operators are not complete even in the presence of the corresponding prelude. Add them as delayed functions to ensure we loop rather than generating an incorrect model.


Halbaroth (approved):
LGTM

PR #1246, opened 2024-10-03
Add js_of_ocaml debug flags

Halbaroth:
Currently, we compile the js worker without debug flags in our makefile.
I got an annoying bug while adapting the worker for Dolmen and without the
dev flags of `dune`, it was impossible to debug. I think it is better to
build it with the profile `dev` by default.

I also add more debugs flags at the toplevel of our worktree.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,20 +183,20 @@ html: doc
 # note that --timeout option is ignored due to the lack of js primitives
 # and the use of input zip file is also unavailable
 js-node:
-	$(DUNE) build $(DUNE_FLAGS) --profile=release $(BJS_DIR)/main_text_js.bc.js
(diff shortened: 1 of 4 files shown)


Halbaroth:
Uhm, it seems that the node version does not compile correctly with the `dev` profile. This issue is already present on `next`:
```
dune build src/bin/js/main_text_js.bc.js --profile release 
node _build/default/src/bin/js/main_text_js.bc.js tests/everything/f1.ae 
File "tests/everything/f1.ae", line 159, characters 1-166: Valid (0.3200) (240 steps) (goal WP_create)
```
but 
``` 
dune build src/bin/js/main_text_js.bc.js --profile dev 
node _build/default/src/bin/js/main_text_js.bc.js tests/everything/f1.ae 
Fatal error: exception Failure("caml_register_global: cannot locate Dynlink_compilerlibs.Binutils")
```

PR #1248, opened 2024-10-03
Disable zip feature in Bytecode mode

Halbaroth:
As Basile noticed, the Javascript code in `My_zip` is dead too. I
removed it and clean a bit the module.

I also produces an appropriate `run error` if we try to use the feature
in Javascript because `camlzip` does not work properly in bytecode.

Currently, the error cannot be trigger because we do not use yet the
Dolmen version of `Solving_loop` in the javascript worker. This will
change in my next PR.


tests/misc/unzip.ae.expected
```diff
@@ -1,2 +1,2 @@
 
-unknown
+unsat
```
Halbaroth, line 2: This change is normal. I changed the test in order to use the same one for both native and SMT language.

src/lib/util/my_zip.ml
```diff
@@ -25,54 +25,31 @@
...
-   let open_in  _  =  failwith s
-   let close_in _ = failwith s
-   let entries  _  =  failwith s
-   let read_entry  _ _  =  failwith s
-   let filename  _  =  failwith s
-   let is_directory  _  =  failwith s
-   end
+let extract_zip_file f =
+  if Stdlib.(Sys.backend_type = Bytecode) then
+    Errors.unsupported_feature "Zip format in javascript mode";
```
bclement-ocp, line 41: I believe this is disabling zip files for non-jsoo bytecode but not jsoo (which I think has [`Other "js_of_ocaml"`](https://github.com/ocsigen/js_of_ocaml/blob/ef42a5635907a58de90c6752f724c1934ef8c6bf/runtime/sys.js#L278) as a backend type). Although I think this test should be removed entirely -- see toplevel comment.

bclement-ocp (commented):
The reason for the difference between the `unzip.ae.zip` and `unzip.smt2.zip` is this:

```console
$ file tests/misc/unzip.ae.zip
tests/misc/unzip.ae.zip: Zip archive data, at least v1.0 to extract, compression method=store

$ file tests/misc/unzip.smt2.zip
tests/misc/unzip.smt2.zip: Zip archive data, at least v2.0 to extract, compression method=deflate
```

The `unzip.ae.zip` file is not compressed (`method=store`) while the `unzip.smt2.zip` file is compressed with zlib (`method=deflate`). Since zlib is not available with `js_of_ocaml`, only uncompressed zip files are supported in `js_of_ocaml`.

I am fine keeping the check (with the proper test, see inline comment) but I think we should remove the check for `js_of_ocaml` when reading zip files entirely -- if the zip file is uncompressed, it will just work; if it is compressed, we will fail due to a missing primitive anyway (it's not the nicest way to fail, but we don't expect people to actually use the jsoo version as a binary -- only through the worker).

PR #1249, opened 2024-10-07
Use Dolmen in the JS worker

Halbaroth:
This PR removes the dependency of the Javascript worker to the legacy
frontend in order to remove it completely. In order to avoid code
duplication as it was done with the legacy frontend, we reuse the
`Solving_loop` module in the worker. This requires some modifications:

- Add a new argument to the main function of `Solving_loop` to give
  different kind of sources instead of reading the file at the path
  `Options.get_file ()`. The worker used a raw string as input. I reused
  the type `Dolmen_loop.State.source`.

- Add a new argument to the main function of `Solving_loop` to pass the
  selector hook to the constructor of the environment of the SAT solver.
  This hook is not used by the worker to select instantiations but just to
  record them. I have no strong opinion to preserve this feature but I
  try to keep all the feature of the Javascript worker.

- Remove the unused `sat_env` argument of the constructor `init_env` in
  `Frontend`. This removal is mostly motivated by the fact the semantic
  of `init_env = Some env` + `selector_inst = Some f` is not clear at
  all.

- Monomophize the `status` type in `Frontend`. This status is only used
  for printting purposes but we kept the SAT environment in some of its
  constructors. Unfortunately, the type of the SAT can change at the
  very beginning of the main function of `Solving_loop`, which means it
  is not easy to add a new argument for the `hook_status`. This hook is
  required to print the result of AE in the worker.


src/bin/common/solving_loop.ml
```diff
@@ -429,21 +431,54 @@ let main () =
...
+      | Some (Dl.Logic.Smtlib2 _) | None -> true
+      | _ -> false
+    in
+    let src =
+      match src with
+      | `Stdin when is_incremental ->
+        set_output_format ".smt2";
+        `Stdin
+      | `Stdin ->
+        `Raw (filename, Compat.In_channel.input_all stdin)
```
bclement-ocp: I believe this is no longer necessary since Gbury/dolmen#201 and can be replaced with:
```suggestion
      | `Stdin -> `Stdin
```

(And the definition for `is_incremental` removed)

src/bin/common/solving_loop.ml
```diff
@@ -429,21 +431,54 @@ let main () =
...
+      | `File path when Filename.check_suffix path ".zip" ->
+        Filename.(chop_extension path |> extension) |> set_output_format;
+        let content = AltErgoLib.My_zip.extract_zip_file path in
+        `Raw (Filename.chop_extension filename, content)
+      | `File path ->
+        Filename.extension path |> set_output_format;
+        let cin = open_in path in
+        let content = Compat.In_channel.input_all cin in
+        close_in cin;
+        `Raw (filename, content)
```
bclement-ocp: Note to self: not sure why we had this, I think this can just be `` `File path`` (provided that we call `set_output_format` first)?

src/bin/js/worker_js.ml
```diff
@@ -259,11 +146,12 @@ let main worker_id content =
       Worker_interface.diagnostic =
         Some [Format.asprintf "%a" Errors.report e]
     }
-  | _ ->
+  | exn ->
     let res = Worker_interface.init_results () in
+    let msg = Fmt.str "Unknown error: %s" (Printexc.to_string exn) in
```
bclement-ocp, line 151: :heart:

src/lib/frontend/frontend.ml
```diff
@@ -39,11 +39,11 @@ let unused_context name context =
...
   | Some s -> not (Util.SS.mem name s)
 
-type 'a status =
+type status =
   | Unsat of Commands.sat_tdecl * Ex.t
   | Inconsistent of Commands.sat_tdecl
-  | Sat of Commands.sat_tdecl * 'a
-  | Unknown of Commands.sat_tdecl * 'a
+  | Sat of Commands.sat_tdecl
+  | Unknown of Commands.sat_tdecl
```
bclement-ocp, line 46: I am slightly surprised that we were never using these environments, but if that is the case, there is no point to keep them anyways.

bclement-ocp (approved):
Overall looks good to me --- I think we can simplify handling of input files with gbury/dolmen#201 and maybe get rid of `Compat.In_channel.input_all` entirely. If you have time to look into it it'd be great, but it's not important and I'm OK merging as is as well.

src/lib/util/compat.ml
```diff
@@ -153,3 +153,99 @@ module Type = struct
       match A.Id with B.Id -> Some Equal | _ -> None
   end
 end
+
+module In_channel = struct
+  open In_channel
```
bclement-ocp: Ah yes unfortunately that one is not going to work since the module itself was added in 4.14; bummer.

Halbaroth:
> Overall looks good to me --- I think we can simplify handling of input files with [Gbury/dolmen#201](https://github.com/Gbury/dolmen/pull/201) and maybe get rid of `Compat.In_channel.input_all` entirely. If you have time to look into it it'd be great, but it's not important and I'm OK merging as is as well.

Unfortunately, we still need `input_all` for the zip format. We may create a temporary file and read it instead of loading all the file in a string.

src/lib/util/compat.ml
Halbaroth: outch!

src/lib/frontend/frontend.ml
Halbaroth: Me too. Our API is such an onion with too many layers, we didn't notice it before.

src/bin/js/worker_js.ml
Halbaroth: Actually, this is not sufficient to catch some errors because the worker is running into Lwt monad and the backtrace is sometimes lost.

bclement-ocp:
> > Overall looks good to me --- I think we can simplify handling of input files with [Gbury/dolmen#201](https://github.com/Gbury/dolmen/pull/201) and maybe get rid of `Compat.In_channel.input_all` entirely. If you have time to look into it it'd be great, but it's not important and I'm OK merging as is as well.
> 
> Unfortunately, we still need `input_all` for the zip format. We may create a temporary file and read it instead of loading all the file in a string.

I think for the zip format we use `My_zip.extract_zip_file` which does not use `input_all`? We are using `input_all` for the case of non-zip file, but I believe that passing `` `File path`` to Dolmen instead of reading the file ourselves should work.

src/bin/js/worker_js.ml
bclement-ocp: It is still better for the ones we do catch — I'll take an improvement over what we had before even if it is not perfect!

Issue #898, opened 2023-10-18
Optimization for BV theory


Halbaroth:

Halbaroth:
Duplicate of #884

PR #1251, opened 2024-10-07
Get rid of the legacy frontend finally!

Halbaroth:
This PR removes:
- the deprecated legacy frontend (parser and typechecker),
- the deprecated AB-Why3 plugin which depends on the legacy parser,
- the associated opam packages and dependencies (including `shell.nix`
  and our Makefile),
- all the reference in the documentation to these features,
- simplify `gentest` and rename the tag `dolmen` and `models` into `default` in `tests/` and
  remove it when it is useless,
- the option `frontend` in `Options`.

We keep the `--frontend` option but now it is a no op command and we
output an appropriate message if users still use it.


Halbaroth:
I rollbacked modifications I have done in the `tests` directory and `gentest` script. I merged `models` and `dolmen` into an new `default` tag. 

Notice that the +2662 comes from the fact I rename `D_cnf` into `Cnf`. There is absolutely no new line in this file but Github didn't manage to produce a proper diff. This happens also for some test files.

src/bin/common/parse_command.ml
```diff
@@ -1188,6 +1185,8 @@ let parse_output_opt =
     Term.(const Output.set_dump_models $ dump_models_on)
   in
 
+  let set_frontend _ = () in
+
   let set_frontend =
     Term.(const set_frontend $ frontend)
```
```suggestion
  let set_frontend =
    Term.(const ignore $ frontend)
```

removes one level of indirection for the reader

src/lib/reasoners/theory.ml
```diff
@@ -619,7 +619,7 @@ module Main_Default : S = struct
     | Unknown ->
       (* In the current implementation of optimization, the function
          [CC_X.optimizing_objective] cannot fail to optimize the objective
-         function [obj]. First of all, the legacy parser only accepts
+         function [obj]. First of all, the parser only accepts
```
bclement-ocp: I believe this is false given that #898 works. I think it's fine to say that `CC_X.optimizing_objective` is required to actually optimize the objective function as an internal invariant and raise an internal error here if it fails.

tests/models/bitv/bvxor-1.smt2
```diff
@@ -0,0 +1,7 @@
+(set-logic ALL)
+(set-option :produce-models true)
+(declare-const x (_ BitVec 8))
+(declare-const y (_ BitVec 8))
+(assert (= x (bvxor #b10100101 y)))
+(check-sat)
+(get-model)
```
bclement-ocp: I think this is duplication with `bvxor-1.default.smt2` (same for many others); is it intended?

bclement-ocp (commented):
Feels good to finally see this go! 

> Notice that the +2662 comes from the fact I rename D_cnf into Cnf. There is absolutely no new line in this file but Github didn't manage to produce a proper diff. This happens also for some test files.

Hm that is unfortunate. I think this is because the `Cnf` module already exists before this PR. It's not only a (minor) issue for viewing the PR but I think it will also mess up git facilities for tracking a specific line across renames, I'd prefer if we could work around it.

Anyways this module should not be called `Cnf` because it doesn't actually perform anything related to CNF-conversion (and the previous `Cnf` module did not either). What about calling it `Convert` or `Translate` instead? This would make more sense and also solve the +2k issue.

tests/models/bitv/bvxor-1.smt2
Halbaroth: Nope, I duplicate some tests by accident as you can see with my last commit. I removed it.

src/lib/reasoners/theory.ml
Halbaroth: Indeed, this documentation is outdated. I will update it now. Notice that the point of this comment was not to explain we have not implemented the feature for some theory but our implementation of `optimizing_objective` for the `Int` and `Real` theories will always produce an answer even if this answer is not correct. Calling `optimizing_objective` on a split involving unsupported theory will produce `None`:
- `None` means that no theory knows how to optimize this value,
- `Unknown` means the split has no already be optimized.

I will clarify this point in the documentation of `Rel.optimizing_objective`.

tests/models/adt/small.smt2
```diff
@@ -0,0 +1,11 @@
...
+(set-logic ALL)
+(declare-datatypes ((data1 0) (data2 0)) (
+  ((cons1 (d1 data1)) (cons2 (d2 data2)))
+  ((cons3 (d3 data1)) (cons4 (d4 Int) (d5 Int) (d6 Bool))
+                      (cons5 (d7 Bool)))
+))
+(declare-const a data1)
+(assert ((_ is cons1) a))
+(check-sat)
+(get-model)
```
bclement-ocp: Looks like there are still duplicates (with `small.default.smt2`)?

bclement-ocp (commented):
Looks like there are still some duplicated tests — everything else is OK!

bclement-ocp (commented):
I hate to be a bother but there are still duplicate tests :sweat_smile: 

I think the following should solve it:

```console
$ find tests -name '*.default.smt2' -exec sh -c 'rm -f ${0%.default.smt2}.smt2' {} \;
$ find tests -name '*.default.ae' -exec sh -c 'rm -f ${0%.default.ae}.ae' {} \;
$ find tests -name '*.default.expected' -exec sh -c 'rm -f ${0%.default.expected}.expected' {} \;
```

Halbaroth:
Done!

bclement-ocp (dismissed):
And with that the net difference is over 10k lines removed :tada: 

(Although it was already over 9000)

tests/models/issues/719.err.smt2
```diff
@@ -0,0 +1,19 @@
...
+    (=> (and (<= 0 p) (< p q) (<= q (- s 1))) (<= (select a p) (select a q)))))
+(assert (
+  forall ((p Int) (q Int))
+    (=> (and (<= (+ s 1) p) (< p q) (<= q 10)) (<= (select a p) (select a q)))))
+(assert (not (
+  forall ((p Int) (q Int))
+    (=> (and (<= 0 p) (< p q) (<= q 10)) (<= (select a p) (select a q))))))
+(check-sat)
+(get-model)
+; (get-model) should fail because the problem is UNSAT.
```
bclement-ocp: I think this one is the last duplicate actually which was not caught by the script because the new file is called `.default.err` (and not `err.default`)!

Halbaroth: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAH

bclement-ocp (approved):
Approving conditionally to the removal of 719.err ; you can click the big green button once that one is removed.

PR #1255, opened 2024-10-09
Better `is_enum_constr` test in Adt_rel

Halbaroth:
As we do not use `Hstring.t` to identify ADT constructors, we can use Dolmen types of constructors to determine their arity.

### DIFF
--- a/src/lib/reasoners/adt_rel.ml
+++ b/src/lib/reasoners/adt_rel.ml
@@ -169,19 +169,15 @@ module Domains = struct
 
   let filter_ty = is_adt_ty
 
-  let is_enum_constr DE.{ id_ty; _ } =
-    let monomorphic_case id_ty =
-      match DE.Ty.view id_ty with
-      | `Arrow (l, _) -> Compat.List.is_empty l
-      | `App _ -> true
-      | _ ->
-        (* This case cannot occur as [id_ty] denotes the type of a monomorphic
-           ADT constructor. *)
-        assert false
-    in
-    match DE.Ty.view id_ty with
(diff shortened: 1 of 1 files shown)


bclement-ocp (changes requested):
I don't think looking at the type of the constructor is the right approach here. It has implicit assumptions (for instance it assumes that the argument is an ADT constructor, which is only documented by an internal comment in the function) and it is not immediately obvious to me why the check does what it says on the tin.

I'd much prefer keeping the existing approach of checking the type definition, except that we can now do it directly from the constructor and also the check is in constant time, because Dolmen uses an array rather than a list:

```ocaml
let is_enum_constr DE.{ builtin; _ } =
  match builtin with
  | B.Constructor { adt ; case } ->
    begin match DE.Ty.definition adt with
    | Some Adt { cases; _ } ->
        Array.length cases.(case).dstrs = 0
    | _ -> (* assert ? *) false
    end
  | _ -> false
```

bclement-ocp (approved):

Halbaroth:
I put an assertion instead returning `false` in both cases. I have no strong opinion on this point but it is clearly an internal invariant and the name of the function makes clear it expects an ADT constructor as argument. I add a comment.

bclement-ocp:
I suggested returning `false` for a non-`Constructor` because it reduces the scope of the invariant, which is generally good for complexity/comprehension, but I'm OK with either.

Halbaroth:
I keep your version because we check that `term_cst` in constructors of Alt-Ergo are Dolmen constructors in the smart constructor `mk_constr` of `Expr`. So we do not need to test it later.

PR #1256, opened 2024-10-09
Use an exception to leave `Solving_loop.process_source`.

### LINKED ISSUE #1250: Supporting `exit on error` in Javascript mode
We use the system call `exit` in `Solving_loop` in order to halt quickly Alt-Ergo in case of fatal error or recoverable error if we turn on the flag `exit_on_error`. Unfortunately, this system call is not available in the Javascript runtime. 

This issue can be tempered to some extent by turn off the flag `exit_on_error` in JS mode as it is done in #1249 but `exit` is still called for fatal errors. 

I am not sure what is the best solution here. We may raise an exception in the main function of `Solving_loop` instead of calling `exit` and catch it immediately in both `Main_text` an `Main_text_js`. In the former, we call `exit` and in the latter we print a message and restart the worker.

Halbaroth:
This function cannot call the system call `exit` because this system call is not supported in browsers (it works in Node).

This commit uses a new exception `Exit_on_error` and catch it at the toplevel of binaries.

Solve issue #1250 

I didn't create a new error in `Errors` because these errors are caught in `Solving_loop` to raise `Exit_on_error`.

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -32,7 +32,7 @@ module DO = D_state_option
 module Sy = Symbols
 module O = Options
 
-exception Exit_on_error of int
+exception Exit_with_code of int
 
 type solver_ctx = {
   ctx    : Commands.sat_tdecl list;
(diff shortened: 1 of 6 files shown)


src/bin/common/solving_loop.ml
```diff
@@ -32,6 +32,8 @@ module DO = D_state_option
 module Sy = Symbols
 module O = Options
 
+exception Exit_on_error of int
```
bclement-ocp: This is also used to exit with a return code of `0`, which is not an error — I'd suggest renaming it to `Exit_with_code`.

bclement-ocp (approved):

PR #1258, opened 2024-10-09
Remove the Typed module

Halbaroth:
This PR removes the output AST of the legacy typechecker. This AST was still used by the compilation of match expressions in `Expr`.


src/lib/frontend/translate.ml
```diff
@@ -761,55 +761,139 @@ let handle_patt_var id (DE.{ term_descr; _ } as term)  =
...
+               p_list_vars args
+               print v;
+           | Typed.Var x ->
+             Printer.print_dbg  ~flushed:false ~header:false
+               "| %a -> %a@ " Var.print x print v;
+        )cases;
+      Printer.print_dbg ~header:false
+        "end@ result is: %a" print res;
+     end *)
+
```
```suggestion
```

src/lib/frontend/translate.ml
```diff
@@ -761,55 +761,139 @@ let handle_patt_var id (DE.{ term_descr; _ } as term)  =
...
-              | Some dstr -> dstr :: acc
-              | _ -> assert false
-          ) dstrs []
-        | _ ->
-          Fmt.failwith
-            "Expected a constructor for an algebraic data type but got\
-             something else for the definition of: %a"
-            DE.Ty.Const.print adt
-      end
+module Match = struct
```
bclement-ocp: Nit: if using local modules for actual implementation work there is an abstraction boundary that should be ideally documented and at least typed.

```suggestion
module Match : sig
  type pat
  val mk_pat : DE.term -> pat
  val make : Expr.t -> (pat * Expr.t) list -> Expr.t
end = struct
```

bclement-ocp (approved):
Apologies, I thought I had already reviewed this one before my vacation but looks like I forgot to finish the review.

Overall looks good — I think the `Expr` module was a good boundary for the `mk_match` abstraction but I'm OK as is now that we have a single frontend.

Halbaroth:
> Overall looks good — I think the `Expr` module was a good boundary for the `mk_match` abstraction but I'm OK as is now that we have a single frontend.

I had the same concerns when I have done this modification. `Expr` is an internal AST of the library and Dolmen expressions will become the new input AST of the library. Moving this feature from `Expr` to `Translate` means we cannot use it directly on AE expressions as library developers, but we do not use it. Actually we could use it in models because it would be simpler to produce a match expression for model terms then compiles it. My opinion is that we do not need to focus on this detail now and we can move back a part of the code in `Expr` later if we want to produce match expressions from AE expressions. Keeping all the match compilation code in `Translate` now is just simpler in the current state of the code base.

bclement-ocp:
> I had the same concerns when I have done this modification. [..] My opinion is that we do not need to focus on this detail now and we can move back a part of the code in Expr later

This argument could have been made to avoid moving the code out of `Expr` in the first place ;)

PR #1263, opened 2024-10-25
Simplify type variables

Halbaroth:
This PR changes three thing:
1. We do not need a value field in the type variable of Alt-Ergo. These values were only used by the unification of the 
types in the legacy typechecker. 
2. Add a proper type of the substitution of types in `Ty`. I add two different functions in the new module `Subst` to update a substitution. The function `Subst.update` updates a substitution without checking if it is compatible with a previous bind in the map. The function `Subst.try_bind` updates a substitution too but raises `TypeClash` if a previous bind in the map is incompatible with the new value.
4. Use directly type variables from Dolmen. Notice that I didn't remove some caches in `Translate` for type variables but I think we can simplify this part too. I prefer doing it in a separate PR.


src/lib/structures/ty.ml
```diff
@@ -311,18 +261,64 @@ let rec equal t1 t2 =
...
+  let is_id sbt = TvMap.is_empty sbt
+
+  let eval sbt tv =
+    match TvMap.find tv sbt with
+    | ty -> ty
+    | exception Not_found -> Tvar tv
+
+  let update tv ty sbt =
+    match ty with
+    | Tvar tv' when DE.Ty.Var.equal tv tv' -> sbt
```
bclement-ocp, line 278: This should be an error, no? `x → x` is never a valid substitution?

src/lib/structures/ty.mli
```diff
@@ -193,24 +180,62 @@ val trecord :
...
+  (** [try_bind tv ty sbt] tries to bind [tv] with [ty]. The function
+      succeeds if [tv] is not in the domain of [sbt] or the current
+      value of [tv] in [sbt] is equal to [ty].
+
+      If the current value of [tv] is not compatible with [ty], raises
+      the exception {!exception TypeClash}.
+
+      If the previous value of [tv] in [sbt] is equal to [ty] for [equal],
+      the returned substitution is physically equal to [sbt]. *)
+
```
bclement-ocp, line 218: This could be made much simpler by simply exposing a function `find` that raises `Not_found` if the variable is not in the substitution.

bclement-ocp (changes requested):
I started the review but I think that while I agree with points 1 and 3 in the description (removing mutability, using Dolmen variables), I don't think point 2 is a good idea.

Substitutions are well understood as maps from variables to terms and I think we should just use that; wrapping that in a module with a hard-to-understand interface (`try_bind`) seems to only bring unneeded complexity.

Can you update the PR to remove the `Subst` module and I'll finish my review (I think most of the comments were going to be about that module anyway)?

src/lib/structures/ty.ml
Halbaroth: If we consider a substitution as a total map from type variables into types, this is the default binding.

Halbaroth:
> I started the review but I think that while I agree with points 1 and 3 in the description (removing mutability, using Dolmen variables), I don't think point 2 is a good idea.
> 
> Substitutions are well understood as maps from variables to terms and I think we should just use that; wrapping that in a module with a hard-to-understand interface (`try_bind`) seems to only bring unneeded complexity.
> 
> Can you update the PR to remove the `Subst` module and I'll finish my review (I think most of the comments were going to be about that module anyway)?

During the last dev meeting, we agreed that this modification was not appropriate. As often, the mathematical way to present a data structure is not the best way to implement it. It is common to consider a substitution as a total map from as set `V` to a superset `E` (here `V = variables` and `E = expressions`). After pushing this PR, I notices that this modification cannot be done for expression substitutions too because currently AE term variables are not typed, so we have no canonical injection between term variables and terms.

src/lib/structures/ty.mli
Halbaroth: Actually, we do not need to expose this function because it is only used in the matching function of `Ty`. So we can use directly the find function.

src/lib/structures/ty.ml
bclement-ocp: Ah yes sorry, I mixed things up with rewrite rules.

bclement-ocp: (Although given the way we use substitutions I think that they are intended to be destructive so that the binding `x → x` would always be incorrect, but that is a meta-property)

Halbaroth:
Close in favor of #1267

PR #1265, opened 2024-11-05
Remove deprecated cli options and preludes

Halbaroth:
This PR removes some deprecated features in `v2.6.0`.


bclement-ocp (changes requested):
There does not seem to be any way to set the `debug_warnings` flag anymore; it should be removed.

Halbaroth:
Oups I forgot it :)
It is done.

bclement-ocp (approved):

PR #1266, opened 2024-11-07
Remove labels on Symbols and Xliteral

Halbaroth:
This feature is unused.

Also clean up the printer of Xliteral.


bclement-ocp (approved):
Did you check the new spacings introduced in the Xliteral printer on a few problems? We had issues in the past where such changes made logs utterly unreadable.

Halbaroth:
I checked, and it should be fine now. As usual, `@ ` and `@,` must always be used inside boxes. `Log.debug` wraps the formatter string into a box.

bclement-ocp (approved):

PR #1267, opened 2024-11-12
Use Dolmen type variables

Halbaroth:
This PR is open to replace #1263: 
- Remove the useless mutable field in type variables. 
- Use directly the Dolmen type variables from Dolmen.


src/lib/structures/expr.ml
```diff
@@ -335,7 +335,7 @@ module SmtPrinter = struct
 
   (* This printer follows the convention used to print
      type variables in the module [Ty]. *)
-  let pp_tyvar ppf v = Fmt.pf ppf "A%d" v
+  let pp_tvar ppf v = Fmt.pf ppf "A%a" DE.Ty.Var.print v
```
```suggestion
  let pp_tvar ppf v = Fmt.pf ppf "%a" DE.Ty.Var.print v
```

`"A"` is already added to the name of the variable in `fresh_tvar`, and there is no need to add a prefix to variables from the problem (once we use them).

bclement-ocp (approved):
OK for me to merge with the requested change.

src/lib/structures/expr.ml
Halbaroth: Indeed, it should work because it seems that we always call `Ty.fresh_tvar()` in `Translate` even if we do not need to call it on some type variables emitted by Dolmen. Otherwise, we could see type variables without this prefix. I remove the printer.

bclement-ocp (approved):

PR #1268, opened 2024-11-13
Documentation of Arrays_rel

Halbaroth:
While writing this documentation, I read the paper `Generalized, Efficient Array
Decision Procedures` of de Moura and Bjorner.

Before writing this documentation, I believed that we only implement the basic
inferences rules, which is exposed in Fig 1 of the paper. These rules are:

1. (idx)    `a == store(b, i, v) ===> a[i] = v`
2. (down)  `a == store(b, i, v), w == a'[j], a ~ a' ===> i = j \/ a[j] = b[j]`
3. (up)     `a == store(b, i, v), w = b'[j], b ~ b' ===> i = j \/ a[j] = b[j]`
4. (ext)   `a, b know ===> a = b \/ a[k] <> b[k] (k fresh name)`

(I use the double equality symbol `==` to denote syntactical equality and `~` for the
congruence relation.)

In fact, we implement a slightly different version of these rules. The `down`
, `up` and `ext` inference rules are implemented by respectively `get_of_set`,
`get_and_set` and `extensionality` functions in `Arrays_rel` module.

The `idx` axiom is not implemented by a dedicated function. Instead, different
instantiations of this axiom are produced by `get_of_set`, `get_and_set` and
`get_from_set`:

1a. (idx a)  `a[i] known, a ~ store(b, j, v) ==> i <> j \/ a[i] = v`
1b. (idx b)  `store(b, i, v), store(a, j, w) known, a ~ b ==> store(a, j, w)[j] = w`.

Technically, our implementation does not send directly instantiated axioms to the
SAT solver. Consider for instance the axiom `down` and assume that all its hypotheses
are satisfied. We produce the deduction `i = j \/ a[j] = b[j]` but we do not send it
to the SAT solver. Instead, we add the equality `i = j` in the set `env.split` in
order to split on its negation later. If we split on it or we see `i <> j` in `assume`,
we send `a[j] = b[j]` to the SAT solver.

Our implementation suffers from a number of issues:
1. We produce a lot of redundant axioms (see section A and section B of the paper).
For instance, we do not need to propagate `a[i] = v` if we already know that
`a'[i'] = v'` for with `a ~ a'`, `i ~ i'` and `v ~ v'`. The same goes for other axioms.
2. We store all the relevant terms in huge maps which is not efficient.
3. We transform literal expressions into literal semantic values.

To be more efficient, our implementation only should follow equivalent classes
of read/write terms and relevant indices. When some of these classes merge, we
may have to instantiate axioms. This new version could be implemented using a union-find
in `Arrays_rel` itself but as I explained during the last dev meeting, this modification
will conflict with our plan for the union-find in UF (using Basile's store).

### DIFF
--- a/src/lib/reasoners/arrays_rel.ml
+++ b/src/lib/reasoners/arrays_rel.ml
@@ -35,6 +35,14 @@ module Ex = Explanation
 
 module LR = Uf.LX
 
+(* In this module, we follow the below conventions in the comments:
+   - A get term refers to a term of the form `(select a i)` for an array `a` and
+     an index `i`.
+   - A set term refers to a term of the form `(store a i v)` for an array `a`
+     and an index `i`.
+   - A relevant index is any term that appears as an index in get or set terms.
+   - A get or set term is known if it has been encountered in [assume]. *)
+
 let src = Logs.Src.create ~doc:"Arrays_rel" __MODULE__
 module Log = (val Logs.src_log src : Logs.LOG)
 
@@ -56,30 +64,25 @@ module Conseq =
       let compare (lt1,_) (lt2,_) = E.compare lt1 lt2
     end)
 
-(* map k |-> {sem Atom} d'egalites/disegalites sur des atomes semantiques*)
 module LRmap = struct
   include LR.Map
   let find k mp = try find k mp with Not_found -> Conseq.empty
   let add k v ex mp = add k (Conseq.add (v,ex) (find k mp)) mp
 end
 
-(* Set of reads *)
+(* Set of get terms. *)
 type gtype = {g:Expr.t; gt:Expr.t; gi:Expr.t}
 module G = Set.Make
     (struct type t = gtype let compare t1 t2 = E.compare t1.g t2.g end)
 
-(* Set of writes *)
+(* Set of set terms. *)
 type stype = {s:E.t; st:E.t; si:E.t; sv:E.t}
 module S = Set.Make
     (struct type t = stype let compare t1 t2 = E.compare t1.s t2.s end)
 
-(* map t |-> {set(t,-,-)} qui associe a chaque tableau l'ensemble
-   de ses affectations *)
 module TBS = struct
   include Map.Make(E)
   let find k mp = try find k mp with Not_found -> S.empty
-
-  (* add reutilise find ci-dessus *)
   let add k v mp = add k (S.add v (find k mp)) mp
 end
 
@@ -87,22 +90,16 @@ let timer = Timers.M_Arrays
 
 module H = Ephemeron.K1.Make (Expr)
 
-(* A read is a term of the form `(select a i)` for some array `a` and index `i`
-   and a write is a term of the form `(store a i v)`.
-
-   A relevant index is a term that appears as an index in read/write terms.
-
-   We say that a read/write is known if we have seen this term in [assume]. *)
 type t = {
   gets : G.t;
-  (* Set of all the known reads. *)
+  (* Set of all the known get terms. *)
 
   tbset : S.t TBS.t;
-  (* Set of all the known writes. There are indexed by array. *)
+  (* Set of all the known set terms, indexed by array. *)
 
   split : LRset.t;
   (* Set of equalities or disequalities on relevant indices.
-     There are the hypotheses of consequences in the field [conseq].
+     These are the hypotheses of consequences in the field [conseq].
 
      We split on these literals in [case_split].
 
@@ -123,16 +120,12 @@ type t = {
      in [get_of_set]. *)
 
   new_terms : E.Set.t;
-  (* Set of read and write terms produced by the theory. These terms
+  (* Set of get and set terms produced by the theory. These terms
      are supposed to be sent to the instantiation engine. *)
-
-  (* size_splits : Numbers.Q.t; *)
-  (* Limit the number of case splits performed on indices. This field
-     is currently unused. *)
 
   cached_relevant_terms : (G.t * S.t TBS.t) H.t;
   (* Weak cache used to accelerate the exploration of new terms in order to
-     find new read or write terms. *)
+     find new get or set terms. *)
 }
 
 
@@ -216,7 +209,8 @@ and cached_relevant_terms env t =
     H.add env.cached_relevant_terms t r;
     r
 
-(* Search all the reads and writes as subterms of literals [la]. *)
+(* Search for all the set and get terms within the subterms of literals in the
+   list [la]. *)
 let new_terms env la =
   let fct acc r =
     List.fold_left
@@ -265,11 +259,11 @@ let update_env (module Uf : UF) uf dep env acc gi si p p_ded n n_ded =
        `gi <> si` at the same time. *)
     assert false
 
-(* Produce a formula to propagate a read over a store.
+(* Produce a formula to propagate a get term over a set term.
 
-   More precisely, assume that [gtype] represents the read `(select a i)`.
-   For all writes `(store b j v)` in the equivalence class of `a`, we form the
-   formulas:
+   More precisely, assume that [gtype] represents the get term `(select a i)`.
+   For all set terms `(store b j v)` in the equivalence class of `a`, we create
+   the formulas:
 
      i <> j -> (select a i) = (select b i)
 
@@ -277,7 +271,7 @@ let update_env (module Uf : UF) uf dep env acc gi si p p_ded n n_ded =
 
      i = j -> (select a i) = v.
 
-   Moreover, the read `(select b i)` is sent to the instantiation engine. *)
+   Moreover, the get term `(select b i)` is sent to the instantiation engine. *)
 let get_of_set (module Uf : UF) uf gtype (env, acc) =
   let {g=get; gt=gtab; gi=gi} = gtype in
   E.Set.fold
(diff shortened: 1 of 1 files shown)


src/lib/reasoners/arrays_rel.ml
```diff
@@ -55,20 +55,22 @@ module Conseq =
...
     (struct type t = gtype let compare t1 t2 = E.compare t1.g t2.g end)
 
-(* ensemble de termes "set" avec leurs arguments et leurs types *)
+(* Set of writes *)
 type stype = {s:E.t; st:E.t; si:E.t; sv:E.t}
-module S :Set.S with type elt = stype = Set.Make
+module S = Set.Make
     (struct type t = stype let compare t1 t2 = E.compare t1.s t2.s end)
 
 (* map t |-> {set(t,-,-)} qui associe a chaque tableau l'ensemble
```
hra687261: Missed translation

src/lib/reasoners/arrays_rel.ml
```diff
@@ -85,91 +87,91 @@ let timer = Timers.M_Arrays
...
-        (Conseq.cardinal st);
-      Conseq.iter (fun (a,ex) ->
-          Printer.print_dbg ~header:false
-            "%a : %a" E.print a Ex.print ex
-        ) st
-    end
+   We say that a read/write is known if we have seen this term in [assume]. *)
+type t = {
+  gets : G.t;
+  (* Set of all the known reads. *)
```
hra687261: The documentations talks about reads and writes (referring to `select` and `store`) but the code refers to them as `set`s and `get`s. I think it should be harmonized and only one notation should be used.

Maybe at some point we should simply switch to the standard and use `select` and `store` internally, but I think `select` is used in the `ADT` theory as well.

src/lib/reasoners/arrays_rel.ml
```diff
@@ -85,91 +87,91 @@ let timer = Timers.M_Arrays
...
+
+  tbset : S.t TBS.t;
+  (* Set of all the known writes. There are indexed by array. *)
+
+  split : LRset.t;
+  (* Set of equalities or disequalities on relevant indices.
+     There are the hypotheses of consequences in the field [conseq].
```
hra687261:
```suggestion
     These are the hypotheses of consequences in the field [conseq].
```

src/lib/reasoners/arrays_rel.ml
```diff
@@ -85,91 +87,91 @@ let timer = Timers.M_Arrays
...
+     in [get_of_set]. *)
 
+  new_terms : E.Set.t;
+  (* Set of read and write terms produced by the theory. These terms
+     are supposed to be sent to the instantiation engine. *)
+
+  (* size_splits : Numbers.Q.t; *)
+  (* Limit the number of case splits performed on indices. This field
+     is currently unused. *)
+
```
hra687261: I suppose it can be removed

src/lib/reasoners/arrays_rel.ml
```diff
@@ -258,52 +260,69 @@ let update_env are_eq are_dist dep env acc gi si p p_ded n n_ded =
     let conseq = LRmap.add n n_ded dep conseq in
     { env with split = sp; conseq = conseq }, acc
 
-  | Some _,  Some _ -> assert false
+  | Some _,  Some _ ->
+    (* This case cannot occur because we cannot have `gi = si` and
+       `gi <> si` at the same time. *)
+    assert false
+
+(* Produce a formula to propagate a read over a store.
```
hra687261: Same as the comment above about notation, we should settle one notation and not mix multiple ones (in comments use `read` and `write`, in the code we use `get` and `set`, and here we use `store` as well)

src/lib/reasoners/arrays_rel.ml
```diff
@@ -258,52 +260,69 @@ let update_env are_eq are_dist dep env acc gi si p p_ded n n_ded =
...
-           update_env
-             are_eq are_dist dep env acc gi si p p_ded n n_ded
+           update_env (module Uf) uf dep env acc gi si p p_ded n n_ded
          | _ -> (env,acc)
-    ) (class_of gtab) (env,acc)
+    ) (Uf.class_of uf gtab) (env,acc)
+
+(* Assume that [stype] represents the write `(store b j v)`.
+   For all known writes of the form `(store a i w)` such that `a` and
+   `b` are equal, we form the formula:
```
hra687261: nitpick: this is more of a preference really
```suggestion
   `b` are equal, we create the formula:
```

src/lib/reasoners/arrays_rel.ml
```diff
@@ -410,54 +440,54 @@ let implied_consequences env eqs la =
...
       (fun (a,_,_,_) -> match a with A.Builtin _  -> false | _ -> true) la
   in
-  let env, eqs = extensionality (env, eqs) la class_of in
+  let env, eqs = extensionality (env, eqs) la in
   implied_consequences env eqs la
 
-(* choisir une egalite sur laquelle on fait un case-split *)
 let two = Numbers.Q.from_int 2
 
+(* Choose a equality/disequality between relevant indices to split on it. *)
```
hra687261:
```suggestion
(* Choose an equality/disequality between relevant indices to split on it. *)
```

src/lib/reasoners/arrays_rel.ml
```diff
@@ -410,54 +440,54 @@ let implied_consequences env eqs la =
...
       (fun nb (_,_,_,i) ->
          match i with
          | Th_util.CS (Th_util.Th_arrays, n) -> Numbers.Q.mult nb n
          | _ -> nb
-      )env.size_splits la
-  in
-  {env with size_splits = nb}
+      ) env.size_splits la
+   in
+   {env with size_splits = nb} *)
```
hra687261: To remove?

hra687261 (commented):
Thanks for the documentation (and the edits to the code), it makes the code more readable. I made some comments mainly on the form and the used language.

src/lib/reasoners/arrays_rel.ml
Halbaroth: I agree, it's a mess... There are three notations:
- `set/get` from the code
- `select/store` from the SMT-LIB standard,
- `read/write` from many papers on Array theories.

In my opinion, we should use the SMT-LIB standard but, as you pointed out, `select` is already in use in the ADT theory. I suggest that we keep `set/get` in Array and harmonize notations later.

src/lib/reasoners/arrays_rel.ml
Halbaroth: I didn't translate it because this comment is useless as the meaning of the TBS module is explained in the documentation of the field `tbset`. I suggest to remove it.

src/lib/reasoners/arrays_rel.ml
Halbaroth: I've pushed a new commit to address your comments. Is it better?

src/lib/reasoners/arrays_rel.ml
hra687261: Alright, there is another in line 49 that should also be removed

src/lib/reasoners/arrays_rel.ml
hra687261: Yes! LGTM

src/lib/reasoners/arrays_rel.ml
```diff
@@ -35,6 +35,14 @@ module Ex = Explanation
...
+(* In this module, we follow the below conventions in the comments:
+   - A get term refers to a term of the form `(select a i)` for an array `a` and
+     an index `i`.
+   - A set term refers to a term of the form `(store a i v)` for an array `a`
+     and an index `i`.
+   - A relevant index is any term that appears as an index in get or set terms.
+   - A get or set term is known if it has been encountered in [assume]. *)
+
 let src = Logs.Src.create ~doc:"Arrays_rel" __MODULE__
 module Log = (val Logs.src_log src : Logs.LOG)
```
hra687261, line 48: Unfortunately, GitHub doesn't allow comments on unchanged lines, but the comment in line 49 should be removed

hra687261 (approved):
LGTM! Other than the comment that needs to be remove.

src/lib/reasoners/arrays_rel.ml
```diff
@@ -35,6 +35,14 @@ module Ex = Explanation
 
 module LR = Uf.LX
 
+(* In this module, we follow the below conventions in the comments:
+   - A get term refers to a term of the form `(select a i)` for an array `a` and
+     an index `i`.
+   - A set term refers to a term of the form `(store a i v)` for an array `a`
+     and an index `i`.
+   - A relevant index is any term that appears as an index in get or set terms.
+   - A get or set term is known if it has been encountered in [assume]. *)
```
bclement-ocp, line 44: I'd argue that it would make more sense to call these "select terms" and "store terms" in documentation, with a comment mentioning that the variable names are legacy (or also rename the variables).

src/lib/reasoners/arrays_rel.ml
Halbaroth: I know this issue, it's really annoying!

src/lib/reasoners/arrays_rel.ml
bclement-ocp: I missed the other discussion — ignore this comment.

Halbaroth:
Strangely the CI failed on Ubuntu targets. I remember there is a test that fails sometimes on MacOS because the MacOS runner is too slow but it didn't happen on Ubuntu too? I restarted these tasks.

src/lib/reasoners/arrays_rel.ml
Halbaroth: As you've reviewed the PR, I need your approbation to merge it now :D

bclement-ocp:
> Strangely the CI failed on Ubuntu targets. I remember there is a test that fails sometimes on MacOS because the MacOS runner is too slow but it didn't happen on Ubuntu too? I restarted these tasks.

It is (somewhat) the opposite -- that test is supposed to timeout, but sometimes the runner is fast and it doesn't (and fails with a reported reason of "incomplete" or "proof-search" instead of "timeout"), because our "reproducible-resource-limit" is related to time and not actually reproducible. 

See #1277 for an attempt at a fix.

src/lib/reasoners/arrays_rel.ml
bclement-ocp: Sure, go ahead

PR #1269, opened 2024-11-13
Fix issue 1212

### LINKED ISSUE #1212: Invalid assumptions in array model generation
Alt-Ergo crashes with an assertion failure in the union-find module with the following input:

```smt-lib
; 1212.smt2
(set-logic ALL)
(declare-const b (Array Int Int))
(assert (= (store a 0 0) (store b 0 0)))
(check-sat)
(get-model)
```

Using 0d6300ec27446ea57626a19dc0a7b00a80daf3a7 (current `next` at the time of writing):

```console
$ alt-ergo 1212.smt2
; File "1212.smt2", line 7, characters 1-12: I don't know (0.0147) (8 steps) (goal g_1)

unknown
Fatal error: exception File "src/lib/reasoners/uf.ml", line 1276, characters 14-20: Assertion failed
```

Halbaroth:
During model extraction, we only should store accesses to arrays whose the representative elements are names. Indeed, if the representative is not a name, this array cannot appear in a model.

The embedded case is correctly managed because we generate fresh names that will become the representative elements.

### DIFF
--- a/tests/issues/1212.default.expected
+++ b/tests/issues/1212.default.expected
@@ -1,0 +1,6 @@
+
+unknown
+(
+  (define-fun a () (Array Int Int) (as @a3 (Array Int Int)))
+  (define-fun b () (Array Int Int) (as @a2 (Array Int Int)))
+)
--- a/src/lib/reasoners/uf.ml
+++ b/src/lib/reasoners/uf.ml
@@ -1158,11 +1158,12 @@ end
 
 type cache = {
   array_selects: Expr.t E.Table.t E.Table.t;
-  (** Stores all the get accesses to arrays. *)
+  (** Stores all the get accesses to array names. *)
 
   abstracts: Expr.t Shostak.HX.t;
   (** Stores all the abstract values generated. This cache is necessary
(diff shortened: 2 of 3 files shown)


tests/issues/1212.default.expected
```diff
@@ -0,0 +1,6 @@
+
+unknown
+(
+  (define-fun a () (Array Int Int) (as @a3 (Array Int Int)))
+  (define-fun b () (Array Int Int) (as @a2 (Array Int Int)))
+)
```
bclement-ocp, line 6: The generated model is wrong: it does not ensure that `a` and `b` are equal everywhere except on `0` (this is now a simpler reproducer for #929). The `1212.default.smt2` file should include a comment that the model is wrong and point back to #929.

hra687261 (approved):

bclement-ocp (dismissed):
I had the above comment marked as pending for a while (sorry! GitHub UI strikes again). I think this is still the case -- can you add a comment in the file so that we don't later think that we break things when we fix #929?

Halbaroth:
Sorry, I didn't follow notifications on Alt-Ergo for a long time. I will fix it and merge the PR.

PR #1276, opened 2025-01-27
Add command line option to record backtraces

hra687261:
I find it pretty useful for debugging especially crashes/assertion failures during development

### DIFF
--- a/src/bin/common/parse_command.ml
+++ b/src/bin/common/parse_command.ml
@@ -321,6 +321,10 @@ module Debug = struct
            ~doc:medium_doc),
     Arg.(value & opt_all enum_conv [] & info ["ddd"; "dddebug"] ~docv
            ~doc:full_doc)
+
+  let backtrace_flag =
+    Arg.(value & flag & info ["b"; "backtrace"]
+           ~doc:"Turn on recording of exception backtraces.")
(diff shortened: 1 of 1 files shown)


Halbaroth:
You can use the environment variable `OCAMLRUNPARAM` to turn on backtrace recording as follows:
```
OCAMLRUNPARAM=b alt-ergo ...
```
See https://ocaml.org/manual/5.3/runtime.html for more information.
Your solution looks good to me but it adds an extra flag which is not absolutely necessary. Let's me know if you really want it.

hra687261:
Tbh I find `alt-ergo -b` a lot more practical than using an env variable with `OCAMLRUNPARAM=b alt-ergo`, but yes it is not necessary, I guess it is more a matter of taste.
I don't know, any opinions @bclement-ocp and @Gbury ?

Halbaroth:
If you need to record backtraces for several commands, you can type `export OCAMLRUNPARAM=b` in your shell. I don't have a strong opinion on this, it's simply a matter of taste.

bclement-ocp (approved):
This is simple enough that if it's a convenience, however small, to at least one developer I'm OK with it.

(I believe test failures are unrelated)

PR #1277, opened 2025-01-27
test: Make sure qfbv-timeout actually times out

bclement-ocp:


Halbaroth (approved):

Halbaroth:
We need to implement a reproducible `reproducible-resource-limit` with steps.

bclement-ocp:
Yes -- I thought we had an issue for it, but I can't find it at the moment.

Looks like the fix is not good since this still fails on ubuntu/4.08.1. It's tricky because we want to test timeout errors during assume, not necessarily search. Let's just disable the test for now.

Halbaroth:
> Yes -- I thought we had an issue for it, but I can't find it at the moment.

Sure, I will try to fix it this week when I find the time.

Issue #1270, opened 2024-12-04
Unknown syntax in model produced by Alt-Ergo 2.6.0


manmatteo:
Since Alt-Ergo 2.6.0, we are getting literals starting with `.` in the produced models, and I am not sure how they should be interpreted (currently Why3 considers them invalid syntax). For example in this file:

```smt
(set-logic ALL)
(set-option :produce-models true)
(declare-sort t 0)
(declare-fun q (t) Int)
(declare-fun o (Int) t)
(assert (forall ((i Int)) (or (> 0 0) (= i (q (o i))))))
(check-sat)
(get-model)
```

I obtain the model
```smt
unknown
(
  (define-fun q ((arg_0 t)) Int
   (ite (= arg_0 .k1) 1 (ite (= arg_0 .k0) 0 (- 1))))
  (define-fun o ((arg_0 Int)) t
   (ite (= arg_0 1) (as @a2 t) (ite (= arg_0 0) (as @a3 t) (as @a4 t))))
)
```

The definition of `o` contains things like `(as @a2 t)`, that I expect. The definition of `q` contains `.k0` instead, which seem to play a similar role but, for instance, skips the type annotation.

Gbury:
In the SMT-LIB standard, identifiers starting with either a `.` or a `@` are reserved for use by solvers in models (or more generally in the output produced by solvers). This helps solvers more easily generate names with the guarantee they they do not clash with existing names from the input. So from the point of view of why3, these should be regular identifiers. That being said, in the output you showed, the identifiers that start with a dot appear to not be bound or defined, so that looks like an internal error of alt-ergo (probably a missing substitution before printing a model).

manmatteo:
I see. It is strange that we only started observing this problem with Alt-Ergo 2.6.0 then. As for the fact that they are not bound, isn't it the same also for the `@` identifiers in the output above? I am used to seeing them appear without being declared, and I think the model parser in Why3 already deals with this.

Gbury:
Right, so there is a difference between identifiers starting with a `.` and those starting with `@`: identifiers with a dot should be regular identifiers, the dot is only there to prevent conflicts with names from the input problem, so in particular these identifiers should be defined.

[...]

Issue #1213, opened 2024-08-21
Fresh names in polymorphic model


Halbaroth:
Consider the following input file (which is a polymorphic version of our `list.models.smt2`):
```smt2 
(set-option :produce-models true)
(set-logic ALL)
(declare-datatypes ((List 1)) (
  (par (T) (
    (nil)
    (cons (head T) (tail (List T)))
  ))
))
(declare-sort T 0)
(declare-const a (List T))
(declare-const b (List T))
(declare-const c (List T))
(declare-const d (List T))
(assert (= (tail a) b))
(assert ((_ is cons) c))
(assert (= (tail c) d))
(check-sat)
(get-model)
```
Alt-Ergo outputs:
```
(
  (define-fun a () (List T) nil)
  (define-fun b () (List T) nil)
  (define-fun c () (List T) (cons .k2 nil))
  (define-fun d () (List T) nil)
)
```
which is not correct as `.k2` is defined (and as it starts by a dot, it should not appear in models). We expect to get an abstract value instead of `.k2`. 

I tried to fix this bug but it is actually very annoying. A simple fix consists in replace `fresh_name` by `mk_abstract` at some places in relation modules but it makes sense to use `fresh_name` here. Most of the times, there are not supposed to appear in models! 

I think that a decent fix consists in applying a substitution on representative terms before generating model terms. This substitution replaces fresh leaves by abstract ones and we have to ensure that we never generate two different abstract values for the same fresh ones.

bclement-ocp:
> I think that a decent fix consists in applying a substitution on representative terms before generating model terms. This substitution replaces fresh leaves by abstract ones and we have to ensure that we never generate two different abstract values for the same fresh ones.

This should be the job of `Uf.assign_next` / `X.assign_value`, which they are not doing properly because our definition of model terms (called "value terms" in the SMT-LIB standard) is wrong: only abstract constants should be treated as such.

Halbaroth:
Of course, I tried to fix this bug by modifying `Uf.assign_next` and `X.assign_value` but it is not sufficient. I will investigate later.

bclement-ocp:
Quick proof of concept (the code for array models needs to be adapted):

```patch
diff --git a/src/lib/reasoners/records.ml b/src/lib/reasoners/records.ml
index cb8e1170..b14b3d56 100644
--- a/src/lib/reasoners/records.ml
+++ b/src/lib/reasoners/records.ml
[...]

PR #1280, opened 2025-01-28
Replace fresh terms by abstract values in all subterms

Halbaroth:
This PR fixes the issue #1270. More precisely,
1. Add a new argument to the function `X.to_model_term`. The purpose of `to_model_term` is to generate a model term from constant semantic values. Unfortunately, these model terms are not exactly the model values of the SMT-LIB standard because they can contain fresh terms. The new argument of `X.to_model_term` is a function of type `E.t -> E.t` that is used to create abstract values from fresh terms. A cache is hidden in the closure of this function to prevent from duplicating abstract values for the 
same fresh terms. 
3. Improve and update some comments.
4. Remove the `term_values` field in `Models`. This field was only used to cache the result of `model_repr_of_term` in `Uf`. 
    I replaced it by a local cache in the closure of `extract_concrete_model`.

Now the output of the example of #1270 is:
```smt2 
unknown
(
  (define-fun q ((arg_0 t)) Int
   (ite (= arg_0 (as @a2 t)) 1 (ite (= arg_0 (as @a4 t)) 0 (- 1))))
  (define-fun o ((arg_0 Int)) t
   (ite (= arg_0 1) (as @a3 t) (ite (= arg_0 0) (as @a5 t) (as @a7 t))))
) 
```

### DIFF
--- a/src/lib/frontend/models.ml
+++ b/src/lib/frontend/models.ml
@@ -27,13 +27,11 @@ let constraints = ref MS.empty
 type t = {
   propositional : Expr.Set.t;
   model : ModelMap.t;
-  term_values : Expr.t Expr.Map.t
 }
 
 let empty = {
   propositional = Expr.Set.empty;
   model = ModelMap.empty ~suspicious:false [];
-  term_values = Expr.Map.empty;
 }
 
 module Pp_smtlib_term = struct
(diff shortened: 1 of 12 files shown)


Halbaroth:
I realized that the solution we found for #1213 should fix #1270 too without modifying the signature of any modules. I tested it on 
the file of Matteo, it works! I will finish your patch. I keep this PR open.

PR #1281, opened 2025-01-30
Use constant terms as identifiers in names and variables

Halbaroth:
This PR changes the way we construct identifiers for names (in `Symbols`) and variables (in `Var`).

We plan to completely replace the AE identifiers with identifiers from Dolmen. This PR serves as an intermediate step from string-based identifiers to Dolmen identifiers.

- All the identifiers are now managed in the `Id` module.
- Constant terms are used for declared or defined identifiers in the input file.
- Internal identifiers are still pre-mangled strings. Replacing them with Dolmen identifiers requires careful investigation, 
  particularly in the AC(X) implementation.
- I haven't removed `Id.typed` yet. After this PR, the user identifiers will be typed, so we do not need to store them separately. If we use the internal type of the identifier, we will need to implement a conversion function from Dolmen type to Alt-Ergo type, which is not trivial. I prefer keeping this type and remove it after transitioning from Alt-Ergo types to Dolmen types.
- The new API of `Id` prevents us from creating duplicate fresh/fresh_ac/abstract identifiers.
- I still check that user identifiers do not start with `.` or `@`. In Before this PR, we mangled the identifier if it happened. In this PR, we crash. I don't have a strong opinion on this behavior, it was just simpler to implement this way.
- I use identifiers for term variables too. I preserved the internal representation because:
  1. the notion of local variables is AE-specific, whereas the local name in Dolmen identifiers is a distinct concept.
  2. After this PR, variables are typed but what is the type of `Underscore`? May be `forall a. a`?

This PR have been tested on ae-format (+2-2).


Halbaroth:
Should be ready for reviewing. I am running some tests to be sure that I did not break anything with variables.

src/lib/structures/id.ml
```diff
@@ -16,45 +16,139 @@
...
+
+let hash i =
+  match i with
+  | Term_cst { tcst; _ } ->
+    Dolmen.Std.Expr.Term.Const.hash tcst
+  | Hstring { hs; _ } ->
+    (* NB: Internal identifiers are pre-mangled, which means that we do not
+       need to take the name space into consideration when hashing. *)
+    Hstring.hash hs
+
```
bclement-ocp: There are probably a lot of collisions here because both Dolmen and Alt-Ergo use increasing integers starting from 0 for `Hstring` and `Id` hashes.

Gbury: That actually reminds me that, even though it's often tempting (and dolmen falls into that trap), using exactly the unique id as a hash is not always a very good choice: it might be good in the case where you expect all (or most) of the values to be in the hastbl, but especially with hashtables that uses hash modulo the table size, there may be some surprising collisions, particularly if there is some pattern to which values are added to the table (e.g. if only values with an even unique id are added to the table, or other similar situations).

bclement-ocp: Yup, it is good when usage is dense (the hash table effectively becomes a dynarray -- e.g. for hash consing) but I'm partial myself to hashing the integer identifier in `hash` as a default and exposing it so that people that know they are in the dense scenario can use it instead.

Halbaroth: Uhm, that could indeed be the source of the slowdown. I will try another hash function.

PR #1283, opened 2025-02-03
Use OCaml 5.3.0 in CI

Halbaroth:


Halbaroth:
I don't understand why there is still a pending check with OCaml 5.0.0.

bclement-ocp:
> I don't understand why there is still a pending check with OCaml 5.0.0.

Because OCaml 5.0.0 is marked as required.

bclement-ocp (approved):

bclement-ocp:
I have updated the settings on GitHub to require 5.3.0 instead of 5.0.0.

PR #1284, opened 2025-02-04
Use Dolmen locations in recoverable/fatal errors

Halbaroth:
This PR includes two modifications:
1. Replacing Alt-Ergo locations from `Loc` to `Dolmen.Std.Loc.loc`.
2. Using Dolmen locations in recoverable or fatal errors in `Solving_loop` (when it makes sense to print such locations).


tests/cram.t/run.t
```diff
@@ -123,21 +127,25 @@ Now we check that we have a proper error message when optimizing with the
...
+  (error "File "<stdin>", line 1, character 16-28:
+          the selected solver does not support optimization")
   
   unknown
 
   $ echo '(set-option :produce-models true) (set-logic ALL) (check-sat) (get-objectives)' | alt-ergo -i smtlib2 -o smtlib2 --continue-on-error --sat-solver Tableaux 2>/dev/null
   
   unknown
-  (error "the selected solver does not support optimization")
+  (error "File "<stdin>", line 1, character 62-78:
```
bclement-ocp: It is not correct to have quotes inside strings in SMT-LIB format; they need to be replaced with double quotes `""`.

src/lib/frontend/translate.ml
```diff
@@ -1762,15 +1758,13 @@ let rec is_pure_term t =
   | Sy.Op Tite -> false
   | _ -> List.for_all is_pure_term xs
 
-let make dloc_file acc stmt =
+let make file acc stmt =
   let rec aux acc (stmt: _ Typer_Pipe.stmt) =
+    let loc = Dolmen.Std.Loc.loc file stmt.loc in
```
bclement-ocp: Nit: Calling this `st_loc` would make it a bit less verbose to output the commands below.

src/bin/common/solving_loop.ml
```diff
@@ -815,19 +826,19 @@ let process_source ?selector_inst ~print_status src =
...
+            recoverable_error ~loc
+              "%a: No model produced, cannot execute get-assignment."
+              DStd.Loc.fmt loc;
             st
         end
 
-      | {contents = `Other (custom, args); loc; _} ->
-        handle_custom_statement loc custom args st
+      | {contents = `Other (custom, args); loc = l; _} ->
+        handle_custom_statement ~loc:l custom args st
```
```suggestion
      | {contents = `Other (custom, args); loc; _} ->
        handle_custom_statement ~loc custom args st
```

bclement-ocp (changes requested):
Looks sensible overall.

I think we might want to have a less verbose output for errors reported in `(error)` where using the Emacs format doesn't make that much sense; something terser like CVC5's `file:line.columns: message` would be better suited.  We can iterate on the error messages themselves, however the escaping issue needs to be fixed (`"` should be displayed as `""`).

src/bin/common/solving_loop.ml
Halbaroth: The variable `st_loc` already contains the location of the statement but the value `Dolmen.Loc.compact` is buggy (even its type is incorrect!), so we cannot transform a full location in `handle_custom_statement` into a compact location. Guillaume will address this issue in the next Dolmen release. 

As a workaround, I passed the compact location of the statement itself here and used a different name for clarity. Actually the types are different, so shadowing wasn't that bad.

I applied your suggestion and add todo comment too.

tests/cram.t/run.t
Halbaroth: I imported the function `Dolmen.Std.Loc.fmt` and adapted it. Yet, if the file name contains double quotes, they won't be escaped. I can write a function to handle them if you think it is important.

bclement-ocp: Thanks! I think that since this is a format that we are using to interoperate with tools (not just users) we should be robust here -- the escaping should ideally be done generically in `print_smtlib_err`. I will make a separate PR.

bclement-ocp (approved):

PR #1286, opened 2025-02-06
Remove get file

Halbaroth:
One step after another, we get rid of the global state!

This PR is rebased on #1284, only the last commit is relevant for reviewing.

### DIFF
--- a/src/bin/common/parse_command.mli
+++ b/src/bin/common/parse_command.mli
@@ -28,9 +28,10 @@
 (** {1 Parse_command module used at start-up to parse the command line} *)
 
 (** Only exported function. Calling it will parse the command line options
-    and set them accordingly for the rest of the execution *)
-val parse_cmdline_arguments : unit -> unit
+    and set them accordingly for the rest of the execution.
 
+    Returns the path of the input file. *)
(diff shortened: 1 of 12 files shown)


src/bin/common/parse_command.mli
```diff
@@ -28,9 +28,10 @@
 (** {1 Parse_command module used at start-up to parse the command line} *)
 
 (** Only exported function. Calling it will parse the command line options
-    and set them accordingly for the rest of the execution *)
-val parse_cmdline_arguments : unit -> unit
+    and set them accordingly for the rest of the execution.
 
+    Returns the path of the input file. *)
+val parse_cmdline_arguments : unit -> [`Stdin | `File of string]
```
bclement-ocp: I think it would make sense to create a type for this rather than just returning the path. I can see us wanting to return more information from parsing than just the input file, might as well lay up the ground work.

Something like:

```ocaml
type parse_result = { path : [ `Stdin | `File of string ] }
```

which we can extend later.

bclement-ocp (approved):
LGTM (only the last commit, #1284 is reviewed separately). I'd prefer using a separate type for the info returned after command line parsing, but am fine with doing that change later when it's needed.

PR #1294, opened 2025-02-12
Check the status of SatML in Satml_frontend

### LINKED ISSUE #1293: Obvious inconsistency ignored after optimization in CDCL-Tableaux
As noticed in #1023, the SAT solver environment is unsat after the optimization. Consider the simplified example: 
```ocaml 
let env = SAT.empty () in 
SAT.declare env x; 
SAT.assume env (x <= 10);
SAT.maximize env x;
SAT.check_sat ();           (* A *)
SAT.assume env (x <= 10);   (* B *)
SAT.check_sat ();
```
After `A`, the environment of the SAT solver is unsat because it contains both the clause `x <= 10` and `x > 10` at the decision 
level 0. We expect that the last `assume` raises an exception. As explained in PR #1291, the `is_unsat` field of the SatML environment is `true` after `A`, but no `Unsat` exception is raised at `B`!

Now replace the command `B` by `SAT.assume env (x + x <= x + 10)`. Then `B` raises `Unsat` as expected. 
The issue comes from the wrapper of `SatML.assume` in `Satml_frontend`. As the formula `x <= 10` is already in the `gamma` set, `Satml_frontend` does not send twice the formula to `SatML` but it doesn't check the status of the SatML environment too:
https://github.com/OCamlPro/alt-ergo/blob/837293ef8d4c55eafdcc152089866a24db4fca91/src/lib/reasoners/satml_frontend.ml#L763-L773

I believe that checking the status of SatML at the beginning of `SAT.assume` and `SAT.check_sat` is the correct fix, but I would like to get a second opinion.

Note: `CDCL` is also affected.

Halbaroth:
As explained in the issue https://github.com/OCamlPro/alt-ergo/issues/1293, Satml_frontend does not send to SatML a formula already present in its lazy cnf. In particular, `assume` and `unsat` of `Satml_frontend` can answer `unknown` for obviously unsat environment. After this commit, we always check the status of SatML in `assume` and `unsat`.

If the solver is already unsat, the command `unsat` should return the unsat core. As SatML does not yet support this feature, I returned an empty explanation.

This PR fixes #1293

### DIFF
--- a/src/lib/reasoners/satml.ml
+++ b/src/lib/reasoners/satml.ml
@@ -84,6 +84,9 @@ type conflict_origin =
 (* not even the final one *)
 let vraie_form = E.vrai
 
+type status =
+  | Sat
+  | Unsat of Atom.clause list option
 
 module type SAT_ML = sig
   (*module Make (Dummy : sig end) : sig*)
@@ -141,6 +144,7 @@ module type SAT_ML = sig
 
   val optimize : t -> Objective.Function.t -> unit
 
+  val status : t -> status
 end
 
 module MFF = FF.Map
@@ -2289,4 +2293,8 @@ module Make (Th : Theory.S) : SAT_ML with type th = Th.t = struct
     else
       Vec.replace (fun fns -> fn :: fns) env.objectives
         (Vec.size env.objectives - 1)
+
+  let[@inline always] status env =
+    if env.is_unsat then Unsat env.unsat_core
+    else Sat
 end
--- a/src/lib/reasoners/satml.mli
+++ b/src/lib/reasoners/satml.mli
@@ -36,6 +36,10 @@ type conflict_origin =
   | C_bool of Atom.clause
   | C_theory of Explanation.t
 
+type status =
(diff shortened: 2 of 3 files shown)


bclement-ocp:
I'm not sure I understand what this has to do with the lazy CNF, can you clarify that point?

It seems that the cases where we are checking for `SAT.status` in `Satml_frontend` ultimately go through to `cdcl_assume`, which in turn calls `SAT.assume`. It seems to me that it would be more robust to raise `Unsat` in `SAT.assume` if the environment is already known to be unsat (`SAT.assume` is already raising `Unsat` if it finds a contradiction through propagation only), because that way we don't require caller to check unsat-status manually.

Halbaroth:
Right, the issue is not related with the lazy cnf, but the `gamma` field in the environment of `Satml_frontend`. This field prevents from asserting twice the same formula in `pre_assume` as I explained in #1293. I agree that we can directly raise `Unsat` in Satml. I close this PR and open a new one as the patch is too different.

PR #1299, opened 2025-02-19
Remove get_first_interpretation adn get_every_interpretation

Halbaroth:
As noticed by Basile, this feature is broken in presence of optimization and no one used it. This PR removes these options.

I kept the cli option but it is now a no-op.

### DIFF
--- a/src/bin/common/parse_command.ml
+++ b/src/bin/common/parse_command.ml
@@ -423,7 +423,7 @@ let mk_limit_opt age_bound fm_cross_limit timelimit_interpretation
     `Ok()
 
 let mk_output_opt
-    interpretation objectives_in_interpretation unsat_core
+    model_generation objectives_in_interpretation unsat_core
     output_format model_type () () () ()
   =
   set_infer_output_format (Option.is_none output_format);
@@ -435,7 +435,7 @@ let mk_output_opt
     | None -> Value
     | Some v -> v
   in
-  set_interpretation interpretation;
+  set_model_generation model_generation;
   set_objectives_in_interpretation objectives_in_interpretation;
   set_unsat_core unsat_core;
   set_output_format output_format;
@@ -912,28 +912,17 @@ let parse_output_opt =
      --interpretation last) to determine the interpretation value. *)
   let interpretation, dump_models, dump_models_on, frontend =
     let interpretation =
-      let doc = Format.sprintf
-          "Best effort support for counter-example generation. \
-           $(docv) must be %s. %s shows the first computed interpretation. \
-           %s compute an interpretation before every decision, \
-           and %s only before returning unknown. \
-           Note that $(b, --max-split) limitation will \
-           be ignored in model generation phase."
-          (Arg.doc_alts
-             ["none"; "first"; "every"; "last"])
-          (Arg.doc_quote "first") (Arg.doc_quote "every")
-          (Arg.doc_quote "last") in
-      let docv = "VAL" in
-      let interpretation =
-        Arg.enum
-          [ "none", INone
-          ; "first", IFirst
-          ; "every", IEvery
-          ; "last", ILast
-          ]
+      let doc =
+        Fmt.str
+          "This option allowed choosing when the model generation occurs. \
+           It have been removed in Alt-Ergo 2.7.0. If model generation is \
+           enabled, it is performed as the final step, which was the default \
+           in previous versions."
       in
-      Arg.(value & opt interpretation INone &
-           info ["interpretation"] ~docv ~docs:s_models ~doc)
+      let docv = "VAL" in
+      let deprecated = "this option is deprecated and is ignored." in
+      Arg.(value & opt string "dummy" &
+           info ["interpretation"] ~docv ~docs:s_models ~doc ~deprecated)
     in
     let produce_models =
       let doc =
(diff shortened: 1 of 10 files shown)


src/bin/common/parse_command.ml
```diff
@@ -912,28 +912,17 @@ let parse_output_opt =
...
-        Arg.enum
-          [ "none", INone
-          ; "first", IFirst
-          ; "every", IEvery
-          ; "last", ILast
-          ]
+      let doc =
+        Fmt.str
+          "This option allowed choosing when the model generation occurs. \
+           It have been removed in Alt-Ergo 2.7.0. If model generation is \
```
```suggestion
           It has been removed in Alt-Ergo 2.7.0. If model generation is \
```

src/bin/common/parse_command.ml
```diff
@@ -980,15 +969,11 @@ let parse_output_opt =
            info ["dump-models-on"] ~docv ~docs:s_models ~doc)
     in
 
-    let mk_interpretation interpretation produce_models dump_models =
-      match interpretation with
-      | INone when produce_models || dump_models -> ILast
-      | interpretation -> interpretation
+    let mk_interpretation _interpretation produce_models dump_models =
```
```suggestion
    let mk_model_generation _interpretation produce_models dump_models =
```

src/bin/common/parse_command.ml
```diff
@@ -980,15 +969,11 @@ let parse_output_opt =
...
+      produce_models || dump_models
     in
     Term.(
-      const mk_interpretation $ interpretation $
-      produce_models $ dump_models
-    ),
+      const mk_interpretation $ interpretation $ produce_models $ dump_models),
```
```suggestion
      const mk_model_generation $ interpretation $ produce_models $ dump_models),
```

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1198,10 +1186,12 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
             env.last_forced_normal <- env.last_forced_normal - 1
         in
-        if not updated then
-          update_model_and_return_unknown
-            env (Options.get_last_interpretation ())
-            ~unknown_reason:Incomplete; (* may becomes ModelGen *)
+        if not updated then (
+          if Options.get_model_generation () then update_model env;
+          Options.Time.unset_timeout ();
+          (* may becomes ModelGen *)
```
bclement-ocp: I don't think it can become `ModelGen` anymore — if the unknown-reason becomes ModelGen it should be set by `update_model`, not `i_dont_know`.

src/lib/util/options.mli
```diff
@@ -209,14 +198,8 @@ val set_inline_lets : bool -> unit
...
-(** Set [interpretation] accessible with {!val:get_interpretation}
-
-    Possible values are :
-    {ol {- First} {- Before every instantiation}
-     {- Before every decision and instantiation}
-     {- Before end}}
-*)
-val set_interpretation : interpretation -> unit
+(** Set [produce_models] accessible with {!val:get_produce_models} *)
+val set_model_generation : bool -> unit
```
bclement-ocp: Either the comment should be updated to say `model_generation` and `get_model_generation`, or the code should be updated to be called `set_produce_models` and `get_produce_models` (I think the second option makes more sense --- might as well standardise on the SMT-LIB names).

Halbaroth: I named it `set_produce_models` first, I have no strong opinion on its name.

bclement-ocp: When in doubt about naming and there is an appropriate name for the SMT-LIB standard we should pick that one I think, so let's go with `set_produce_models` even if that is a slightly awkward name.

src/lib/reasoners/satml_frontend.ml
```diff
@@ -1000,28 +1000,18 @@ module Make (Th : Theory.S) : Sat_solver_sig.S = struct
...
+      (* also performs case-split and pushes pending atoms to CS *)
+      let declared_ids = env.declare_top in
+      let model, objectives =
+        SAT.compute_concrete_model ~declared_ids env.satml
+      in
+      env.last_saved_model <- Some model;
+      env.last_saved_objectives <- Some objectives;
+    with
+    | Ex.Inconsistent (_expl, _classes) as e -> raise e
+    | Util.Timeout -> i_dont_know env (Timeout ModelGen)
```
bclement-ocp, line 1014: I believe this line was removed because if we raise `I_dont_know` here the solver ends up in an inconsistent state and performing further commands can lead to incorrect output or crashes — is there a reason to add it back?

Halbaroth: On `next`, if `Timeout` is raised during model generation, we catch it in `unsat_rec` and raise `I_dont_know` with `Timeout ProofSearch` as unknown reason, so the solver already ends up in an inconsistent state.

bclement-ocp: Ah, that is true — we actually check for timeout outside of the solver, from the frontend (which is a design wart we should fix someday, but it does mean this line is OK).

Halbaroth: I added it because you pointed out that we cannot change the unknown reason as suggested by Steven. I don't understand what is wrong here.

Halbaroth: uhm, I see. It is because of the lazy value again...

bclement-ocp (approved):

PR #1300, opened 2025-02-19
Remove the lazy value in `extract_concrete_model`

Halbaroth:
The code is harder to understand with this lazy value and most of the model generation code does not run if `Options.get_interpretation ()` is not true.


Halbaroth:
Added to #1291

PR #1304, opened 2025-02-25
Create a custom ADT for Th_util.answer

Halbaroth:
An option type is not self-documenting code.

### DIFF
--- a/src/lib/reasoners/th_util.mli
+++ b/src/lib/reasoners/th_util.mli
@@ -25,7 +25,18 @@
 (*                                                                        *)
 (**************************************************************************)
 
-type answer = (Explanation.t * Expr.Set.t list) option
+(** Type of answers returned by theory queries for ground literals. *)
+type answer =
+  | Entailed of { ex : Explanation.t; classes : Expr.Set.t list }
+  (** The literal is entailed by the theory environment:
+      - The explanation [ex] justifies why the literal holds in the
+        environment.
+      - [classes] is used only for debugging purposes. *)
+
+  | Unknown
+  (** One does not encounter a contradiction when assuming the negation of the
+      literal. It does not imply that the literal is false in the environment,
+      as the decision procedures might be incomplete. *)
 
 type theory =
   | Th_arith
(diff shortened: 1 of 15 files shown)


src/lib/reasoners/th_util.mli
```diff
@@ -25,7 +25,18 @@
...
+  | Entailed of { ex : Explanation.t; classes : Expr.Set.t list }
+  (** The literal is entailed by the theory environment:
+      - The explanation [ex] justifies why the literal holds in the
+        environment.
+      - [classes] is used only for debugging purposes. *)
+
+  | Unknown
+  (** One does not encounter a contradiction when assuming the negation of the
+      literal. It does not imply that the literal is false in the environment,
+      as the decision procedures might be incomplete. *)
```
bclement-ocp: This comment seems misleading — even with a complete decision procedure, if a literal is not entailed, it does not imply that its negation is entailed (there might be models where the literal is true and other models where it is false).

Incompleteness means that the literal might still be entailed even though we were not able to prove that, so I'd suggest something along the line of:


> The theory couldn't prove that the literal is entailed by the current environment. This might mean one of several things:
>        
>  - The negation of the literal is entailed by the current environment, or
>         
>  - The truth value of the literal is independent of the current environment (it is true in some models and false in other models), or
>            
> - The literal is entailed by the current environment but we failed to prove so (in which case the decision procedure used is incomplete).

src/lib/reasoners/th_util.mli
```diff
@@ -25,7 +25,18 @@
 (*                                                                        *)
 (**************************************************************************)
 
-type answer = (Explanation.t * Expr.Set.t list) option
+(** Type of answers returned by theory queries for ground literals. *)
+type answer =
+  | Entailed of { ex : Explanation.t; classes : Expr.Set.t list }
+  (** The literal is entailed by the theory environment:
+      - The explanation [ex] justifies why the literal holds in the
+        environment.
```
bclement-ocp: Might be worth insisting here that this means:
 - `ex` holds in the current environment
 - `ex => lit` is a tautology

bclement-ocp (commented):
Thanks! This makes things more readable. Please update the comments on `Th_util.answer` to be reworked a bit for clarity and correctness (see inline comments) and we should be good to merge.

bclement-ocp (approved):
Thought I'd approved this one, sorry!

PR #1305, opened 2025-02-26
Add a test to cover corner cases in `add_predicate`

Halbaroth:
The `Instances` module does not handle ground predicates the same way as other predicates. These corner cases are not covered by our non-regression tests (add an assertion if you do not believe me).

This PR adds a test in AE format to cover these cases.

Note that we cannot cover them with a SMT format test, as we do not differentiate functions and predicates in `Translate` for this format.


bclement-ocp (approved):
Can you clarify the

PR #1307, opened 2025-02-28
Update the installation documentation

Halbaroth:
- Explain how to install Alt-Ergo binary with `dune install`.
- Simplify the Makefile targets for JavaScript artifacts.
- Remove documentation about AB-Why3 plugin and mention it was deprecated and removed.
- Reorganize a bit the documentation of the installation and move developer-specific details to the developer documentations.
- Remove the documentation about `install` and `install-all` targets, as no one uses them and the user cannot specify a prefix directory with them. I kept them for the CI.


docs/sphinx_docs/requirements.txt
```diff
@@ -5,4 +5,5 @@
 sphinx >= 4.4.0
 myst-parser
 sphinx_rtd_theme
+sphinx-markdown-tables
```
bclement-ocp, line 8: I think we had issues with this plugin in the past — did you check that the rendering of the documentation was OK?

bclement-ocp (approved):
LGTM. Would be nice to tweak the makefile to provide standard targets sometimes (`make install` etc.) but that's not required.

Please do check that the sphinx-markdown-table plugin does not mess up existing documentation pages though, I think we had issues with it in the past.

Halbaroth:
> Please do check that the sphinx-markdown-table plugin does not mess up existing documentation pages though, I think we had issues with it in the past.

Sure, I checked before submitting this PR and it works fine.
Screenshot 2025-03-24 at 10-13-47 Contributing — Alt-Ergo Documentation documentation

bclement-ocp:
> > Please do check that the sphinx-markdown-table plugin does not mess up existing documentation pages though, I think we had issues with it in the past.
> 
> Sure, I checked before submitting this PR and it works fine. Screenshot 2025-03-24 at 10-13-47 Contributing — Alt-Ergo Documentation documentation

I'm not talking about the new table, I'm talking about existing tables that we have on other pages. 

Actually — why do we need this new plugin for tables? I think they are already supported by `myst-parser`?

Halbaroth:
I tried to use the markdown syntax for the table without this plugin and it didn't work.

Halbaroth:
Strange, it seems to work now \O/. I will remove it.

PR #1313, opened 2025-03-25
Move guard creation from Satml_frontend to Satml

bclement-ocp:
This reduces the amount of calls to `new_vars` from outside the `Satml` module to 1, making it easier to get rid of that function and fold it into `assume` for a simpler and more robust API.


Halbaroth (approved):
LGTM :)

PR #1318, opened 2025-04-15
Step limit fixes

bclement-ocp:
Backport of #1245 #1316 #1317 for 2.6.1 bugfix release.

### DIFF
--- a/src/lib/reasoners/uf.ml
+++ b/src/lib/reasoners/uf.ml
@@ -474,6 +474,9 @@ module Env = struct
     let rec fixpoint () =
       Options.exec_thread_yield ();
       (try SetRL.iter apply_rule rls with Exit -> ());
+      (* AC(X) bugs can trigger an infinite loop here, so increment a steps
+         counter to be safe. *)
+      Steps.incr Ac;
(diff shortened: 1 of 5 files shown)


src/lib/reasoners/uf.ml
```diff
@@ -474,6 +474,9 @@ module Env = struct
     let rec fixpoint () =
       Options.exec_thread_yield ();
       (try SetRL.iter apply_rule rls with Exit -> ());
+      (* AC(X) bugs can trigger an infinite loop here, so increment a steps
+         counter to be safe. *)
+      Steps.incr Ac;
```
Halbaroth, line 479: Nit: we could mention the issue here.

Halbaroth (approved):

src/lib/reasoners/uf.ml
bclement-ocp: I'd rather avoid linking to any specific issue because that implies the code could be removed when the issue is fixed, but any AC(X) bug (more specifically any bug related to AC(X) order) could trigger an infinite loop here.

PR #1320, opened 2025-04-22
refactor(bv): Simplify bv2nat mapping using right shifts

bclement-ocp:
The bv2nat mapping is able to record an integer expression for each bit-vector extraction, but we only need to record arithmetic right shifts since we encode an extraction `bv<i, j>` as `(bv asr j) - (bv asr i) * 2^(j - i + 1)`.

This ensures we can't accidentally leave bogus extractions in the map.

### DIFF
--- a/src/lib/reasoners/bitv_rel.ml
+++ b/src/lib/reasoners/bitv_rel.ml
@@ -118,7 +118,7 @@ module BV2Nat = struct
 
      In order to minimize the number of polynomial variables, we compute the
      integer version of an extraction of [x] from bit [i] to [j] (inclusive) as
-     [0 <= bv2nat(x asr i) - bv2nat(x asr j) * 2^(j-i+1) < 2^(j-i+1)].
+     [0 <= bv2nat(x) asr i - bv2nat(x) asr j * 2^(j-i+1) < 2^(j-i+1)].
 
      Note: We currently have no way of defining the bv2nat and int2bv of an
      arbitrary semantic value structurally, so we resort to dynamically creating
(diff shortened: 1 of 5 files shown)


src/lib/reasoners/bitv_rel.ml
```diff
@@ -396,84 +398,89 @@ module BV2Nat = struct
...
+        (* bv2nat(st) * 2^(size - sz) + bv2nat(bv') = p *)
+        let (bv2nat_st, st_ex), t = find_or_init_st st t in
+        let size' = size - sz in
+        let p' = T.Ints.(p - bv2nat_st *$$ Z.(one lsl size')) in
+        loop ~ex:(Ex.union ex st_ex) ~size:size' bv' p' t
+    in loop ~ex ~size bv p t
 
   (* Add the equality [nat_r = bv2nat(bv_r)]. *)
   let add_bv2nat ~ex nat_r bv_r t =
-    match BitvNormalForm.normal_form bv_r with
```
Halbaroth, line 449: What is the motivation of this change? If I understand well, you replace the normal form of `Bitv_rel` by the normal form of `Bitv`?

Halbaroth (approved):

src/lib/reasoners/bitv_rel.ml
bclement-ocp: (Sorry, realized I have left this comment as "pending" — hopefully it still makes some sense…)

It's not really useful to compute the `BitvNormalForm` here, the loop in `add_bv2nat_abstract` already processes this and it's simpler.

The `BitvNormalForm` distinguishes between constants / variables (concatenation of constants and exactly one variable) / composites (arbitrary concatenations), which is useful when doing reductions. But here we don't really care about this distinction, we just need to be able to iterate on the components of the concatenation to decompose the `bv2nat` expression — I think I failed to recognise this when writing the code initially and it ended up more convoluted than necessary.

PR #1321, opened 2025-04-23
Centralize handling of disjunction order in `mk_or`

bclement-ocp:
Clauses have an additional `is_impl` boolean flag tracking whether they originated from an implication. This flag is used in two places:

 - To determine the order in which the Tableaux solver decides on literals;

 - In the `resolution_of_disj` function used for backward/forward triggers (also only used by the Tableaux solver).

The logic for determining the order between disjunction operands used to be partially in the `mk_or` smart constructor and partially in the Tableaux solver itself. This patch moves all of that logic into the `mk_or` smart constructor.

Note that the CDCL solver currently ignores this order and always sorts disjunction arguments (and ignores `is_impl`) using `Expr.compare` after flattening, but this makes it easier to bring both solvers in line in the future.

### DIFF
--- a/src/lib/reasoners/fun_sat.ml
+++ b/src/lib/reasoners/fun_sat.ml
@@ -79,11 +79,11 @@ module Make (Th : Theory.S) = struct
         if Options.get_no_decisions_on_is_empty () then delta, []
         else
           List.partition
-            (fun (a, _,_,_) -> Options.get_can_decide_on a.E.origin_name) delta
+            (fun (a, _,_) -> Options.get_can_decide_on a.E.origin_name) delta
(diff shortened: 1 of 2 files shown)


bclement-ocp:
> Note that the CDCL solver currently ignores this order and always sorts disjunction arguments (and ignores is_impl) using Expr.compare after flattening, but this makes it easier to bring both solvers in line in the future.

This seems maybe false according to benchmarks. Marking as draft until I investigate.

bclement-ocp:
OK; I now think this simplification attempt is misguided in its current incarnation. Closing for now.

PR #1340, opened 2026-03-31
Bump minimal version of cmdliner to 2.0

Halbaroth:
Changes:
- Bump the minimal version of `cmdliner` to 2.0.
- Bump the minimal version of `js_of_ocaml` to 6.2.0. This version is
  compatible both with cmdliner < 2.0 and cmdliner 2. The latest release
  of js_of_ocaml contains a bug
  (see https://github.com/ocsigen/js_of_ocaml/issues/2209). 
  We must use js_of_ocaml 6.2.0 until the next release of the compiler.
  Unfortunately, they dropped support for OCaml compiler < 4.13, which means we cannot compile our JavaScript artifact  for the old versions of the compiler.
- Bump `ocp-indent` to 1.9.0 as 1.8.1 depends on `cmdliner` < 2.0.
- Update OCaml versions in the CI (4.14.1 -> 4.14.3 and 5.0 -> 5.4.0).
- Bump the nixpkgs version to 25.11. The release `24.05` didn't contain this patch https://github.com/NixOS/nixpkgs/commit/69ca7055ae35b751e62099e51200fcb280b8f448.

### DIFF
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -26,15 +26,15 @@
         "version": "dev"
     },
     "nixpkgs": {
-        "branch": "release-24.05",
+        "branch": "release-25.11",
         "description": "Nix Packages collection",
         "homepage": null,
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "b260a95463a8f5f9094527c05329d79527e5a4aa",
-        "sha256": "0qlnwrha950kflbnfsl39mrdkkll3663bssjlxrrm7w1qnbbbp1n",
+        "rev": "e4f687fc1caadcb540453a2ce7ea9c9f2763434d",
+        "sha256": "089xw68v2sk4i3pra095qaqqs31j5b482j4bl7lw9qwrl9bni91n",
         "type": "tarball",
-        "url": "https://github.com/NixOS/nixpkgs/archive/b260a95463a8f5f9094527c05329d79527e5a4aa.tar.gz",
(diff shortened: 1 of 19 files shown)


nix/sources.json
```diff
@@ -26,15 +26,15 @@
         "version": "dev"
     },
     "nixpkgs": {
-        "branch": "release-24.05",
+        "branch": "nixos-unstable",
```
bclement-ocp: I'd prefer if we kept on a stable branch if possible -- I think 25.11 should have cmdliner 2?

bclement-ocp:
I think the opam lockfiles need some care :sob:

nix/sources.json
Halbaroth: Sadly, https://search.nixos.org/packages?channel=25.11&query=cmdliner

I will override the package.

Halbaroth:
Hallelujah! It is ready for a second pass :)
There is two commits:
- Update the workflows with many fixes for the CI
- Bump the minimal version of cmdliner to 2.0

src/bin/js/dune
```diff
@@ -31,7 +31,7 @@
  )
  (modules worker_js options_interface)
  (modes byte js)
- (js_of_ocaml (flags --no-source-map +toplevel.js +dynlink.js))
+ (js_of_ocaml (flags +toplevel.js +dynlink.js))
```
bclement-ocp, line 34: I'm not going to ask 🙈

bclement-ocp (approved):
🚀

PR #1342, opened 2026-04-01
Add support for ae.pow_int, ae.pow_real

### LINKED ISSUE #1341: Access to exponentiation with operation the SMT-LIB input syntax
Hi, in SPARK we followed Why3 and moved from Alt-Ergo native input file format to the SMT-LIB input format when adopting Alt-Ergo-2.6. However, we noticed that it is not possible to access the exponentiation operation with the SMT-LIB syntax and we have some regressions because of this 😞 

Is there any deep reason for not exposing the exponentiation operation via the SMT-LIB syntax? I crafted a tentative patch for the Dolmen frontend to expose the `**` operation, which is the syntax recently adopted by https://smt-lib.org/theories-Ints.shtml. With this patch alt-ergo seems to work, which would suggest that the lack of support was only a limitation in the Dolmen library 😕 

Here is a sample input file:
```
(set-logic ALL)
(set-info :smt-lib-version 2.6)
(assert
  (not (forall ((base Int)) (=> (= base 65536) (= base (** 2 16))))))
(check-sat)
```

I did some basic sanity tests:
* changing the operands of `**` to incorrect values turns the answer from `unsat` to `unknown`
* introducing typing errors, e.g. by replacing ``2 with `2.0` causes a nice typing error to appear
* changing `Int` to `Real` and suffixing all numbers with `.0` still gives `unsat`

[dolmen-power.patch](https://github.com/user-attachments/files/26401704/dolmen-power.patch)

bclement-ocp:
Fixes #1341

Note that in the future we will be able to use the "**" operator from SMT-LIB once support lands in Dolmen and a new release of Dolmen is made.

### DIFF
--- a/src/lib/frontend/translate.ml
+++ b/src/lib/frontend/translate.ml
@@ -426,6 +426,10 @@ let smt_fpa_builtins =
       term_app env s (smt_round 53 1074)
     | Id { ns = Term ; name = Simple "ae.float128" } ->
       term_app env s (smt_round 113 16494)
+    | Id { ns = Term ; name = Simple "ae.pow_int" } ->
+      term_app env s DE.Term.Int.pow
+    | Id { ns = Term ; name = Simple "ae.pow_real" } ->
+      term_app env s DE.Term.Real.pow
     | Dl.Typer.T.Id id -> begin
         match DStd.Id.Map.find_exn id other_builtins env s with
         | e -> e


Halbaroth:
Can you rebase on master please to run the CI again?

bclement-ocp:
I was on vacation last week — rebased now (also rebased #1343).

Halbaroth (approved):

PR #1344, opened 2026-04-01
Bump minimal version of cmdliner to 2.0

Halbaroth:
See #1340


Halbaroth:
I tested `nix-shell` and it works as expected :)

bclement-ocp (approved):

Halbaroth:
It looks like there are some workflows left over from the old CI. It seems like a bug in GitHub.

PR #1346, opened 2026-04-27
Create alt-ergo-js package

Halbaroth:
This package installs:
- The byte code version of Alt-Ergo as `bin/alt-ergo.bc`,
- The JavaScript version of Alt-Ergo in `share/alt-ergo-js/alt-ergo.js`,
- The JavaScript worker and the example in `share/alt-ergo-js` too.

You can test the result as follows:
```console
make js
python -m http.server -d www
```

This commit removes the `js-deps` target as we can run `opam install . --deps-only`.


bclement-ocp (approved):
Should the opam lockfile be updated?

Halbaroth:
Sorry, I didn't see your message and I noticed that the documentation of the JavaScript artifacts were outdated after reading the last message of Claude. I updated the documentation about the installation and integrated it in the Sphinx documentation. 

I don't merge because you didn't review this version of the code. Sorry for that...

src/bin/common/dune
```diff
@@ -4,7 +4,7 @@
 )
 
 (library
- (package alt-ergo)
+ (public_name alt-ergo.common)
```
bclement-ocp, line 7: I forgot to note --- we don't need `package alt-ergo` because that is implied by `public_name alt-ergo.common`, is that right?

bclement-ocp (approved):
The changes looks OK but I still have the same question about opam lockfiles :)

src/bin/common/dune
Halbaroth: Indeed, and dune complains:
```
File "src/bin/common/dune", line 7, characters 10-18:
7 |  (package alt-ergo)
              ^^^^^^^^
Error: This library has a public_name, it already belongs to the package
alt-ergo
```

Halbaroth:
I tested the lock file as follows:
```console
opam sw create foo 4.14.3
opam pin -n .
opam install . --locked
dune build
``` 
It works, so we don't need to update the lock file.

PR #1349, opened 2026-04-29
Use local exception to early exit few functions

### LINKED ISSUE #1348: Uncaught `Exit` exception in Alt-Ergo JS
While working on #1325, I found a bug on both `v2.6.x` and `next`. 

To reproduce the bug:
1. Save the following file as `input.psmt2`: [input.txt](https://github.com/user-attachments/files/27204001/input.txt)
3. Build Alt-Ergo with jsoo in **release** mode:   
    ```console
    dune build --release _build/default/src/bin/js/main_text_js.bc.js
    ln -sf _build/default/src/bin/js/main_text_js.bc.js ./alt-ergo.js
     ```
3. Run Alt-Ergo with `node` on `input.psmt2`: `node ./alt-ergo.js ./input.psmt2`

Alt-Ergo immediately stops without error and exit status 0. If you run it with native Alt-Ergo, it will loop.

Alt-Ergo raises `Exit` to early leave some functions and the frontend catches it to invoke `Stdlib.exit`:
```ocaml
  let handle_exn st bt = function
    | Dolmen.Std.Loc.Syntax_error (_, `Regular msg) ->
      recoverable_error "%t" msg; st
    | Util.Timeout ->
      Printer.print_status_timeout None None None None;
      exit_as_timeout ()
    | Errors.Error e ->
      recoverable_error "%a" Errors.report e;
      st
    | Exit -> raise (Exit_with_code 0)
    | _ as exn -> Printexc.raise_with_backtrace exn bt
  in
```
Unfortunately, `Stdlib.Exit` is used by several functions and in particular in `Satml_types.mk_or`, we have the following exception handler:
```ocaml
  let rec mk_or hcons l =
    try
      ...
      let delta_u = match delta_inv with
        | [] -> delta_inv
        | e::l ->
          let _, delta_u =
            List.fold_left
              (fun ((c,l) as acc) e ->
                 if complements c e then raise Exit;
                 if equal c e then acc
                 else (e, e::l)
              )(e,[e]) l
          in
          delta_u
      in
      ...
    with Exit -> vrai
```
On both native and bytecode, the handler catches `Exit` as expected but jsoo 6.2.0 fails to translate this code correctly in JS. You can apply this patch if you don't believe me ;)
```patch
diff --git a/src/lib/structures/satml_types.ml b/src/lib/structures/satml_types.ml
index 7e87404bcf..e6b76e70de 100644
--- a/src/lib/structures/satml_types.ml
+++ b/src/lib/structures/satml_types.ml
@@ -830,6 +830,8 @@ module Flat_Formula : FLAT_FORMULA = struct
         try Some (common, List.rev_map (diff_list common) ands)
         with Not_included -> assert false
 
+  exception Kamoulox
+
   let rec mk_or hcons l =
     try
       let so, nso =
@@ -855,7 +857,7 @@ module Flat_Formula : FLAT_FORMULA = struct
           let _, delta_u =
             List.fold_left
               (fun ((c,l) as acc) e ->
-                 if complements c e then raise Exit;
+                 if complements c e then raise Kamoulox;
                  if equal c e then acc
                  else (e, e::l)
               )(e,[e]) l
@@ -883,7 +885,12 @@ module Flat_Formula : FLAT_FORMULA = struct
         | Some (com,ands) ->
           let ands = List.rev_map (mk_and hcons) ands in
           mk_and hcons ((mk_or hcons ands) :: com)
-    with Exit -> vrai
+    with Exit | Kamoulox -> vrai
+
+  let mk_or x y =
+    match mk_or x y with
+    | exception Kamoulox -> vrai
+    | r -> r
 
   (* translation from E.t *)
```

Halbaroth:
Using `Stdlib.Exit` to early exit loops is acceptable if we have backtraces. Unfortunately, backtraces are not recorded without instrumentating code for jsoo. Debugging #1348 was a nightmare because of this. 

I believe that we should use local exceptions in this case:
```ocaml
let foo = 
  let exception Early_exit in 
  fun ... -> ...
```

As a bonus, this allows us to use more descriptive names for our exceptions.

### DIFF
--- a/src/bin/common/solving_loop.ml
+++ b/src/bin/common/solving_loop.ml
@@ -158,6 +158,13 @@ type solve_res =
 
 exception StopProcessDecl
 
+let parse_solver_name s =
+  match s with
+  | "tableaux" -> Some Util.Tableaux
+  | "cdcl" | "satml" -> Some Util.CDCL
+  | "cdcl_tableaux" | "satml_tableaux" | "default" -> Some Util.CDCL_Tableaux
+  | _ -> None
+
 let interactive_prompt st =
   match D_loop.State.get D_loop.State.logic_file st with
   | { source = `Stdin; mode = (None | Some `Incremental); _ }
@@ -296,7 +303,6 @@ let process_source ?selector_inst ~print_status src =
     | Errors.Error e ->
       recoverable_error "%a" Errors.report e;
       st
-    | Exit -> raise (Exit_with_code 0)
     | _ as exn -> Printexc.raise_with_backtrace exn bt
   in
   let finally ~handle_exn st e =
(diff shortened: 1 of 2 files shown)


Halbaroth:
~~I add an extra patch to fix #1348. We cannot use jsoo 6.0.0 --> 6.3.2 and jsoo < 6.0.0 is not compatible with OCaml 4.14.3.~~

Halbaroth:
@bclement-ocp I removed the hotfix as we decided during the last dev meeting. We can close the PR or merge it, as you wish.

bclement-ocp (approved):
Looks good to me — I like the trick of scoping a `let exception` within the toplevel binding before an anonymous function!

Halbaroth:
> Looks good to me — I like the trick of scoping a let exception within the toplevel binding before an anonymous function!

It’s a consequence of the trauma from the last debugging session.

PR #1353, opened 2026-05-22
Make :reproducible-resource-limit standards-compliant

### LINKED ISSUE #1279: Make `reproducible-resource-limit` use steps
We need to implement a reproducible `reproducible-resource-limit` with steps.

_Originally posted by @Halbaroth in https://github.com/OCamlPro/alt-ergo/issues/1277#issuecomment-2615744369_

This should be part of an overhaul of the timeout/steps limit system. See also #1244

bclement-ocp:
Fixes #1279

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@
 
    The reproducible resource limit was previously using a time-based limit,
    which does not respect the reproducibility criterion outlined in the SMT-LIB
-   standard. It now defines a per-goal step limit, which apply in addition to
+   standard. It now defines a per-goal step limit, which applies in addition to
    the **global** step limit (defined with `--steps-bound`), if any.
 
    We encourage users of the global step limit (`--steps-bound`) to switch to
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,19 @@
 ## unreleased
 
+ - Added support for the `:reproducible-resource-limit` and
+   `--reproducible-resource-limit` options, following the SMT-LIB standard
+   (#1353)
+
+   The reproducible resource limit was previously using a time-based limit,
+   which does not respect the reproducibility criterion outlined in the SMT-LIB
+   standard. It now defines a per-goal step limit, which apply in addition to
+   the **global** step limit (defined with `--steps-bound`), if any.
+
+   We encourage users of the global step limit (`--steps-bound`) to switch to
+   the reproducible resource limit instead.
+
+## v2.6.3
+
  - Bump minimal cmdliner version to 2.0 (#1340)
 
 ## v2.6.2
--- a/src/bin/common/parse_command.ml
+++ b/src/bin/common/parse_command.ml
@@ -395,7 +395,7 @@ let mk_internal_opt
   `Ok(gc_policy)
 
 let mk_limit_opt age_bound fm_cross_limit timelimit_interpretation
-    steps_bound timelimit timelimit_per_goal
+    steps_bound reproducible_resource_limit timelimit timelimit_per_goal
   =
   let set_limit t d =
     match t with
@@ -421,7 +421,7 @@ let mk_limit_opt age_bound fm_cross_limit timelimit_interpretation
     Steps.set_steps_bound steps_bound;
     set_timelimit timelimit;
     set_timelimit_per_goal timelimit_per_goal;
-    `Ok()
+    `Ok Solving_loop.{ reproducible_resource_limit }
 
 let mk_output_opt
     produce_models objectives_in_interpretation unsat_core
@@ -563,7 +563,7 @@ let get_verbose_t =
   Arg.(value & flag & info ["v"; "verbose"] ~doc)
 
 let mk_opts file () () debug_flags ddebug_flags dddebug_flags backtrace
-    rule () halt_opt (gc) () () () () () () () () =
+    rule () halt_opt (gc) limits () () () () () () () =
   Debug.mk ~verbosity:1 debug_flags;
   Debug.mk ~verbosity:2 ddebug_flags;
   Debug.mk ~verbosity:3 dddebug_flags;
@@ -591,7 +591,7 @@ let mk_opts file () () debug_flags ddebug_flags dddebug_flags backtrace
     );
 
     Gc.set { (Gc.get()) with Gc.allocation_policy = gc };
-    `Ok (Some path)
+    `Ok (Some (Solving_loop.{ path; limits}))
   end
 
 let mk_output_channel_opt regular_output diagnostic_output =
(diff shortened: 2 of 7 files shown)


bclement-ocp:
Hm I actually want to try something a bit better for the implementation of `with_steps_limit`, let me update the PR.

bclement-ocp:
OK this version should be better @Halbaroth — I've moved the call to `with_steps_limit` down inside `solve`, which ensures that it's after the call to `reset_refs` (which calls `reset_steps`) otherwise weird things would be happening. I've also made sure that `with_steps_limit` always returns an `Error` if the limit was reached.

It's not that useful right now because we actually update the internal state of the solver through mutability so we ignore the result at the moment, but I think it should be a more robust way of checking for the limit in a follow-up PR.

bclement-ocp:
OK I messed something up apparently, I'll revisit this. Marking as draft for now.

Halbaroth:
Ok, I was reviewing it.

bclement-ocp:
Apologies — the previous version had a bug (it would raise an uncaught `Step_limit_reached` on a timeout) and trying to fix that bug broke all the tests :(

bclement-ocp:
OK, I went with a slightly simpler approach (no atomic token to detect when the limit is reached, because we don't actually need it for now) to just implement what should be needed for the `:reproducible-resource-limit` behavior from the SMT-LIB standard. I think this works now :)

src/lib/util/steps.mli
```diff
@@ -81,6 +81,19 @@ val incr_cs_steps : unit -> unit
 (** Disables the step limit during the execution of the continuation. *)
 val apply_without_step_limit : (unit -> 'a) -> 'a
 
+(** [with_step_limit limit ~scope] calls scope with the provided [limit]. The
+    [limit] is local: only steps performed during the call to [scope] are
+    counted towards the [limit] (in particular, any steps that have already been
+    recorded are *NOT* counted towards the [limit]).
+
+    [with_step_limit] is not guaranteed to raise if the limit is reached, as the
+    [Util.Step_limit_reached] exception might be catched by [scope].
```
hra687261:
```suggestion
    [Util.Step_limit_reached] exception might be caught by [scope].
```

tests/cram/reproducible-resource-limit.t
```diff
@@ -0,0 +1,113 @@
...
+  
+  unknown
+  (:reason-unknown (:step-limit 2))
+  
+  unknown
+  (:reason-unknown (:step-limit 2))
+  
+  unsat
+
+The global step limit cannot be overriden and always apply.
```
hra687261:
```suggestion
The global step limit cannot be overridden and always applies.
```

src/lib/util/steps.ml
```diff
@@ -169,6 +172,45 @@ let set_steps_bound i =
   if get_steps () > i  && i >= 0 then invalid_arg "Steps.set_steps_bound";
   steps_bound := i
 
+let with_step_limit limit ~scope =
+  if limit < 0 then invalid_arg "with_step_limit: limit must be nonnegative";
+  (* Note that we may not guaranteed that calling [with_step_limit] with a
+      given value inside two distinct context will result in exactly the same
```
hra687261:
```suggestion
      given value inside two distinct contexts will result in exactly the same
```

CHANGES.md
```diff
@@ -1,5 +1,19 @@
 ## unreleased
 
+ - Added support for the `:reproducible-resource-limit` and
+   `--reproducible-resource-limit` options, following the SMT-LIB standard
+   (#1353)
+
+   The reproducible resource limit was previously using a time-based limit,
+   which does not respect the reproducibility criterion outlined in the SMT-LIB
+   standard. It now defines a per-goal step limit, which apply in addition to
```
hra687261:
```suggestion
   standard. It now defines per-goal step limits, which apply in addition to
```

src/bin/common/solving_loop.ml
```diff
@@ -203,6 +211,11 @@ let process_source ?selector_inst ~print_status src =
       let ftdn_env = FE.init_env ?selector_inst used_context in
       let () =
         try
+          (* At the moment we ignore the [Error] case here: the unknown reason
+             should have already been set internally by the solver when the
+             [Util.Step_limit_reached] exception was raised. *)
```
hra687261, line 216: Which error case?

src/bin/common/solving_loop.ml
```diff
@@ -203,6 +211,11 @@ let process_source ?selector_inst ~print_status src =
...
+          let ( let& ) f scope = f ~scope in
+          let& () = Steps.with_step_limit limit in
```
hra687261, line 218: The operator seems to be only used once (right after its declaration), I don't know if it's worth keeping it, inlining it would make the code more readable.

src/bin/common/parse_command.ml
```diff
@@ -886,6 +886,12 @@ let parse_limit_opt =
     Arg.(value & opt int (get_steps_bound ()) &
          info ["S"; "steps-bound"] ~docv ~doc) in
 
+  let reproducible_resource_limit =
+    let doc = "Set the reproducible resource limit." in
```
hra687261, line 890: Maybe give a bit more details here? (e.g. that what we measure is the number of steps and that if the limit is 0 then there is no limit)

hra687261 (approved):
Looks good to me, other than some typos/superficial remarks.

src/bin/common/solving_loop.ml
bclement-ocp: It'd increase indentation though ;)

It is the "resource binding" operator from [memprof-limits](https://guillaume.munch-maccagnoni.name/software/ocaml/memprof-limits.0.3.0/memprof-limits/Memprof_limits/Resource_bind/index.html) — I think it is useful to have a way to explicitly say "we are creating a scope where resources (here a fixed number of steps) are acquired" rather than a simple function call, but did not want to put it in a module of its own for a single use. I can add a short documentation to explain this.

src/bin/common/solving_loop.ml
bclement-ocp: Old comment from a previous implementation, I'll remove it.

CHANGES.md
bclement-ocp: I think "per-goal step limits" is ambiguous (it hints at there being possibly different (numerical) limits for each goal), so I prefer my phrasing.

hra687261: Good point, in that case:
```suggestion
   standard. It now defines a per-goal step limit, which applies in addition to
```

src/bin/common/solving_loop.ml
hra687261: Ok, I was not familiar with it, I agree it is useful, I don't know if there is a better place to put it that within the code. And yes a comment would be helpful :)

PR #1357, opened 2026-06-05
Associate concrete values to unconstrainted uninterpreted symbols

hra687261:
This is an attempt at a quick fix for the issue reported in https://github.com/formalsec/smtml/issues/623.

The idea is simply to associate a default value to unconstrained uninterpreted symbols (of type int, real, bitv and bool) in the models, I would assume that this should be done earlier and elsewhere (maybe by the theories themselves) but I am not sure.

This notably concerns uninterpreted symbols that are declared but not involved in the problems, or those (booleans) that appear in tautologies and for which concrete values are not decided.

EDIT: if this is the correct approach, it should be done for the remainig types as well.

### DIFF
--- a/src/lib/reasoners/theory.ml
+++ b/src/lib/reasoners/theory.ml
@@ -869,18 +869,19 @@ module Main_Default : S = struct
       ()
 
   let extract_concrete_model ~declared_ids env =
-    let env, defaults =
+    let model = ModelMap.empty ~suspicious:false declared_ids in
+    let env, model =
       List.fold_left
-        (fun (env, defaults) ((id, arg_tys, ret_ty) as ty_id) ->
+        (fun (env, model) ((id, arg_tys, ret_ty) as ty_id) ->
            match arg_tys with
            | [] ->
              let t = E.mk_term (Sy.name (Hstring.view id)) [] ret_ty in
-             add_term env t ~add_in_cs:true, defaults
+             add_term env t ~add_in_cs:true, model
            | _ :: _ ->
              let t = E.fresh_name ret_ty in
              let env = add_term env t ~add_in_cs:true in
-             env, (ty_id, t) :: defaults)
-        (env, []) declared_ids
+             env, ModelMap.set_free_defval ty_id t model)
+        (env, model) declared_ids
     in
     let { gamma_finite; assumed_set; objectives; _ }, _ =
       do_case_split_aux env ~for_model:true
@@ -888,8 +889,7 @@ module Main_Default : S = struct
     lazy (
       CC_X.extract_concrete_model
         ~prop_model:assumed_set
-        ~declared_ids
-        ~defaults
+        ~model
         gamma_finite
     ), objectives
 
--- a/src/lib/reasoners/theory.ml
+++ b/src/lib/reasoners/theory.ml
@@ -869,6 +869,16 @@ module Main_Default : S = struct
       ()
 
   let extract_concrete_model ~declared_ids env =
+    let env =
+      List.fold_left
+        (fun env (id, arg_tys, ret_ty) ->
+           match arg_tys with
+           | _ :: _ -> env
+           | [] ->
+             let t = E.mk_term (Sy.name (Hstring.view id)) [] ret_ty in
+             add_term env t ~add_in_cs:true)
+        env declared_ids
+    in
     let { gamma_finite; assumed_set; objectives; _ }, _ =
       do_case_split_aux env ~for_model:true
     in
--- a/src/lib/reasoners/theory.ml
+++ b/src/lib/reasoners/theory.ml
@@ -869,19 +869,18 @@ module Main_Default : S = struct
       ()
 
   let extract_concrete_model ~declared_ids env =
-    let model = ModelMap.empty ~suspicious:false declared_ids in
-    let env, model =
+    let env, defaults =
       List.fold_left
-        (fun (env, model) ((id, arg_tys, ret_ty) as ty_id) ->
+        (fun (env, defaults) ((id, arg_tys, ret_ty) as ty_id) ->
            match arg_tys with
            | [] ->
              let t = E.mk_term (Sy.name (Hstring.view id)) [] ret_ty in
-             add_term env t ~add_in_cs:true, model
+             add_term env t ~add_in_cs:true, defaults
            | _ :: _ ->
              let t = E.fresh_name ret_ty in
              let env = add_term env t ~add_in_cs:true in
-             env, ModelMap.set_free_defval ty_id t model)
-        (env, model) declared_ids
+             env, (ty_id, t) :: defaults)
+        (env, []) declared_ids
     in
     let { gamma_finite; assumed_set; objectives; _ }, _ =
       do_case_split_aux env ~for_model:true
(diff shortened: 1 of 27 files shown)


Halbaroth:
I cite you: 
> It would be helpful to add the error message/raised exception (and maybe even the stack_trace) to the issue.

Did you find the error?

Alt-Ergo produces an abstract value in models if there is no constraint on it:
```smt2
(set-logic ALL)
(declare-const x Int)
(check-sat)
(get-model)
```
```smt2
unknown
(
  (define-fun x () Int (as @a0 Int))
)
```

hra687261:
> Did you find the error?

> Alt-Ergo produces an abstract value in models if there is no constraint on it:

That's the "issue", we expect concrete values in models in smtml and that's what causes the error. 
(I would assume that a concrete interpretation is also expected smt-lib compliance wise, but idk)

Halbaroth:
Ok!

> EDIT: if this is the correct approach, it should be done for the remainig types as well.

I agree and for the ADT, you need to order constructors to ensure the termination of your function. I wrote a similar code in `Adt_rel` using the module `Nest`.

bclement-ocp:
The way we are doing this kind of made sense when we were creating abstract values for these variables, but if we want to build actual values, I think it would be better to use the existing infrastructure for value generation in models.

I think this should just be a matter of explicitly adding the `declared_ids` (using `add_term`) to the theory at the start of `extract_concrete_model` (before doing the case splits, which is where the generation of the model values actually happen).

ETA: likely we should add the `declared_ids` to the theory iff they have no arguments, and create graphs in the `ModelMap` only for actual functions with arguments, but we should also have a dummy (fresh, not abstract) variable for the default value of the functions, and use that in the `ModelMap` instead of forcing abstract values. That dummy variable also needs to be created in `extract_concrete_model` before doing the splits, or we won't assign an actual value to it.

src/lib/reasoners/theory.ml
```diff
@@ -869,6 +869,16 @@ module Main_Default : S = struct
       ()
 
   let extract_concrete_model ~declared_ids env =
+    let env =
+      List.fold_left
+        (fun env (id, arg_tys, ret_ty) ->
+           match arg_tys with
+           | _ :: _ -> env
```
bclement-ocp: I think in this case we still need to create a dummy value and pass it to the `ModelMap` (this means changing the interface of `CC_X.extract_concrete_model`, maybe simply passing a `ModelMap` directly instead of passing the `declared_ids`).

Otherwise for this input:

```smtlib2
(set-logic ALL)
(declare-fun f (Int) Int)
(check-sat)
(get-model)
```

we have the following output:

```smtlib2
unknown
(
  (define-fun f ((_arg_0 Int)) Int (as @a0 Int))
)
```

but we'd like to also have a concrete value there.

hra687261: Done, but currently using a list of associating defaults (fresh exprs) to function symbols, I'll try to use a `ModelMap` directly

bclement-ocp: I think you can move the `suspicious` test to the `modelMap`  (at the time of printing, it is "suspicious" if it has a definition for one of the "suspicious" functions in the `values` map), it should make this easier.

bclement-ocp: Actually I'm not sure that works, but simply providing a function to set the `suspicious` flag in `Uf` should be enough (with the goal being to avoid creating the `ModelMap` with "wrong" abstract names followed by patching it).

hra687261: That lead to a change of ids for some of the abstract values in the old models, but it seems to work correctly.

bclement-ocp: Sorry, I think I was not clear — the idea would be to build the model directly with the right `Free` variables (in the current implementation, we still create a model with unused abstract variables, then patch it afterwards — the patching just occurs earlier). This requires slightly changing the `ModelMap` API; instead of having `empty` be a function that takes a `suspicious` flag and a list of declared IDs, it would just be a constant ("truly empty" model) and then for each functional declared ID we'd create the "free" graph for that ID during the iteration.

That said, I forgot that we'd need to update the model again in the `UF` module, as you figured out in your patch! I think we can keep the code this way (with the changes suggested above, because we should not create those abstract variables during `ModelMap` creation — those are useless now), but I think the following would be a bit nicer:

 - In `Theory.extract_concrete_model`, convert the `declared_ids` into a list of pairs `(Id.typed, Expr.t)` where the `Expr.t` is the default for the `Id.typed` and is added with `add_term`. (This is similar to your first approach). We only should need to keep the functional values in this list, because the non-functional values are guaranteed to be added in the `Uf` module now.
 - In `Uf.extract_concrete_model`, initialise the `ModelMap` with the `model_repr_of_term` for the `Expr.t` in that list, before calling `compute_concrete_model_of_val`.

What do you think?

hra687261: So if I understand correctly, the main change is to not create a model in  ` Theory.extract_concrete_model` and only create a list of `defaults` (of type `(Id.typed, Expr.t)` mapping uninterpreted function symbols to fresh symbols), the list is passed to `Uf.extract_concrete_model` which will update the model by setting the actual concrete values of the symbols in `defaults` directly? (In addition to the changes to the ModelMap API) 
That looks good to me.

bclement-ocp: Yes exactly — the `ModelMap` would be created in `Uf.extract_concrete_model` from the concrete values of the symbols in `defaults` at that time.

hra687261: It should be good now.

src/lib/structures/modelMap.ml
```diff
@@ -144,13 +144,14 @@ let find k {values; _ } =
...
+let set_free_defval sy v { values; suspicious } =
   let values =
-    List.fold_left
-      (fun values ((_, _, ret_ty) as sy) ->
-         P.add sy (Free (Expr.mk_abstract ret_ty)) values
-      )
-      P.empty declared_ids
+    match P.find_opt sy values with
+    | None | Some (Free _) -> P.add sy (Free v) values
+    | Some (C _) -> values
```
bclement-ocp: I think it's fine for this to be called `add` rather than `set` now? We should not be replacing anything in the model anymore (in fact, I'd argue we should raise an `internal_error` if we do).

bclement-ocp (changes requested):
Looks good overall, just one last small comment 👍 

Can you add a test for the function case as well (e.g. the test I posted in the other thread)?

src/lib/structures/modelMap.ml
hra687261: Indeed! Done!

hra687261:
The test [complete_2.default.smt2](https://github.com/hra687261/alt-ergo/blob/37fc178263d196ac0b6769e6dafa7e29f823018a/tests/models/complete_2.default.smt2) tests that (that's how I noticed the error in https://github.com/OCamlPro/alt-ergo/pull/1357/changes/37fc178263d196ac0b6769e6dafa7e29f823018a#diff-f910108a0334845ef177fe86dbc4277899e70105862eab3328b9b0f31b175ed3)

bclement-ocp (approved):

PR #1358, opened 2026-06-11
Support parsing for SMT-LIB FPA Theory literals

hra687261:
This PR is a first step towards supporting the SMT-LIB standard Floating-Point Arithmetic theory, it adds a type for floats in Alt-Ergo as well as a representation for FPA literals (+/- inf, +/- zero, Nan and finite literal floating-point numbers).
The support for the reasoning over the operations/symbols of the theory will be added in a separate PR.

Out of prudence, it might be better to put this support behind a command line option, at least in the meant time, wdyt?


bclement-ocp:
> Out of prudence, it might be better to put this support behind a command line option, at least in the meant time, wdyt?

I'm fine with having an option for this, but also it should not impact any existing features (we just accept more files that we previously rejected).

src/bin/common/parse_command.ml
```diff
@@ -1358,6 +1359,11 @@ let parse_theory_opt =
     Term.term_result' term
   in
 
+  let smt_lib_fpa =
+    let doc = "Activate support for the SMT-LIB standard theory of \
+               floating-point arithmetic." in
+    Arg.(value & flag & info ["smt-lib-fpa"] ~docs ~doc) in
+
```
bclement-ocp: We are trying to avoid having a bazillion of boolean flags lying around — can you add this to the existing support for the `--enable-theories` flag (see `theories.ml`)? This is already how we deal with enabling/disabling the ADT and AC theories.

(I suggest calling this theory something along the lines of `--enable-theories smt.float` or `--enable-theories smt.floating_point` as that is the [official name](https://smt-lib.org/theories-FloatingPoint.shtml) of the theory, to avoid conflicts with the existing `fpa` theory)

src/lib/frontend/translate.ml
```diff
@@ -900,6 +901,17 @@ let rec mk_expr
...
+          | B.Float Plus_infinity { e; s } when Options.get_smt_lib_fpa () ->
+            E.fp Sy.Plus_infinity e s
+          | B.Float Minus_infinity { e; s } when Options.get_smt_lib_fpa () ->
+            E.fp Sy.Minus_infinity e s
+          | B.Float Plus_zero { e; s } when Options.get_smt_lib_fpa () ->
+            E.fp Sy.Plus_zero e s
+          | B.Float Minus_zero { e; s } when Options.get_smt_lib_fpa () ->
+            E.fp Sy.Minus_zero e s
+          | B.Float NaN { e; s } when Options.get_smt_lib_fpa () ->
+            E.fp Sy.NaN e s
```
bclement-ocp: I think it would be nicer to do: `B.Float cst when Options.get_smt_lib_fpa ()` and then match on `cst`. This way we benefit from pattern-matching exhaustiveness!

src/lib/frontend/translate.ml
```diff
@@ -1380,6 +1392,40 @@ let rec mk_expr
...
               | RoundTowardZero, _ -> mk_rounding ToZero
 
+              | Fp { e; s }, [sign_t; exp_t; sig_t]
+                when Options.get_smt_lib_fpa () ->
+                let bv_z DE.{ term_descr; _ } =
+                  (* TODO: does Dolmen guarantee that the term is a bitvector
+                     literal? *)
+                  match term_descr with
+                  | DE.Cst { builtin = B.Bitv (Binary_lit bs); _ } ->
+                    Z.of_string ("0b" ^ bs)
```
```suggestion
                    Z.of_string_base 2 bs
```

src/lib/frontend/translate.ml
```diff
@@ -1380,6 +1392,40 @@ let rec mk_expr
...
+                  | _ -> invalid_app_term ()
+                in
+                let neg = Z.equal (bv_z sign_t) Z.one in
+                let biased_exp = Z.to_int (bv_z exp_t) in
+                let trail = bv_z sig_t in
+                let max_exp = (1 lsl e) - 1 in
+                let fp_val =
+                  (* TODO: doesn't Dolmen avoid that? *)
```
bclement-ocp: What is "that"?

src/lib/frontend/translate.ml
```diff
@@ -1380,6 +1392,40 @@ let rec mk_expr
               | RoundTowardNegative, _ -> mk_rounding Down
               | RoundTowardZero, _ -> mk_rounding ToZero
 
+              | Fp { e; s }, [sign_t; exp_t; sig_t]
+                when Options.get_smt_lib_fpa () ->
+                let bv_z DE.{ term_descr; _ } =
+                  (* TODO: does Dolmen guarantee that the term is a bitvector
+                     literal? *)
```
bclement-ocp: No, these can be arbitrary bit-vectors! For instance the following SMT-LIB file is accepted by the standard:

```smt-lib
(set-logic ALL)

(declare-const x (_ BitVec 1))
(declare-const y (_ BitVec 2))
(declare-const z (_ BitVec 2))

(declare-const f (_ FloatingPoint 2 3))

(assert (= f (fp x y z)))

(check-sat)
```

We'll need a generic `Fp` symbol / operator in addition to the one for constant values, and a theory to represent these constants (I'd suggest to extract the `fp_val` type into its own module, to be used both in symbols and for the theory — like we use `Z.t` etc. both for integer symbols and in semantic values of the arithmetic theory).

I think it makes sense to do constant-propagate the `fp` value because there are no other ways to write float literals, but that should happen in the `E.fp` smart constructor, not here.

(As this PR only introduces support for float literals, it is fine for that code to stay here for the time being, but we should have a comment explaining this and move/adapt the code later)

src/lib/structures/ty.ml
```diff
@@ -234,6 +242,7 @@ let rec matching s pat t =
     matching (matching s ta1 tb1) ta2 tb2
   | Tint , Tint | Tbool , Tbool | Treal , Treal -> s
   | Tbitv n , Tbitv m when n=m -> s
+  | Tfloat _ , Tfloat _ -> s
```
bclement-ocp: This needs to check that the two types are actually the same (same `eb` and `sb`).

src/lib/structures/ty.ml
```diff
@@ -185,6 +188,9 @@ let rec compare t1 t2 =
   | Tbool, _ -> -1 | _, Tbool -> 1
 
   | Tbitv sz1, Tbitv sz2 -> Int.compare sz1 sz2
+  | Tbitv _, _ -> -1 | _, Tbitv _ -> 1
+
+  | Tfloat _, Tfloat _ -> 0
```
bclement-ocp: This needs to compare the float parameters `eb` and `sb`.

src/lib/structures/expr.ml
```diff
@@ -991,6 +997,8 @@ let real r =
 
 let bitv bt ty = mk_term (Sy.bitv bt) [] ty
 
+let fp v eb sb = mk_term (Sy.Fp v) [] (Ty.Tfloat (eb, sb))
```
bclement-ocp: I think this should be called `float` instead of `fp`, and `fp` should be added as a function that takes 3 bit-vector terms (and forwards to `float` if those terms are bit-vector literals — usually we would not do this, and do the simplification only in the `make` function of the "float" Shostak theory, as we do e.g. for `Real_of_int`, but given that there are no other ways to write fp literals, I think it's fine to do this simplification in the term constructors — although it will impact matching on `fp` terms…).

(See also comment in `translate.ml`; this can be done in a follow-up PR)

bclement-ocp (changes requested):

src/bin/common/parse_command.ml
hra687261: Done. (On the other hide I can't find the official names of the theory in the link you sent?)

src/lib/frontend/translate.ml
hra687261: Agreed, done! But unfortunately, FPA literals are in the same algebraic type as other FPA builtin symbols, so we won't have exhaustiveness checks for the literal builtins (unless we add all symbols). 
Maybe this should be changed in dolmen? I can open a PR, wdyt @Gbury ?

src/lib/frontend/translate.ml
hra687261: The translation from FP literals in their format with bit-vectors to the FP literals NaN/Inf, as is done below those line. I took a quick look and Dolmen does not do that, which I assume was a deliberate choice to be able to print the same thing that was parsed (cc @Gbury ). 
Now that I think of it, instead of doing that at the translation level, there should probably be a proper semantic rewriting somewhere (or just an equality propagation as an axiom, assuming its easy to write such an axiom).

src/lib/structures/ty.ml
hra687261: Done.

src/lib/structures/ty.ml
hra687261: Done.

src/bin/common/solving_loop.ml
```diff
@@ -548,6 +548,13 @@ let process_source ?selector_inst ~print_status src =
         | None -> print_wrn_opt ~loc ~name "integer" value; st
         | Some i -> set_steps_bound i st
       end
+    | ":smt-lib-fpa", Symbol { name = Simple "true"; _ } ->
+      (* TODO: while this "works" for now, its not correct, activating the
+         smt-lib fpa theory should tell AE to load its prelude, but loading
+         preludes is done before these in-file options are processed, so
+         supporting a proper in-file option setting would require making sure
+         that preludes are loaded after parsing these options.  *)
+      Options.set_smt_lib_fpa true; st
```
bclement-ocp: I think we should drop the option (and maybe just keep a comment explaining why it can't be set here) in that case; we shouldn't have a partially working option.

src/lib/frontend/translate.ml
```diff
@@ -900,6 +901,18 @@ let rec mk_expr
...
+          | B.Float cst when Options.get_smt_lib_fpa () ->
+            begin match cst with
+              | Plus_infinity { e; s } -> E.fp Sy.Plus_infinity e s
+              | Minus_infinity { e; s } -> E.fp Sy.Minus_infinity e s
+              | Plus_zero { e; s } -> E.fp Sy.Plus_zero e s
+              | Minus_zero { e; s } -> E.fp Sy.Minus_zero e s
+              | NaN { e; s } -> E.fp Sy.NaN e s
+              | _ ->
+                unsupported "Constant Floating-Point Arithmetic literal %a"
+                  DE.Term.print term
```
bclement-ocp, line 921: Please list the other Float symbols explicitly to have an exhaustive match :) This way we get a warning if new constructors are added in a new version of Dolmen and we can decide whether it goes into "constants" or "functions"!

src/lib/frontend/translate.ml
bclement-ocp: The way we typically deal with such semantic rewriting would be in the `make` function of the `Float` theory! See for instance the treatment of the existing `Float` primitive in `arith.ml`.

It should look something like this, where `BV` is the bit-vector theory from the Shostak module (that part of the code is a bit tricky, I can help you set up the hooks for a new theory if you want!):

```ocaml
let make t =
  let { E.f; xs } = E.term_extract t in
  match f, xs with
  | Op Fp, [sign_t; exp_t; sig_t] -> (
    let sign_r, sign_ctx = X.make sign_t in
    let exp_r, exp_ctx = X.make exp_t in
    let sig_r, sig_ctx = X.make sig_t in
    let extract r = Option.bind (BV.extract r) Bitv.to_Z_opt in
    match extract sign_r, extract exp_r, extract sig_r with
    | Some sign_Z, Some exp_Z, Some sig_Z ->
      ... build floating-point constant ...
    | _ -> X.term_embed t
  )
  | _ -> X.term_embed t
```

But again, I think in a first version of the PR it's OK to do it at the term level (as I stated in another comment, we might actually want to also do this simplification at the term level because there are no other ways to write FP literals otherwise).

src/lib/frontend/translate.ml
bclement-ocp: As noted in another comment: it's fine that the other primitives are not valid as constants (and maybe we should refactor `translate` a little bit to not have this separation), but there is value in listing them explicitly still — this way we get a warning when Dolmen adds a new constructor and we can decide how to treat it instead of it silently being ignored!

src/bin/common/parse_command.ml
bclement-ocp: It is "FloatingPoint": see the heading of the page and the `(theory FloatingPoint` at the start.

hra687261: Ok, I misunderstood, I thought there was a specific naming syntax for the `set-option` command. `smt.float` it is.

src/lib/frontend/translate.ml
hra687261: Done. (Agree on refactoring `translate` and group all symbols by theory, notably since it goes well with the new dolmen)

src/lib/frontend/translate.ml
hra687261: `Fp_val` was extracted into its own module, and a comment was added to the `let mk_fp_literal` function in `Expr` stating how it should be done.

src/lib/structures/expr.ml
hra687261: > I think it's fine to do this simplification in the term constructors — although it will impact matching on fp terms

Done for the renaming, and added a TODO about the quoted text.

src/bin/common/solving_loop.ml
hra687261: Ok, I added it because it seems to be the way in which we test different options in the testsuite (instead of using different dune files/commands with calls to the alt-ergo binary with different cmdline options), I'll see if there is a simple way to do that with dune.
If there isn't wdyt of keeping this before a full support of `set-option` for theories/preludes is implemented? (it's a bit tricky because it requires moving where the preludes are loaded, to after the set-options are parsed, I'll try to do that in a separate PR)

src/lib/frontend/translate.ml
hra687261: Done (also added `ieee_format_to_fp` as a way to parse fp literals)

src/bin/common/solving_loop.ml
bclement-ocp: Ah yeah, we have a kind of clunky system to be able to run things in different environments (by suffixing e.g. with `.default`) but I don't think it works for setting options. I think for parsing tests it is fine to use cram tests, you can set flags there.

src/lib/structures/expr.ml
```diff
@@ -3140,6 +3149,63 @@ module BV = struct
   let bvsge s t = bvsle t s
 end
 
+(** Constructors from the smtlib theory of floating-point numbers.
+
+    https://smt-lib.org/theories-FloatingPoint.shtml *)
+module FP = struct
+
+
+  let mk_fp_literal ~neg ~biased_exp ~trail e s =
```
bclement-ocp: What is called the "trail"? I have never seen that term. It seems to be the significand (implicit, i.e. no hidden bit)?

src/lib/structures/expr.ml
```diff
@@ -3140,6 +3149,63 @@ module BV = struct
...
+    else if biased_exp = 0 && Z.equal trail Z.zero then
+      (* zero exponent + zero significand: signed zero *)
+      (if neg then Fp_value.Minus_zero else Fp_value.Plus_zero)
+    else
+      let hidden_bit = s - 1 in
+      let significand =
+        if biased_exp = 0 then trail
+        else Z.(trail lor (one lsl hidden_bit))
+      in
+      Fp_value.Finite { neg; biased_exp; significand }
```
bclement-ocp: This makes the hidden bit explicit, but I'm not sure why — can't we just keep it hidden (in the `biased_exp`)?

src/lib/structures/expr.ml
```diff
@@ -3140,6 +3149,63 @@ module BV = struct
...
+  let bv_literal_to_z t =
+    match t.f, t.xs with
+    | Sy.Bitv (_, z), [] -> z
+    | _ ->
+      invalid_arg
+        "fp applications are currently only supported for bitvector literals"
+
+  let fp sign_t exp_t sig_t e s =
+    let neg = Z.equal (bv_literal_to_z sign_t) Z.one in
+    let biased_exp = Z.to_int (bv_literal_to_z exp_t) in
```
bclement-ocp: This should catch the potential `Overflow` on `Z.to_int` to convert it into a proper user-facing error.

src/lib/structures/expr.ml
```diff
@@ -3140,6 +3149,63 @@ module BV = struct
...
+    let neg =
+      Z.equal (Z.shift_right bv_z (e + s - 1)) Z.one
+    in
+    let exp_mask = Z.(one lsl e - one) in
+    let biased_exp =
+      Z.to_int (Z.logand (Z.shift_right bv_z exp_shift) exp_mask)
+    in
+    let sig_mask = Z.(one lsl exp_shift - one) in
+    let trail = Z.logand bv_z sig_mask in
+    float (mk_fp_literal ~neg ~biased_exp ~trail e s) e s
```
bclement-ocp: I think this can be simplified at bit:

```suggestion
    let bv_z = bv_literal_to_z bv_t in
    let trail = Z.extract bv_z 0 s in
    let biased_exp = Z.extract bv_z s e in
    let neg = Z.testbit bv_z (e + s - 1) in
    float (mk_fp_literal ~neg ~biased_exp ~trail e s) e s
```

src/lib/structures/fp_value.ml
```diff
@@ -0,0 +1,64 @@
...
+
+(** Literal floating-point values. *)
+
+type t =
+  | Plus_infinity
+  | Minus_infinity
+  | Plus_zero
+  | Minus_zero
+  | NaN
+  | Finite of { neg : bool; biased_exp : int; significand : Z.t }
```
bclement-ocp, line 27: I'm not 100% convinced by this choice of representation. I think that given that our implementation of the Float theory is going to be (initially, at least) driven by the existing implementation of floating-point arithmetic in alt-ergo, it would make more sense for this to be a `Finite of Q.t` (in fact, I'd maybe go as far as representing `+0` this way and keeping a distinct constructor for `-0` only, but I'm not sure it is a good idea at this stage).

src/lib/structures/symbols.ml
```diff
@@ -120,6 +120,7 @@ type t =
   | Int of Z.t
   | Real of Q.t
   | Bitv of int * Z.t
+  | Fp of Fp_value.t
```
bclement-ocp: I think it'd be better for this to be called `Float` as well, and reserve `Fp` for the actual SMT-LIB `fp` function (which is not limited to literals).

src/lib/util/theories.ml
```diff
@@ -18,17 +18,19 @@
 
 (* Note: keep the constructors in the same order as in the definition in
    [prelude_to_int] so that it gets simplified to the identity. *)
-type prelude = Nra | Ria | Fpa
+type prelude = Nra | Ria | Fpa | SmtFloat
```
bclement-ocp, line 21: (note: if we don't have a prelude for the SmtFloat theory yet we could add it to `Theories.t` for the time being, but we'll want a prelude later, so it's fine)

src/lib/structures/expr.ml
hra687261: It's the significand without the hidden bit, I've seen the term "trailing significand field" to refer to that part of the significand (e.g. in https://en.wikipedia.org/wiki/Significand), what would you suggest?

src/lib/structures/expr.ml
hra687261: I think we can, I assumed it would be "more natural" to manipulate the floating point number with its full representation.

src/lib/structures/expr.ml
hra687261: Done.

src/lib/structures/symbols.ml
hra687261: Done.

src/bin/common/solving_loop.ml
hra687261: Done.

src/lib/structures/fp_value.ml
hra687261: Neither am I, but this is the simple/straightforward way to do it for parsing stuff from the smt-lib, the idea is to change/adapt it in the future when the reasoning will be implemented.

src/lib/structures/expr.ml
hra687261: Indeed, that's much simpler! But to not "overflow" it's:
```ocaml
    let trail = Z.extract bv_z 0 (s - 1) in
    let biased_exp = Z.extract bv_z (s - 1) e in
```

src/lib/structures/expr.ml
bclement-ocp: I think don't think there is a "natural" representation of floating points — in fact if there was one I'd argue it is the one from the IEEE standard (i.e. no hidden bit) :)

I'm fine with including the hidden bit if it turns out to be easier, but in that case it should be copiously documented in `Fp_value` as I'd by default expect that to use the semantics from the IEEE/SMT-LIB standard (which does not include the hidden bit). I'd really like to avoid having the same situation that we currently have with `extract`, where the arguments of Alt-Ergo's `Extract` symbol do not mean the same thing as the arguments of the SMT-LIB `extract` primitive.

src/lib/structures/expr.ml
bclement-ocp: I was not aware of that name :) I'm fine with keeping `trail` as is.

Although the page you link seems to say that "mantissa" should refer to the significand without hidden bit (actually, I was more familiar with "mantissa" than with "significand"), I'd also be fine with that.

src/lib/structures/expr.ml
hra687261: I updated the implementation, the current one follows the smt-lib standard (the hidden bit stays hidden). I guess what I meant by "natural" is what feels like the more explicit one with no "special" encoding like the hidden bit. But I agree that following what is most common is best.

tests/cram.t/run.t
```diff
@@ -141,3 +153,22 @@ Tableaux solver.
...
+  
+  unsat
+  
+  unsat
+  
+  unsat
+  
+  unsat
+  
+  unsat
```
bclement-ocp, line 160: General remark (it's fine to keep the test as is in the PR, just a thought about this file): I know we currently have a single big fat cram test file, but I find it's getting quite unwieldy. I think we could try separating these, and maybe also placing the input smt files inline; for instance we could have a file `tests/fp_literals.t`:

```cram
Parsing of positive infinite literals is supported:

  $ alt-ergo -o smtlib2 --enable-theory smt.float 2>/dev/null << EOF
  > (set-logic QF_FP)
  > (set-option :smt-lib-fpa true)
  > 
  > (assert (not (or
  >     (= (_ +oo 8 24) (fp #b0 #b11111111 #b00000000000000000000000))
  >     (= (_ +oo 8 24) ((_ to_fp 8 24) #b01111111100000000000000000000000))
  > )))
  > (check-sat)
  > EOF
  
  unsat
```

etc.

tests/cram.t/run.t
```diff
@@ -76,11 +82,13 @@ The messages above mention `--produce-models`, but we can also use
   unknown
   (error "<stdin>:1.60: Model generation disabled (try --produce-models)")
 
+
```
bclement-ocp: I'm not sure what's up with these new lines, did you add them manually?

src/bin/common/solving_loop.ml
bclement-ocp: Should we remove the `set-option` for now then?

tests/cram.t/run.t
hra687261: Nope, doing a `make runtest` and a `dune promote` added them, not sure why or how to avoid them.

bclement-ocp: Hm, interesting. Which version of `dune` do you have?

tests/cram.t/run.t
```diff
@@ -141,3 +153,22 @@ Tableaux solver.
   
   unknown
   (error "<stdin>:1.62: the selected solver does not support optimization")
+
+
+Testing Alt-Ergo's support for parsing smt-lib FPA literals, this test ought to
+be moved to ./tests/float when we have a proper support for enabling and
+disablling theories in smt-lib files with `set-option`
```
bclement-ocp:
```suggestion
disabling theories in smt-lib files with `set-option`
```

tests/cram.t/run.t
hra687261: So, it turns out it was due to my editor removing trailing spaces when I added them test and saved the file. I think dune saw the lines without the indentation spaces as empty comment lines and decided to add a new empty line after them

tests/cram.t/run.t
hra687261: I agree that the file is getting big, I don't know about putting input smt files inline, I usually find that it makes it hard to debug if a test doesn't pass anymore, also when adding tests it requires copy-pasting the content instead of just copying a file (but it's not really an issue either way)

src/bin/common/solving_loop.ml
hra687261: Yes indeed, that was an omission.

src/lib/structures/expr.ml
hra687261: In fact, I am also more familiar with `mantissa`, I changed it to that.

tests/cram.t/run.t
bclement-ocp: Interesting — I don't find it that much harder to debug, but I usually just rerun the test.

(Although I can see how it would be more annoying if we were to test anything beyond surface-level parsing, but we usually use the cram tests only for that sort of purpose anyway)

bclement-ocp (approved):

tests/cram.t/run.t
hra687261: Yes, I was talking more about the cases in which you'd have to edit the test and so on, but yes for simple stuff, rerunning the test usually suffices.

PR #1359, opened 2026-06-15
matching: Separate trigger inference from actual matching

bclement-ocp:
Currently, the matching environment (from the `Matching` module) is used to hold both the current triggers and the terms to match on. This separates the triggers part from the terms part for improved clarity and a better separation of concerns — it is also one step towards incremental e-matching, which requires keeping triggers (but not the terms to match against!) across instantiation rounds.

Review commit by commit.


Halbaroth (approved):
Thanks for this atomic PR. It was easy to review.

PR #1360, opened 2026-06-15
Switch to ocamlformat

### LINKED ISSUE #634: Switching to ocamlformat

bclement-ocp:
Fixes #634 — it used to be that we couldn't do it due to crashes in ocamlformat on some files (I think `satml_frontend.ml`, but I can't remember), but it seems those have been fixed.

I have used the configuration from the OxCaml repository as that is the one I am familiar with, but I don't really have a strong opinion on the configuration to use. I have also not read the code after formatting — I trust `ocamlformat` to do a reasonably OK job, and if there are places where it doesn't, we can fix it later. I think having a consistent, automatically formatted output matters more than having a few places where it is slightly sub-optimal — that is the whole point of using a tool like `ocamlformat`.

There are some changes to the `dune` files because I used `dune fmt` to format the repository — I don't think that is a big issue, but I can remove these changes if required.


bclement-ocp:
It looks like this is going to wait for a bit again:

 - `ocamlformat` has a slightly different opinion than our `check_style.sh` on what is an appropriate maximum line length apparently, especially in documentation comments. Maybe we should make `check_style.sh` a bit more lenient (or get rid of it completely?).
 - The container we are using for building statically with flambda does not have a recent enough version of `opam` to be able to parse with `with-dev-setup` option…

Halbaroth:
> The container we are using for building statically with flambda does not have a recent enough version of opam to be able to parse with with-dev-setup option…

My secret hack for this issue is to use a Docker image instead of writing the script in the CI but it works only for Linux :/

bclement-ocp:
Actually I think we can just get rid of the `with-dev-setup` marker to be honest, we did not have one for ocp-indent either. (I'll do this when I rebase the PR)

bclement-ocp:
Updated the PR to:

 - Remove the `with-dev-setup` marker, and
 - Remove the `dune` file formatting, and
 - Update the line length limit in ocp-checker to be a bit more lenient (90 chars instead of 80). 80 is the target we give to ocamlformat, but sometimes it overflows a bit. I think 90 as a limit should be fine?

bclement-ocp:
Updated again to change keep documentation comments after declarations when possible, apologies!

shell.nix
```diff
@@ -9,7 +9,7 @@ pkgs.mkShell {
   # see https://nixos.wiki/wiki/Locales
   LOCALE_ARCHIVE = "${pkgs.glibcLocales}/lib/locale/locale-archive";
   nativeBuildInputs = with ocamlPackages; [
-    pkgs.ocamlformat
+    pkgs.ocamlformat_0_28_1
```
Halbaroth, line 12: You don't want to use the latest version of ocamlformat? 
I know the package isn't available on NixOS 25.11 but we can update to NixOS 26.05 ;)

dune-project
```diff
@@ -7,6 +7,9 @@
 ; For plugins and preludes
 (using dune_site 0.1)
 
+(formatting
+ (enabled_for ocaml))
```
Halbaroth, line 11: Formatting the dune files isn't appropriate?

Halbaroth (approved):

shell.nix
bclement-ocp: I do, but when I tried to update to 26.05 I got errors about `http` not being compatible with OCaml 4.14 and did not want to follow that rabbit hole (yet)!

dune-project
bclement-ocp: I initially formatted them, but we have some `dune` files with a bunch of modules explicitly listed (in `src/lib/dune` in particular) and having one per line makes the file quite big and hard to navigate. I'm not necessarily opposed to formatting them, but it seems less useful in practice, so I figured we can remove these lines later if needed.

bclement-ocp:
Merged manually to preserve history and separate the actual reformatting commit from the rest.

PR #1361, opened 2026-06-16
Get rid of ppxes

bclement-ocp:
As mentioned in the previous dev meeting; this will make it easier to experiment with alt-ergo and Flambda2 (reviving #1302 ).

(I *think* we don't have users of ppxes in our dependencies)

Review by commit.

### DIFF
--- a/alt-ergo-lib.opam.locked
+++ b/alt-ergo-lib.opam.locked
@@ -26,6 +26,7 @@ depends: [
   "conf-pkg-config" {= "4"}
   "conf-zlib" {= "1"}
   "cppo" {= "1.8.0"}
+  "crunch" {= "4.0.0"}
   "csexp" {= "1.5.2"}
   "dolmen" {= "dev"}
   "dolmen_loop" {= "dev"}
@@ -58,6 +59,7 @@ depends: [
   "ounit2" {= "2.2.7" & with-test}
   "pp" {= "2.0.0"}
   "pp_loc" {= "2.1.0"}
+  "ptime" {= "1.2.0"}
   "qcheck" {= "0.25" & with-test}
   "qcheck-core" {= "0.25" & with-test}
   "qcheck-ounit" {= "0.25" & with-test}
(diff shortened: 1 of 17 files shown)


bclement-ocp:
I found https://github.com/robur-coop/mcrunch just after writing this PR, maybe we should use that instead of crunch (the API of `mcrunch` is simpler and I like it more).

Halbaroth:
> I found https://github.com/robur-coop/mcrunch just after writing this PR, maybe we should use that instead of crunch (the API of `mcrunch` is simpler and I like it more).

I haven't strong opinion. I use `crunch` in my projects but `mcrunch` looks as good.

src/bin/common/parse_command.ml
```diff
@@ -170,7 +172,37 @@ module Debug = struct
     | Use
     | Commands
     | Optimize
-  [@@deriving eq]
+
+  let to_int = function
```
Halbaroth: Is it always inline?

alt-ergo-lib.opam.locked
```diff
@@ -58,10 +59,7 @@ depends: [
   "ounit2" {= "2.2.7" & with-test}
   "pp" {= "2.0.0"}
   "pp_loc" {= "2.1.0"}
-  "ppx_blob" {= "0.9.0"}
-  "ppx_derivers" {= "1.2.1"}
-  "ppx_deriving" {= "6.0.3"}
-  "ppxlib" {= "0.35.0"}
+  "ptime" {= "1.2.0"}
```
Halbaroth, line 62: Is it necessary?

Halbaroth (approved):
Looks good to me :)

alt-ergo-lib.opam.locked
bclement-ocp: It is a dependency of `crunch` :shrug:

src/bin/common/parse_command.ml
bclement-ocp: I think it should be (it becomes more or less the identity) but my mind might be warped by what Flambda2 does — I'll add `[@inline]` annotations.

PR #1364, opened 2026-06-24
Support reasoning over most the SMT-LIB FPA theory symbols

hra687261:
This PR adds a prelude which axiomatizes the semantics of most of the SMT-LIB FPA theory symbols, and maps the symbols of the theory parsed by dolmen to these axiomatizations, providing support for symbolic reasoning over most of them (except fp.rem, fp.to_fp, fp.of_sbv, fp.of_ubv, fp.to_ubv and fp.to_sbv the support for which will be added in the future in a separate PR).
The PR also propagates for each of the FPA literals for which support was added in #1358 predicates and assertions which allow linking these literals to the axiomatization. It also adds propagation of equalities with literals from asserted predicates, as well as a shostak theory to solve equalities between fp literals among themselves or fp literals and fp variables.

Rebased on https://github.com/OCamlPro/alt-ergo/pull/1371


src/lib/structures/expr.ml
```diff
@@ -3137,14 +3137,162 @@ module FP = struct
...
   let ieee_format_to_fp bv_t e s =
     let bv_z = bv_literal_to_z bv_t in
     let mantissa = Z.extract bv_z 0 (s - 1) in
     let biased_exp = z_to_int (Z.extract bv_z (s - 1) e) in
     let neg = Z.testbit bv_z (e + s - 1) in
-    float (Fp_value.mk_fp_literal ~neg ~biased_exp ~mantissa e) e s
+    float (Fp_value.mk_fp_literal ~neg ~biased_exp ~mantissa ~e ~s) e s
+
+  (* TODO: move this module to somewhere else? fpa_rounding.ml? (might need
+     renaming) *)
```
Halbaroth: Did you forget to remove this comment or to move this module?

src/lib/structures/expr.mli
```diff
@@ -682,9 +682,127 @@ end
 
     https://smt-lib.org/theories-FloatingPoint.shtml *)
 module FP : sig
+  module Names : sig
+    (* float conversion function *)
```
Halbaroth: These comments are not visible for odoc.

src/lib/structures/fp_value.ml
```diff
@@ -50,23 +42,74 @@ let pp ppf = function
   | Plus_zero -> Fmt.pf ppf "+zero"
   | Minus_zero -> Fmt.pf ppf "-zero"
   | NaN -> Fmt.pf ppf "NaN"
-  | Finite { neg; biased_exp; significand } ->
-    Fmt.pf ppf "fp[%b;%d;%s]" neg biased_exp (Z.to_string significand)
+  | Finite q -> Fmt.pf ppf "fp[%s]" (Q.to_string q)
```
Halbaroth: We display the size of the biased and significand in the smtlib printer but loss it in the debug printer.

src/lib/util/theories.ml
```diff
@@ -67,7 +67,11 @@ let pp ppf = function
   | ADT -> Format.fprintf ppf "adt"
   | AC -> Format.fprintf ppf "ac"
 
-let filename = Format.asprintf "<builtins>/%a.ae" pp_prelude
+let filename th =
+  match th with
+  | (Fpa | Ria | Nra) as prelude ->
+    Format.asprintf "<builtins>/%a.ae" pp_prelude prelude
+  | SmtFloat -> "<builtins>/smt-lib-fpa.smt2"
```
Halbaroth: We could create a function `prelude_to_format` and modify `pp_prelude` to eliminate this particular case.

src/preludes/smt-lib-fpa.smt2
```diff
@@ -0,0 +1,1845 @@
+(set-logic ALL)
```
Halbaroth, line 6: For the future developers, we may let a comment to explain how this file has been produced?

src/lib/reasoners/shostak.ml
```diff
@@ -348,19 +372,21 @@ end = struct
...
-    | false, true, false, false -> BITV.fully_interpreted sb
-    | false, false, true, false -> ADT.fully_interpreted sb
-    | false, false, false, true -> AC.fully_interpreted sb
-    | false, false, false, false -> false
+    | true, false, false, false, false -> ARITH.fully_interpreted sb
+    | false, true, false, false, false -> BITV.fully_interpreted sb
+    | false, false, true, false, false -> ADT.fully_interpreted sb
+    | false, false, false, true, false -> FPA.fully_interpreted sb
+    | false, false, false, false, true -> AC.fully_interpreted sb
+    | false, false, false, false, false -> false
```
Halbaroth, line 383: Beautiful Alt-Ergo style <3

src/lib/frontend/translate.ml
```diff
@@ -1159,22 +1189,96 @@ let rec mk_expr ?(loc = Loc.dummy) ?(name_base = "") ?(toplevel = false)
...
+                _ ) ->
+              invalid_app_term ()
+          end
+          (* can't be applied *)
           | (RoundingMode | T _ | Fp _), _
           | ( ( Plus_infinity _ | Minus_infinity _ | Plus_zero _ | Minus_zero _
               | NaN _ ),
-              _ ) ->
+              _ )
+          (* roudning modes are now matched as literals since we use the SMT-LIB
```
Halbaroth: Rounding

src/lib/reasoners/fpa.ml
```diff
@@ -0,0 +1,147 @@
...
+  val embed : r abstract -> r
+end) =
+struct
+  type t = X.r abstract
+
+  type r = X.r
+
+  let name = "Fpa"
+
+  let timer = Timers.M_None
```
Halbaroth: We should create a new timer for the FPA theory.

src/lib/structures/expr.ml
```diff
@@ -3137,14 +3137,162 @@ module FP = struct
...
+    | _ -> mk_term (Sy.name Names.ae_float) [eb; sb; mode; x] Ty.Treal
+
+  let ae_float_literal_prec ~eb ~sb ~mode x =
+    ae_float_literal_prec_aux ~eb:(Ints.of_int eb) ~sb:(Ints.of_int sb) ~mode x
+
+  let ae_float ~eb ~sb ~mode x =
+    match term_view eb, term_view sb with
+    | { f = Sy.Int _; _ }, { f = Sy.Int _; _ } ->
+      ae_float_literal_prec_aux ~eb ~sb ~mode x
+    | _ -> mk_term (Sy.name Names.ae_float) [eb; sb; mode; x] Ty.Treal
```
Halbaroth: We may fail if `eb` or `sb` are not expression of type int?

src/lib/reasoners/fpa_rel.ml
```diff
@@ -0,0 +1,692 @@
...
+module Z = Numbers.Z
+module MX = Shostak.MXH
+module SX = Shostak.SXH
+module Names = E.FP.Names
+
+let src = Logs.Src.create ~doc:"Fpa_rel" __MODULE__
+
+module Log = (val Logs.src_log src : Logs.LOG)
+
+let timer = Timers.M_None
```
Halbaroth: We could create a timer for FPA.

src/lib/reasoners/fpa_rel.ml
```diff
@@ -0,0 +1,692 @@
...
+    then Some Nan
+    else if String.equal name Names.is_zero
+    then Some Zero
+    else if String.equal name Names.is_infinite
+    then Some Infinite
+    else if String.equal name Names.is_positive
+    then Some Positive
+    else if String.equal name Names.is_negative
+    then Some Negative
+    else None
```
Halbaroth, line 164: We may use a hash table to store this function?

src/lib/reasoners/fpa_rel.ml
```diff
@@ -0,0 +1,692 @@
...
+      (* not NaN & not positive -> negative *)
+      apply Negative (True (Ex.union ex_nan ex_p)) d
+    | _ -> d
+
+  let merge ~ex d1 d2 =
+    let apply_if_known p st d =
+      match st with
+      | Unknown -> d
+      | True ex2 -> apply p (True (Ex.union ex ex2)) d
+      | False ex2 -> apply p (False (Ex.union ex ex2)) d
```
Halbaroth: If you add a parameter `~ex` to `apply`, you don't need to reconstruct the values True and False and you avoid to hide the explanation for the merge in one of the status.

src/lib/reasoners/fpa_rel.ml
```diff
@@ -0,0 +1,692 @@
...
+      Printer.print_dbg ~module_name:"Fpa_rel" ~function_name:"apply"
+        "apply %a = %a to %a" pp_pred p pp_status st pp d;
+    let d = upd_pred_status p st d in
+    let d =
+      (* if [p] became true, then the predicates in [exclusive_of p] are must be
+         false *)
+      match st with
+      | True ex ->
+        List.fold_left
+          (fun d p -> upd_pred_status p (False ex) d)
```
Halbaroth: If you appy my suggestion and add an extra `~ex` argument to `apply`,  change this code as follows:
```ocaml
| True ex_st ->
        List.fold_left
          (fun d p -> upd_pred_status ~ex p (False ex_st) d)
...
```

src/lib/reasoners/fpa_rel.ml
```diff
@@ -0,0 +1,692 @@
...
+        | _ -> None)
+      | _ -> (
+        match d.is_infinite with
+        | True ex_z -> (
+          match d.is_positive, d.is_negative with
+          | True ex_p, _ -> Some (Fp_value.Plus_infinity, Ex.union ex_z ex_p)
+          | _, True ex_n -> Some (Fp_value.Minus_infinity, Ex.union ex_z ex_n)
+          | _ -> None)
+        | _ -> None)
+    end
```
Halbaroth: It could be easier to read with a single match on the value of all the predicates?

src/lib/reasoners/fpa_rel.ml
```diff
@@ -0,0 +1,692 @@
...
+    match d.is_nan, d.is_positive, d.is_negative with
+    | Unknown, False ex_p, False ex_n ->
+      (* not positive & not negative -> NaN *)
+      apply Nan (True (Ex.union ex_p ex_n)) d
+    | False ex_nan, Unknown, False ex_n ->
+      (* not NaN & not negative -> positive *)
+      apply Positive (True (Ex.union ex_nan ex_n)) d
+    | False ex_nan, False ex_p, Unknown ->
+      (* not NaN & not positive -> negative *)
+      apply Negative (True (Ex.union ex_nan ex_p)) d
```
Halbaroth: It's a bit confusing to mix propagation and domain merging in a single step. Could you move this code in `deduce_fpval_eq`?

src/lib/reasoners/fpa_rel.ml
```diff
@@ -0,0 +1,652 @@
...
+      let sb = int_lit_view r_sb in
+      if X.is_constant r_mode
+      then { t with ready = E.Map.add term (ready_value eb sb r_mode) t.ready }
+      else
+        { t with
+          apps = E.Map.add term { eb; sb; mode = r_mode } t.apps;
+          watched_args = add_by_arg r_mode term t.watched_args
+        }
+
+  let subst ~ex:_ rr nrr t =
```
Halbaroth: You don't need to track the explanation as these domains cannot generate conflicts?

src/lib/reasoners/fpa_rel.ml
```diff
@@ -0,0 +1,652 @@
...
+    Uf.domains uf
+    |> Uf.GlobalDomains.add (module Domains) Domains.empty
+    |> Uf.GlobalDomains.add (module AEFloat_arg_watch) AEFloat_arg_watch.empty )
+
+let pow2 n = Z.shift_left Z.one n
+
+let mk_eq_fact lhs rhs =
+  Literal.LTerm (E.mk_eq ~iff:false lhs rhs), Ex.empty, Th_util.Other
+
+(* TODO: cache (eb, sb) -> (eb, sb) exprs? *)
```
Halbaroth: Did you plan to do it in this PR?

src/lib/reasoners/fpa_rel.ml
```diff
@@ -0,0 +1,652 @@
...
+
+let type_literal_facts env eb sb =
+  let c = get_literals env eb sb in
+  [ int_literal_fact eb sb "ae.fp.pow2sb" c.pow2sb;
+    int_literal_fact eb sb "ae.fp.max_int" c.max_int_z;
+    real_literal_fact eb sb "ae.fp.max_real" c.max_int_z;
+    real_literal_fact eb sb "ae.fp.pow2sb_real" c.pow2sb;
+    real_literal_fact eb sb "ae.fp.half_pow2sb_real" c.half_pow2sb;
+    real_literal_fact eb sb "ae.fp.abs_err_rne_denom" c.abs_err_rne_denom;
+    real_literal_fact eb sb "ae.fp.abs_err_denom" c.abs_err_denom ]
```
Halbaroth: Technically these terms are created by the FPA theory and should be added to new_terms for the instantiation?

src/preludes/smt-lib-fpa.smt2
Halbaroth, line 6: Should we be concerned about the license of this file? It is essentially a one-one translation of the Why3 theory.

src/lib/structures/expr.ml
hra687261: I forgot to remove the comment indeed, done.

src/lib/structures/expr.mli
hra687261: These are not meant to be part of the doc, but more as a way to organize/differentiate the different groups of functions.

src/lib/structures/fp_value.ml
hra687261: Yes, because the smt-lib printer is expected (I suppose) to produce something that is (at least close) to something parseable in the smt-lib format, while this one is for debugging only. 
On the other hand, I am not sure if for debugging it is better to print a `Q.t` or the mantissa/significand bit-vectors, intuitively, since we represent finite Floats as `Q.t`, I'd say its this one but I am not sure, in practice I've seen that it quickly becomes unreadable anyway.

src/lib/util/theories.ml
hra687261: Not sure I understand what you're suggestion, but I updated it with a simpler version.

src/preludes/smt-lib-fpa.smt2
hra687261: The code used to produce this axiomatization is currently in a private repository, it wasn't necessarily intended to be published but I guess we could make it public. 
In the meantime I added a comment saying how it was produced.

src/lib/reasoners/fpa.ml
hra687261: Done.

src/lib/structures/expr.ml
hra687261: No, though this code is outdated, the idea was to keep applications of `ae.float` as uninterpreted symbol applications when the arguments are not literals, and only replace it with `Sy.Op Float` when we know that they are literals (that is done by the watch module in `fpa_rel`). 
The current version simply adds support for non-literal arguments to `Sy.Op Float` (and replaces all "ae.float" applications with `Sy.Op Float` directly)

src/lib/reasoners/fpa_rel.ml
hra687261: Done.

src/lib/reasoners/fpa_rel.ml
hra687261: I am not sure that hashing a string + a lookup in a table is better that 5 string equality checks, either way, its probably insignificant and an over-optimization.
Ideally, I'd write something `match s with Names.is_zero -> ... | Names.is_infinite -> ...` but that's unfortunately not possible.

src/lib/reasoners/fpa_rel.ml
hra687261: Done.

src/lib/reasoners/fpa_rel.ml
hra687261: Here `ex_st` would be equal to `ex` since the status would be created from `ex` anyway, so it would be redundant, but since it is in the state, we can remove the `ex` argument from both `upd_pred_status` and `merge_status`.

src/lib/reasoners/fpa_rel.ml
hra687261: I don't think that a match over 5 values is significantly simpler, but treating `is_nan` separately, then all other predicates does indeed seem to improve readability.

src/lib/reasoners/fpa_rel.ml
hra687261: `apply` is meant to do a complete propagation, i.e. if we propagate `p1(d)` and it implies `p2(d)`, we want to do both at the same time instead of risking having a state where `p1(d)` is done but not `p2(d)`, that simplifies merging because we have a guarantee of completeness, the merge function would be more complicated otherwise. Also, if there is a contradiction, we will also find it earlier thanks to the guarantee that propagations are complete.
Also `deduce_fpval_eq` is supposed to take a complete domain and deduce equalities from it, since we only propagate equalities once, it makes sens to take domains that are already complete at that point to make deductions from them, and not take an incomplete one over which we will have to make additional propagations.

src/lib/reasoners/fpa_rel.ml
hra687261: Indeed, this one simply tracks arguments and substitutes them when they becomes literals, then propagates equalities, but it might actually still be necessary to pass the explanation down to the deduced equalities, I will do that. (Not exactly sure, the explanation is probably attached to the semantic values already no?)
Now, that I think of it, since `Sy.Op Float` now supports symbolic arguments for the rounding mode, its not clear that this domain is still necessary.

src/lib/reasoners/fpa_rel.ml
hra687261: No, simply passing exprs as arguments instead of literals avoids having to call `E.Ints.of_int` multiple times, it shouldn't be necessary or significantly helpful to cache them.

src/lib/reasoners/fpa_rel.ml
hra687261: Indeed, added it.

src/preludes/smt-lib-fpa.smt2
hra687261: Not sure, though the transformation is pretty significant, I wouldn't call it a "one-one translation". It's also not clear how the license applies, since the original axiomatization was technically generated by Why3 and not part of its source code.

src/lib/frontend/translate.ml
```diff
@@ -155,33 +155,39 @@ let builtin_term t = Dl.Typer.T.builtin_term t
...
+    fun map ->
+      map
+      |> DStd.Id.Map.add
+           { name = DStd.Name.simple "RoundingMode"; ns = Sort }
+           (fun env s ->
+             builtin_ty
+             @@ Dolmen_type.Base.app0
+                  (module Dl.Typer.T)
+                  env s DE.Ty.roundingMode)
+      |> add_constrs )
```
bclement-ocp, line 190: We should remove the rounding mode type from `fpa_rounding.ml` if we don't use it anymore (and if we do, replace it with `DE.Ty.roundingMode`).

src/lib/frontend/translate.ml
```diff
@@ -369,6 +380,9 @@ let smt_fpa_builtins =
     | Id { ns = Term; name = Simple "int.pow2" } ->
       Dl.Typer.T.builtin_term
       @@ Dolmen_type.Base.term_app_cst (module Dl.Typer.T) env int_pow2_cst
+    | Id { ns = Term; name = Simple "ae.float" } ->
+      Dl.Typer.T.builtin_term
+      @@ Dolmen_type.Base.term_app_cst (module Dl.Typer.T) env ae_float_cst
```
bclement-ocp, line 390: Ah, it is unfortunate that we need to expose both `ae.round` (with `_` constant args) and `ae.float` :(

src/lib/frontend/translate.ml
```diff
@@ -447,6 +461,8 @@ let rec dty_to_ty ?(update = false) ?(is_var = false) dty =
   | `Pi (tyvl, ty) ->
     if update then Cache.store_tyvl ~is_var tyvl;
     aux ty
+  | `App (`Builtin (B.Float B.Float.RoundingMode), []) ->
+    Fpa_rounding.fpa_rounding_mode
```
bclement-ocp: I think this should go above, along with the other `App` cases (likely just after `B.Unit`).

src/lib/frontend/translate.ml
```diff
@@ -456,6 +472,11 @@ and handle_ty_app ?(update = false) ty_c l =
   match Cache.find_ty ty_c with
   | Tadt (hs, _) -> Tadt (hs, tyl)
   | Text (_, s) -> Text (tyl, s)
+  | Tvar { path = Absolute { name; _ } | Local { name; _ }; _ } as tv
+    when Compat.List.is_empty tyl && String.equal name E.FP.Names.t ->
+    (* To get the generic abstract float type from the axiomatization which is
+       stored as type varialbe `ae.fp.t`. *)
+    tv
```
bclement-ocp, line 487: I'm not sure I understand this. Why don't we define `ae.fp.t` directly here (like we do e.g. for rounding mode) and make it available to the axiomatization?

src/lib/frontend/translate.ml
```diff
@@ -779,6 +804,11 @@ let rec mk_expr ?(loc = Loc.dummy) ?(name_base = "") ?(toplevel = false)
         | B.Adt (Constructor _) ->
           let ty = dty_to_ty term_ty in
           E.mk_constr tcst [] ty
+        | B.Float RoundNearestTiesToEven -> mk_rounding NearestTiesToEven
+        | B.Float RoundNearestTiesToAway -> mk_rounding NearestTiesToAway
+        | B.Float RoundTowardPositive -> mk_rounding Up
+        | B.Float RoundTowardNegative -> mk_rounding Down
+        | B.Float RoundTowardZero -> mk_rounding ToZero
```
bclement-ocp: Please keep these within the global match on `B.Float` below (moving guards on `when get_smt_lib_fpa` to the constants).

src/lib/frontend/translate.ml
```diff
@@ -1159,22 +1189,96 @@ let rec mk_expr ?(loc = Loc.dummy) ?(name_base = "") ?(toplevel = false)
...
+              E.FP.round_to_integral ~e ~s ~mode:(mk mode) (mk x)
+            | Of_real { e; s }, [mode; x] ->
+              E.FP.of_real ~e ~s ~mode:(mk mode) (mk x)
+            (* arithmetic without rounding mode *)
+            | Abs { e; s }, [x] -> E.FP.abs ~e ~s (mk x)
+            | Neg { e; s }, [x] -> E.FP.neg ~e ~s (mk x)
+            | Rem { e = _; s = _ }, [_x; _y] ->
+              (* TODO: rem is not currently in the axiomatization, its semantics
+                 need to be either axiomatized or implemented in Alt-Ergo. *)
+              unsupported_app_term ()
```
bclement-ocp: It should become an uninterpreted function in that case, not raise an error!

src/lib/frontend/translate.ml
```diff
@@ -1159,22 +1189,96 @@ let rec mk_expr ?(loc = Loc.dummy) ?(name_base = "") ?(toplevel = false)
...
+            (* real conversion *)
+            | To_real { e; s }, [x] -> E.FP.to_real ~e ~s (mk x)
+            (* TODO: FP <-> FP and BV <-> FP conversion *)
+            | To_fp { e1 = _; s1 = _; e2 = _; s2 = _ }, [_; _]
+            | ( ( Of_sbv { m = _; e = _; s = _ }
+                | Of_ubv { m = _; e = _; s = _ }
+                | To_ubv { m = _; e = _; s = _ }
+                | To_sbv { m = _; e = _; s = _ } ),
+                [_] ) ->
+              unsupported_app_term ()
```
bclement-ocp: These should become uninterpreted as well.

src/lib/frontend/translate.ml
```diff
@@ -482,13 +503,17 @@ let mk_ty_decl (ty_c : DE.ty_cst) =
...
-    let ty = Ty.text ty_params ty_c in
-    Cache.store_ty ty_c ty
+  | None | Some Abstract -> (
+    match ty_c with
+    | { path = Absolute { name; _ }; _ }
+      when Options.get_smt_lib_fpa () && String.equal name E.FP.Names.t ->
+      (* Storing the generic abstract float type from the axiomatization as a
+         type variable `ae.fp.t`, so that axioms defined for it can apply for
+         specific instances of the Float(es,sb) type. *)
+      Cache.store_ty ty_c (Ty.named_tvar E.FP.Names.t)
```
bclement-ocp, line 521: I'm not sure I understand this either. My understanding was that for now we would map all the `Float(eb, sb)` types to `ae.fp.t` (with an assertion saying that the value is a valid float for these `eb` and `sb`), but it seems you are keeping the type around?

I think if we want to make `ae.fp.t` a variable (that in practice ranges over the fpa types) we should do that explicitly using `par` instead of this hack.

src/lib/reasoners/arith.ml
```diff
@@ -289,18 +289,35 @@ struct
...
+        let aux_func e =
+          let res, _, _ = Fpa_rounding.float_of_rational prec exp mode e in
+          res
+        in
+        mk_partial_interpretation_1 aux_func coef p ty t x, ctx
+      | None when Options.get_smt_lib_fpa () ->
+        (* If the rounding mode is not a literal and [Options.get_smt_lib_fpa
+           ()] is true, treat it as an uninterpreted function and wait for
+           fpa_rel to update it when the rounding mode is substituted with a
+           literal. *)
```
bclement-ocp: I think we can do this unconditionally

src/lib/reasoners/fpa.ml
```diff
@@ -0,0 +1,147 @@
...
+
+  let embed r = match X.extract r with Some v -> v | None -> Alien r
+
+  let is_mine v = X.embed v
+
+  let make t =
+    let { E.f; ty; _ } = E.term_view t in
+    match f with
+    | Sy.Float fp_val -> is_mine (Literal (ty, fp_val)), []
+    | _ -> assert false
```
bclement-ocp: This should be a proper `internal_error`

src/lib/reasoners/fpa.ml
```diff
@@ -0,0 +1,147 @@
...
+
+  let type_info = function Alien r -> X.type_info r | Literal (ty, _) -> ty
+
+  let equal s1 s2 =
+    match s1, s2 with
+    | Alien r1, Alien r2 -> X.equal r1 r2
+    | Literal (ty1, v1), Literal (ty2, v2) ->
+      Ty.equal ty1 ty2 && Fp_value.compare v1 v2 = 0
```
bclement-ocp: Why not have `Fp_value.equal`?

src/lib/reasoners/fpa.ml
```diff
@@ -0,0 +1,147 @@
...
+    | _ -> false
+
+  let hash = function
+    | Alien r -> X.hash r
+    | Literal (ty, v) -> Ty.hash ty * 17 lxor Hashtbl.hash v
```
bclement-ocp: We use `acc + <prime> * b` to combine hashes elsewhere, please use the same pattern (I don't think it matters too much, but if it doesn't matter too much then consistency is good).

src/lib/reasoners/fpa.ml
```diff
@@ -0,0 +1,147 @@
...
+    | Alien _, _ -> 1
+    | _, Alien _ -> -1
+    | Literal (ty1, v1), Literal (ty2, v2) ->
+      let c = Ty.compare ty1 ty2 in
+      if c <> 0 then c else Fp_value.compare v1 v2
+
+  let print ppf = function
+    | Alien r -> X.print ppf r
+    | Literal (Ty.Tfloat (eb, sb), v) -> Fp_value.pp_smtlib eb sb ppf v
+    | Literal _ -> assert false
```
bclement-ocp: There is a bunch of places where we `assert false` if the type is not a `Tfloat`, we should just store the `eb` and `sb` in the `Literal`, no?

src/lib/structures/expr.ml
```diff
@@ -2805,22 +2805,22 @@ let z_to_int z =
...
-  | Int n -> n
+  | Some (Int n) -> n
   | _ -> Fmt.failwith "The given term %a is not an integer" print t
 
 let rounding_mode_view t =
   match const_view t with
-  | RoundingMode m -> m
-  | _ -> Fmt.failwith "The given term %a is not a rounding mode" print t
+  | Some (RoundingMode m) -> Some m
+  | _ -> None
```
bclement-ocp, line 2813: We don't actually use `const_view` elsewhere, I think we could simply inline the test for `Constr c when Ty.equal ty Fpa_rounding_mode.fpa_rounding_mode` in the `rounding_mode_view` function and simplify things a tiny bit.

src/lib/structures/expr.mli
bclement-ocp: I think they would be helpful as documentation sections (`(** {2 Float conversion functions} )` etc)!

src/lib/structures/expr.mli
```diff
@@ -682,9 +681,138 @@ end
...
+
+    val abs_err_denom : string
+  end
+
   val fp : t -> t -> t -> int -> int -> t
 
   val ieee_format_to_fp : t -> int -> int -> t
+
+  (* arithmetic with rounding mode *)
+  val add : e:int -> s:int -> mode:t -> t -> t -> t
```
bclement-ocp: The `e` and `s` labelled arguments should maybe be explained somewhere? Is there a reason they are called `e` and `s` rather than `eb` and `sb` following the SMT-LIB?

src/lib/structures/fp_value.ml
bclement-ocp: Ideally for debug we would print as a float, i.e. a real with a dot (decimal would be nice but hexadecimal is easy to do I believe)!

src/lib/structures/fp_value.ml
```diff
@@ -50,23 +42,74 @@ let pp ppf = function
...
+  let neg = Q.sign q < 0 in
+  (* (_, m, e) with m*2^e = |q|, e = max(floor(log2|q|) + 1 - sb, -min_exp). *)
+  let _, m, e =
+    Fpa_rounding.float_of_rational sb min_exp Fpa_rounding.NearestTiesToEven
+      (Q.abs q)
+  in
+  (* hidden_bit = 2^(sb-1), m >= hidden_bit -> normal *)
+  let hidden_bit = Z.shift_left Z.one (sb - 1) in
+  let biased_exp, significand =
+    if Z.compare m hidden_bit >= 0
```
bclement-ocp, line 85: I think we can simply check `Z.testbit m (sb - 1)`?

src/preludes/smt-lib-fpa.smt2
bclement-ocp: I agree with Pierre here, I think that the code generating the prelude should be part of the repository (given the readability of the generated file we are likely to want to make modifications by changing the Why3 theory + the generator code rather than manually editing the prelude).

bclement-ocp (commented):
Thanks :) Did a first pass with some comments, but did not look at the prelude and some calculations in detail yet.

src/preludes/smt-lib-fpa.smt2
Halbaroth: Alt-Ergo is distributed under a unfree license but Why3 is a free software under the terms of LGPL 2.1 https://gitlab.inria.fr/why3/why3/-/blob/master/LICENSE?ref_type=heads

This file is a kind of derivative work of the original file in Why3.

Halbaroth: We may store your code in a Gitlab repository and add a link in this file or in a HACKING.md file?

src/lib/structures/expr.mli
hra687261: Done.

src/lib/structures/fp_value.ml
hra687261: We could simply convert to float then print with "%.17g", but we would lose precision on float > 64, so maybe the hex format is best. 
The current version should be good.

src/preludes/smt-lib-fpa.smt2
hra687261: Ok, then I suggest doing that in a separate PR because this one is already larger than intended as is, and the code will need some work before being rendered public.

hra687261: > This file is a kind of derivative work of the original file in Why3.

That is unclear, since:

> the original axiomatization was technically generated by Why3 and not part of its source code.

(Though it is generated from a whyml module which is part of the Why3 source code ...)

Anyway, what do you suggest?

src/lib/frontend/translate.ml
hra687261: Done. But unfortunately, rounding modes in Dolmen are put in the same ADT as other FP symbols, so it makes matching on rounding modes awkward (we always have to skip the cases of the non-rounding mode symbols). 
I will open a PR on Dolmen to put rounding modes in their own enum type (with the hope that it will be accepted and be part of the next Dolmen release).

src/lib/frontend/translate.ml
hra687261: Indeed, IIRC indexed operators expect literals as indices not symbolic values, so we have to. But maybe in the future we can deprecate `ae.round` for `ae.float`, or rename `ae.float` to `ae.round` and make the precisions arguments normal arguments instead of indices (because its a better name, the operator in alt-ergo should probably renamed to something like `FloatRound` instead of just `Float`)

src/lib/frontend/translate.ml
hra687261: Done.

src/lib/frontend/translate.ml
hra687261: Done.

src/lib/frontend/translate.ml
hra687261: Done. 
But IMO, if we don't support any reasoning over something, it would be better to have a failure than have surprising behavior. 
From a user's pov, it would be confusing if the solver is unable to solve simple queries, or its able to solve a query but not a similar with maybe a couple of additional symbols, if the user's mind does not go to AE's support for those specific symbols, they might end up wasting a lot of time looking for the solution elsewhere. But, support for them is meant to be added before the next is release anyway, so it should not be an issue.

src/lib/frontend/translate.ml
hra687261: Done.

src/lib/frontend/translate.ml
hra687261: The idea is indeed to make `ae.fp.t` a variable, so that axioms we write over it would match over any type `Float (eb, sb)` (we want to keep the types `Float (eb,sb)` as is because its more convenient/practical, tho I suppose we could cheat by having the type parameters embedded into `ae.fp.t` but still somehow have a generic one that matches over all of them (?) not sure how that would work).
But yes, the proper solution would be to use `par` in the axiomatization.

src/lib/reasoners/arith.ml
hra687261: Done.

src/lib/reasoners/fpa.ml
hra687261: Done.

src/lib/reasoners/fpa.ml
hra687261: Done.

src/lib/reasoners/fpa.ml
hra687261: Done.

src/lib/reasoners/fpa.ml
hra687261: Indeed, done.

src/lib/structures/expr.ml
hra687261: We use also in `int_view` but inlining indeed simplifies things, done.

src/lib/structures/expr.mli
hra687261: This just follows what the Dolmen constructors used, `e` and `s` probably as short for `eb` and `sb`.
I changed them.

src/lib/frontend/translate.ml
hra687261: The idea was to "match" the `ae.fp.t` sort and replace it with a variable so that the axioms work with a generic float type. But this is no longer needed since we now use polymorphism in the axiomatization, so the generic float type is a variable from the start. (https://github.com/OCamlPro/alt-ergo/pull/1371)

src/preludes/smt-lib-fpa.smt2
hra687261: Here's the separate PR that contains the prelude generation code: https://github.com/OCamlPro/alt-ergo/pull/1371 and on which this PR is rebased

PR #1366, opened 2026-07-08
Bump the OCaml compiler versions in the CI

Halbaroth:
This commit introduces three changes:
- Bump 4.14.3 to 4.14.4 and 5.4.0 to 5.5.0,
- Use major version for the upper bound and minor version for the 4.14,
- Quote compiler versions as recommended by the documentation of setup-ocaml.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,16 +27,9 @@ jobs:
           - ubuntu-latest
           - windows-latest
         ocaml-compiler:
-          - "4.11.2"
+          - "4.13.1"
           - "4.14"
           - "5.5"
-        exclude:
-          # OCaml 4.11.2 is not available on macOS + arm
-          - os: macos-latest
-            ocaml-compiler: "4.11.2"
-          # Opam 2.2 requires at least OCaml 4.13
-          - os: windows-latest
-            ocaml-compiler: "4.11.2"
 
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
 permissions: {}
 
 env:
-  OCAML_DEFAULT_VERSION: 4.14.3
+  OCAML_DEFAULT_VERSION: "4.14"
   # Add OPAMYES=true to the environment, this is usefill to replace `-y` option
   # in any opam call
   OPAMYES: true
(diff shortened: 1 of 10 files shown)


bclement-ocp (commented):
I'm OK with bumping the compiler versions that we use and with restricting to patch versions (i.e. `4.14` instead of `4.14.3`), but I'm not sure I'm confident with the major `5` bound — if a new OCaml release accidentally breaks something, that would also brick our CI.

I think new OCaml releases are infrequent enough that we can keep `5.5` (I agree that patch releases are very likely fine) and bump manually it when there is a new minor release?

Halbaroth:
The minimal version for `js_of_ocaml` 6.4.0 is OCaml 4.13. We could bump the minimal version to 4.13 or 4.14?

bclement-ocp:
I'd be fine bumping the minimal version to 4.13 / 4.14 but then we need to do it in the opam files as well — otherwise we risk accidentally using APIs that are not available on older OCaml versions. (I think if we bump the minimal version it should be 4.13 as that is what is still in debian oldstable)

Although do we really need to test the `js_of_ocaml` build on all versions? I think it's fine to only test it on 4.14 (or even on 5.x).

Halbaroth:
> /home/runner/work/alt-ergo/alt-ergo/_build/default/src/bin/js/main_text_js.bc.js:16468
      b ||= a.length - 1 < (ofs + len | 0) ? 1 : 0;
        ^^^

NOOOOOOOOOOOOOOOO... again a bug of jsoo?

bclement-ocp:
> > /home/runner/work/alt-ergo/alt-ergo/_build/default/src/bin/js/main_text_js.bc.js:16468
> > b ||= a.length - 1 < (ofs + len | 0) ? 1 : 0;
> > ^^^
> 
> NOOOOOOOOOOOOOOOO... again a bug of jsoo?

Apparently ||= is only supported since nodejs 15 (released in 2020) but we are still using an ancient nodejs 14 (?!) in the CI.

bclement-ocp:
Is this still relevant @Halbaroth?

Halbaroth:
> Is this still relevant @Halbaroth?

Sorry, I forgot to finish it. There is an issue with the installation of dependencies on Windows but the log of the jobs are removed. I rerun them to get the error.

bclement-ocp:
No worries! However we should finish this so that we avoid more drift :)

I see you set NodeJS version to 16.9, which is still very old; is there a reason not to use at least version 22 (earliest version that's still supported by nodejs: https://nodejs.org/en/about/previous-releases )?

Halbaroth:
> I see you set NodeJS version to 16.9, which is still very old; is there a reason not to use at least version 22 (earliest version that's still supported by nodejs: https://nodejs.org/en/about/previous-releases )?

It was the minimal version that worked with our code. I upgrade to 22.

The error on Windows looks like a race condition between processes for a common file. I tried to install the same version of the compiler on our Windows remote host and I got another error...  installation of old compilers doesn't work
on Windows. I excluded this version.

hra687261:
It seems that the current version of dolmen (next release) is incompatible with 4.11.2 because it uses `String.exists` (cf. https://github.com/OCamlPro/alt-ergo/actions/runs/35351168786/job/105619383975?pr=1371) which is only available in ocaml `>= 4.13`, so its an additional reason to bump the minimal version of the compiler tested in CI (maybe to 4.14.4?)

bclement-ocp:
As stated above, I don't think we should go beyond what is supported on Debian LTS, which is still 4.13 for the next two years.

PR #1368, opened 2026-08-31
Rely on Z.t and Q.t to build integer and rational literals when possible

hra687261:
Since we now rely on Dolmen as a front-end we should not need to manually parse integer and real literals, notably because Dolmen uses the `Minus` unary operator to represent negative integer/real literals, so AFAIK we only need to parse positive integer/real literals.
And since we use `Z.t` and `Q.t` to represent integer/real literals, we can have negative literals directly instead of rewriting things as `0 - x`. We can also build literals using `Z.t` and `Q.t` directly instead of using strings.

### DIFF
--- a/src/lib/reasoners/intervalCalculus.ml
+++ b/src/lib/reasoners/intervalCalculus.ml
@@ -1898,7 +1898,7 @@ let model_from_simplex sim is_int env uf =
       then main_vars, slake_vars
       else round_to_integers main_vars, round_to_integers slake_vars
     in
-    let fct = if is_int then E.int else E.real in
+    let fct = if is_int then E.Ints.of_Q else E.Reals.of_Q in
     List.fold_left
       (fun acc (v, q) ->
         assert ((not is_int) || Q.is_int q);
@@ -1907,7 +1907,7 @@ let model_from_simplex sim is_int env uf =
           (* may happen because of incremental simplex on rationals *)
           acc
         else
-          let t = fct (Q.to_string q) in
+          let t = fct q in
           let r, _ = X.make t in
           if get_debug_interpretation ()
           then
@@ -1939,7 +1939,7 @@ let model_from_infinite_domains =
 
 let mk_const_term c ty =
   match ty with
-  | Ty.Tint -> E.Ints.of_Z (Q.to_z c)
+  | Ty.Tint -> E.Ints.of_Q c
   | Ty.Treal -> E.Reals.of_Q c
   | _ -> assert false
 
@@ -2120,8 +2120,8 @@ let best_interval_of optimized env p =
 
 let mk_const_term ty s =
   match ty with
-  | Ty.Tint -> E.int (Q.to_string s)
-  | Ty.Treal -> E.real (Q.to_string s)
+  | Ty.Tint -> E.Ints.of_Q s
+  | Ty.Treal -> E.Reals.of_Q s
   | _ -> assert false
 
 let integrate_mapsTo_bindings sbs maps_to =
--- a/src/lib/reasoners/intervalCalculus.ml
+++ b/src/lib/reasoners/intervalCalculus.ml
@@ -1898,7 +1898,7 @@ let model_from_simplex sim is_int env uf =
       then main_vars, slake_vars
       else round_to_integers main_vars, round_to_integers slake_vars
     in
-    let fct = if is_int then E.Ints.of_Q else E.Reals.of_Q in
+    let fct = if is_int then E.Ints.of_Q_exn else E.Reals.of_Q in
     List.fold_left
       (fun acc (v, q) ->
         assert ((not is_int) || Q.is_int q);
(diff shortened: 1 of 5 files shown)


Gbury:
One small note: while currently most languages tend to only have positive literals and parse/type things such as `-5` as the application of the unary minus to a positive literal, this is mainly because it's usually simpler to specify, it's not an absolute rule of dolmen, but rather a coincidence of the currently supported languages. In other term, in the future `dolmen` may emit negative literals (if a language appears in the future that does things this way), so it's probably better to be able to correctly parse those (though also, at one point, `dolmen` may also start to directly emit literals using zarith because of the far-fetched wonders of the smtlibv3....)

hra687261:
Good remark, the new code relies on `Z.of_string` and `Q.of_string`, which (empirically by testing) seem to parse correctly negative literals. The main change of this PR is using negative literals in Alt-Ergo expressions instead of writing as `0 - literal` which was done before. But I am not sure if there is a good reason to prefer `0 - literal` over `- literal`.

> though also, at one point, dolmen may also start to directly emit literals using zarith because of the far-fetched wonders of the smtlibv3

That would be pretty convenient IMO :) what's the reason why it wasn't done before?

Gbury:
> Good remark, the new code relies on `Z.of_string` and `Q.of_string`, which (empirically by testing) seem to parse correctly negative literals. The main change of this PR is using negative literals in Alt-Ergo expressions instead of writing as `0 - literal` which was done before. But I am not sure if there is a good reason to prefer `0 - literal` over `- literal`.

I agree that it's better to use negative literals when possible, it' simpler.

> 
> > though also, at one point, dolmen may also start to directly emit literals using zarith because of the far-fetched wonders of the smtlibv3
> 
> That would be pretty convenient IMO :) what's the reason why it wasn't done before?

Well, the idea was to not force users to use `zarith` and let them use whatever big number library they want (event e.g. using regular integers with checks/assertions to prevent overflows), and it's okay because until now, you didn't need to do any kind of computation on big numbers[1]. However, the smtlib3 plans to have dependant types with full reduction of all interpreted and defined constants, which basically means the model evaluation functions would essentially become a dependency of the typechecker (yay for the insane complexity !....).

[1] : there is one small exception: the bitv/floats sizes, but if one of those ever became too big to represent on a regular integer, I'm not sure any kind of reasonable solving could be done.

src/lib/reasoners/intervalCalculus.ml
```diff
@@ -1898,7 +1898,7 @@ let model_from_simplex sim is_int env uf =
       then main_vars, slake_vars
       else round_to_integers main_vars, round_to_integers slake_vars
     in
-    let fct = if is_int then E.int else E.real in
+    let fct q = if is_int then E.Ints.of_Q q else E.Reals.of_Q q in
```
bclement-ocp: This should not be eta-expanded because we want the `if is_int` test to occur only once (at the definition of `fct`) rather than for each call.

bclement-ocp (changes requested):
> We can [also] build literals using Z.t and Q.t directly instead of using strings.

This looks good to me!

>  And since we use Z.t and Q.t to represent integer/real literals, we can have negative literals directly instead of rewriting things as 0 - x.

Please make this a separate PR. I suspect we might find regressions and this needs thorough testing on our benchmark set first…

src/lib/reasoners/intervalCalculus.ml
hra687261: Well spotted! Done.

src/lib/structures/expr.ml
```diff
@@ -2845,13 +2846,15 @@ end
...
     https://smtlib.cs.uiowa.edu/theories-Ints.shtml *)
 module Ints = struct
-  let of_int n = int (string_of_int n)
+  let of_Z n = mk_term (Int n) [] Tint
 
-  let ( ~$ ) = of_int
+  let ( ~$$ ) = of_Z
 
-  let of_Z n = int (Z.to_string n)
+  let of_Q q = of_Z (Numbers.Q.to_z q)
```
bclement-ocp: Somehow missed that earlier — I think I'd rather have callers use `Numbers.Q.to_z` explicitly rather than expose `E.Ints.of_Q`. Or at least call this `of_Q_exn` (but I think it is better to leave the Q -> Z conversion where we have context to justify it).

bclement-ocp (approved):
LGTM, just not very fond of having `Ints.of_Q`.

src/lib/structures/expr.ml
hra687261: I went with the `of_Q_exn` approach for now, the main reason I added the function is because its needed in a few places and its more convenient to write `E.Ints.of_Q(_exn)` rather than `fun x -> E.Ints.of_Z (Numbers.Q.to_z x)`, but either way, I don't have a strong opinion so I can change it, wdyt?

bclement-ocp: As per above `of_Q_exn` is fine with me :rocket:

hra687261: Alright! Merging it!

